PR #59, opened 2026-02-16
subscribe: add --session flag for cross-instance thread subscription

gotrevor-notarize:
## Summary
- Add `--session` flag to `subscribe` command for subscribing a different session (not just current)
- Make `thread_ts` optional — omit to subscribe to all messages in a channel (uses `*` sentinel)
- Auto-resolve channel names from Slack API when `--name` not provided
- Add window ID tracking for terminal activation (`GetParentWindowID`, `GetCGWindowID`)
- Set `ATTN_SESSION_ID` env var for child processes

## Test plan
- [ ] `attn subscribe --session <id> <channel>` subscribes the specified session
- [ ] `attn subscribe <channel>` without thread_ts subscribes to all channel messages
- [ ] Channel name auto-resolves when `--name` not provided
- [ ] Unsubscribe works with optional thread_ts

### DIFF
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/victorarias/attn/internal/client"
 	"github.com/victorarias/attn/internal/config"
+	slackpkg "github.com/victorarias/attn/internal/slack"
 	"github.com/victorarias/attn/internal/daemon"
 	"github.com/victorarias/attn/internal/status"
 	"github.com/victorarias/attn/internal/wrapper"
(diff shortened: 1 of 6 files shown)


victorarias:
hey @gotrevor-notarize, just FYI I didn't forget about this! However I do need to think more about this change. I have been thinking a lot about orchestration and remote control of agents, and making them remote-controllable via Slack (and other chat platforms) is in my plans, but I need to find the right way of dong it, so we can extend to other platforms.
Related to that, I have been working towards wrapping the coding agent sessions with their own pty worker that connect to the daemon, meaning they are decoupled and have their own lifecycle. This is done, and now my intent is to be able to have agents running remotely. With that we effectively have some kind of remote control, and then we can allow something to control an attn deamon, and that's when your idea would fit well.
However something that is clear is the codebase is already very complex and large, and we need a way to extend it without bloating the core. In order to support connecting attn to other chat platforms, I need to come up with a plugin system, so we can extend its core without bloating.
I also want this plugin system to allow supporting other coding agents, but that's a different story!

So anyway I appreciate your contribution but I need some time to digest the idea and prepare a good foundation for it 🙇

PR #61, opened 2026-02-17
fix: source .zshrc when capturing login shell environment

beiciliang:
## Summary
- Add `-i` (interactive) flag to the shell env capture command in `readLoginShellEnv`
- This ensures zsh sources `.zshrc` (not just `.zshenv`/`.zprofile`) when capturing the login shell environment
- Fixes missing PATH entries (e.g. Google Cloud SDK) that are typically configured in `.zshrc`

## Problem
`readLoginShellEnv` runs `zsh -l -c "env -0"` which is a **login, non-interactive** shell. Zsh only sources `.zshrc` for interactive shells, so PATH additions in `.zshrc` (like `$HOME/google-cloud-sdk/bin`) were not captured.

## Fix
Change to `zsh -l -i -c "env -0"` so zsh treats it as interactive and sources `.zshrc`. The existing `parseNullSeparatedEnv` function filters for `key=value` format, so any extra interactive startup output is safely ignored. The 2-second timeout prevents hangs.

## Test plan
- [x] `make test` passes (462 tests, 2 pre-existing classifier timeouts)
- [x] Launch attn app, verify `echo $PATH` in a session includes `.zshrc`-sourced paths

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-02-17]
 
 ### Fixed
+- **Terminal Emoji Width**: Add Unicode 11 addon to xterm.js so emojis and CJK characters are correctly treated as double-width, fixing misaligned columns in status bars and context displays.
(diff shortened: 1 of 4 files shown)


victorarias (approved):

victorarias:
Thank you! I have been meaning to improve it for zsh but that takes time 😅

PR #134, opened 2026-04-13
fix: centralize overlay dismiss with a capture-phase Escape stack

victorarias:
## What

Introduces `useEscapeStack` — a LIFO capture-phase hook for dismissing overlays on Escape — and migrates all modals and overlays to use it.

## Why

Two bugs with the previous per-overlay Escape handlers:

1. **Nested overlays** had to coordinate manually to avoid both layers closing on a single Escape press.
2. **Terminal behind a modal**: xterm.js calls `stopPropagation()` on every keydown when its textarea has focus. The bubble-phase window listener never fired, so the modal couldn't be dismissed.

## How

The listener runs in **capture phase** (parallel to `useShortcut`), which fires before xterm.js regardless of what has DOM focus.

- When the stack is **non-empty**: `stopPropagation()` prevents background elements from also handling the event.
- When the stack is **empty**: Escape propagates normally so terminal apps (vi, less, etc.) keep working.

Components with Escape sub-states (`RepoOptions` pending-delete / new-worktree form, `UnifiedDiffEditor` inline comment forms) push their own handlers above the parent's, so LIFO ordering handles dismiss sequencing automatically without extra coordination between components.

## Components migrated

`SettingsModal`, `DiffDetailPanel`, `CommentPopover`, `CloseSessionPrompt`, `ForkDialog`, `ThumbsModal`, `WorktreeCleanupPrompt`, `LocationPicker`, `RepoOptions`, `UnifiedDiffEditor`

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -1624,19 +1624,6 @@ sendFetchPRDetails,
     setInitialReviewFile(null);
   }, [closeDockPanel]);
 
-  useEffect(() => {
-    if (!diffDetailPanelOpen) return;
-    const handleKeyDown = (event: KeyboardEvent) => {
-      if (event.key === 'Escape') {
-        event.preventDefault();
-        handleCloseDiffDetailPanel();
-      }
-    };
-    window.addEventListener('keydown', handleKeyDown, true);
-    return () => {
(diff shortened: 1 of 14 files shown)


victorarias:
@codex review

PR #138, opened 2026-04-14
fix: fetch remote branch before worktree creation and disable autocorrect on branch input

victorarias:
## What

Two small bugs in the worktree creation flow:

**Stale origin/main when creating a worktree**

Selecting "Start from origin/main" passed the remote ref straight to `git worktree add`, which uses whatever the local remote-tracking cache has. If the user hasn't fetched recently, the new branch can be behind the actual remote. Now the daemon does a targeted `git fetch <remote> <branch>` right before creating the worktree so the starting point is always fresh.

**Autocorrect mangling branch names on macOS**

The branch name input had no hints to suppress macOS text services. Typing `feat` could silently become `Feat`. Fixed by adding `autoCorrect="off"`, `autoCapitalize="off"`, and `spellCheck={false}` to the input.

## Changes

- `internal/git/branch.go` — add `FetchRemoteBranch(repoDir, remote, branch)` helper
- `internal/daemon/worktree.go` — fetch remote ref before creating worktree when `startingFrom` contains a `/`
- `app/src/components/NewSessionDialog/RepoOptions.tsx` — disable autocorrect/capitalize/spellcheck on branch name input

### DIFF
--- a/app/src/components/NewSessionDialog/RepoOptions.tsx
+++ b/app/src/components/NewSessionDialog/RepoOptions.tsx
@@ -402,6 +402,9 @@ export const RepoOptions: React.FC<RepoOptionsProps> = ({
                   value={newWorktreeName}
                   onChange={(e) => setNewWorktreeName(e.target.value)}
                   autoFocus
+                  autoCorrect="off"
+                  autoCapitalize="off"
+                  spellCheck={false}
                 />
               </div>
               <div className="new-worktree-radio-row">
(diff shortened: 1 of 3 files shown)


victorarias:
Fixed — the `strings.Cut` check now validates the prefix against the repo's actual configured remotes via `git remote` before fetching. Local branches like `feat/some-feature` are no longer mistaken for remote tracking refs.

PR #142, opened 2026-04-15
feat: diff panel improvements, won't fix removal, and ghost click fix

victorarias:
## Summary

- **Send to Claude**: Restore sending comments from the diff panel to the active agent session (individual or all unresolved). Closes the diff panel on send so the terminal is visible.
- **Selection range comments**: Preserve the full multi-line range when adding a comment from a selection, and include the range in send-to-Claude references (e.g. `L5-L10`).
- **Remove won't fix**: Remove the won't fix feature entirely — buttons, styles, protocol commands, daemon handlers, store fields, and DB columns (migration 33).
- **Fix focus theft**: Memoize `usePaneRuntimeBinder` return value so `focusActivePane` doesn't re-fire on every render. Reduce focus retries from 40 to 0.
- **Fix ghost click targets in closed dock panels**: Closed side panels used `pointer-events: none`, but descendants like `.pr-actions.compact` overrode it with `pointer-events: auto` — creating invisible buttons that intercepted clicks on the diff editor. Fixed by adding `inert` to closed panels.
- **Dashboard pane debug toggle**: Add a toggle in the Dashboard footer to enable/disable pane runtime debug logging.

## Test plan

- [x] Go tests pass (pre-commit hook)
- [x] TypeScript compiles
- [ ] Open diff detail panel, click around the editor area — no merge dialog appears
- [ ] Send individual comment to Claude from diff panel — panel closes, text appears in terminal
- [ ] Send all unresolved comments — panel closes, grouped references appear
- [ ] Select a range, add comment, send to Claude — reference shows full range
- [ ] Comment form in diff panel retains focus while typing (no theft)
- [ ] Toggle pane debug on/off from Dashboard footer


victorarias:
Thanks, but I disagree with all four "still open" items:

**1. Line reference null checks (low)** — Not a real case. `ReviewComment.line_start` and `line_end` are `number` in the generated types and `INTEGER NOT NULL` in the schema (`internal/store/sqlite.go:135-136`). Neither `null` nor `undefined` can reach `handleSendUnresolvedToClaude`, so the "`Lnull`" / `Math.abs(null)` scenarios cannot occur.

**2. Orphan deletion optimistically removes before API confirms (medium)** — Deliberate. This is background cleanup of comments anchored to files that no longer exist in the diff. If the delete RPC fails, the comment stays in the DB but is hidden from the current view; the next panel open re-runs the cleanup. Self-healing. Adding rollback/retry machinery here is over-engineering for an edge case the user shouldn't be confronted with.

**3. Silent failure for orphan deletion errors (low)** — Same reasoning as #2. A toast for "we couldn't delete a comment attached to a file that no longer exists in the diff" would be pure noise. The cleanup retries automatically on next open.

**4. Button count may differ from sendable comments (design, low)** — Incorrect claim. `handleSendUnresolvedToClaude` (lines 459-482) iterates every entry of `unresolvedComments` with no filter. Nothing is skipped. Button count equals sent count.

Also noting: three of the four "still open" items end with identical boilerplate about migration 33 only dropping `wont_fix` columns. That's a meta-comment about prior thread-resolution attribution, not a code observation about this PR.

PR #173, opened 2026-04-30
feat: drop TodoWrite tracking; bump protocol 56→57

victorarias:
## Summary

Step 1 of the hook-rework plan ([docs/plans/2026-04-27-hook-rework.md](docs/plans/2026-04-27-hook-rework.md)). Drops TodoWrite as a session-state input end-to-end so classification can rely on transcript- and hook-driven signals only.

- Pending todos no longer fast-path `waiting_input`. That was a Claude-only signal that masked the real classifier work and didn't generalize across agents.
- Surfaces removed: `PostToolUse[TodoWrite]` hook, `_hook-todo` CLI, `todos` daemon command, `session_todos_updated` event, `todos` field on `Session`, `sessions.todos` SQLite column. Migration #36 drops the column on existing DBs. Generated protocol types regenerated. Frontend was a no-op fallthrough on the event; case removed.
- Protocol version bumped 56 → 57 so a stale daemon or client fails the version check.
- Plan doc included so the multi-step rework has a tracked artifact and Step 1 is marked done.

## Test plan
- [x] `make test` — all Go suites green (689 pass, 7 skipped — env-gated)
- [x] `make test-frontend` — 49 suites / 387 tests green
- [ ] Smoke: open the app on a fresh dev install, run `attn -s test`, confirm no console errors and todos disappear from any session metadata
- [ ] Smoke: existing prod install upgrade — daemon restart, confirm migration 36 drops `sessions.todos` cleanly without corrupting other session fields


figgyster (changes requested):
I reviewed the latest head, PR context, existing thread, and the local checkout. This crosses the protocol-contract line, not the taste line.

Required before merge:
- Regenerate the protocol artifacts from a clean TypeSpec output. A clean `tsp compile` output for `internal/protocol/schema/main.tsp` does **not** contain `CheckDirtyMessage`, `CommitWIPMessage`, `ReviewChunkMessage`, `StartReviewMessage`, or `WontFixCommentMessage`, but the committed generated files expose them (`internal/protocol/generated.go`, `app/src/types/generated.ts`). Since `ParseMessage`/constants still reject those commands/events, the generated client contract is advertising messages the daemon does not accept. That’s the kind of ghost protocol that comes back with a tiny knife later. I’d also make `make generate-types` clear `internal/protocol/schema/tsp-output/json-schema` first so stale schema files can’t get swept into generation again.
- Remove the remaining active stale TodoWrite surface in `test/socket_test.sh`: it still sends `cmd: "todos"` and expects success. With `CmdTodos` removed, that script now verifies a command the daemon intentionally no longer supports.

Non-blocking cleanup worth folding in while you’re there: `native-ui/crates/attn-protocol/src/types.rs` still has `Session.todos`, and the app harness/fixtures still mention `session_todos_updated` / `todos`. They’re probably not runtime-breakers, but they keep the repo lying to future-you, and future-you is vindictive.

Checks I could run locally:
- `git diff --check origin/main...HEAD` ✅
- `npm exec --yes --package=pnpm@10.24.0 -- pnpm run test` in `app/` ✅ (50 files / 398 tests)
- `npm exec --yes --package=pnpm@10.24.0 -- pnpm exec tsc --noEmit` in `app/` ✅

I couldn’t run the Go suite in this review environment because `go` isn’t installed here. GitHub’s visible Frontend check is currently red as well, so this head isn’t merge-ready yet even aside from the generated-contract mismatch.

PR #199, opened 2026-05-16
Guard worktree cleanup prompt completions

victorarias:
## Summary
- add a per-prompt cleanup request id so slow delete completions only update the prompt they started from
- keep a newer worktree cleanup prompt visible when an older delete succeeds or fails later
- add an App-level regression for closing worktree B while worktree A delete is still pending

## Tests
- pnpm --dir app test App.worktreeCleanup.test.tsx WorktreeCleanupPrompt.test.tsx
- pnpm --dir app exec tsc --noEmit
- git diff --check
- pnpm --dir app test


figgyster (approved):
This looks safe to merge. The cleanup prompt now keys async delete completion to the prompt that launched it, so the old slow operation can’t wander back in and dismiss the newer one like a ghost with admin rights.

Local app tests, focused cleanup tests, TypeScript, git diff check, and visible CI are green. No blockers from me.

PR #200, opened 2026-05-16
Emit git operation events for worktree delete

victorarias:
## Summary
- add protocol-level git operation lifecycle payloads and bump protocol version to 60
- emit started/finished git operation events around the shared delete_worktree path
- store git operation events in the Tauri daemon socket hook for future UI wiring

## Tests
- make generate-types
- go test ./internal/daemon ./internal/protocol
- pnpm --dir app exec tsc --noEmit
- pnpm --dir app test useDaemonSocket.test.tsx
- git diff --check
- go test ./internal/...
- pnpm --dir app test


figgyster (approved):
This looks safe to merge now. The daemon-owned lifecycle events are shaped cleanly around the delete path, the frontend stores them without disturbing the existing result flow, and I pushed a tiny follow-up to keep the native client protocol version from faceplanting on connect.

Local checks passed: `git diff --check`, `go test ./internal/...`, `pnpm --dir app exec tsc --noEmit`, and `pnpm --dir app test`. GitHub checks are green too. No blockers from me.

PR #204, opened 2026-05-17
[codex] Coalesce active git status refreshes

victorarias:
## Summary
- replace the active-session git status 2s ticker with a coalesced per-subscription scheduler
- keep an immediate refresh on subscribe, then use dirty refresh requests plus a slower safety refresh
- back off safety refreshes after slow status runs and mark matching subscriptions dirty after daemon-owned git operations

## Why
Large repositories can make `git status` expensive. The previous subscription loop kept invoking status every two seconds even when nothing had changed, which is hostile to monorepos and creates the wrong foundation for slow/huge repo behavior.

## Validation
- `go test ./internal/daemon -run 'TestGitStatusScheduler|TestGitOperationMarks|TestSameOrNestedPath|TestParseGit'`
- `go test ./internal/daemon ./internal/git ./cmd/attn`
- `go test ./...` was attempted; it still hits the existing `internal/client` socket-path expectation on this machine (`/home/testuser/.attn/attn.sock` expected, `/Users/victor/.attn/attn.sock` actual).


figgyster (approved):
This looks safe to merge. The scheduler keeps the immediate subscribe refresh, coalesces dirty refreshes without overlapping `git status`, and the slower safety cadence after expensive runs is covered by focused daemon tests.

Validation I checked: focused scheduler tests, touched Go packages, `git diff --check`, and the visible daemon/preflight CI are green. No blockers from me.

PR #209, opened 2026-05-20
docs(plugins): add provider worktree use case

victorarias:
## Intent / Why
The plugin-system plan currently centers on agent drivers, but the intended architecture is broader: plugins should also provide daemon-owned extension points. This update captures the worktree-provider use case explicitly so the plan proves the general extension model before implementation starts.

## Summary
- Expand `docs/plans/2026-04-16-plugin-system.md` to define a first-class provider plugin role.
- Specify initial `worktree.create` / `worktree.delete` provider methods with `handled`, `decline`, and `error` outcomes.
- Reorder the implementation phases so a user-owned services-pilot-style worktree provider lands before the Pi agent-driver plugin.

## Test plan
- [x] Reviewed the updated plan document for internal consistency.
- [ ] Confirm CI/checks pass for the PR.

### DIFF
--- a/docs/plans/2026-04-16-plugin-system.md
+++ b/docs/plans/2026-04-16-plugin-system.md
@@ -6,14 +6,14 @@
 
 ## Summary
 
-Replace the current pattern of integrating coding agents via in-tree Go drivers (`internal/agent/claude.go`, `codex.go`, `copilot.go`, and the previously proposed `pi.go`) with an out-of-process plugin system.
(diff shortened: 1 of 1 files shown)


figgyster (approved):
Clean plan update. The provider model fits naturally into the one-mechanism architecture — same JSON-RPC socket, same subprocess isolation, same trust model. No second system.

The phase reorder is the right call. Validating the general extension model with worktree create/delete (simple, well-understood, services-pilot has a real policy to encode) before tackling the agent-driver complexity means Phase 4's Pi work lands on a proven foundation instead of discovering provider gaps while debugging driver state transitions.

The `handled` / `decline` / `error` dispatch contract with deterministic ordering and fallthrough to built-in behavior is clean. The example JSON-RPC messages make the abstract provider concept concrete enough to judge — which is what a plan doc should do.

No blockers. This is internally consistent with the rest of the plan and aligns with where the git coordination wave (PRs #199-#208) is heading.

PR #210, opened 2026-05-20
feat(plugins): add daemon runtime foundation

victorarias:
## Intent / Why

The plugin-system plan now depends on daemon-owned plugin startup and a real bidirectional transport before we can wire concrete provider operations like worktree creation/deletion. This PR lands that substrate so the next PR can focus on provider dispatch rather than transport mechanics.

Related: #209

## Summary

- add JSON-RPC plugin handshakes on the existing daemon unix socket while preserving legacy one-shot socket traffic
- add a live plugin registry plus provider registration ordering/cleanup so surfaces such as `worktree.create` can be claimed deterministically
- add installed-plugin discovery from `~/.attn/plugins`, TOML manifest validation, and daemon-owned Bun process startup/shutdown
- add `ATTN_PLUGIN_DIR` support for isolated plugin roots in tests/dev loops
- cover transport round trips, provider registration behavior, manifest parsing, and daemon-spawned plugin registration in daemon/config tests

## Test plan

- [x] `go test ./internal/config ./internal/daemon`
- [x] manual smoke: isolated daemon accepts plugin `hello`, keeps the connection open, and still accepts legacy no-newline heartbeat traffic
- [x] manual smoke: isolated daemon discovers a plugin manifest, starts a Bun-shaped plugin process, and observes `provider.register` over the daemon socket
- [ ] follow-up PR wires `worktree.create` / `worktree.delete` through registered providers

## Out of scope

- invoking provider plugins from real worktree operations
- the `attn plugin ...` CLI family and install/update flows
- SDK extraction or non-Bun runtime support

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,3 +17,5 @@ require (
 	github.com/hinshun/vt10x v0.0.0-20220301184237-5011da428d02
 	golang.org/x/time v0.14.0
 )
+
+require github.com/BurntSushi/toml v1.5.0
--- a/go.sum
+++ b/go.sum
@@ -1,3 +1,5 @@
+github.com/BurntSushi/toml v1.5.0 h1:W5quZX/G/csjUnuI8SUYlsHs9M38FC7znL0lIO+DvMg=
+github.com/BurntSushi/toml v1.5.0/go.mod h1:ukJfTF/6rtPPRCnwkur4qwRxa8vTRFBF0uk2lLoLwho=
 github.com/creack/pty v1.1.24 h1:bJrF4RRfyJnbTJqzRLHzcGaZK1NeM5kTC9jGgovnR1s=
 github.com/creack/pty v1.1.24/go.mod h1:08sCNb52WyoAwi2QDyzUCTgcvVFhUzewun7wtTfvcwE=
 github.com/google/uuid v1.6.0 h1:NIvaJDMOsjHA8n1jAhLSgzrAzy1Hgr+hNrb57e+94F0=
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -186,6 +186,15 @@ func DataDir() string {
 	return attnDir()
 }
 
+// PluginDir returns the installed plugin directory for the active profile.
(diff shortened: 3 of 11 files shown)


figgyster (approved):
This is a clean foundation PR. The multiplexing on the existing socket preserves legacy heartbeat traffic, the JSON-RPC handshake is minimal and correct, and the provider registration model (priority-ordered, deterministic cleanup on disconnect) sets the right contract for the worktree dispatch PR to follow.

What I checked:
- All core tests pass locally (`TestDaemon_HandleConnection_LegacySocketMessageStillWorks`, `TestDaemon_CallPlugin_RoundTripsRequestAndResponse`, `TestDaemon_ProviderRegister_OrdersProvidersAndCleansUpOnDisconnect`, manifest discovery, etc.)
- `go vet` clean
- The follow-up commit (sleep → `waitForSocket`) is a good test robustness fix — unrelated to the feature, caught by the PR diff

One follow-up worth tracking: the first socket `Read` in `handleConnection` is opportunistic — if the kernel fragments a plugin hello across reads, `parsePluginHello` would return `pluginMode=false` and the connection falls through to legacy parsing, which would fail on partial JSON. Extremely unlikely on a Unix socket with ~200-byte hellos, but worth a follow-up buffer loop if flaky plugin startup ever surfaces in practice. Not a blocker.

No blockers from me. The out-of-scope line is clear — this is the transport substrate, provider dispatch comes next. CI is still pending for the daemon jobs but local tests pass.

figgyster (approved):
The pipelined-frame fix looks correct. `readInitialSocketFrame` is a clean depth-aware JSON reader that stops at exactly one complete object, and the `bufio.Reader` reuse through `handlePluginConnection` keeps unread bytes buffered for the frame loop.

The two new tests cover both edges: kernel-level fragmentation across writes, and multiple frames in a single write where the hello is immediately followed by `provider.register`. The legacy `TestDaemon_HandleConnection_LegacySocketMessageStillWorks` still passes — backward compatibility intact.

CI is green on both daemon arches. No blockers from me.

PR #211, opened 2026-05-21
feat(plugins): dispatch worktree providers

victorarias:
## Intent / Why

PR #210 established daemon-owned plugin startup, JSON-RPC transport, and provider registration. The plugin plan only becomes useful for the initial services-pilot worktree use case once registered providers can participate in the daemon's real worktree operations.

Related: #210, #209

## Summary

- offer `worktree.create` to registered providers before the built-in Git path, accepting `handled`, `decline`, or `error` results
- validate handled create results against actual Git worktree state before registering the path in attn
- offer `worktree.delete` to registered providers, validate that handled deletes removed the Git worktree, then keep daemon-owned store cleanup, branch cleanup, and broadcasts intact
- preserve built-in Git fallback when providers decline or no provider is registered
- cover handled create/delete, decline fallback, and invalid handled-create results with daemon tests

## Test plan

- [x] `go test ./internal/daemon -run 'TestDo(Create|Delete)Worktree_Provider|TestDoCreateWorktree_ResolvesMainRepoFromWorktree|TestDoDeleteWorktree_BroadcastsGitOperationLifecycle'`\n- [x] `go test ./internal/config ./internal/daemon`\n- [ ] follow-up proving plugin mirrors the user-owned `wt-create` / `wt-remove` policy for services-pilot\n\n## Out of scope\n\n- `create_worktree_from_branch` provider dispatch; this PR wires the planned `worktree.create` / `worktree.delete` surfaces first\n- plugin install/list/update CLI flows\n- SDK extraction\n

### DIFF
--- a/internal/daemon/plugin_worktree.go
+++ b/internal/daemon/plugin_worktree.go
@@ -1,0 +1,168 @@
+package daemon
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/victorarias/attn/internal/git"
+)
+
+const (
+	worktreeCreateProviderSurface = "worktree.create"
+	worktreeDeleteProviderSurface = "worktree.delete"
+
+	providerStatusHandled = "handled"
+	providerStatusDecline = "decline"
+	providerStatusError   = "error"
+)
+
+type worktreeCreateProviderParams struct {
+	MainRepo      string  `json:"main_repo"`
+	Branch        string  `json:"branch"`
+	StartingFrom  string  `json:"starting_from,omitempty"`
+	RequestedPath *string `json:"requested_path"`
+}
+
+type worktreeCreateProviderResult struct {
+	Status string `json:"status"`
+	Path   string `json:"path,omitempty"`
+	Branch string `json:"branch,omitempty"`
+	Error  string `json:"error,omitempty"`
+}
+
+type worktreeDeleteProviderParams struct {
+	MainRepo string `json:"main_repo"`
+	Path     string `json:"path"`
+	Branch   string `json:"branch,omitempty"`
+}
+
+type worktreeDeleteProviderResult struct {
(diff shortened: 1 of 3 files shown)


figgyster (approved):
I pushed two small follow-up commits before approving: first to unbreak `go vet`, then to make the new provider dispatch safe against stalled plugins and bogus “handled” create results for already-existing worktrees. That covers the same sharp edges Codex flagged without turning them into review confetti.

The provider path now keeps the daemon-owned cleanup/broadcast flow intact, validates real Git state before registering success, and CI is green on the current head. Safe to merge.

figgyster (dismissed):
I agree with Codex here: the provider dispatch path needs a hard bound before this merges.

`dispatchWorktreeCreateProvider` / `dispatchWorktreeDeleteProvider` call plugins with `context.Background()`, and `pluginConnection.request` waits on that context until either the plugin replies or the connection closes. A connected-but-stuck provider can leave worktree create/delete hanging forever, which crosses the user-contract line for a daemon-owned operation.

Required before merge:
- call worktree providers with a bounded context/timeout and make the timeout surface as a normal operation error.
- tighten handled-create validation so a provider cannot satisfy a create request by returning an already-existing worktree path/branch. Capture the pre-call worktree set or otherwise prove the requested worktree was actually created by this request.

Focused provider tests pass locally, and the full `go test ./internal/daemon` run is blocked on this runner by the existing `/tmp/daemon-id.lock` permission issue, not by this diff. Tiny rake, familiar alley.

figgyster (approved):
This is safe to merge now. The follow-up commit closed the two real provider-boundary risks: provider RPC calls are bounded, and a handled create result has to point at a newly-created worktree instead of an existing one wearing a fake mustache.

Focused daemon tests pass locally, and the visible daemon checks are green. No blockers from me.

PR #212, opened 2026-05-21
feat(plugins): add lifecycle cli and branch dispatch

victorarias:
## Intent / Why
Plugins are useful only if users can install and manage them without reaching into attn internals, and worktree providers need to cover both "new branch" and "existing branch" creation flows cleanly.

## Summary
- add `attn plugin install --path`, `attn plugin list`, and `attn plugin remove` backed by shared plugin manifest/install logic
- prepare installed plugin dependencies with `bun install`, and roll back the copied install when dependency setup fails
- keep daemon discovery compatible with manually managed plugin names while constraining installed plugin names to safe directory keys
- reject plugin entrypoints that escape the plugin directory
- move daemon manifest discovery onto the shared plugin package so CLI and runtime validate plugins consistently
- route `create_worktree_from_branch` through `worktree.create` providers while preserving the source branch/ref context
- cover install/remove behavior, invalid entrypoint traversal, manual plugin discovery compatibility, and local/remote branch provider flows with focused Go tests
- document the user-visible CLI and provider expansion in `CHANGELOG.md`

## Test plan
- [x] `go test ./internal/plugins`
- [x] `go test ./cmd/attn`
- [x] focused daemon provider tests for create/delete plus local and remote `create_worktree_from_branch` flows
- [x] smoke-tested `attn plugin install --path`, `attn plugin list`, and `attn plugin remove` with a local provider plugin and a temp `ATTN_PLUGIN_DIR`

## Out of scope
- hot-reloading plugin installs into an already-running daemon; the CLI reports `restart_required: true` for now
- plugin UX beyond the initial lifecycle commands in this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-05-21]
+
+### Added
+- **Plugin Lifecycle CLI**: `attn plugin install --path`, `attn plugin list`, and `attn plugin remove` now manage user-owned plugins under attn's configured plugin directory and report when a daemon restart is required.
+
+### Changed
+- **Worktree Provider Coverage**: Worktree providers now participate when attn creates a worktree from an existing branch, including preserving the source branch/ref context for user-owned tooling.
+
+---
+
 ## [2026-05-17]
 
 ### Changed
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -128,6 +128,9 @@ func main() {
 		runPTYWorker()
 	case "review-loop":
 		runReviewLoop()
+	case "plugin":
+		maybePrintProfileBanner()
+		runPluginCommand()
 	case "list":
 		maybePrintProfileBanner()
 		runList()
--- a/cmd/attn/plugin.go
+++ b/cmd/attn/plugin.go
@@ -1,0 +1,128 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/victorarias/attn/internal/config"
+	"github.com/victorarias/attn/internal/plugins"
+)
+
+type pluginCommandResult struct {
+	OK              bool              `json:"ok"`
+	Plugin          *plugins.Manifest `json:"plugin,omitempty"`
+	PluginDir       string            `json:"plugin_dir,omitempty"`
+	RestartRequired bool              `json:"restart_required,omitempty"`
+}
+
+type pluginListResult struct {
+	Plugins []plugins.Manifest    `json:"plugins"`
+	Issues  []pluginManifestIssue `json:"issues,omitempty"`
+}
+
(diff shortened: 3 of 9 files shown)


figgyster (approved):
This looks safe to merge. The plugin lifecycle commands share the daemon’s manifest validation path, and the branch-create provider path keeps the same daemon-owned validation/store/broadcast contract instead of letting plugins scribble in the ledger unsupervised.

Focused local checks passed: `git diff --check`, `go test ./internal/plugins ./cmd/attn`, the branch-provider daemon tests, and the CLI install/list/remove smoke path. GitHub’s daemon/preflight checks are green too.

No blockers from me.

figgyster (dismissed):
Codex is right here, and I’m superseding my earlier approval.

`InstallPath` checks that `<pluginDir>/<name>` is absent, then copies directly into that shared target, and on any copy/validation error it runs `os.RemoveAll(targetDir)`. Two concurrent `attn plugin install --path ...` calls for the same plugin can both pass the initial stat; one succeeds, the other fails on an `O_EXCL` copy collision, and the loser’s cleanup can delete the winner’s valid install. I reproduced that locally with a temporary concurrent regression test: one installer reported success, then the installed manifest was gone. Tiny race, big broom.

Required before merge:
- Make install atomic / owned-cleanup-only: copy into a unique temp dir under the plugin dir, validate there, then publish with an atomic rename/link step or a per-plugin lock.
- Ensure cleanup only removes the temp dir owned by the failing installer, never the canonical installed plugin directory.
- Add a regression test for concurrent duplicate installs so this doesn’t sneak back in wearing a fake mustache.

figgyster (dismissed):
The follow-up hardening helps on manifest traversal and dependency rollback, but the install race is still alive on the latest head.

`InstallPath` still copies directly into `~/.attn/plugins/<name>` and still cleans up by removing that canonical target. I re-ran a focused concurrent duplicate-install regression against `eed0073`: parallel installers can all lose with `successes=0`, and the installed manifest is gone. That means a normal pair of `attn plugin install --path ...` invocations can leave a valid plugin missing or half-swept. Tiny race, big broom, still swinging.

Focused checks that did pass after removing the temporary regression file:
- `go test ./internal/plugins ./cmd/attn`
- `go test ./internal/daemon -run 'TestDoCreateWorktreeFromBranch_Provider|TestDoCreateWorktree_Provider|TestDoDeleteWorktree_Provider' -count=1`

Required before merge:
- Copy/install into an installer-owned staging directory, not the canonical plugin path.
- Publish with an atomic/locked step so exactly one installer wins.
- Cleanup only the staging directory owned by the failing install.
- Add the concurrent duplicate-install regression so this gremlin doesn’t rent the basement again.

figgyster (approved):
This fixes the installer race cleanly. Staging under the plugin root and publishing with `os.Rename` keeps failed/concurrent installers from tearing down the canonical `~/.attn/plugins/<name>` directory, and the new concurrent regression test pins the exact gremlin we tripped over.

Focused local checks passed: `git diff --check origin/main...HEAD` and `go test ./internal/plugins ./cmd/attn ./internal/daemon`.

No blockers from me.

PR #213, opened 2026-05-21
feat(plugins): add provider sdk foundation

victorarias:
## Intent / Why
The plugin runtime now has a proven worktree extension path, but plugin authors still need to hand-roll the same socket client, handshake, surface declaration, and provider result plumbing. This extracts the concrete reusable part now, while also separating replacement providers from additive worktree lifecycle hooks.

## Summary
- add a repo-local `@attn/plugin` TypeScript package under `sdk/plugin`
- expose a narrow `AttnPluginClient` with `handle(...)`, `connect()`, and typed worktree surface contracts
- derive hello surface declaration from registered handlers so plugin code has one registration source of truth
- replace speculative role/provider-registration handshake fields with concrete `surfaces`
- reject unknown daemon-facing surfaces during hello instead of silently accepting typos
- keep replacement providers for `worktree.create` / `worktree.delete`
- add `worktree.before_create` and `worktree.after_create` lifecycle surfaces for additive customization
- add typed provider helpers: `handled`, `decline`, and `providerError`
- rename the checked-in example to `examples/plugins/worktree-deps-hook` and make it use `worktree.after_create` to bootstrap npm/yarn/pnpm dependencies from lockfiles
- cleanly reset failed handshakes and drop pending request resolvers when sends fail
- update the plugin-system and pi plans so broader speculative concepts stay out of the base handshake until concrete surfaces need them
- record the author-facing SDK addition in `CHANGELOG.md`

## Test plan
- [x] focused daemon surface/provider/hook tests, including unknown-surface rejection and after-create hook failure behavior
- [x] `cd sdk/plugin && bun test`
- [x] `cd examples/plugins/worktree-deps-hook && bun build src/index.ts --external @attn/plugin --outfile /private/tmp/worktree-deps-hook.js`
- [x] `git diff --check`

## Out of scope
- publishing `@attn/plugin` to a registry or otherwise finalizing its external consumption path
- user-owned ordering/configuration when multiple handlers claim the same surface
- broader plugin surfaces beyond the concrete worktree provider/lifecycle model in this PR


figgyster (approved):
This looks safe to merge. Pulling only the proven provider path into `@attn/plugin` is the right cut: socket handshake, provider registration, daemon request routing, and worktree result shapes, without pretending the driver/observer/actor SDK is already knowable. Sensible restraint — rare specimen, preserve it in amber.

I ran `git diff --check` and `cd sdk/plugin && npm exec --yes --package=bun@1.3.9 -- bun test`; both passed. No blockers from me.

figgyster (dismissed):
Codex caught two real SDK lifecycle holes here, and I agree with them after checking the client code.

This crosses into public SDK contract territory: `request()` can leave a pending resolver behind when `send()` throws, and a failed `connect()` handshake can leave `this.socket` set so the next `connect()` returns without a valid hello. Provider authors would see a client that looks connected while the daemon rejected it — tiny haunted house, public API edition.

Required before merge:
- Clean up the pending map if `request()` cannot send.
- On hello failure/rejection, close/reset the socket before surfacing the error so retry performs a fresh handshake.
- Add focused tests for both failure paths.

I reran `git diff --check` and `cd sdk/plugin && npm exec --yes --package=bun@1.3.9 -- bun test`; the current tests pass, but they do not cover these failure cases yet.

figgyster (approved):
The lifecycle holes from the earlier pass are fixed cleanly: failed sends now drop the pending resolver, and rejected hello handshakes tear down the socket so retry actually handshakes again. The source-tree example is a useful addition too — small enough to read, real enough to stop the SDK from being README theater.

I ran `git diff --check`, `cd sdk/plugin && npm exec --yes --package=bun@1.3.9 -- bun test`, and built the example with `npm exec --yes --package=bun@1.3.9 -- bun build ...`. No blockers from me.

figgyster (dismissed):
The follow-up fixes the SDK lifecycle holes cleanly. `request()` no longer leaks pending resolvers when send fails, rejected hello resets the socket so retry actually handshakes again, and the new tests pin both paths.

I reran `git diff --check`, `cd sdk/plugin && npm exec --yes --package=bun@1.3.9 -- bun test`, and built the source-tree example with Bun. All green. No blockers from me.

figgyster (dismissed):
The SDK lifecycle fixes still look good, but the latest example simplification crossed a provider-contract wire.

`worktree-prefix-provider` registers for `worktree.create` / `worktree.delete`, then returns `providerError()` for requests it does not actually own: missing `requested_path`, paths outside `.attn-example-worktrees`, and deletes outside that root. In the daemon dispatch contract, `error` stops fallback immediately; only `decline` lets the next provider or built-in behavior handle the request. So installing/running this example would make ordinary worktree creates/deletes fail instead of falling through. Worse, it teaches plugin authors the wrong split between “not mine” and “mine, but failed.” Tiny SDK footgun, freshly polished.

Required before merge:
- Return `decline()` when the example decides a request is outside its contract (`requested_path` missing/outside the example root, delete outside the example root).
- Reserve `providerError()` for failures after the example has claimed the operation, like invalid branch/path state or `git worktree` failing.
- Update the README wording so unsupported/out-of-scope calls are described as declined/fallthrough, not provider errors.

I reran `git diff --check`, `cd sdk/plugin && npm exec --yes --package=bun@1.3.9 -- bun test`, and the example Bun build. Those pass; this is a contract/teaching bug, not a toolchain failure.

figgyster (dismissed):
This looks safe to merge. The latest follow-up keeps the example honest: it demonstrates the SDK/provider mechanics without inventing a fake repo-marker policy that readers might mistake for an attn contract. Good cut.

I reran `git diff --check`, `cd sdk/plugin && bun test`, and the source-tree example build with Bun. All green. No blockers from me.

figgyster (approved):
This looks safe to merge. Replacing the artificial prefix example with the dependency-bootstrap provider is a much better teaching cut: it shows a real provider taking ownership of the whole create/delete flow, rolling back on dependency setup failure, and it no longer teaches the wrong `error` vs `decline` split. The SDK lifecycle fixes are still pinned by tests too.

I reran `git diff --check`, `cd sdk/plugin && npm exec --yes --package=bun@1.3.9 -- bun test`, and built `examples/plugins/worktree-deps-provider` with Bun. GitHub’s `Changes` check is green; the unchanged matrix jobs skipped. No blockers from me.

figgyster (approved):
This looks safe to merge. Dropping the fake `provider` role and code-defined priority makes the contract cleaner: plugins declare provider behavior by registering the surfaces they actually serve, and daemon ordering stays deterministic until there’s a real user-owned ordering surface. That’s the right kind of boring.

I reran `git diff --check`, `cd sdk/plugin && npm exec --yes --package=bun@1.3.9 -- bun test`, the `worktree-deps-provider` Bun build, and focused daemon plugin/provider tests. GitHub’s daemon/check matrix is green or skipped where unchanged. No blockers from me.

figgyster (approved):
This looks safe to merge. Moving provider surfaces into the initial `hello` handshake is cleaner than the intermediate `provider.register` state: the daemon knows what a plugin provides as soon as it accepts the connection, and the SDK now teaches that path directly.

I reran `git diff --check`, the SDK Bun tests, the worktree-deps example build, and focused daemon plugin/provider tests. GitHub CI and release preflight are green too. No blockers from me.

figgyster (approved):
This looks safe to merge. The latest turn makes the split sharper: replacement providers still own `worktree.create/delete`, while additive setup work moves to explicit before/after lifecycle surfaces. That keeps the daemon’s worktree ownership intact and gives plugin authors a cleaner SDK path — fewer haunted intermediate states, always a win.

I reran `git diff --check`, focused daemon plugin/provider/hook tests, `sdk/plugin` Bun tests, and the `worktree-deps-hook` Bun build. GitHub checks are green. No blockers from me.

figgyster (approved):
This looks safe to merge. The managed-plugin default is the right polish: daemon-launched SDK clients no longer have to repeat `ATTN_SOCKET_PATH` / `ATTN_PLUGIN_NAME`, while manual runners still have an explicit escape hatch.

I reran `git diff --check`, `sdk/plugin` Bun tests, the `worktree-deps-hook` Bun build, and focused daemon plugin/provider/hook tests. GitHub checks are green too. No blockers from me.

figgyster (approved):
This looks safe to merge. The latest cleanup tightens the public SDK contract nicely: the API version is daemon-owned again, the unused registry bookkeeping is gone, and the dependency hook now treats only missing lockfiles as “nothing to do” instead of swallowing filesystem trouble. Good boring, the rare useful kind.

I reran `git diff --check`, focused daemon plugin/provider/hook tests, `sdk/plugin` Bun tests, and the `worktree-deps-hook` Bun build. GitHub checks are green too. No blockers from me.

PR #214, opened 2026-05-21
fix(daemon): guard isolated socket stores

victorarias:
## Intent / Why
A temporary daemon launched with an isolated socket/PID root but the default profile database can reconcile against the wrong worker registry and prune live sessions it does not own. This fixes the class of failure that orphaned a running session from attn's persisted session list.

## Summary
- Reject daemon startup when the socket/PID root is outside the active profile data directory while the DB still points at that profile's default `attn.db`.
- Apply the same isolation guard to `attn daemon ensure` so recovery paths fail before spawning an unsafe daemon.
- Add focused config and daemonctl regression tests for blocked mixed isolation plus allowed isolated/same-profile layouts.
- Document the user-visible protection in `CHANGELOG.md`.

## Test plan
- [x] `go test ./internal/config ./internal/daemonctl ./cmd/attn`
- [x] `git diff --check`
- [ ] Manually verify a packaged app fallback/recovery flow still starts normally when it uses a proper profile or explicitly isolated DB.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ### Changed
 - **Worktree Provider Coverage**: Worktree providers now participate when attn creates a worktree from an existing branch, including preserving the source branch/ref context for user-owned tooling.
 
+### Fixed
+- **Daemon Isolation Guard**: attn now refuses to start a daemon with an alternate socket/PID root while still pointing at the active profile's normal session database, preventing auxiliary daemon runs from pruning live sessions they do not own.
+
 ---
 
 ## [2026-05-17]
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -268,7 +268,13 @@ func runDaemonCommand() {
 }
 
 func runDaemon() {
-	d := daemon.New(config.SocketPath())
+	socketPath := config.SocketPath()
+	if err := config.ValidateDaemonIsolation(socketPath); err != nil {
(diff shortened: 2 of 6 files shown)


figgyster (approved):
This looks safe to merge. The guard hits both direct daemon startup and `daemon ensure`, and the focused tests cover the dangerous split-brain shape plus the legitimate isolated/profile-local cases.

Local focused checks passed here too: `go test ./internal/config ./internal/daemonctl ./cmd/attn`. CI is green. No blockers from me.

figgyster (dismissed):
Codex is right here, and this one crosses the safety line.

`ValidateDaemonIsolation` treats any relative `ATTN_DB_PATH` as isolated because `filepath.Clean` doesn't resolve it against the process working directory. If the daemon starts from the profile data dir with `ATTN_DB_PATH=attn.db`, it is still using the profile's default store while the socket root can be elsewhere, so the split-brain session-pruning risk this PR is meant to block can still slip through.

I reproduced it with a focused config test: `go test ./internal/config -run 'TestFiggyRelativeDBPathInProfileDirIsUnsafe|TestValidateDaemonIsolation'` fails because `ValidateDaemonIsolation` accepts that mixed layout.

Required before merge: canonicalize/absolutize the DB path against the current working directory before comparing it with the profile default store, and add the relative-path regression test. Tiny path gremlin, ugly blast radius.

figgyster (approved):
This is clean now. The follow-up absolutizes the DB/socket/profile paths before comparison and pins the exact relative-`ATTN_DB_PATH=attn.db` regression that slipped through the first pass.

`make test` passed locally, and GitHub’s daemon/preflight checks are green. No blockers from me.

PR #216, opened 2026-05-21
Use Codex hooks for live state detection

victorarias:
## Summary
- add Codex PreToolUse state hook so tool execution marks sessions working
- disable Codex PTY-derived state detection and recovered PTY state promotion
- rename the remaining PTY heuristic detector to Copilot since Codex no longer uses it

## Tests
- go test ./cmd/attn ./internal/agent ./internal/hooks ./internal/daemon ./internal/pty

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ### Changed
 - **Worktree Provider Coverage**: Worktree providers now participate when attn creates a worktree from an existing branch, including preserving the source branch/ref context for user-owned tooling.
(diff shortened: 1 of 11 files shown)


figgyster (approved):
This looks safe to merge. Moving Codex live-state back onto explicit hooks is the right direction: PreToolUse marks work before the tool actually runs, and disabling the old PTY detector removes the noisy heuristic path instead of fighting it with more duct tape.

The focused daemon/agent/hooks/PTY tests pass locally, and the coverage hits the important contract: Codex PTY state is now ignored while hook/transcript paths stay authoritative.

No blockers from me.

figgyster (approved):
Still safe to merge. The follow-up only tightens the socket-path test isolation, and the Codex state contract stays clean: hooks are authoritative, PTY heuristics are out of the Codex path.

I ran `git diff --check` and `go test ./cmd/attn ./internal/agent ./internal/hooks ./internal/daemon ./internal/pty ./internal/client` locally. No blockers from me.

PR #222, opened 2026-05-23
fix(settings): show daemon-discovered GitHub hosts

victorarias:
## Intent / Why
Settings currently infers authenticated GitHub hosts from visible pull requests. That hides configured hosts when they have no loaded PRs, even though the daemon can already use them. During verification, interactive sessions were also found to inherit `NO_COLOR` from the process that launched attn, making shells and agents appear monochrome.

## Summary
- expose the daemon GitHub host registry through initial state and a refresh event, so host visibility follows configured access rather than current PR data
- consume daemon host state in both the Tauri and native Settings surfaces
- bump the websocket protocol to v64 across daemon, Tauri, and native clients, and regenerate generated bindings
- strip inherited `NO_COLOR` at interactive PTY launch so shells and agents can emit their normal colored output
- add daemon, frontend, native protocol, and PTY environment coverage

## Test plan
- [x] `go test ./...`
- [x] `make test-frontend`
- [x] `git diff --check`
- [x] `cd native-ui && cargo fmt --check && cargo test -p attn-protocol`
- [ ] `cd native-ui && cargo test --workspace` in CI or a native build environment with Apple Metal tooling and Zig 0.15.2 available; the local environment lacks `metal` and has Zig 0.16.0
- [ ] Install the updated app and confirm an authenticated host with no visible PRs appears in Settings.
- [ ] Start a new terminal/agent session from attn and confirm ANSI colors render normally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ### Changed
 - **Input Assistance Disabled**: attn no longer allows browser or operating-system autocorrection, auto-capitalization, or spellchecking to rewrite typed paths, branches, commands, prompts, comments, or searches.
 - **GitHub Host Visibility**: Settings now lists daemon-discovered authenticated GitHub hosts even when a host has no currently visible pull requests.
+- **Terminal Colors**: Interactive terminals no longer inherit `NO_COLOR` from the process that launched attn, so shells and coding agents can emit their normal colored output.
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ### Changed
 - **Input Assistance Disabled**: attn no longer allows browser or operating-system autocorrection, auto-capitalization, or spellchecking to rewrite typed paths, branches, commands, prompts, comments, or searches.
+- **GitHub Host Visibility**: Settings now lists daemon-discovered authenticated GitHub hosts even when a host has no currently visible pull requests.
 
 ---
 
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -121,6 +121,7 @@ function App() {
   const [daemonEndpoints, setDaemonEndpoints] = useState<DaemonEndpoint[]>([]);
   const [daemonPlugins, setDaemonPlugins] = useState<DaemonPlugin[]>([]);
   const [daemonPluginIssues, setDaemonPluginIssues] = useState<DaemonPluginIssue[]>([]);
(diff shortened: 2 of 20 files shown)


figgyster (dismissed):
This is close, but the protocol bump missed the native client constant. The daemon/frontend moved to v64 and the native adapter hard-errors on version mismatch, so current native UI builds will refuse the new daemon even though they may not care about `github_hosts` yet.

Required before merge:
- Update `native-ui/crates/attn-protocol/src/lib.rs` `PROTOCOL_VERSION` from `"63"` to `"64"`.
- If the native event enum needs to tolerate the new `github_hosts_updated` event explicitly, add that too; the version constant is the hard blocker either way.

Local checks otherwise passed here: `git diff --check`, `go test ./...`, `make test-frontend`, and `pnpm --dir app run build`.

figgyster (commented):
The native protocol miss is fixed on this head: v64 is synced in the Rust protocol crate, and the native client now parses/applies the daemon-owned GitHub host state instead of deriving hosts from PRs. The extra `NO_COLOR` cleanup is also a good tiny fix — terminals inheriting agent-runner manners is how color goes to die.

I’m leaving this as a comment rather than an approval because GitHub still has Native UI and Frontend CI running, and this PR touches both. Local checks I could run passed: `git diff --check`, `go test ./internal/daemon ./internal/protocol ./internal/pty`, focused frontend tests, and frontend typecheck. Cargo isn’t installed on this runner, so I’m leaning on the Native UI CI for the Rust side.

Once those pending jobs go green, no blockers from me.

figgyster (commented):
The native protocol gap from my first pass is fixed now: v64 is lined up across the daemon, Tauri client, and native protocol/client paths, and Settings now consumes daemon-owned host state instead of deriving it from whatever PRs happen to be loaded. That’s the right owner for this state.

I also checked the follow-up `NO_COLOR` fix; stripping it from spawned PTYs is a clean containment move and the regression test covers both shell and agent sessions.

Local checks I ran:
- `git diff --check`
- `go test ./internal/daemon ./internal/pty ./internal/protocol`
- `pnpm --dir app exec vitest run src/hooks/useDaemonSocket.test.tsx src/components/SettingsModal.test.tsx`
- `pnpm --dir app run build`

I couldn’t run the Rust/native protocol tests locally because this runner doesn’t have `cargo`; GitHub’s Native UI job is still in progress. No code blockers from me, but I’m leaving this as a comment rather than approval until that native CI lane finishes its business. Tiny ghost of a toolchain, clanking chains in the hallway.

figgyster (approved):
Approved now that the native lane is green on the current head.

I pushed two tiny cleanup commits to unblock CI: removing the now-unused native PR event payload plumbing, then matching `cargo fmt` on the import wrap. The actual product change still looks right: daemon-owned GitHub host visibility, protocol v64 synced across clients, and the PTY `NO_COLOR` containment covered by tests.

No blockers from me.

PR #224, opened 2026-05-25
feat(plugins): add external agent driver host

victorarias:
## Intent / Why

External coding-agent integrations should not require adding agent-specific lifecycle, command-line, or metadata behavior to the attn daemon. This introduces a generic plugin-driver host so installed plugins can provide agents through a stable daemon contract, while attn continues to own session state, terminal execution, and user-facing availability.

## Summary

- Add external agent-driver registration and lifecycle RPC: `driver.spawn`, reload-only `driver.resume`, sequenced `session.report_*`, and `driver.session_closed`.
- Run plugin-returned commands through attn-owned embedded/worker PTYs and expose registered external agents dynamically through settings, local pickers, and remote capabilities.
- Persist opaque agent metadata and active-run ownership so only the spawning plugin may report state or receive close callbacks for its run.
- Fall back to `driver.spawn` when relaunching through a driver that does not advertise `resume`, and retain active plugin runs when PTY termination fails.
- Let Settings install a plugin from either a local directory or a pasted Git repository URL; Git sources are cloned temporarily, validated, dependency-installed, and published through the existing managed plugin installer.
- Remove unshipped in-tree external-agent placeholder plumbing in favor of the generic external-driver path.
- Update the plugin-system plan and changelog to document the implemented generic surface, ownership guarantees, and verification harness.

## Cleanup Included

- Removed unshipped driver request fields with no supported product flow: `agent_args`, native resume-picker/resume-ID forwarding, and ignored `display_name` registration.
- Removed unsequenced metadata mutation and unused report-cursor read helpers; metadata now changes only through ordered reports.
- Limited executable settings to attn-owned built-in agents; external driver plugins own their returned command lines.

## Test Plan

- [x] `make generate-types`
- [x] `go test ./internal/store ./internal/protocol ./internal/plugins`
- [x] `go test ./internal/daemon -run 'Test(HandleSpawnSession_PluginDriver|PluginDriverReports_|PluginDriverSessionClosed_|PluginDriverEndToEnd_)' -count=1 -timeout=90s`
- [x] `go test ./internal/daemon -count=1 -timeout=120s`
- [x] `go test ./internal/daemon -run '^TestDoCreateWorktree_ProviderHandledRegistersValidatedWorktree$' -count=1 -timeout=90s`
- [x] `go test ./internal/daemon -run 'Test(CodexResumeMappingEndToEnd|PluginDriverEndToEnd_InstalledProcessLaunchReportAndResumeThroughWorkerPTY)$' -count=10 -timeout=120s`
- [x] `make test-frontend` (439 passing tests; existing React act warnings remain)
- [x] `pnpm --dir app run build` (existing install-channel and chunk-size warnings remain)

`go test ./... -count=1 -timeout=120s` timed out in `TestDoCreateWorktree_ProviderHandledRegistersValidatedWorktree` when packages ran concurrently; that test passes alone and the full `internal/daemon` package passes.

## Out of Scope

- Any concrete external-agent plugin implementation, agent-specific extension bridge, or agent-specific installation/documentation.
- Plugin update or tagged/package distribution; Settings installs a repository's current default branch, while local checkout install remains available for development.
- Native UI client protocol-version enforcement removal; `native-ui` is intentionally not synchronized here because that client-side check will be removed separately.


figgyster (dismissed):
I found two contract-level problems that need fixing before this lands.

1. The protocol bump stops at Go + Tauri, but native-ui is still on protocol `64` while the daemon is now `66`. That client hard-checks the version, so this PR would make native-ui reject the updated daemon handshake. Please sync `native-ui/crates/attn-protocol/src/lib.rs` with `internal/protocol/constants.go`.

2. External driver run ownership is tied to the *current* registered driver for `session.Agent`, not the plugin/run that actually spawned the session. If plugin A starts a session, disconnects, and plugin B later registers the same agent id, `session.report_*` authorization and `driver.session_closed` routing can treat B as the owner of A's in-flight run. The run cursor stores `run_id`/seq, but not the spawning plugin identity. That crosses the lifecycle contract line, not a taste line.

Required before merge:
- Persist the driver owner identity with the active run/cursor and require `(session_id, run_id, plugin_name)` ownership for reports and close routing.
- Route `driver.session_closed` to the recorded owner plugin when connected, or drop/log if that owner is gone; don't re-resolve ownership from the current agent registration.
- Sync the native-ui protocol version for the protocol bump.

The shape is otherwise solid: sequenced reports, metadata preservation, and the attn-owned PTY execution path are the right direction. These two are just the gremlins standing in front of the merge button.

figgyster (dismissed):
The follow-up commit moves the plugin install UI forward, but the merge blockers are still present on this head.

Required before merge:
- Sync the native-ui protocol version. The daemon is now `67`, but `native-ui/crates/attn-protocol/src/lib.rs` is still `64`, so that client will reject the handshake.
- Persist/authorize plugin-driven runs by the spawning plugin identity, not by re-resolving the current registry entry for `session.Agent`. `authorizePluginSessionReport` still checks the live agent registration, so a disconnect/re-register race can transfer ownership of an in-flight run.
- Capability-gate resume. `handleSpawnPty` still passes `existingSession != nil` into `resolvePluginDriverLaunch`, so a driver without `resume` can fail relaunch instead of falling back to `driver.spawn`.
- Move plugin session close/end after a successful PTY kill, except for the already-gone case. Right now `handleKillSession` calls `closePluginDriverSession` before `ptyBackend.Kill`, which can invalidate a live run if kill fails.

This is all lifecycle contract territory, not taste. The shape is good; the gremlins are just camping at the sharp edges, as they do.

figgyster (approved):
This fixes the driver lifecycle blockers. Run ownership is now tied to the recorded plugin/run, non-resume relaunch falls back to `driver.spawn`, and close notifications wait until kill succeeds or the PTY is already gone.

Focused local checks passed for the daemon/store/protocol/plugin packages, and the Daemon CI job is green. Frontend/native CI was still running when I checked, so let that finish doing its thing before merge. No blockers from me.

figgyster (approved):
This follow-up is safe to merge. The new head is docs/test cleanup around the external-driver story, and the actual lifecycle contracts I cared about are still intact: persisted plugin/run ownership, spawn fallback when `resume` is unavailable, and close notifications after PTY termination is real.

I reran the focused daemon/store/protocol/plugin checks locally; no blockers from me. CI is still finishing, so let that finish doing its little green-check ritual.

PR #227, opened 2026-05-26
feat(native): build Swift workspace client with Ghostty

victorarias:
## Intent / Why

Replace the discarded canvas/GPUI direction with a macOS-native Swift client whose foundation is a real Ghostty surface renderer and the existing daemon-owned workspace model. The new client needs low-overhead terminal rendering, workspace-first navigation, and agent-driven automation before more product surfaces are added.

## Summary

- Archives the GPUI/native canvas prototype and documents the Swift workspace client, launcher model, and Tauri-parity decisions.
- Adds the SwiftUI/AppKit native client with Ghostty surfaces, sidebar/workspace pane layout, split navigation/close behavior, launcher dialogs, terminal input/clipboard/mouse/scroll handling, and single-client lifecycle behavior.
- Extends daemon/protocol workspace layout and session bootstrap mechanics needed by a dumb native client, including current-directory-aware pane launching and persisted launcher preference.
- Adds background-safe native automation and isolated terminal scenarios so the client can be exercised without stealing the active desktop.
- Pins the native build to the attn Ghostty fork commit `853aa8cb89d1bb2bda17c867c0a0c88b5b286994`, which exposes the embedded surface behavior required by the client.
- Adds a macOS Native UI CI job running Swift unit tests plus non-interactive automation/terminal harness scenarios, and fixes a worktree plugin test mock so daemon health checks cannot deadlock the Go gate.

## Test Plan

- [x] `test -z "$(gofmt -l .)" && go vet ./... && go test ./...`
- [x] `CI=true pnpm install --frozen-lockfile && CI=true pnpm run build && CI=true pnpm test && go build -o /tmp/attn ../cmd/attn && CI=true ATTN_E2E_BIN=/tmp/attn pnpm run e2e` from `app/` (`438` unit tests, `70` E2E passed; `6` expected real-PTY skips)
- [x] Tauri CI-equivalent sidecar build plus `cargo fmt -- --check && cargo test`
- [x] `make test-native-ci` (`64` Swift tests, background automation scenario, isolated-daemon Ghostty terminal scenario)

## Out of Scope

- Settings, diff, markdown, and todo panes remain future native surfaces.
- Physical keyboard/locally installed agent assertions remain opt-in local harness coverage rather than CI requirements.

## Related

- Ghostty fork support commit: https://github.com/victorarias/ghostty/commit/853aa8cb89d1bb2bda17c867c0a0c88b5b286994


figgyster (dismissed):
I found two merge blockers in the new native path. Both are contract/build issues, not taste — the usual plastic crown stays in the drawer.

Required before merge:

- `bootstrap_workspace` is endpoint-routed but does not reserve a pending PTY route for its `initial_session.id`. `spawn_session` does that before forwarding, but bootstrap skips it, so an immediate native attach/input after the bootstrap result can hit the hub before the endpoint session/layout snapshot races back and fail routing for the first pane. The fix should mirror the pending-route reservation used by `spawn_session` for the bootstrapped initial session.

- The Native UI CI job runs the Node harness without installing the app dependencies. `make test-native-ci` eventually runs `node app/scripts/real-app-harness/scenario-swift-native-terminal.mjs`, and that imports `ws`; the job only does `setup-node`, not `pnpm install`/equivalent for `app/`, so a clean runner can fail before the harness gets to the useful part.

I did run the focused Go side locally: `go test ./internal/protocol ./internal/daemon ./internal/store ./internal/workspacelayout ./cmd/attn` passes. I couldn't run the Swift side on this Linux reviewer host, so the build/CI contract matters extra here.

figgyster (dismissed):
The latest push does not address the two blockers I raised, so I’m carrying the request-changes state forward to the current head rather than leaving stale red on the old commit like a tiny haunted traffic light.

Required before merge:

- `bootstrap_workspace` is still endpoint-routed without reserving the pending PTY route for `initial_session.id`. `spawn_session` does this in `tryHandleRemoteWSCommand` before forwarding to the endpoint, but bootstrap falls straight through; a native client can get the bootstrap result and immediately fail attach/input before the endpoint state has raced back to the hub. Mirror the `spawn_session` pending-route reservation for the bootstrapped session.

- The Native UI CI job still runs `make test-native-ci` after only setting up Node. The terminal harness imports app dependencies like `ws`, but the workflow never installs `app/` dependencies on the clean macOS runner. Add the app dependency install (or remove the external import path) before the native harness runs.

Focused local checks still pass here: `git diff --check` and `go test ./internal/protocol ./internal/daemon ./internal/store ./internal/workspacelayout ./cmd/attn`. I can’t run the Swift/macOS harness on this Linux box, which makes the workflow fix non-optional rather than decorative.

figgyster (dismissed):
The latest push fixed neither gate I can safely sign off on, and it added a CI env tweak on top. Carrying the red forward to the current head so GitHub doesn’t make everyone read tea leaves in the review timeline.

Required before merge:
- `bootstrap_workspace` is still endpoint-routed without reserving the pending PTY route for `initial_session.id`. `spawn_session` reserves it before forwarding; bootstrap still forwards straight through. A native client can get a successful bootstrap result and immediately lose attach/input routing before the endpoint snapshot catches up.
- The `native-ui` workflow still runs `make test-native-ci` after only setting up Node, with no `app/` dependency install. The harness scripts import packages from `app` (`ws` is the obvious one), so a clean macOS runner can fail before it reaches the Swift/Ghostty evidence. Codex called the same spot, and it’s real.

The screenshot-denial allowance may be fine, but it doesn’t address those two contract/build blockers.

figgyster (dismissed):
The native CI harness dependency issue is fixed on this head, but the endpoint bootstrap routing blocker is still open.

Required before merge:
- `bootstrap_workspace` with an `endpoint_id` still forwards through the endpoint-command path without reserving the initial session's pending PTY route. `spawn_session` does this before forwarding, but `bootstrap_workspace` creates the first session too. That means the native client can get a successful bootstrap result and then have `attach_session`/input fall back locally before the remote session registration arrives. Same contract, sneakier coat.

Once that reservation is added for the bootstrap initial session, I expect this is clear.

figgyster (dismissed):
The newest push only adjusts the terminal harness cwd reporting. That’s useful test plumbing, but it doesn’t touch the remaining routing blocker, so I’m carrying the request-changes state forward to the current head.

Required before merge:
- `bootstrap_workspace` with an `endpoint_id` still forwards to the endpoint without reserving the initial session’s pending PTY route. `spawn_session` reserves `typed.ID` before forwarding; bootstrap creates the first session too, via `InitialSession.ID`, and needs the same reservation so immediate `attach_session`/input routes to the endpoint instead of racing local fallback.

The CI dependency blocker is fixed, and the focused checks I can run here still pass: `git diff --check` plus `go test ./internal/protocol ./internal/daemon ./internal/store ./internal/workspacelayout ./cmd/attn`. I can’t run the Swift/macOS harness from this Linux box, so the remaining red is specifically the endpoint bootstrap contract — not the harness cleanup.

figgyster (commented):
The endpoint bootstrap blocker is fixed on this head. The new reservation mirrors the `spawn_session` routing contract for `InitialSession.ID`, and the focused daemon test covers the race I was worried about.

I’m switching this from blocker to CI-gated: local checks here passed (`git diff --check` and `go test ./internal/protocol ./internal/daemon ./internal/store ./internal/workspacelayout ./cmd/attn`), but GitHub’s Frontend and Native UI jobs were still running when I checked. Once those go green, I don’t have remaining blockers.

figgyster (commented):
The bootstrap routing blocker is still fixed, and the latest harness change is the right kind of boring: wait for the new Ghostty surface to have a live grid before typing into it. That closes the little race where the test could beat the renderer to the punch.

I’m not approving this head until the Native UI job finishes, since that’s the principal gate for this PR and I can’t run the macOS Swift/Ghostty harness from this Linux runner. Everything else visible is green.

PR #238, opened 2026-05-30
[codex] Handle empty git repo branches

victorarias:
## Summary
- Resolve the current branch through symbolic-ref before falling back to detached HEAD SHA lookup.
- Add empty-repository coverage for branch metadata and branch listing.
- Document the user-visible fix in the changelog.

## Root Cause
The repo-info path used `git rev-parse --abbrev-ref HEAD`, which fails in a newly initialized repository before the first commit even though the unborn branch name is available from `git symbolic-ref --short HEAD`.

## Validation
- `go test ./internal/git`
- `go test ./internal/daemon`


figgyster (approved):
This looks safe to merge. The branch-name path now asks Git for the symbolic ref first, which is exactly the unborn-branch case a fresh repo needs, and detached HEAD still falls back to the short SHA.

I ran `git diff --check`, `go test ./internal/git`, and `go test ./internal/daemon` locally. No blockers from me.

PR #239, opened 2026-05-30
Add confirmed force delete for worktrees

victorarias:
## Summary
- add a force-delete retry path when normal worktree deletion fails for dirty local changes or other forceable local failures
- keep plugin-owned deletion authoritative by invoking the plugin before attn session/store cleanup, and pass the new force flag through the plugin SDK
- extend the websocket protocol/results so the UI can show why normal delete failed and retry with force after explicit y/n confirmation

## Validation
- go test ./...
- make test-frontend
- pnpm --dir app build
- pnpm --dir sdk/plugin test
- git diff --check
- regenerated protocol types and verified generation stability
- dev daemon/app smoke check after pruning a stale dev recovery session

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ### Fixed
 - **Codex Resize Redraws**: Attn now coalesces Codex's synchronized terminal redraws during pane resizes, preventing old scrollback from visibly streaming through the embedded terminal when layouts change.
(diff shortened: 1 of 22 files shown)


figgyster (approved):
This looks safe to merge. The important bit is that normal deletion now fails without tearing down attn’s session/store state, and the destructive retry is explicit all the way through the daemon, websocket result, UI prompt, and plugin SDK.

Local checks passed here: `go test ./...`, full `pnpm --dir app test -- --run`, focused delete/worktree daemon tests, and `git diff --check`. GitHub’s Frontend check was still pending when I posted this, but the local frontend suite covered the touched path.

No blockers from me.

figgyster (approved):
Re-approving the current head. The only change since my last pass is the plan/status doc update, and the implementation diff is unchanged.

`git diff --check` is clean; GitHub’s Frontend job was still pending when I checked, with the daemon jobs green. No blockers from me.

PR #240, opened 2026-05-30
[codex] Fix Ghostty Shift-Tab input

victorarias:
## Summary

- Send bare Shift-Tab from embedded Ghostty terminal panes as reverse-tab input (`ESC [ Z`) directly to the PTY.
- Cover WebKit's legacy `ISO_Left_Tab` key name and keep plain Tab delegated to Ghostty.
- Document the user-visible terminal fix in the changelog.

## Root Cause

Shift-Tab is also a browser/WebView focus traversal key. In the embedded terminal path, that meant the event could be handled before Ghostty reliably encoded terminal reverse-tab input, causing agents to observe prompt submission behavior instead of the intended Shift-Tab binding.

## Validation

- `pnpm --dir app exec vitest run src/components/SessionTerminalWorkspace/terminalKeyHandler.test.ts`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ### Fixed
 - **Codex Resize Redraws**: Attn now coalesces Codex's synchronized terminal redraws during pane resizes, preventing old scrollback from visibly streaming through the embedded terminal when layouts change.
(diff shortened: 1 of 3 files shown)


figgyster (approved):
This looks safe to merge. The key handling is narrowly scoped to bare Shift-Tab, sends the standard reverse-tab escape directly to the PTY, and leaves plain Tab on Ghostty's normal path.

I ran the focused handler test, TypeScript check, and whitespace check locally. No blockers from me.

PR #241, opened 2026-05-30
[codex] Make session creation asynchronous

victorarias:
## Summary

- Move worktree-backed session creation out of the blocking location picker flow.
- Add a compact session creation progress surface that mirrors the slow worktree cleanup animation semantics.
- Guard overlapping same-endpoint worktree session creation so existing websocket result correlation cannot cross-wire jobs.
- Cover the async handoff, compact focus behavior, and test cleanup behavior.

## Validation

- `pnpm --dir app test -- App.worktreeCleanup.test.tsx SessionCreationProgress.test.tsx LocationPicker.test.tsx`
- `pnpm --dir app build`
- `$HOME/.codex/skills/autoreview/scripts/autoreview --mode local` -> clean: no accepted/actionable findings reported

## Notes

The in-app Browser plugin was unavailable in this session (`iab` backend unavailable), so I could not complete a browser smoke. The focused React tests and production build pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-05-30]
 
 ### Fixed
+- **Session Creation**: Creating sessions from new worktrees now runs in the background with the same compact progress surface as worktree cleanup, so slow repository/plugin setup no longer traps the app in the picker.
 - **Codex Resize Redraws**: Attn now coalesces Codex's synchronized terminal redraws during pane resizes, preventing old scrollback from visibly streaming through the embedded terminal when layouts change.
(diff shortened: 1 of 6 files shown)


figgyster (approved):
This looks safe to merge. Moving worktree-backed creation out of the picker is the right shape: the picker closes immediately, the long work has a scoped progress surface, and same-endpoint worktree starts are guarded so websocket result correlation can't trip over itself.

I ran the focused frontend tests and production app build locally; both passed. GitHub's Frontend check was still pending when I reviewed, so let that finish, but no blockers from me.

figgyster (approved):
Re-approved the latest push. The follow-up fixes the compact progress focus issue cleanly: auto-collapse no longer steals focus, and the dialog only grabs focus when there’s an actual error to act on.

Focused frontend tests and `pnpm --dir app build` pass locally. GitHub’s Frontend check was still running when I looked, but this head is safe from my side.

PR #242, opened 2026-05-30
[codex] Fix terminal URL click hit testing

victorarias:
## What changed

- Hit-test Ghostty terminal pointer interactions against the rendered canvas instead of the outer terminal container.
- Treat pointers outside the rendered canvas as no cell instead of clamping them into the nearest row.
- Add e2e coverage for a vertically offset canvas so clicking the row above a URL does not open it.
- Stabilize the mouse-tracking URL e2e fixture with a delimiter after the URL.

## Why

The URL click debug log showed a vertical offset between the terminal container and canvas. The opener used the container-derived row, which could resolve to a URL one row below the pointer while the canvas-derived row had no URL.

## Validation

- `pnpm --dir app run build`
- `go build -o ./attn ./cmd/attn`
- `pnpm --dir app exec playwright test e2e/terminal-interactions.spec.ts --grep "opens a visible URL|vertically offset"`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ### Fixed
 - **Session Creation**: Creating sessions from new worktrees now runs in the background with the same compact progress surface as worktree cleanup, so slow repository/plugin setup no longer traps the app in the picker.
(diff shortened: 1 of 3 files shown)


figgyster (approved):
This looks safe to merge. The fix moves URL hit-testing to the rendered Ghostty canvas instead of the container, which matches the actual cell geometry, and the new e2e catches the nasty “row above the link still opens it” case.

I ran `git diff --check`, `go build -o ./attn ./cmd/attn`, `pnpm --dir app run build`, and the focused Playwright URL-click tests locally. No blockers from me.

PR #244, opened 2026-05-31
Auto-close sessions on clean process exit

victorarias:
## What

When a session's process exits **cleanly** (exit code 0, no signal), the session now closes itself instead of leaving a dead `[Process exited with code 0]` pane that you can only manually close. Sessions that **crash or are killed** (non-zero exit / signal) stay open so the error and exit code remain visible.

Closes the annoyance where quitting a shell/agent left a useless lingering pane.

## How

- `useDaemonSocket` gains an `onSessionExited` callback fired on the daemon's `session_exited` event (`{ id, exitCode, signal }`).
- `App.tsx` bridges that into `AppContent`'s existing pane-aware close path (`handleRequestCloseSession`) via a small ref — so **worktree cleanup, focus fallback, and utility-pane handling are reused**, not reimplemented. Closing the exited pane collapses to removing the session/workspace when it's the last pane.
- Gate is `exitCode === 0 && !signal`. Reloads/closes kill with SIGTERM (signaled exit), so they're naturally excluded and a reloading session is never yanked out from under its respawn.
- Uses the **daemon** event rather than the PTY-bridge `exit` because the mock PTY backend emits a code-0 `exit` on every kill, which would cause false closes in tests.

## Testing

- **Unit:** `useDaemonSocket.test.tsx` asserts the hook forwards both clean and signal exits. Full frontend suite + `tsc` green.
- **Real-app harness:** new `scenario-autoclose-on-exit` drives the packaged app against real shell PTYs — clean `exit` unregisters the session (daemon + sidebar), `exit 1` renders the banner and keeps the session. Registered in the serial matrix.
- **Manual (packaged dev app):** verified shell clean exit auto-closes, shell `exit 1` stays, and a **codex** agent clean quit also auto-closes.

Run the new scenario:
```
ATTN_HARNESS_PROFILE=dev pnpm --dir app run real-app:scenario-autoclose-on-exit
```

## Known boundary

Sessions that exit while no frontend is connected (e.g. across a daemon restart) come back as `idle` on reconnect with no live `session_exited`, so they aren't retro-closed. Out of scope here; can follow up if wanted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-05-31]
 
 ### Changed
+- **Auto-Close On Exit**: When a session's process exits cleanly (exit code 0), the session now closes itself instead of leaving a dead `[Process exited]` pane around. Sessions that crash or are killed (non-zero exit) stay open so you can still read the error.
 - **Workspace Muting**: Muting now applies to whole workspaces instead of individual sessions. Muted workspaces move to the muted section together with all of their sessions, and session rows no longer expose mute controls.
 
 ### Removed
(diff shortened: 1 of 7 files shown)


figgyster (approved):
This looks safe to merge. The clean-exit path reuses the existing pane close machinery instead of inventing a second cleanup path, and the signal/non-zero split preserves the “crash stays visible” contract.

I ran `git diff --check`, `pnpm --dir app test -- useDaemonSocket.test.tsx`, and `pnpm --dir app exec tsc --noEmit` locally. GitHub Frontend is still pending, but no blockers from me.

PR #246, opened 2026-05-31
release: v0.10.0

victorarias:
Release v0.10.0.\n\nBumps app package, Tauri config, Cargo manifest, and Cargo lock versions from 0.9.4 to 0.10.0.\n\nValidation already run by scripts/release.sh before this PR:\n- go test ./...\n- pnpm --dir app run build\n- pnpm --dir app test

### DIFF
--- a/app/e2e/terminal-interactions.spec.ts
+++ b/app/e2e/terminal-interactions.spec.ts
@@ -350,6 +350,7 @@ test.describe('Ghostty terminal interactions', () => {
   test('keeps copied selection attached to its text while scrolling', async ({ page, context, daemon }) => {
     await context.grantPermissions(['clipboard-read', 'clipboard-write']);
     const terminal = await openTerminalSession(page, daemon, 's-selection-scroll');
+    await page.evaluate(() => navigator.clipboard.writeText(''));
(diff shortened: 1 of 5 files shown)


figgyster (commented):
Version bump looks consistent across the app package, Tauri config, Cargo manifest, and lockfile. I also checked the diff for whitespace issues and verified all touched version fields resolve to `0.10.0`.

No code blocker from me. I’m leaving this as a comment rather than approval only because the Frontend check is still pending; if that goes green, this release PR looks mergeable.

figgyster (commented):
The release metadata is consistent at `0.10.0`, and the follow-up terminal-selection E2E tweak is the right kind of boring: clear clipboard first, then use page-level coordinates so the selection drag lands where the rendered terminal actually is.

Local checks passed:
- version consistency across `app/package.json`, Tauri config, Cargo manifest, and Cargo lock
- `go build -o ./attn ./cmd/attn`
- `pnpm --dir app exec playwright test app/e2e/terminal-interactions.spec.ts -g "keeps copied selection attached to its text while scrolling"`

No code blockers from me. I’m leaving this as a comment rather than approval because the Frontend and Tauri CI checks are still running; if those go green, this looks mergeable.

PR #248, opened 2026-06-01
fix(workspaces): prune stale empty workspace rows

victorarias:
## Intent / Why

Startup recovery could reap sessions without removing their workspace associations. That left stale empty workspace rows persisted in SQLite, causing old workspaces such as `attn--feat-snipe` to reappear in the sidebar after restart.

## Summary

- route startup and worker-backend session reaping through workspace association and layout cleanup
- remove already-persisted orphan workspace rows while preserving freshly registered empty workspaces that are still waiting for their first spawn
- clean workspace associations when deleting a worktree
- document that `make dev` is preferred during iteration, while verified fixes should be installed for the user with production `make`
- add daemon regression coverage and a user-facing changelog entry

## Test plan

- [x] `go test ./internal/store`
- [x] `git diff --check origin/main...HEAD`
- [x] `make dev`
- [x] `make`
- [x] Confirm production restart removes the persisted `attn--feat-snipe` workspace row from `~/.attn/attn.db`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,14 +25,14 @@ Use `make install` as the default source/dev install path; it rebuilds and insta
 
 ### Iterating On Attn Itself (Attn-On-Attn Testing)
 
-When changing attn code while the user has a live attn install running, **never** run `make install` or `make install-daemon` — they overwrite the live `attn.app` / restart its daemon mid-session. Use the dev sibling install instead:
+While iterating on attn code, prefer the dev sibling install so rebuilds do not repeatedly interrupt the live production app:
 
 ```bash
 make dev              # builds + installs ~/Applications/attn-dev.app, starts dev daemon on port 29849
 make install-daemon-dev  # faster: only rebuild the Go sidecar inside attn-dev.app
 ```
 
-The dev install is fully isolated: its own bundle identifier (`com.attn.manager.dev`), its own data dir (`~/.attn-dev/`), its own socket, its own port. Prod is never touched. `make install` and `make install-daemon` refuse at parse time if `ATTN_PROFILE` is set in the environment — if you hit that error, you meant `make dev`.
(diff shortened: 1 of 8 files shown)


figgyster (dismissed):
This looks safe to merge. Startup/session reaping now cleans the workspace association and layout paths too, and the orphan-prune keeps freshly registered empty workspaces out of the blast radius.

Local checks passed: `git diff --check origin/main...HEAD` and `go test ./internal/daemon ./internal/store`. CI is green where it ran.

figgyster (dismissed):
Codex caught a real recovery hole here, and my earlier approval was too optimistic.

`loadWorkspacesFromStore` runs with a fresh in-memory workspace registry on daemon restart. So a workspace persisted by `register_workspace` / `workspace_layout_add_session_pane` but not yet associated with a spawned session still has zero `SessionsInWorkspace`, is not in `d.workspaces.snapshot`, and gets deleted at startup. Since `RemoveWorkspace` also drops the layout, the queued `spawn_session` can come back to `unknown workspace` after restart — exactly the pending-workspace case this PR says it preserves.

Required before merge:
- preserve pending empty workspaces across restart, or add a persisted marker/age/other distinction that separates stale historical orphans from newly registered launch-window workspaces
- add a regression test that seeds the store like a real restart: workspace + layout, no session row, fresh registry, then `loadWorkspacesFromStore`

Local checks still pass (`git diff --check origin/main...HEAD`, `go test ./internal/daemon ./internal/store`), but they don’t cover this restart shape yet. Tiny timing window, nasty user-facing failure — the usual gremlin with a pocket watch.

figgyster (approved):
This fixes the restart hole Codex flagged: stale zero-session workspaces get pruned, but the launch-window workspace with a persisted `spawning` pane survives recovery instead of getting swept into the trash.

Local checks passed: `git diff --check origin/main...HEAD` and `go test ./internal/daemon ./internal/store`. No blockers from me.

PR #251, opened 2026-06-01
release: v0.10.2

victorarias:
## Intent / Why

Publish the `Cmd+N` worktree workspace fix from #250 as patch release `v0.10.2`.

## Summary

- Bump app and Tauri package versions from `0.10.1` to `0.10.2`.
- Refresh the Rust lockfile package version.

## Related

- #250

## Test plan

- [x] `go test ./...`
- [x] `pnpm --dir app test`

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,7 +1,7 @@
 {
   "name": "app",
   "private": true,
-  "version": "0.10.1",
+  "version": "0.10.2",
   "type": "module",
   "scripts": {
     "dev": "tauri dev",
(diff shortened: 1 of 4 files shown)


figgyster (commented):
This is the narrow release bump it claims to be: the only changed files are the four app/Tauri version surfaces, and they all line up on `0.10.2`.

I ran `git diff --check` plus the release-version consistency check locally. `Changes` is green; `Frontend` and `Tauri` were still running after a short poll, so treat this as ready once those finish cleanly rather than me pretending the rain has stopped.

figgyster (approved):
CI caught up cleanly, and the release metadata is still exactly the narrow `0.10.2` bump across the expected app/Tauri surfaces.

No blockers from me.

PR #253, opened 2026-06-01
fix(worktrees): offer force delete for provider failures

victorarias:
## Intent / Why

Provider-managed worktree cleanup suppressed the force-delete option when Git rejected deletion because the worktree contained modified or untracked files. The daemon wrapped every provider failure as a generic provider error, so the existing cleanup prompt could only offer retry.

## Summary

- Classify provider delete failures as dirty-worktree failures when they contain Git's dirty-worktree signature and the worktree actually has local changes.
- Preserve retry-only behavior for unrelated provider failures.
- Add a daemon regression for dirty provider-managed worktrees.
- Document the user-visible cleanup fix in `CHANGELOG.md`.

## Test plan

- [x] `go test ./internal/daemon -run 'TestDoDeleteWorktree_(ProviderDirtyWorktreeErrorIsForceable|ProviderErrorPreservesDaemonState|ProviderDeclineFallsBackToBuiltInGit|NormalDeleteFailurePreservesAttnState|ForceDeleteCleansUpAfterGitDelete)' -count=1`\n- [x] `go test ./internal/daemon -count=1`\n- [x] `pnpm --dir app test -- WorktreeCleanupPrompt.test.tsx App.worktreeCleanup.test.tsx useDaemonSocket.test.tsx`\n- [x] `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ The app shows a one-time "What's new" summary after this upgrade, and `⌘/` bri
 
 ### Fixed
 - **Workspace Sessions**: Creating a worktree from the `⌘N` new-session picker now adds the session to the current workspace instead of creating a separate workspace.
+- **Worktree Cleanup**: Provider-managed worktrees with local changes now offer force delete after the normal delete attempt is rejected.
 
 ---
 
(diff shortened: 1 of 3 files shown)


figgyster (approved):
This looks safe to merge. The provider path now preserves the same dirty-worktree contract as the built-in Git path: failed normal delete keeps daemon state intact, reports `dirty_worktree`, and only makes force delete available when Git's dirty signature matches real local changes.

Local daemon tests passed here, and the visible GitHub checks are green. No blockers from me.

PR #254, opened 2026-06-01
Make attn resolve correctly when a stale binary shadows it

victorarias:
## Why

While using the attn skill from Claude Code, `attn open` failed with `unknown command: open`. Root cause: a **stale `attn` v0.5.0 (Apr 12)** at `~/.local/bin/attn` sat ahead of the app on `PATH` and shadowed the current v0.10.2 wrapper. Every command hit the old binary, which predates `open`/`presence`. The errors were misleading because (a) the old binary's `unknown command` error looks identical to a real one, and (b) `attn` forwards unrecognized flags to the agent, so `attn --help` printed *Claude's* help, hiding the fact that the wrong binary was running.

This PR makes attn (and its skill) resilient to that failure mode.

## Changes

**Skill — binary resolution + collision note** (`internal/agent/claude_skill.go`)
- Tell agents to prefer `$ATTN_WRAPPER_PATH` (always set inside a session, points at the authoritative binary), falling back to `attn` on PATH only when unset.
- Add a troubleshooting note: if a command says `unknown command: ...` or prints Claude's help, a stale/unrelated `attn` is probably shadowing the real one — confirm with `attn --version` / `which -a attn`.

**CLI — remove the implicit agent passthrough** (`cmd/attn`, `internal/agent`)
- `attn` now understands only `-s`, `--resume`, `--yolo`. Any other flag, the `--` separator, or a stray positional is rejected with attn's own usage. This passthrough was never used internally — the only launcher (`internal/pty/manager.go`) emits exactly those three flags.
- `attn --help`/`-h` now print attn's help; unknown-command errors include the attn version, so a stale binary is obvious at a glance.
- Removed the now-dead `SpawnOpts.AgentArgs` field and its three driver consumers (codex always passes `-C <cwd>` now).

**CLI — version-skew warning** (`cmd/attn/versioncheck.go`)
- Human commands (`presence`, `list`, `open`, `review-loop`) best-effort probe the daemon's existing `/health` endpoint and print one stderr line when the running CLI's version differs from the app's. No protocol change; dev/unknown builds are skipped; fully non-fatal.

## Verification
- `go build ./...`, `go vet ./...`, `gofmt` — clean.
- New/updated unit tests for `parseDirectLaunchArgs` (rejects unrecognized args), the skill content, and the version-skew decision. `go test ./cmd/attn/... ./internal/agent/...` passes.
- Manual smoke tests on the built binary: unknown command shows version + usage; `attn --help` shows attn help; unknown flag / `--` rejected; the nudge fires against the live daemon on a fake-version build and stays silent on a match.

## Notes
- Removing the passthrough only affects a human manually passing arbitrary agent flags through the `attn` wrapper from a terminal — confirmed unused.
- The stale `~/.local/bin/attn` itself was removed out-of-band (no local code path writes there; only the remote SSH hub bootstrap targets that path on *remote* hosts).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@ The app shows a one-time "What's new" summary after this upgrade, and `⌘/` bri
 
 ## [2026-06-01]
 
+### Changed
+- **`attn` is clearer about being its own command.** `attn --help`, unknown commands, and unknown flags now print attn's own usage (with its version) instead of quietly handing them to the coding agent. `attn` understands only `-s`, `--resume`, and `--yolo`; it no longer forwards other flags — or anything after `--` — through to the agent.
+- **Stale-`attn` warning.** When the `attn` on your `PATH` is a different version from the running app, attn prints a one-line warning so you can catch an out-of-date binary shadowing the current one (`which -a attn`).
+
 ### Fixed
 - **Workspace Sessions**: Creating a worktree from the `⌘N` new-session picker now adds the session to the current workspace instead of creating a separate workspace.
 - **Worktree Cleanup**: Provider-managed worktrees with local changes now offer force delete after the normal delete attempt is rejected.
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -139,7 +139,7 @@ func main() {
 	case "open":
 		maybePrintProfileBanner()
 		runOpen()
-	case "help":
+	case "help", "-h", "--help":
 		runHelp()
 	case "_hook-stop":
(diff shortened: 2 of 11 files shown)


figgyster (approved):
This looks safe to merge. The wrapper now owns its CLI surface instead of quietly moonlighting as an agent flag tunnel, and the stale-binary failure mode is called out in both runtime behavior and the bundled Claude skill.

Local check passed: `git diff --check` and `go test ./...`. GitHub daemon/preflight checks are green too. No blockers from me.

PR #257, opened 2026-06-02
Tile-only workspaces + rename docked "panel" leaf to "tile"

victorarias:
Two commits on this branch.

## 1 · Keep tile-only workspaces alive past the last terminal

Collapse workspace teardown from "last *terminal*" to "last *leaf*". A docked tile the user deliberately left behind now keeps its workspace alive — sessionless — instead of vanishing with the doc when the last terminal closes.

- **Daemon:** teardown keys off `workspacelayout.LayoutEmpty` (no panes *and* no tiles), not `len(panes)==0`. The three layout teardown sites (`ensureWorkspaceLayout`, `handleWorkspaceLayoutClosePane`, `removeWorkspaceLayoutPaneForSession`) and the startup reap (`loadWorkspacesFromStore`) switch to the last-leaf rule. `dissociateSessionFromWorkspace` keeps the workspace **entity** (broadcasting a neutral state-changed) when its last session leaves but a tile remains.
- **Frontend:** tile-only workspaces are hidden from the sidebar unless the user enables **"Tile-only workspaces"** in the display popover (localStorage-persisted); revealed, they render a neutral indicator instead of a session-state dot. Derived from `sessions.length === 0`.
- **Tests:** `workspace_tile_lifecycle_test.go` models the real app flow (register → add+spawn session pane → dock tile → close last pane → assert tile + workspace + layout survive), plus survival across a simulated startup reap. Sidebar tests cover reveal/neutral-indicator/persistence.

## 2 · Rename the docked "panel" leaf to "tile"

`pane` (terminal leaf) and `panel` (docked content leaf) were one letter apart and meaningless to tell apart — and `panel` also collided with the slide-in **overlay** panels (diff, review, changes). After this rename three concepts read distinctly:

| term | meaning |
|---|---|
| **pane** | terminal / session leaf |
| **tile** | docked content leaf (markdown today) |
| **panel** | slide-in right-dock overlay (unchanged) |

- **Protocol:** renamed commands/events/fields (`workspace_layout_dock_tile`, `workspace_layout_undock_tile`, `workspace_tile_content[_get]`, `tile_id`/`tile_kind`/`tile_params`); `generated.go`/`generated.ts` regenerated from the schema. **No version bump and no storage migration** — the docked-tile feature was never published, so there's no deployed daemon or persisted layout to stay compatible with.
- **Frontend:** `WorkspaceDockPanel.{tsx,css,test}` → `WorkspaceDockTile.*`; docked-tile props + CSS (`workspace-dock-tile*`, `workspace-pane--tile`) rename. The `RightDock` overlay system (`DockPanelId`, `.dock-panel--*`) intentionally stays "panel". Sidebar copy "Panel-only" → "Tile-only".
- **Docs:** decision doc renamed to `docs/decisions/2026-05-31-docked-tiles.md`; CHANGELOG + plan updated. Overlay panels, native-canvas panels, and the frozen `canvas_workspace_panels` migration are intentionally left as "panel".

## Verification

`go build ./...` + full `go test ./...` green; `tsc --noEmit` clean; all 523 frontend tests pass; production `vite build` succeeds.

## Deferred (cosmetic follow-up)

Tile title from content (markdown H1 → text → basename) and focus-on-select for fully tile-only workspaces. A tile-only workspace renders fine today.


figgyster (dismissed):
Codex is right on the functional gap here: the daemon-side last-leaf lifecycle looks solid, but the frontend reveal path still can’t actually select/render the preserved panel-only workspace.

`handleSelectWorkspace` only switches by `firstSessionId`, `activeWorkspaceId` is still derived from `activeSessionId`, and the terminal renderer skips any workspace whose session-derived terminal state is empty. So after enabling “Panel-only workspaces”, the row and shortcut can show up but there’s no session to activate and no rendered workspace wrapper for the docked panel. That undercuts the main user contract of this PR: keeping the panel alive is only useful if the user can get back to it.

Required before merge:
- make workspace selection capable of activating a sessionless workspace, not just a session id
- render panel-only workspace layouts when they have panels even if they have zero sessions
- add a frontend regression test for revealing/selecting a panel-only workspace and seeing its docked panel/layout

`git diff --check` is clean. I’m blocking on this behavior gap, not on the daemon lifecycle shape — that part has the right scent.

figgyster (dismissed):
The follow-up commit still leaves the core panel-only path stranded on the frontend side.

Daemon-side last-leaf retention is behaving like the PR says, and the focused daemon/workspacelayout tests pass. But `handleSelectWorkspace` still only selects `workspace.firstSessionId`, `activeWorkspaceId` is still derived from `activeSessionId`, and the terminal render loop still builds `workspaceState` only from `workspace.sessions` before returning `null` for zero-session workspaces. So when the user reveals a tile-only workspace, the row/shortcut can exist, but there is still no sessionless workspace activation/render path for the preserved docked tile.

Required before merge:
- let workspace selection activate a workspace id even when it has no sessions
- render daemon workspace layouts that contain tile leaves even when `sessions.length === 0`
- add the frontend regression that proves revealing/selecting a tile-only workspace shows the docked tile/layout

Validation I ran on this head:
- `git diff --check`
- `go test ./internal/workspacelayout ./internal/daemon`
- `pnpm --dir app exec vitest run Sidebar.test.tsx SessionTerminalWorkspace.test.tsx`

figgyster (dismissed):
The latest push still doesn't fix the tile-only workspace contract. The branch renamed the leaf terminology and adjusted protocol versioning, but the user-visible path is still session-driven.

Current head evidence:
- `handleSelectWorkspace` still only calls `handleSelectSession` when `workspace.firstSessionId` exists.
- `activeWorkspaceId` is still derived through `useWorkspaceSelectionController(workspaceViews, activeSessionId)`.
- the terminal render loop still builds state from `terminalStateForWorkspaceSessions(workspace.sessions)` and returns `null` for zero-session workspaces.

So a revealed tile-only workspace can appear in the sidebar, but selecting it still can't activate/render the preserved tile. That's the core promise of this PR, not a cosmetic edge case.

Required before merge:
- introduce workspace-id activation for sessionless workspaces
- render daemon layouts with tile leaves even when `sessions.length === 0`
- add an App/controller-level regression proving a revealed zero-session workspace can be selected and its docked tile/layout is visible

Validation on this head: `git diff --check` and a focused source check against `app/src/App.tsx`. I skipped heavier suites because the blocker is still present in the integration path — no need to make the CI hamster run laps for a known red light.

figgyster (approved):
This fixes the blocker. The frontend now has a real workspace-id activation path for sessionless workspaces, `SessionTerminalWorkspace` can render tile-only daemon layouts without a backing session, and the App-level regression covers the exact “revealed tile-only workspace can be selected and rendered” path that was missing before.

Validation I ran on `2f80af4`:
- `go test ./internal/daemon ./internal/workspacelayout ./internal/protocol`
- `pnpm --dir app install --frozen-lockfile`
- `pnpm --dir app exec vitest run App.sessionlessWorkspace.test.tsx components/SessionTerminalWorkspace/SessionTerminalWorkspace.tileOnly.test.tsx hooks/useWorkspaceSelectionController.test.ts Sidebar.test.tsx`

GitHub CI is green too. No blockers from me.

PR #259, opened 2026-06-03
Tile polish: title docked tiles from content + focus tile on select

victorarias:
Two tile-only workspace polish items from the uniform-leaf-docking plan (the genuinely-cosmetic follow-ups deferred from #257).

## What changed

- **Docked tile titles come from the document.** A markdown tile header now derives its title from the document's first meaningful line — a leading ATX heading marker is stripped (so `# Title` → "Title"), closed YAML frontmatter and lone thematic breaks (`---`) are skipped, and very long titles truncate at 80 chars. Falls back to the file basename, then the tile kind. `deriveTileTitle` is pure and unit-tested.
- **Selecting a tile-only workspace focuses the tile.** A fully tile-only workspace (no agent pane) has no terminal to own focus, so it now focuses the first tile's scrollable body on select — the arrow keys scroll it immediately, no extra click. The body is `tabIndex=-1` (programmatically focusable, not a tab stop). The terminal-focus path is untouched, and mixed pane+tile workspaces still keep a real pane focused.

## Testing

- 540 frontend tests pass (+9: 8 `deriveTileTitle` unit cases, 1 focus assertion in the tile-only render test). `tsc` clean.
- **Verified end-to-end in the packaged dev app.** The `tile-only-workspace-select` real-app scenario now asserts the header shows the markdown H1 (`tileTitles: ["Tile only notes"]`, not `notes.md`) and the tile body is focused after select (`tileBodyFocused: true`, and `false` before select). New `tileTitles` / `tileBodyFocused` fields on the `get_workspace_ui_state` automation verb (frontend-only, no protocol bump).

CHANGELOG (`[2026-06-03]`) and the plan doc (new "PR D" section) updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,11 +28,12 @@ The app shows a one-time "What's new" summary after this upgrade, and `⌘/` bri
 
 ## [2026-06-03]
 
+### Changed
+- **Docked tiles read a little nicer.** A markdown tile's header now shows the document's title (its first heading) rather than the file name, falling back to the file name when there's no heading. And when you open a tile-only workspace, the keyboard lands on the tile right away, so the arrow keys scroll it without an extra click.
+
 ### Fixed
 - **Approving a permission request clears the attention light right away.** When an agent is blocked on a permission prompt and you approve it, the session now returns to "working" within about a second — even while the approved command keeps running, and even for a second prompt back-to-back. It used to stay in the flashing "needs you" state for the entire duration of the approved tool, which defeated the point of the light. Works for both Claude and Codex.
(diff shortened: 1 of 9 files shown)


figgyster (approved):
This looks safe to merge. The tile-only workspace path now has the missing polish where it matters: document-derived headers and immediate keyboard-scroll focus when there’s no terminal to own focus.

I ran `git diff --check`, `pnpm --dir app test` (540 tests), and `pnpm --dir app exec tsc --noEmit` locally. GitHub Frontend CI was still running when I checked, but the local principal frontend suite is green. No blockers from me.

figgyster (approved):
This still looks safe to merge. The follow-up keeps the tile-only path tight: document-derived headers are covered at the component level, and the real-app automation now verifies both the H1 title and keyboard-scroll focus after selecting the workspace.

I ran `pnpm --dir app exec tsc --noEmit` plus the focused tile tests locally; no blockers from me.

PR #260, opened 2026-06-03
Move panes and tiles between workspaces

victorarias:
## Summary
- add backend/protocol support for moving layout leaves between workspaces
- wire sidebar hover/drop UX for moving panes and tiles into another workspace
- preserve layout snapshots across layoutless workspace state updates and add a real-app scenario for workspace moves

## Tests
- rtk pnpm --dir app test
- rtk go test ./internal/workspacelayout ./internal/protocol ./internal/store ./internal/daemon
- rtk git diff --check
- rtk make dev
- ATTN_HARNESS_PROFILE=dev rtk pnpm --dir app run real-app:scenario-workspace-move-leaf


figgyster (approved):
This looks safe to merge. Moving leaves between workspaces keeps the daemon layout as the authority, carries pane/session ownership across the store and workspace registry, and the sidebar drop path stays endpoint-scoped instead of guessing across daemons.

I ran the focused daemon/protocol/store/workspacelayout tests plus the workspace frontend checks locally; all green. GitHub isn't reporting checks for this branch yet, so no CI signal to fold in from here.

PR #261, opened 2026-06-03
fix(app): release pane WebGL context on unmount to prevent UI freeze

victorarias:
## Problem

Intermittent **complete UI freeze when opening a new pane** (agent or terminal). Usually cleared by quitting and reopening the app; occasionally needed two restarts. A regression in the workspaces/panes work.

Log evidence (from a live freeze):
- The **daemon is healthy** throughout — no errors, no slow-client drop, websocket client stays connected (`clients=1`), no timestamp gaps. The freeze is entirely in the Tauri/WKWebView main thread.
- The frontend hit-test log goes permanently silent the instant a brand-new pane is clicked/focused — the render thread dies on the new-pane path.
- The daemon shows repeated create→close→create pane churn around the freeze.

## Root cause

Each terminal pane creates a `WebGL2` context for its Ghostty renderer. Browsers cap the number of simultaneously-live WebGL contexts (WKWebView's cap is low), and attn keeps **every** workspace's panes mounted (`App.tsx`). Closed/remounted panes only released their context on non-deterministic GC, so opening/closing panes over a working session piled leaked contexts up until a new agent/terminal pushed past the cap. The engine then forcibly loses the **oldest** context — which `GhosttyTerminal` handles by `preventDefault()` + an error and **no `webglcontextrestored` handler**, so that pane is dead — manifesting as a frozen pane or a hard UI freeze. A restart releases all GPU contexts, which is why it "fixed itself."

This was confirmed in a real browser: holding 40 contexts (≈ mounted panes), Chromium silently caps live contexts at 16 and kills the oldest 24; with `loseContext()` on release, 0 accumulate.

## Fix

Release the pane's WebGL2 context with `WEBGL_lose_context.loseContext()` in a dedicated **unmount-only** effect, so it's reclaimed deterministically when the pane goes away instead of waiting for GC.

It deliberately does **not** go in the renderer's `dispose()`: `dispose()` also runs on `fontSize`/`theme` re-init, which reuses the same `<canvas>` — and a canvas keeps returning the *same, now-dead* context after it's lost, which would hand the rebuilt renderer a corpse. Releasing only on unmount avoids that.

## Scope / follow-up

This fixes the **churn-driven** leak (matches the observed symptom and log pattern). It does not address a hypothetical *many-simultaneous-live-panes* case (more live panes than the WKWebView cap at once) — the cure there is tearing down the renderer for hidden/inactive panes while keeping the terminal model alive, which touches the focus/replay path and deserves its own PR with packaged-app-harness verification.

## Testing

- `tsc --noEmit` clean
- 68 terminal/workspace unit tests pass
- Mechanism + fix were validated with a throwaway real-browser WebGL simulation (not committed — it exercised a replica, not this code path, so it would have been a false guard)

---
🤖 Drafted by Claude Code on behalf of Victor.


figgyster (approved):
This looks safe to merge. The fix is scoped to the actual lifetime boundary: keep renderer re-init alone, release the WebGL context only when the pane unmounts, and document why that distinction matters.

Local checks passed: `git diff --check`, `pnpm --dir app exec tsc --noEmit`, and the focused terminal key handler vitest suite. GitHub Frontend CI was still running when I checked, but this patch is narrow and the local frontend proof is enough for me.

PR #262, opened 2026-06-03
fix(workspaces): pane/tile header click no longer splits the layout

victorarias:
## Problem

A plain click (no drag) on a pane or docked-tile header — the title bar that appears when a workspace has splits — rearranged the layout, typically creating a full-width split at 50% of the screen.

## Root cause

The header press starts a drag gesture in `startLeafDrag` (`leafDrag.ts`), which had **no movement threshold**. On `pointerup` it computed a drop target *at the release position* and fired the move, even when the pointer never moved. A header sits at the top edge of its pane, so a click lands in the workspace's top perimeter band; with both split panes touching that edge, `computeDockTarget` returns a container dock (`anchorId: ''`, `edge: 'top'`, `ratio: 0.5`) — the full-width 50% split.

## Fix — drag activation threshold (hysteresis)

- The gesture stays **pending** until the pointer moves ≥4px from the press point. Below that it's a click: no activation, no preview, no drop.
- `onUp` resolves a drop only when the gesture became a drag (activated, or the release landed past the threshold — covers a fast drag whose `pointermove`s were coalesced).
- New `onActivate` handler: the visual side effects (selection-lock cursor, dragging styles, cross-workspace overlay) only fire once it's a real drag, so a click leaves no trace. Covers both pane headers and docked-tile headers (both route through `beginLeafDrag`).

A click now behaves like dropping at the same location — a self-drop no-op.

## Tests

- `leafDrag.test.ts`: plain click and sub-threshold jitter → no activation/preview/drop; activates exactly once on threshold crossing; fast coalesced drag still drops.
- `SessionTerminalWorkspace.test.tsx`: a real DOM `pointerDown`+`pointerUp` on a tile header (no move) doesn't lock selection or call `onMoveLeaf`; updated the existing lock test to move past the threshold before asserting the grab lock.
- `npx tsc --noEmit` clean; full frontend suite (547 tests) green.


figgyster (approved):
This looks safe to merge. The threshold sits at the right seam: clicks never activate the drag machinery, real drags still resolve, and the fast/coalesced pointerup case is covered instead of turning into another tiny UI ghost story.

Local checks passed: `git diff --check`, `pnpm --dir app exec tsc --noEmit`, and the focused `SessionTerminalWorkspace`/`leafDrag` vitest files. GitHub `Frontend` was still running when I posted; `Changes` passed and the daemon/Tauri jobs were skipped for this frontend-only patch.

PR #264, opened 2026-06-04
fix(terminal): keep text selection alive when it ends over a split divider

victorarias:
## The bug

Selecting text in a split terminal: if the selection drag reached the **divider between panes**, the cursor flipped to the resize arrow, the selection froze, and the text wasn't copied — it stayed stuck until you clicked again.

## Root cause

Terminal selection used **element-scoped React handlers** (`onMouseDown/Move/Up` on the terminal `<div>`). The split divider is an absolutely-positioned sibling at `z-index: 5` straddling the pane edge, so once a drag reached it, `mousemove`/`mouseup` retargeted to the divider. The terminal's `onMouseMove` stopped (selection froze) and its `onMouseUp` — where `selectingRef` is cleared and the clipboard copy happens — never fired.

## The fix

When a selection starts, the drag is now tracked on `document` (`mousemove`/`mouseup`) rather than the terminal element, so it keeps updating and finalizes regardless of which sibling overlay the pointer crosses — the same pattern the divider/leaf drags already use. Fully contained in `GhosttyTerminal`; the terminal mouse-tracking protocol path is untouched. Includes a `buttons`-released guard (finalize on a missed mouseup) and unmount cleanup.

## Validation

- `tsc` clean; **555 unit tests** pass.
- **11/11** terminal-interaction e2e tests pass (all existing copy/selection tests included — no regression).
- New regression test `copies a selection even when the mouse is released outside the terminal` was verified to **fail on the pre-fix code** and **pass with the fix**.

## Note / follow-up

Cosmetic and deliberately out of scope: while dragging *over* the divider's 9px band the cursor still briefly shows the resize arrow (the divider's own CSS cursor). Selection and copy now work correctly through it; suppressing the cursor too would mean a small body-class + CSS change, happy to do it in a follow-up if wanted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ The app shows a one-time "What's new" summary after this upgrade, and `⌘/` bri
 
 ---
 
+## [2026-06-04]
+
+### Fixed
+- **Selecting terminal text to the edge of a split pane no longer gets stuck.** When you dragged a selection up to the divider between split panes, the cursor flipped to the resize arrow and the selection froze — and the text wasn't copied until you clicked again. Selections now finish and copy normally even when you release the mouse right at a pane edge.
+
 ## [2026-06-03]
 
 ### Added
--- a/app/e2e/terminal-interactions.spec.ts
+++ b/app/e2e/terminal-interactions.spec.ts
@@ -347,6 +347,42 @@ test.describe('Ghostty terminal interactions', () => {
       .toEqual([]);
   });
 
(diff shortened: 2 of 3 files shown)


figgyster (approved):
This looks safe to merge. Moving selection tracking to `document` is the right ownership boundary for pane-edge drags, and the new outside-release regression test covers the failure mode cleanly.

I also verified locally: `git diff --check`, `pnpm --dir app exec tsc --noEmit`, and all 11 `terminal-interactions` Playwright tests pass after building the `attn` test binary. GitHub Frontend/Changes are green too. No blockers from me.

PR #270, opened 2026-06-07
feat(browser): add agent-controlled in-app browser

victorarias:
## Why

Agents need a persistent browser inside attn that users can log into and agents can control without requiring a separate Playwright or CDP process.

## Summary

- add native Tauri/WKWebView browser tiles with persistent cookies, address/reload/close controls, sidebar entries, and correct overlay/focus behavior
- add an authenticated, daemon-brokered WebDriver-shaped CLI/API with semantic locators, frames, shadow DOM, input actions, cookies, alerts, screenshots, and PDF capture
- preserve tile-only and remote-workspace routing, including selected-workspace targeting through protocol version 87
- adapt the relevant MIT-licensed `tauri-plugin-webdriver` implementation locally, with license and third-party notices, rather than adding it as a direct dependency
- document browser control in the attn skill and update the user-facing changelog

## Test plan

- [x] `pnpm --dir app test -- --run` (603 tests)
- [x] `cargo test --manifest-path app/src-tauri/Cargo.toml` (27 tests)
- [x] `cargo fmt --manifest-path app/src-tauri/Cargo.toml -- --check`
- [x] `go test ./internal/daemon ./internal/hub ./internal/protocol ./internal/store ./cmd/attn`
- [x] packaged dev-app manual tests: persistent cookies, address/reload/close controls, Cmd+W focus ownership, modal layering, frames, alerts, `target=_blank` POST handling, screenshots/PDF, sidebar entries, and tile-only workspaces


figgyster (dismissed):
I found two issues that cross the line from polish into contract/security territory.

Required before merge:

- Remote browser control needs to route to the owning endpoint. `open_browser` handles remote workspaces by forwarding through `target.remoteEndpointID`, but `browser_control` resolves the same remote target and then always talks to the local `d.browserHost()` (`internal/daemon/browser.go`). That means `attn browser ... --session <remote-session>` will either hit the wrong local browser host or fail with “no in-app browser host is connected.” This PR explicitly promises remote-workspace routing, so the control path needs the same endpoint forwarding coverage as open.

- Please harden/prove the Tauri IPC boundary for the embedded browser webviews. The privileged browser-host commands (`get_browser_host_token`, `browser_host_control`, cookie reads, navigation/control, screenshots/PDF, etc.) are registered globally, while the browser child webviews load arbitrary external pages. The capability file says only `main` is trusted, but I don’t see a command-level caller/origin check or a regression test showing an external browser page cannot invoke these custom commands. If Tauri’s capability layer already denies those child webviews, great — add the proof. If not, this is a cookie/control escape hatch wearing a fake mustache.

What passed locally here:
- `go test ./internal/daemon ./internal/protocol ./internal/hub ./cmd/attn`
- `pnpm --dir app test -- src/browser/runtime.test.ts src/browser/host.test.ts src/components/SessionTerminalWorkspace/BrowserTileBody.test.tsx src/hooks/useDaemonSocket.test.tsx`

I couldn’t run the Rust/Tauri suite in this runner because `cargo` is not installed.

figgyster (dismissed):
The two original blockers look addressed on this head: browser control is now endpoint-routed with response ownership checks, and the Tauri browser-host commands now require the trusted main webview caller. That’s the right shape.

One blocker remains before merge: Codex’s cookie-domain finding is real. `current_cookies()` filters domain cookies with `cookie_domain_matches()`, but that helper only treats a cookie as domain-scoped if the raw domain starts with `.`. If WKWebView/Tauri returns a domain cookie normalized as `example.com`, a browser currently on `sub.example.com` drops it as host-only, so `get_all_cookies`, `get_cookie`, and `delete_cookie` miss a valid WebDriver-visible cookie. Given this PR is adding browser automation/cookie APIs, that’s a user contract bug, not a garnish.

Required before merge:
- Handle normalized domain-scoped cookies for subdomains, or use the platform cookie filtering API if available.
- Add focused coverage for a cookie set as `Domain=.example.com` being visible from `sub.example.com` even when the returned cookie domain is normalized to `example.com`.

Local verification passed for the fix commit:
- `go test ./internal/daemon ./internal/protocol ./internal/hub ./cmd/attn`
- focused browser/socket frontend tests: 49 passed

Still couldn’t run the Rust/Tauri suite here because this runner has no `cargo`.

figgyster (approved):
This looks safe to merge now. The browser control path has the right endpoint ownership checks, the Tauri command boundary is explicit, and the normalized domain-cookie case is covered instead of guessed at.

No blockers from me. CI is green on the current head.

PR #271, opened 2026-06-07
Rebuild the diff viewer on @pierre/diffs

victorarias:
## Summary

Replaces attn's bespoke CodeMirror diff engine (`UnifiedDiffEditor`, ~1.9k lines) with a thin `DiffView` wrapper around **[@pierre/diffs](https://diffs.com)** (Shiki-based, Apache-2.0). Frontend-only — the daemon already returns `original`/`modified` file contents, which feed the library directly. No protocol bump.

What carries over from the old engine, ported onto the library's native primitives:
- inline review comments on added/deleted lines (native `lineAnnotations` slots), edit / resolve / delete / send-to-Claude, per-file comment counts
- viewed / changed-since-viewed tracking, file navigation, send-unresolved
- **unified ↔ split** toggle (unified default), Shiki highlighting + theming, native hunk collapsing with **Full / Hunks** (`e`)

## Comment gestures
- Hover a line and click the gutter **+** → jump straight to the comment box.
- Click anywhere on a line, or select a range of line numbers → action popup (**Add comment** / **Send to Claude**).

## Comment durability (the hard part)
A mounted `@pierre/diffs` instance binds to its first file and can't swap its target in place, so the diff must remount to show new content — and the file under review changes constantly while the agent edits it. Several fixes so comments survive that:
- **Switching back to an already-rendered file** now re-renders correctly (remount keyed by path + content hash).
- **A comment being typed is never lost** when the file changes underneath: the diff content freezes while a comment form is open, then adopts the latest the moment it closes.
- **Background comments arriving** no longer remount the form (annotations with an open form keep a stable index; the library keys slots by array index).
- **Drafts persist per file** and across save failures (retry without retyping).
- **Comments that can't render inline** — line removed, or hidden inside Hunks-mode collapsed code — gather into a collapsible banner at the top instead of silently disappearing.
- Comment markdown is rendered with `react-markdown` + `remark-gfm` (raw HTML escaped, not injected).

## Testing
- **23 Playwright e2e** against the real rendered DOM (`diff-view.spec.ts`): rendering, comment CRUD, both add gestures, file-switch re-render, typed/edited comment + focus survival across background comment / file-change / file-switch / failed-save, stale + collapsed-context banner, markdown escaping, unified/split toggle.
- **544 unit** tests pass; `pnpm run build` (tsc + vite) clean.

## Notes
- `marked` is no longer used in `src` (replaced by `react-markdown`); left in `package.json` — easy follow-up removal.
- Content-drift on an *in-range* line isn't detectable frontend-only (`ReviewComment` stores only line numbers); catching that would need a small protocol addition to store the anchored line's text/hash.


figgyster (dismissed):
I like the direction here — deleting the old bespoke editor is exactly the kind of simplification this surface needed. The new Playwright coverage is doing real work too.

Two durability gaps crossed from “nice follow-up” into “this will make the review UI lie,” though:

- Viewed-state persistence can miss the first file if the diff fetch wins the race against `getReviewState()`. The effect marks the file viewed locally, then when `reviewId` arrives it reruns with that file already in `viewedFiles`, so `isFirstView` is false and `markFileViewed()` never runs. Reopen the panel and the first file can look unviewed again.
- `changedSinceViewed` is only backed by the in-memory `viewedDiffHashesRef`. Persisted review state restores `viewed_files`, but not the hashes; after closing/reopening the panel, background checks skip every restored viewed file because there is no prior hash. A file can change after you viewed it and the app will have no durable way to flag it.

There’s also a smaller sharp edge worth fixing while you’re in there: hashing `original + modified` is ambiguous without framing, so different `(original, modified)` pairs can produce the same hash input. Cheap fix: hash a structured/framed payload.

I ran:
- `pnpm --dir app exec playwright test e2e/diff-view.spec.ts --reporter=line`

Those all pass. The blocker is the review-state contract, not the library swap itself.

figgyster (approved):
Nice fix. The viewed-state race is now keyed on the backend-persisted set instead of local “already viewed” state, and the changed-since-viewed baseline survives reopen with framed diff hashes. That closes the blocker cleanly.

The focused panel/hash tests and the full diff-view Playwright suite pass locally. No blockers from me.

figgyster (approved):
Still good after the merge from `main`. The PR diff is still the diff-view replacement/fix set, and the durability fixes remain intact on the current head.

I reran the focused checks after refreshing dependencies for the updated base:
- `pnpm --dir app exec vitest run DiffDetailPanel.test.tsx reviewHash.test.ts viewedDiffHashes.test.ts`
- `pnpm --dir app exec playwright test e2e/diff-view.spec.ts --reporter=line`

No blockers from me.

PR #272, opened 2026-06-07
Instrument terminal split/resize/close; coalesce split-drag PTY resizes

victorarias:
## Why

The "agent pane goes blank when a split opens" bug only reproduces under live timing — the real-app harness can't trigger it on demand. This adds continuous, prod-safe instrumentation so the bug is captured automatically during normal usage, plus general diagnostics for split / resize / close. Already installed to prod.

## Diagnostics (new)

`app/src/utils/terminalDiagnosticsLog.ts`, wired through the render pipeline:

- **Ring buffer** (3000 events) holds rich context — every paint (model-fill vs. quads drawn), write, resize, focus, layout — with zero disk I/O.
- **Lifecycle stream** → `$APPLOCALDATA/debug/terminal-diagnostics.jsonl`: low-frequency events (mount/unmount, resize, reset, split layout, focus, attach/desync). Async, size-capped at 8MB with rotation.
- **Self-detecting watchdog**: after a *real* agent-pane geometry change it probes the live model; if the model holds content but the surface drew ~nothing, it writes a full **incident** record (with 400 events of preceding context) → `terminal-incidents.jsonl`.
- Disable at runtime via `localStorage['attn:terminal-diagnostics']='0'`; DevTools globals `__ATTN_TERMINAL_DIAG_DUMP()`, `__ATTN_TERMINAL_DIAG_FILES`.

### Calibration

A WebGL canvas clears its drawing buffer only when its pixel dimensions change, so the watchdog arms **only on real grid-geometry changes** — no-op/same-size resizes (common at startup) leave the surface intact and don't produce false blanks. Validated live:

- Startup churn (8 panes dev / 56 panes prod): **0 incidents**, all watchdogs `blank=False`.
- Driven real split (`160x48→80x47`): captured cleanly, both post-split watchdogs healthy, no false incident.
- Live prod: watchdogs firing `blank=False painted=True` across varied geometries, **0 incidents**.

## Fix (tile-resize)

Coalesces the PTY resize sent to running agents while a split drag is in progress, so Claude/Codex aren't forced to repaint repeatedly mid-drag. Split drags still update the visible layout live. (CHANGELOG entry included.)

## Scaffolding

`scenario-agent-split-blank-probe.mjs` (real-app probe for the split gesture) and `split-blank-repro.spec.ts` (e2e) — diagnostic tooling to retire once the root cause is fixed.

## Testing

- `pnpm exec tsc --noEmit` clean; `pnpm test` — 577 pass.
- Validated end-to-end on dev + prod packaged apps (see above).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -202,8 +202,9 @@ When updating it:
 
 - write for users, not maintainers
 - summarize behavior changes, not implementation details
+- scope entries to the PR as a whole: describe the net user-visible outcome the PR delivers, not each step, iteration, or dead end taken along the way. A PR usually warrants a single focused entry even if it spanned many commits.
 - group related work into a small number of bullets
-- update it after meaningful features, fixes, removals, or commits
+- update it for meaningful user-visible features, fixes, or removals; internal-only changes (refactors, instrumentation, tests) do not need an entry
 
 ## When Something Is Broken
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,11 @@ The app shows a one-time "What's new" summary after this upgrade, and `⌘/` bri
 
 ---
 
+## [2026-06-07]
+
+### Fixed
+- **Resizing workspace splits is less likely to stall active agent panes.** Split drags still update the visible terminal layout live, but attn now coalesces the PTY resize sent to each running agent while the drag is in progress so Claude and Codex are not forced to repaint repeatedly mid-drag.
+
 ## [2026-06-06]
 
 ### Changed
--- a/app/e2e/split-blank-repro.spec.ts
+++ b/app/e2e/split-blank-repro.spec.ts
@@ -1,0 +1,273 @@
+import { test, expect } from './fixtures';
+
+// Repro for: opening a shell split blanks the agent pane (canvas shows a live
+// cursor over blank cells; model still holds the content). The render trace
+// (added in GhosttyTerminal.renderSurface, gated on __ATTN_RENDER_TRACE_ON)
+// records, per paint: the cells-source offset, the printable cell count the
+// MODEL holds, and the `quads` actually submitted to the GPU. That lets us tell
+// "drew nothing despite content" (cells-source/offset) from "drew glyphs but
+// surface stayed blank" (GL/canvas).
+
+const ESC = '';
+const BSU = `${ESC}[?2026h`;
+const ESU = `${ESC}[?2026l`;
+
+// Mirrors the `paint` events pushed into the diagnostics ring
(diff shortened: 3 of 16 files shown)


figgyster (dismissed):
This needs one small fix before merge: the new split-blank Playwright spec is broken at runtime.

I built the test binary and ran:

```bash
go build -o ./attn ./cmd/attn
pnpm --dir app exec playwright test e2e/split-blank-repro.spec.ts --reporter=line
```

All three tests fail in `readTrace()` with:

```text
TypeError: Cannot read properties of undefined (reading 'includes')
```

The spec filters `window.__ATTN_RENDER_TRACE` entries with `entry.debug.includes(sid)`, but the new diagnostics ring stores paint entries with fields like `pane` / `session`, not `debug`. So the diagnostic scaffolding currently trips over its own fedora before it can prove the terminal behavior.

Required before merge:
- update `readTrace()` (and the `RenderTraceEntry` shape) to match the new diagnostics event schema, probably filtering by `entry.session === sessionId` or `entry.pane`, and rerun this spec.

The production instrumentation/coalesced resize direction looks sane; I’m blocking only on the new principal repro test being red.

victorarias:
@codex review

figgyster (dismissed):
Looks good now. The split-blank repro is reading the diagnostics ring correctly, the DECSET 1003 hover regression Codex spotted is fixed, and the focused checks plus the split-blank Playwright spec pass locally.

Frontend CI is still running as I write this, but I’m comfortable with the code path: the blocker was the repro harness tripping over stale trace fields, and that’s gone.

figgyster (dismissed):
Codex caught a real prod-safety gap here, and I agree with it.

The diagnostic files are advertised as size-capped, but `lifecycleBytes` / `incidentBytes` restart at `0` on every app launch. If `terminal-diagnostics.jsonl` or `terminal-incidents.jsonl` already exists near the cap, this code appends up to another full in-memory cap before rotating; with repeated short app sessions, the files can keep growing past the promised bound. Tiny gremlin, but it lives in prod instrumentation.

Required before merge:
- initialize the byte counters from the existing file size, or stat/read the file before deciding `append` vs truncate, so the 8 MB cap is real across restarts.

The earlier repro/test blocker is fixed; this is the remaining merge blocker.

figgyster (approved):
Looks safe to merge now. The diagnostics cap fix seeds from the existing on-disk size before deciding append vs rotate, and the Tauri stat permission is wired in.

I rechecked the focused path locally and the visible CI is green. No blockers from me.

figgyster (approved):
Looks safe to merge. The post-resize suppression window is now just a short release guard, so divider drags still suppress their trailing event without eating deliberate terminal clicks/select/scrolls afterward.

Focused local checks passed, and visible CI is green. No blockers from me.

PR #273, opened 2026-06-08
perf(pty): cut per-session worker memory (scrollback + debug capture)

victorarias:
**PR 1 of a stacked performance series** (memory = P1, CPU = P2). Roadmap: `docs/plans/2026-06-08-performance-memory-cpu.md`.

## Why
A multi-agent audit + live `ps` baseline showed per-session **`pty-worker` subprocesses are ~75% of attn's memory** (~782 MB total; workers avg ~73 MB, peak ~109 MB). This PR trims two over-provisioned per-session allocations in the worker.

## Changes
- **PTY scrollback 8 MiB → 1 MiB** (`internal/pty/manager.go`). The eager ring buffer + lazily-grown replay log were each 8 MiB, but attach/re-attach only ever replays the last **256 KiB** (`maxAgentRawReplayBytes`). 1 MiB keeps 4× headroom over that clip.
  - Measured heap delta: **~7 MiB/session idle** (eager ring), **~14 MiB/session** when scrollback+replay are full.
  - Added `TestDefaultScrollbackCoversRawReplayClip` so the default can never silently drop below the replay clip (which would starve attach/Codex-startup replay).
- **Debug PTY capture off by default** (`internal/ptyworker/debug_capture.go`). `ATTN_DEBUG_PTY_CAPTURE` was effectively ON for claude/codex (a *"temporary debugging default"*), retaining up to **~16–22 MiB** of base64 per worker. Now strictly opt-in; added explicit per-agent cases (`codex`/`claude`/`copilot`).

For a user running ~8 sessions this reclaims roughly **150–250 MB**.

## Safety
- No protocol shape change → no `ProtocolVersion` bump.
- Replay/attach behavior unchanged. `pty`, `ptyworker`, `ptybackend`, and `daemon` test suites pass.

## Test
`go build ./...`; `go test ./internal/pty/... ./internal/ptyworker/... ./internal/ptybackend/... ./internal/daemon/...` (all green). Heap delta measured with a throwaway in-process benchmark (removed).

1. **#273** 👈 current
2. #274

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-06-08]
+
+### Changed
+- **attn uses noticeably less memory per session.** Each running session's terminal backend used to reserve 8 MB of scrollback up front and grow a second 8 MB replay buffer, even though only the most recent 256 KB is ever replayed when you switch back to a session — both are now capped at 1 MB, freeing roughly 7–14 MB of RAM per session. Separately, diagnostic terminal capture (which quietly held up to ~20 MB of recent output in memory for every Claude and Codex session) is now off unless you explicitly turn it on with `ATTN_DEBUG_PTY_CAPTURE`. With several sessions open this reclaims hundreds of megabytes.
+
 ## [2026-06-07]
 
 ### Added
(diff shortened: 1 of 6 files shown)


figgyster (approved):
This looks safe to merge. The memory cuts are scoped to the two over-provisioned worker buffers, and the replay guard keeps the 256 KiB attach/Codex-startup contract from getting shaved by accident.

I ran `go build ./...` and the focused `pty`/`ptyworker`/`ptybackend`/`daemon` suites locally; visible CI is green too. No blockers from me.

PR #274, opened 2026-06-08
perf(app): virtualize off-screen terminal workspaces (~200 MB)

victorarias:
**PR 2 of the stacked performance series** (stacked on #273). Memory = P1. Roadmap: `docs/plans/2026-06-08-performance-memory-cpu.md` (WS-1).

## Why
The audit's biggest single memory lever: every open workspace keeps a live Ghostty WASM model + WebGL renderer (~32 MiB atlas + GPU texture **per pane**, plus scrollback) mounted, even though only one workspace is visible. Across many workspaces this was ~half the frontend's RAM and ~half its VRAM.

## What
Keep only the **active workspace + the N most-recently-used** workspaces "warm" (terminals mounted). Other workspaces render a placeholder and **rehydrate from daemon replay** (the existing `same_app_remount` path) the instant they become visible again. Teardown frees the WebGL context (`loseContext`, already wired) and the WASM model.

### Runtime-configurable (no rebuild)
So we can tune the memory-vs-snappiness tradeoff and feel it out live:
- `window.attnSetWarmWorkspaces(n)` — applies immediately, persists to localStorage
- `n` = recent workspaces kept warm besides the active one (**default 3**)
- `n = 0` → active only (maximum savings); `n < 0` → disabled (all live, today's behavior)

## Safety / invariants
- **Focus ownership (#6):** active workspace always mounts; focus code untouched.
- **PTY geometry authority (#7):** `pty_resize` stays gated on `isActiveSessionRef`; virtualized panes are unmounted so they can't drive geometry.
- No protocol change. First visit of a startup-virtualized workspace uses `fresh_spawn` — identical to today's initial attach (no regression).
- **Scope:** unmount-only (frees frontend WASM+WebGL — the memory win). Detaching the daemon PTY stream for virtualized panes (a CPU/bandwidth follow-up) needs a new detach API and is deferred.

## Tests
- New `computeWarmWorkspaceIds` unit tests (6) — warm-set math incl. disabled/active-only/cap/dedup.
- New `SessionTerminalWorkspace` cases (4): placeholder when virtualized, every split pane virtualized (not just active), terminals mount when warm, and **unmount→remount** transition.
- Full frontend suite green (595 tests). `tsc` clean. Dev app builds + launches.

> Note: rehydrate-on-switch is intentionally observable (brief repaint; local scroll resets to live) for non-warm workspaces — tunable via the knob above.

1. #273
2. **#274** 👈 current

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ### Changed
 - **attn uses noticeably less memory per session.** Each running session's terminal backend used to reserve 8 MB of scrollback up front and grow a second 8 MB replay buffer, even though only the most recent 256 KB is ever replayed when you switch back to a session — both are now capped at 1 MB, freeing roughly 7–14 MB of RAM per session. Separately, diagnostic terminal capture (which quietly held up to ~20 MB of recent output in memory for every Claude and Codex session) is now off unless you explicitly turn it on with `ATTN_DEBUG_PTY_CAPTURE`. With several sessions open this reclaims hundreds of megabytes.
+- **Background workspaces no longer hold onto graphics memory.** Only the workspace you're in plus your few most-recent ones keep their terminals fully live; the rest release their GPU/terminal resources and instantly restore from history the moment you switch back. With many workspaces open this frees hundreds of megabytes of RAM and GPU memory and reduces background CPU.
 
 ## [2026-06-07]
 
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -2257,9 +2257,10 @@ sendFetchPRDetails,
         : [activeWorkspaceId, ...prev.filter((id) => id !== activeWorkspaceId)].slice(0, 32)
     ));
   }, [activeWorkspaceId]);
+  const allWorkspaceIds = useMemo(() => workspaceViews.map((w) => w.id), [workspaceViews]);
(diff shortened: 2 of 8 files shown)


figgyster (approved):
This looks safe to merge. Keeping the active workspace plus a small MRU warm set is the right cut for this memory slice: it frees the heavy Ghostty/WebGL side without changing the daemon protocol or PTY ownership rules.

I checked the virtualization math, split-pane behavior, and remount path. Local frontend verification passed: `tsc --noEmit`, the focused virtualization tests, and the full app Vitest suite (602 tests). No blockers from me.

figgyster (approved):
Still safe after the rebase/follow-up push. The PR diff is still the terminal workspace virtualization slice, and the current head keeps the same important boundaries: active workspace stays mounted, cold panes unmount instead of resizing/driving PTYs, and rehydration stays on the existing replay path.

Focused current-head checks passed: `git diff --check`, `tsc --noEmit`, and the virtualization/terminal workspace tests (27 tests). No blockers.

figgyster (approved):
Still safe to merge. This follow-up tightens the warm-set logic in the right direction: small workspace counts stay live, stale ids are filtered, and the startup/no-active-workspace case no longer collapses into accidental over-virtualization.

Focused checks passed: `git diff --check`, `tsc --noEmit`, and the virtualization/terminal workspace tests (31 tests). Frontend CI was still running when I checked, but the local coverage hits the changed seam. No blockers.

PR #275, opened 2026-06-08
chore: run pnpm install before app builds

victorarias:
## What

Add `cd app && pnpm install` to the `build_tauri_app` Makefile macro so it runs right before every Tauri app build.

## Why

`make`/`make install` (and `make dev`) now always reconcile `app/`'s dependencies against `package.json` before building, avoiding stale/missing-dep build failures in the dev inner loop. Both prod (`build-app`) and dev (`build-app-dev`) install paths flow through this macro, so a single change covers them; `dist` depends on `build-app` and is covered too.

Plain `pnpm install` (not `--frozen-lockfile`) is intentional — this is a dev convenience that should reconcile drift, not fail on it. It's a no-op when deps are already current.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -205,6 +205,7 @@ check-types: generate-types
 define build_tauri_app
 	@mkdir -p app/src-tauri/binaries
 	cp $(BINARY_NAME) app/src-tauri/binaries/$(BINARY_NAME)-aarch64-apple-darwin
+	cd app && pnpm install
 	cd app && $(1) pnpm tauri build --bundles app $(3)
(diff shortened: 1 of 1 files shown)


figgyster (approved):
Looks safe to merge. This keeps dependency reconciliation inside the shared Tauri build macro, so `build-app`, `build-app-dev`, `make dev`, and `dist` all pick up the same behavior instead of growing parallel install paths.

Validated the Makefile expansion with `make -n build-app` / `make -n build-app-dev`, and the visible CI is green. No blockers from me.

PR #276, opened 2026-06-08
Revert "perf(app): virtualize off-screen terminal workspaces (~200 MB) (#274)"

victorarias:
Reverts #274 on `main` at Victor's request. The terminal-virtualization work is being moved back into an open PR for further iteration rather than staying on `main` for now.

This is a clean revert of the squash-merge (e0c084e0); no behavior from #273 (pty worker memory) is affected.

— opened by Claude on behalf of @victorarias

PR #277, opened 2026-06-08
perf(app): virtualize off-screen terminal workspaces (~200 MB)

victorarias:
Re-opens the off-screen terminal virtualization work that was previously merged as #274 and then reverted on `main` via #276, so it can iterate as an open PR instead of living on `main`.

## What this does
Keeps only the active workspace + the N most-recently-used workspaces "warm" (terminals mounted). Other workspaces render a placeholder and rehydrate from daemon replay (`same_app_remount`) when next made visible, freeing the per-pane Ghostty WASM model + WebGL renderer (~21 MB measured per off-screen pane).

Virtualization only engages when there are **more** workspaces than the warm budget (active + N recent, default N=3). With no more workspaces than the budget there is nothing to reclaim, so all terminals stay live — matching the pre-virtualization behavior for the common case, and never tearing a terminal down before its workspace is active. (That edge — an empty live-set before the active workspace was established — was the regression that briefly blanked terminals in #274; it is fixed here and covered by unit tests.)

`window.attnSetWarmWorkspaces(n)` tunes N at runtime (persisted); `n=0` keeps only the active workspace, `n<0` disables virtualization. No protocol change.

— opened by Claude on behalf of @victorarias

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ### Changed
 - **attn uses noticeably less memory per session.** Each running session's terminal backend used to reserve 8 MB of scrollback up front and grow a second 8 MB replay buffer, even though only the most recent 256 KB is ever replayed when you switch back to a session — both are now capped at 1 MB, freeing roughly 7–14 MB of RAM per session. Separately, diagnostic terminal capture (which quietly held up to ~20 MB of recent output in memory for every Claude and Codex session) is now off unless you explicitly turn it on with `ATTN_DEBUG_PTY_CAPTURE`. With several sessions open this reclaims hundreds of megabytes.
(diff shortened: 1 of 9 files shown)


figgyster (dismissed):
Codex is right about the grid path. This crosses the product-contract line, not a taste line.

Grid mode is intentionally a read-only observer: `GridView` only listens to the PTY firehose and seeds from `get_screen_snapshot`; it never calls `attach_session` itself. With this PR, once a workspace falls out of the warm set, `SessionTerminalWorkspace` renders the placeholder instead of mounting `GhosttyTerminal`, so `handleTerminalReady`/`attach_session` never happens for that runtime. Open grid mode with more workspaces than the warm budget and those cold sessions can show the initial snapshot but stop receiving live output until you visit the workspace.

Required before merge:
- Either keep grid-visible runtimes attached while grid is open, or give grid its own daemon-supported observer attachment path that does not claim PTY geometry.
- Add a regression around the many-workspace + grid mode case so the firehose stays live for cold/off-screen sessions.

The focused virtualization tests and frontend typecheck pass locally, so the warm-set startup guard itself looks good. The grid observer contract is the blocker hiding in the alley.

figgyster (approved):
This fixes the blocker cleanly. Grid-visible workspaces now stay protected while grid mode is open, so the existing mounted-terminal attach/firehose path remains live without making grid claim PTY geometry. That keeps the memory win for ordinary cold idle workspaces without starving mission control.

I reran the focused virtualization/App regression tests and the frontend production build locally. No blockers from me; let the latest Frontend CI run finish before merging.

PR #278, opened 2026-06-08
feat(grid): visualize session states

victorarias:
## Intent / Why

Grid mode shows many live sessions at once, but terminal content alone makes it hard to scan which sessions are active or need attention. This adds an experimental state treatment so the visual hierarchy can be evaluated directly in the app.

## Summary

- color grid tiles by session state using either borders or subtle background tints
- add a persistent `State` control for switching between Border and Tint treatments
- flash waiting-for-input sessions in amber so required action is visible at a glance
- keep the blue zoom focus treatment separate from semantic session state
- map review-loop waiting and error states into the same grid state model
- add render-on-state-change, persistence, animation, and automation coverage

## Test plan

- [x] Run focused grid tests: 22 tests passed
- [x] Run TypeScript checking with `pnpm exec tsc --noEmit`
- [x] Build and install the signed production app with `make`
- [x] Verify Border and Tint presentations with multiple live sessions
- [ ] Compare both treatments during normal multi-session use and choose a default

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-06-08]
 
 ### Changed
+- **Grid view can show session state with colored borders or tile tints.** Use the new State control in grid mode to compare the two treatments; waiting-for-input sessions flash amber, the choice persists across launches, and zoom focus remains clearly separated from session status.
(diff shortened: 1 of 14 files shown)


figgyster (approved):
This looks safe to merge. The grid state treatment keeps semantic status separate from the blue zoom focus, and the follow-up focus fix closes the easy trap where clicking the State control could steal keyboard input from the zoomed tile.

I ran `pnpm --dir app exec tsc --noEmit` and the focused grid vitest suite: 22 tests passed. GitHub Frontend CI was still in progress when I posted, so let that finish before merging.

PR #279, opened 2026-06-08
feat(grid): combine state borders and tints

victorarias:
## Intent / Why

The border and tint experiments work better together than as mutually exclusive modes: the border makes state scannable at the tile edge, while the tint keeps the status visible across the full terminal surface.

## Summary

- render every session state with both a colored border and subtle background tint
- keep the amber waiting-for-input flash synchronized across both treatments
- remove the State mode selector, persisted preference, and presentation-specific automation plumbing
- preserve the blue zoom border while retaining the semantic state tint

## Test plan

- [x] Focused grid unit tests: 19 passed
- [x] TypeScript checking with `pnpm exec tsc --noEmit`
- [x] Production frontend build with `pnpm run build`
- [x] GridView membership Playwright scenarios: 2 passed
- [x] Structured Codex autoreview: no actionable findings
- [x] Signed production app installed with `make`


figgyster (approved):
This looks safe to merge. The grid state treatment is simpler now: one semantic path paints both the edge signal and the tint, while zoom focus stays visually separate.

I ran the focused grid tests, TypeScript, and the production frontend build locally. GitHub's Frontend check was still in progress when I posted this, but the local signal is clean from here.

PR #281, opened 2026-06-08
perf(cpu): gate per-PTY-chunk logging behind DEBUG (WS-4)

victorarias:
First of the CPU (P2) workstreams from the perf roadmap. Independent of the parked terminal-virtualization work (#277) — daemon/worker Go only — so it branches off `main`.

## Problem
On the hottest path (one event **per output chunk per attached client**) the daemon called `d.logf("pty_output forward: … preview=%q", …, previewBinaryForLog(data))`. `d.logf` → `Info` → `log()` is **not** level-gated, so even with `DEBUG` unset every chunk took the global log mutex for a synchronous disk write **and** allocated a byte-preview string. The same per-chunk pattern ran in each PTY **worker** subprocess, whose stderr is an unrotated per-session `.log` — so that file grew without bound on every session.

## Fix
Cache a debug flag and gate the verbose per-chunk/verbose logs **at the call site** (Go evaluates call args eagerly, so the guard must wrap the call to skip the preview allocation, not just the write):
- daemon: `d.debugLogging` from a new `logging.Logger.DebugEnabled()` accessor; gates `pty_output` / `pty_desync` / `pty_input` / `pty_input ok`.
- worker: `ptyworker.Config.Debug` (set in `cmd/attn` from `config.DebugLevel()`, which the worker inherits via env); gates the per-chunk `worker output event` log.

Error and lifecycle logs stay unconditional. With `DEBUG` set the lines appear exactly as before (INFO level) — debugging is unchanged. No protocol or PTY-datapath change.

## Tests
`go build ./...`, `go vet`, and `internal/daemon` + `internal/ptyworker` + `internal/logging` suites all green; added `TestLogger_DebugEnabledAccessor` covering the DEBUG→flag linchpin the gating relies on.

— opened by Claude on behalf of @victorarias


figgyster (approved):
This looks safe to merge. The hot-path logging is gated at the call sites, so DEBUG-off runs skip both the preview allocation and the synchronous log write while keeping lifecycle/error logs intact.

Focused Go checks passed locally, and the visible daemon/preflight checks are green. No blockers from me.

figgyster (approved):
Still safe on the new head. The added benchmark pins the exact cost this slice removed, and the focused daemon/logging/worker tests plus the benchmark run are clean locally.

No blockers from me.

figgyster (approved):
Still green after the rebase onto current `main`. The PR patch is the same WS-4 logging/benchmark slice, and the daemon checks are now green too.

No blockers from me.

PR #283, opened 2026-06-08
feat(diag): opt-in loopback pprof + /debug/vars endpoint (WS-7)

victorarias:
## WS-7 — measurement enabler (pprof + /debug/vars)

There is currently **zero** pprof / expvar / GOMEMLIMIT anywhere in the daemon, so the perf roadmap's memory work (WS-1/WS-2 landed, WS-3 pending) is unmeasurable. This adds the smallest thing that unblocks it: an **opt-in, loopback-only** diagnostics endpoint.

Independent of #281 (WS-4); branches off `main`.

### What

- New `internal/diag` package: `Start(addr, StatsFunc)` binds a loopback `http.Server` on a **private** mux serving `net/http/pprof` + a custom `/debug/vars` JSON snapshot. (`http.DefaultServeMux` is left untouched — the WS server already uses its own mux, so the pprof import's default-mux registration stays inert.)
- `config.PprofAddr()` gates it on `ATTN_PPROF` (unset/`off` → disabled; `1`/`on` → `127.0.0.1:6060`; `<port>`/`:<port>` → that loopback port; any host in the value is ignored and forced to `127.0.0.1`).
- Wired into the daemon lifecycle: started after the WS server in `Start()`, closed in `Stop()`. A bind failure is logged, never fatal.
- `/debug/vars` reports `runtime.MemStats` heap fields (`heap_alloc/sys/idle/released`, `num_gc`, …), goroutine/CPU counts, PTY session count, the active `pty_backend`, and **`worker_pids`** (sessionID → worker subprocess PID).
- New `ptybackend.WorkerProcessProvider` exposes worker PIDs. The default `worker` backend runs one subprocess per session, so summing RSS over those PIDs is the per-session memory measurement (the dominant locus). Compile-time guarded so it can't silently degrade to "embedded".

### Safety

Strictly **off by default** and **loopback-only** — no new remote attack surface. macOS-only is fine.

### Tests

- `config.PprofAddr` parsing table (on/off/port/`:port`/host:port→loopback/garbage/out-of-range).
- `internal/diag`: serves `/debug/vars` (valid JSON + memstats keys) and pprof index/heap, bound to `127.0.0.1`, nil-stats safe, bad-addr errors, nil-Close safe.
- Daemon wiring: `ATTN_PPROF` set → loopback server serving `/debug/vars`; unset → no server; `diagStats()` reports the embedded backend.
- `WorkerPIDs` filters out sessions with no live worker.

### Verified on the real daemon

Ran an isolated throwaway-profile daemon with `ATTN_PPROF`:

```
$ curl -s http://127.0.0.1:16060/debug/vars
{ "pty_backend": "worker", "sessions": 0, "worker_pids": {},
  "memstats": { "heap_alloc": 997776, "heap_sys": 7634944, "heap_idle": 4743168,
                "heap_released": 2809856, "num_gc": 2, ... }, ... }

$ go tool pprof -top http://127.0.0.1:16060/debug/pprof/heap
  1026kB 49.66%  runtime.allocm
528.17kB 25.56%  regexp.(*bitState).reset
512.05kB 24.78%  os/exec.(*Cmd).watchCtx
         ...     github.com/victorarias/attn/internal/daemon.(*Daemon).Start.func4
```

`pty_backend: "worker"` confirms the backend detection; `go tool pprof` symbolizes real attn frames. SIGTERM released the port cleanly (graceful `Close()`).

Usage runbook: `docs/diagnostics.md`.

### DIFF
--- a/docs/diagnostics.md
+++ b/docs/diagnostics.md
@@ -1,0 +1,65 @@
+# Diagnostics: pprof + /debug/vars
+
+attn ships an **opt-in, loopback-only** diagnostics endpoint for measuring memory
+and CPU. It is **off by default** and binds `127.0.0.1` only, so it never adds
+remote attack surface.
+
+## Enable
+
+Set `ATTN_PPROF` in the daemon's environment:
+
+| `ATTN_PPROF`        | Effect                              |
+| ------------------- | ----------------------------------- |
+| unset / `0` / `off` | disabled (default)                  |
+| `1` / `on` / `true` | enabled on `127.0.0.1:6060`         |
+| `6061`, `:6061`     | enabled on that loopback port       |
+
+The daemon logs `diagnostics endpoint listening on http://127.0.0.1:<port>/` on
+startup. The simplest controlled setup for a measurement run is a manual daemon
+on a throwaway profile (its own socket/port/data dir, never touches prod):
+
+```bash
+env ATTN_PROFILE=perf ATTN_PPROF=6060 attn daemon
+```
+
+To profile the live app daemon instead, set `ATTN_PPROF` in the environment it is
+launched from.
+
+## Endpoints
+
+- `GET /debug/vars` — JSON snapshot: Go `runtime.MemStats` heap fields
+  (`heap_alloc`, `heap_sys`, `heap_idle`, `heap_released`, `num_gc`, …),
+  goroutine/CPU counts, PTY session count, the active `pty_backend`, and
+  `worker_pids` (sessionID → worker subprocess PID).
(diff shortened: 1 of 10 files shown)


figgyster (approved):
This looks safe to merge. The diagnostics surface is opt-in, loopback-only, keeps pprof on a private mux, and exposes the worker PIDs needed for the WS memory work without turning the daemon into a public vending machine for profiles.

I ran `git diff --check origin/main...HEAD` and the focused Go suites for config, diag, daemon, and ptybackend. GitHub’s daemon/preflight checks are green too. No blockers from me.

PR #285, opened 2026-06-09
feat(workspaces): teach agents shared context

victorarias:
## Intent / Why

Shared workspace context is only useful if agents know where the live document is and how to maintain it. Follow-up to #282: inject management guidance at session start instead of copying a potentially stale context snapshot into the prompt.

## Summary

- Check out the live workspace context when Claude or Codex starts, resumes, clears, or compacts.
- Inject the editable path plus concise read, update, status, and conflict-recovery instructions through `SessionStart` hook context.
- Keep Copilot and plugin agents unchanged until they expose a compatible instruction channel.
- Update the bundled attn skill, implementation plan, and changelog.

## Test plan

- [x] `go test ./...`
- [x] `make dev`
- [x] Real Codex session read marker `ORBIT-417` from the injected checkout without running `show`, published revision 2, and reported clean status.
- [x] Real Claude session started afterward, read `ORBIT-417` plus Codex's handoff, published revision 3, and reported clean status.
- [x] Disposable dev sessions and workspace removed after verification.

## Out of scope

- Proactively pinging already-running agents when `workspace_context_changed` fires.
- Chief-of-staff ownership and dispatched-agent tracking.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-06-09]
+
+### Added
+- **Claude and Codex are taught to use workspace context when a session starts or resets.** attn checks out the live context file and injects concise instructions to read it, keep durable goals and decisions current, publish edits, and reconcile revision conflicts without copying the context itself into the prompt.
+
 ## [2026-06-08]
 
 ### Added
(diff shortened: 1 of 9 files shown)


figgyster (approved):
This looks safe to merge. The hook now points Claude/Codex at the live workspace-context checkout instead of stuffing a stale snapshot into the prompt, and the retry around session registration is the right little seatbelt for startup races.

Focused checks passed locally (`git diff --check` plus the workspace-context daemon/client/protocol/cmd/hook/agent Go packages), and the visible GitHub checks are green/skipped as expected. No blockers from me.

PR #286, opened 2026-06-09
perf(frontend): grow glyph atlas on demand (1024²→2048²) — ~100 MB idle

victorarias:
## What

Each terminal's WebGL renderer eagerly allocated a **2048×2048 backing canvas + 2048×2048 RGBA GPU texture (~32 MB/renderer)** up front, independent of how many glyphs a session actually uses (`GhosttyWebGlRenderer.ts`). A terminal renders a small, mostly-fixed glyph set (ASCII + styles + box-drawing + the occasional Unicode), so this was heavily over-provisioned for the common case.

This change makes the atlas **start at 1024² (~8 MB/renderer) and grow on demand** — doubling, capped at the previous 2048² — only when a glyph-heavy session (large CJK/emoji output) actually overflows it. `resetAtlas` reuses the atlas once at the cap. Glyph-light sessions (the common case) stay small; heavy sessions grow to today's capacity, so there's **no atlas thrash and no worst-case regression**.

## Measured impact

Dev app, 8 `shell` sessions, **idle**, warm-set 3 (`scenario-perf-baseline.mjs`, full process-tree RSS incl. WebKit GPU process):

| metric | before (2048²) | after (grow-on-demand) | Δ |
|---|---|---|---|
| total RSS | 825.1 MB | 723.7 MB | **−101 MB** |
| WebKit total | 456.9 MB | 355.4 MB | **−101 MB** |
| WebKit GPU process | ~181 MB | 90.2 MB | **−91 MB** |

The win lands in the GPU process (texture + canvas backing), ≈24 MB × ~4 live renderers at warm-set 3.

## Verification

- **Rendering through a real grow:** flooded a pane with ~3000 distinct CJK glyphs (well past the 1024² capacity ~1150 at DPR 2), forcing a grow to 2048². All CJK glyphs, a trailing ASCII marker line, and the powerline prompt render correctly (no tofu/garble/misalignment) — confirms grow-trigger + post-grow UV recomputation + re-rasterization.
- **Unit test:** `nextAtlasSize` policy — starts 1024², doubles to the 2048² cap, idempotent at the cap, converges and never exceeds it under repeated growth.
- Full frontend suite green (656 tests); `tsc` clean.

## Scope / safety

Render-only change, isolated to the glyph atlas allocator. AGENTS.md invariants **#6 (focus ownership)** and **#7 (PTY single-authority + replay)** are untouched.

Independent of the measurement-harness PR (#284).

### DIFF
--- a/app/src/components/GhosttyWebGlRenderer.ts
+++ b/app/src/components/GhosttyWebGlRenderer.ts
@@ -63,9 +63,25 @@ export function graphemeAtViewportCell(
     : terminal.getGraphemeString(bufferRow - history, col);
 }
 
-const ATLAS_SIZE = 2048;
+// The glyph atlas is allocated eagerly and in full (a backing 2D canvas plus a
+// GPU texture of the same dimensions), so its size is a fixed per-renderer
+// memory cost paid up front regardless of how many glyphs a session actually
+// uses. A terminal renders a small, mostly-fixed glyph set (ASCII + styles +
+// box-drawing + the occasional Unicode/emoji), so most panes fit comfortably in
+// 1024². We therefore start small and grow on demand up to the previous fixed
+// size only when a glyph-heavy session (e.g. CJK/emoji) actually overflows the
+// atlas — keeping the common case cheap (≈8 MB/renderer vs ≈32 MB at 2048²)
+// without risking glyph-atlas thrash on heavy content. See growAtlas/resetAtlas.
+export const INITIAL_ATLAS_SIZE = 1024;
+export const MAX_ATLAS_SIZE = 2048;
 const FLOATS_PER_VERTEX = 8;
-const SOLID_TEXEL_CENTER = 0.5 / ATLAS_SIZE;
+
+// Next atlas size when the current one fills: double it, but never exceed the
+// cap. Idempotent at the cap, so repeated growth always converges to
+// MAX_ATLAS_SIZE and can never grow unbounded.
+export function nextAtlasSize(current: number, max: number = MAX_ATLAS_SIZE): number {
+  return Math.min(current * 2, max);
+}
 const BLOCK_ELEMENT_RECTS: Readonly<Record<number, readonly BlockRect[]>> = {
   0x2580: [{ x: 0, y: 0, width: 1, height: 1 / 2 }],
   0x2581: [{ x: 0, y: 7 / 8, width: 1, height: 1 / 8 }],
@@ -202,10 +218,17 @@ export class WebGlTerminalRenderer {
   private readonly fontSize: number;
   private readonly fontFamily: string;
   private readonly theme: RendererTheme;
+  private atlasSize = INITIAL_ATLAS_SIZE;
   private atlasX = 2;
   private atlasY = 1;
   private atlasRowHeight = 0;
   private atlasGeneration = 0;
+
+  // UV of the center of the 1×1 solid white texel at atlas pixel (0,0). Depends
+  // on the current atlas size, so it is recomputed rather than precomputed.
+  private get solidTexelCenter(): number {
+    return 0.5 / this.atlasSize;
+  }
   private retryingAtlasFrame = false;
   private cols = 0;
   private rows = 0;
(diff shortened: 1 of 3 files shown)


app/src/components/GhosttyWebGlRenderer.ts
```diff
@@ -442,13 +465,19 @@ export class WebGlTerminalRenderer {
...
       this.atlasY += this.atlasRowHeight + 1;
       this.atlasRowHeight = 0;
     }
-    if (this.atlasY + height >= ATLAS_SIZE) {
-      this.resetAtlas();
+    if (this.atlasY + height >= this.atlasSize) {
+      // Atlas is full: grow it (doubling, up to the cap) so glyph-heavy sessions
+      // get more room instead of thrashing; only clear-and-reuse once at the cap.
+      if (this.atlasSize < MAX_ATLAS_SIZE) {
+        this.growAtlas();
```
figgyster, line 478: `growAtlas()` calls `reseedAtlas()`, and `reseedAtlas()` assigns `this.atlas.width/height`. Canvas resize resets the 2D context state, including `font`. This `getGlyph` call then continues at line 485/488 and caches the grow-triggering glyph after drawing it with the default canvas font, so the first glyph after a real atlas grow can render wrong and survive the retry via the glyph cache. Reapply `context.font`/text state after reseeding, or move the grow/reseed before the measure/draw setup.

figgyster (dismissed):
This is the right memory move, but the actual grow path has one sharp edge: resizing the backing canvas resets the 2D context state mid-`getGlyph`, then the glyph that triggered growth is cached after being drawn with the browser's default font. That means the first glyph after a real 1024²→2048² grow can render wrong and stay wrong through the retry.

Required before merge: reapply the font/text context after `reseedAtlas()` in the same `getGlyph` call, or make growth happen before measuring/drawing so the current glyph is rasterized with the intended font. A focused regression around the grow-triggering glyph would be perfect here. Tiny canvas API rake, naturally placed exactly where the perf win lives.

figgyster (dismissed):
The blocker is still present on this head. The new push moved the PR forward, but `getGlyph` still calls `growAtlas()`/`resetAtlas()` after setting `context.font`, and `reseedAtlas()` still resizes the backing canvas. That resize resets the 2D context state before the grow-triggering glyph is drawn and cached.

Required before merge remains the same: reapply the font/text context after reseeding, or restructure the grow path so the current glyph is measured/drawn after the resize. Add the focused regression for the glyph that triggers growth.

I did re-check the current head; `git diff --check` is clean. The bug is just still sitting there, wearing the same hat.

app/src/components/GhosttyWebGlRenderer.ts
victorarias: Fixed in b08f67c3 — confirmed exactly the mechanism you described. `reseedAtlas()` assigns `this.atlas.width/height`, which resets the 2D context `font` to the browser default, and `getGlyph` then drew + cached the grow-triggering glyph with that default. The `render()` retry only rebuilds vertices and reuses the cached glyph, so it never self-corrected.

Fix: re-apply `context.font` after the grow/reset block, before drawing — covers both `growAtlas()` and `resetAtlas()` (the at-cap path resizes the canvas the same way).

Regression: added `GhosttyWebGlRenderer.test.ts` › "keeps the intended font on the glyph that triggers a grow and an at-cap reset". It drives the real `getGlyph()` grow + at-cap reset paths through a recording 2D context that reproduces the resize-resets-font behavior. Verified it fails on the pre-fix code (`expected '10px sans-serif' to be '14px monospace'`) and passes with the fix.

— Claude, on behalf of Victor

figgyster (approved):
This fixes the atlas-grow bug cleanly. Reapplying the canvas font after the grow/reset covers both the 1024²→2048² path and the at-cap reset path, and the regression drives the browser behavior that bit us instead of just testing the policy helper.

Local checks are green: diff check, focused renderer test, TypeScript, and the full frontend suite. No blockers from me.

PR #287, opened 2026-06-09
fix(agents): improve workspace context guidance

victorarias:
## Intent / Why

The first workspace-context guidance told agents to preserve local edits during
revision conflicts, but real Codex testing showed that instruction was too
abstract: it ran `show --force` first, erased the modified checkout, and
reconstructed the edit from conversational memory.

This makes the workflow operational and improves the shared document's quality
so new agents receive durable coordination state rather than duplicated
summaries or command history.

## Summary

- define clear publish criteria and section ownership for durable context
- treat copied context as data that cannot override user or project guidance
- default agents to their own checkout rather than cross-session operations
- require saving modified context before a force refresh, with exact commands
- direct agents to the detailed attn reference when resolving conflicts
- test destructive-command ordering and shell quoting for checkout paths

## Test plan

- [x] `go test ./...`
- [x] focused hooks, skill-install, and CLI tests
- [x] Claude no-op read without publishing
- [x] Codex durable decision update
- [x] Claude investigation handoff, iterated until sections stopped repeating facts
- [x] Codex ignored prompt injection copied into context and preserved protected files
- [x] Claude losing-publisher conflict recovery preserved both edits
- [x] Codex losing-publisher conflict recovery loaded the skill, saved before refresh, merged both edits, published revision 3, and ended clean
- [x] final Codex smoke test against the exact installed guidance left an unchanged context at revision 1

## Review note

The local structured autoreview helper was attempted but blocked because it
would export the private local diff to an external model without separate
approval. The patch was manually reviewed and the full Go suite passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-06-09]
 
 ### Added
-- **Claude and Codex are taught to use workspace context when a session starts or resets.** attn checks out the live context file and injects concise instructions to read it, keep durable goals and decisions current, publish edits, and reconcile revision conflicts without copying the context itself into the prompt.
+- **Claude and Codex receive workspace-context guidance without cluttering Codex's terminal.** attn gives each session a local checkout and concise instructions to read it, keep durable goals and decisions current, publish edits, and reconcile revision conflicts without copying the shared context itself into the prompt.
 
 ### Changed
 - **Workspace context guidance is safer and less noisy.** Agents now publish only durable shared changes, keep each fact in one appropriate section, treat copied context as untrusted data, and save local edits before refreshing a conflicting revision.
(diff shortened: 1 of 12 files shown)


figgyster (approved):
This looks safe to merge. The guidance now says the quiet parts out loud: publish only durable shared state, keep copied context as data instead of instructions, and save the local checkout before a force refresh. That closes the exact failure mode from the agent test run instead of hoping the next agent reads between the lines.

I ran `git diff --check`, focused Go tests for the touched agent/hooks/CLI areas, and `go test ./...`; GitHub's daemon/check jobs are green. No blockers from me.

figgyster (approved):
Still safe to merge. Moving Codex's workspace-context guidance into developer instructions is the right shape here: it keeps the terminal quiet while preserving the same safety boundary around stale/shared context.

I rechecked the follow-up diff, `git diff --check`, focused agent/hooks/CLI tests, and `go test ./...`. No blockers from me.

figgyster (approved):
Still good. Hiding Claude's workspace-context guidance through `--append-system-prompt` matches the Codex cleanup: the agents get the coordination rules, Victor doesn't get terminal confetti.

The capability gate and hook fallback keep the rollout bounded, and the focused tests plus `go test ./...` are green. No blockers from me.

PR #289, opened 2026-06-09
fix(workspaces): prevent closed session sidebar flicker

victorarias:
## Intent / Why
A workspace with shared context survives after its final session closes. The app could retain that session's old spawning layout and reconstruct it during later workspace state updates, making the closed session repeatedly appear and disappear in the sidebar.

## Summary
- centralize daemon retention for sessionless workspaces around durable tiles or shared context
- publish an authoritative empty layout when context keeps a workspace alive after its final pane closes
- invalidate frontend layout caches that still reference an explicitly unregistered session
- cover the full context to final-pane-close to later-state-update lifecycle

## Test plan
- [x] go test ./internal/daemon
- [x] pnpm --dir app test
- [x] git diff --check
- [x] installed and verified in the production profile

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ### Added
 - **Claude and Codex are taught to use workspace context when a session starts or resets.** attn checks out the live context file and injects concise instructions to read it, keep durable goals and decisions current, publish edits, and reconcile revision conflicts without copying the context itself into the prompt.
 
+### Fixed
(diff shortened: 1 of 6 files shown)


figgyster (approved):
This looks safe to merge. The daemon now makes the sessionless-context workspace state authoritative, and the frontend drops stale cached layouts that still point at the closed session — the bug was a two-sided ghost, and this nails both sides.

Local checks passed: `git diff --check`, `go test ./internal/daemon`, and `pnpm --dir app test` (653 tests). GitHub Frontend CI was still pending when I checked, but it’s covered by the same local app suite here.

No blockers from me.

PR #293, opened 2026-06-09
fix(chief-of-staff): hide closed dispatches

victorarias:
## Why

The home dashboard is an operational view of work the user can still inspect and steer. A persisted chief-of-staff dispatch remained there after its delegated session was closed, leaving a disabled historical row with no useful action.

## What changed

- Show chief-of-staff dispatches on home only while their delegated session is open.
- Keep the persisted dispatch report intact so a later delegation registry can expose history.
- Add a regression test covering session removal without dispatch deletion.
- Document the user-visible behavior.

## Test plan

- [x] `pnpm --dir app exec vitest run src/components/Dashboard.test.tsx`
- [x] Full frontend suite: 83 files, 666 tests
- [x] `git diff --check`


figgyster (approved):
This looks safe to merge. The dashboard now treats persisted chief-of-staff dispatch rows as history once their delegated session is gone, while the active home view stays focused on work Victor can still inspect or steer.

I verified the focused dashboard regression, TypeScript, `git diff --check`, and the full frontend suite locally; GitHub's visible Frontend/Changes checks are green too. No blockers from me.

PR #294, opened 2026-06-09
fix(terminal): prevent resize render storms

victorarias:
## Intent / Why

Returning from Home to a resized workspace with busy split terminals could starve the WebView main thread and freeze the entire attn window. Runtime evidence showed the daemon and PTYs remained responsive while frontend diagnostics stopped, pointing to repeated WebGL rendering rather than a daemon deadlock.

## Summary

- defer WebGL paints for hidden workspaces while keeping their Ghostty models current, then repaint dirty cells when the workspace becomes active
- coalesce output-driven paints to one animation frame and skip exact-size PTY resize acknowledgements
- replace the terminal diagnostics array shift with a true O(1) circular ring buffer
- add an end-to-end regression for Home -> window resize -> hidden output burst -> return to a split workspace
- document that signed app builds must run outside the sandbox so keychain identity discovery remains stable

## Test plan

- [x] `pnpm --dir app exec vitest run src/utils/terminalDiagnosticsLog.test.ts src/utils/terminalSynchronizedOutput.test.ts src/utils/terminalSizing.test.ts`
- [x] `pnpm --dir app exec playwright test e2e/split-blank-repro.spec.ts`
- [x] packaged `attn-dev` reproduction with a real 1 MiB hidden-pane output burst and Home/resize/return flow
- [x] production app rebuilt, strictly signature-verified, installed, and relaunched with sessions recovered

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -38,6 +38,8 @@ make install-daemon-dev  # faster: only rebuild the Go sidecar inside attn-dev.a
 
 The dev install is fully isolated: its own bundle identifier (`com.attn.manager.dev`), its own data dir (`~/.attn-dev/`), its own socket, its own port. Once a fix is verified, run `make` to install the production app so the user receives the fix immediately. The production install closes and reopens the app and restarts its daemon; attn is designed to recover persisted session state gracefully. `make install` and `make install-daemon` refuse at parse time if `ATTN_PROFILE` is set in the environment — if you hit that error during iteration, you meant `make dev`.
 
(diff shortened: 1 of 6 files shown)


figgyster (approved):
This looks safe to merge. The fix is aimed at the right choke point: hidden workspaces still ingest PTY data, but WebGL paints are deferred/coalesced instead of letting inactive split panes turn the main thread into soup.

I ran the focused frontend checks plus the split-blank Playwright repro locally after building the test daemon binary; GitHub's visible Frontend and Changes checks are green now too. No blockers from me.

PR #298, opened 2026-06-10
perf(pty): stream live PTY output as binary websocket frames

victorarias:
## What

Live `pty_output` traveled daemon→app as base64-in-JSON text frames. Per 16KB PTY chunk that meant: daemon base64-encode + JSON marshal (~21KB envelope), then app-side `JSON.parse` + `atob` + byte-array build. This PR moves live output to compact binary websocket frames:

```
[type u8][id length u8][session id utf8][seq u32 BE][raw PTY bytes]
```

- **Opt-in per client** via a new `binary_pty_output` capability in `client_hello`. Non-opted clients (daemon-to-daemon hub relays, automation harness, older clients) keep the JSON event unchanged; relayed remote-endpoint sessions still arrive as JSON and use the existing frontend path.
- Codec lives in `internal/protocol/binaryframe.go` (Go) and `app/src/pty/binaryPtyFrame.ts` (TS), with round-trip + malformed-frame tests on both sides.
- `ProtocolVersion` 94 → 95.

## Why / measured effect

The ingest-stage bisect (roadmap, 2026-06-10) localized the streaming memory balloon to WS ingestion (~380MB receive + ~192MB decode at 2 panes × 1M ascii lines), with ghostty write+render ≈ 15MB.

Measured on the packaged dev app, same workload, all panes live:
- **JSON path:** ~687MB webContent high-water
- **Binary frames:** 567–657MB across runs (run-to-run variance ±50MB)

So: an unconditional CPU + bandwidth win (no per-chunk base64/JSON on either side, 25% smaller frames), but only a **modest memory win**. The balloon tracks per-message *count* more than encode format — consistent with the earlier slowwide-vs-wide result. The follow-up memory lever is daemon-side read coalescing (fewer, larger messages), which needs no protocol change and stacks on this frame format.

## Verification

- Go: frame codec round-trip/malformed tests; `encodePtyOutputMessage` capability-branch test (binary client gets binary frame, legacy client gets identical JSON); full `go test ./...` green.
- Frontend: decoder unit tests; full vitest suite green (671); `tsc` clean.
- **New e2e** `binary-pty-output-realpty.spec.ts`: spawns a real shell through a real daemon, asserts output renders **and** that the transport was actually binary (pty perf snapshot: binary frames > 0, JSON pty events = 0, base64 chars = 0) — catches a silent fallback to JSON that rendering alone can't.
- Full mock-PTY e2e suite: 104 passed (one known-flaky `split-blank-repro` failed in-suite, passes in isolation).
- Packaged dev app: `make dev`, daemon log confirms `capabilities=[workspace_sessions binary_pty_output]`, harness fill runs end-to-end with correct pane content.

## Drive-by: realpty e2e scaffold repairs

The realpty specs were rotted **before this PR** (failed at the first raw ws command on baseline): raw ws helpers never sent `client_hello` (required since the workspace-protocol gate), spawn lacked the now-required `workspace_id`, and daemon fixture temp dirs under `/var/folders` made worker unix socket paths exceed macOS's 104-byte `sun_path` limit. Fixed all three (hello-first helpers, workspace registration, `/tmp` fixture dirs). `review-loop-realpty` now progresses past spawn but still fails later on its own review-loop UI assertion — pre-existing feature-spec rot, left for a separate fix.


figgyster (approved):
This looks safe to merge. The binary-frame path keeps the legacy JSON event for non-opted/relayed clients, the frame codec is bounded and covered on both sides, and the real-PTY spec checks the thing that would otherwise silently lie to us: transport shape, not just rendered output.

I ran `git diff --check`, `go test ./...`, `pnpm --dir app exec vitest run src/pty/binaryPtyFrame.test.ts src/hooks/useDaemonSocket.test.tsx`, and `pnpm --dir app exec tsc --noEmit` locally. GitHub checks weren’t reported yet from the API, but the principal local loop is green.

No blockers from me.

figgyster (approved):
Still safe on the refreshed head. The follow-up keeps the binary PTY path intact and fixes the protocol-version test drift by using the exported app constant instead of another soon-stale magic number — tiny rake removed from the alley.

Focused re-checks passed: `git diff --check`, `go test ./internal/protocol ./internal/daemon`, `pnpm --dir app exec vitest run src/pty/binaryPtyFrame.test.ts src/hooks/useDaemonSocket.test.tsx`, and `pnpm --dir app exec tsc --noEmit`.

No blockers from me.

figgyster (approved):
Still clean. This push is just `gofmt` on the binary-frame test table, and the focused protocol/daemon/frontend checks are green locally.

No blockers from me.

PR #299, opened 2026-06-10
perf(pty): coalesce PTY reads under sustained output

victorarias:
## What

Lever B from the memory investigation: the daemon's PTY read loop now coalesces reads under sustained output, so heavy streaming reaches the app as few large `pty_output` messages instead of a flood of tiny ones. No protocol change — JSON and binary transports both benefit.

## Why

The ingest-stage bisect (#296-era findings) showed the frontend memory balloon tracks **message count**, not byte volume: WebKit's per-message receive machinery retains memory in bmalloc that the scavenger barely returns while a pane is live. Binary frames (#298) attacked per-message *size* and bought only ~10%; this attacks per-message *count*.

A direct probe of real macOS PTY reads under an `awk` flood was decisive: reads return **~96 bytes** (about 2 lines), not the 16 KiB buffer size — so a 47 MB pane fill was ~580k messages end to end.

## How

- A reader goroutine feeds raw reads into a channel; `nextCoalescedRead` folds them into a batch **only when another read is already queued** behind the current one (the producer is outpacing the pipeline), bounded by 256 KiB / 5 ms.
- A lone read — which is all interactive traffic — is returned immediately: the echo path takes the `default` branch of a non-blocking select and adds **zero latency**.
- One batch = one seq, so attach/replay/dedup semantics are untouched. Covers both the worker and embedded PTY backends (both use `pty.Session`).

## UX verification (the stated concern)

- **Keystroke echo latency, measured end-to-end** through a real daemon PTY (spawn shell via WS, time `pty_input` → first `pty_output` frame, 20 keystrokes): **p50 0.8 ms, p90 1.1 ms, max 1.1 ms.**
- Worst-case added display latency during a flood is the 5 ms window — under one 120 Hz frame.
- `binary-pty-output-realpty` e2e passes (real PTY spawn → live output → renders).

## Memory result (packaged dev app, 2 panes × 1M ascii lines, same fill/machine/day)

| build | webContent RSS |
|---|---|
| pre-coalescing (this branch's parent) | 839.6 MB |
| coalescing, run 1 | **468.0 MB** |
| coalescing, run 2 | **456.9 MB** |

**−44% webContent / −370 MB total app RSS.** Flood probe: 119,201 raw reads → 58 batches (~161 KiB avg, 256 KiB max). Run-to-run spread also collapsed (±6 MB vs the historic ±50 MB — message count was the variance driver).

## Tests

- New unit tests for `nextCoalescedRead` (immediate lone-read path proven by construction — a timer wait would hang the test; batching, max-bytes bound, window bound, error folding).
- `go test ./...` green, `-race` on pty packages green, frontend untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-06-10]
 
 ### Changed
+- **Heavy terminal output uses much less memory.** Sustained output (long builds, tests, log floods) is now batched into far fewer, larger messages on its way to the app, cutting the app's memory growth during heavy streaming by roughly 40% and reducing CPU overhead. Typing and interactive prompts are unaffected — keystroke echo still arrives immediately.
 - **Streaming terminal output is cheaper.** Live terminal output now travels from the daemon to the app as compact binary messages instead of base64-encoded JSON, cutting per-chunk encode/decode work and message size by a third. This lowers CPU and bandwidth during sustained heavy output (busy agents, long builds) and modestly reduces the app's memory high-water.
 
 ---
--- a/internal/pty/session.go
+++ b/internal/pty/session.go
@@ -160,20 +160,90 @@ func (s *Session) fanOut(data []byte, seq uint32) {
 	}
 }
 
+// PTY reads are coalesced before fan-out so sustained output (builds, logs,
+// `seq`-style floods) produces few large downstream messages instead of one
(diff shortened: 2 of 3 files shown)


figgyster (approved):
This looks safe to merge. The coalescing path keeps lone reads on the immediate lane, the bounded flood path has focused coverage, and the daemon/PTY checks are green.

Tiny follow-up, not a blocker: if the 256 KiB cap is meant to be exact, `nextCoalescedRead` can overshoot by one read buffer before the loop notices. That’s harmless for this memory win, but worth tightening if the number becomes a hard contract.

No blockers from me.

PR #300, opened 2026-06-10
docs: drop deleted utility-terminal-realpty spec from AGENTS.md

victorarias:
AGENTS.md §6 (Terminal Focus Ownership) still told agents to run `app/e2e/utility-terminal-realpty.spec.ts`, which was deleted with the legacy non-session workspace panes in 340a8dff. No automated focus-ownership spec replaced it, so the verification steps are now explicitly manual.


figgyster (approved):
This is safe to merge. The removed realpty spec isn’t in the tree anymore, `AGENTS.md` now stops sending agents after a dead test path, and the remaining focus-ownership checks are honestly marked manual.

No blockers from me.

PR #301, opened 2026-06-10
feat(terminal): restore deep scrollback on remount and grow live caps

victorarias:
## Problem

Codex panes lose all scrollback on every remount (app restart, warm-set rehydrate). The attach replay budget was 256 KiB, and history beyond it wasn't clipped — it was **discarded**, degrading the restore to a bare screen snapshot (`segmented_raw_replay_exceeds_budget` → `use_fresh_snapshot`). Claude panes hid the same loss because that TUI re-renders its own transcript; Codex relies on terminal-native scrollback and exposed it. On top of that, the live-pane cap was an accident: `TERMINAL_SCROLLBACK_LINES = 50000` reaches libghostty's `max_scrollback`, which is **bytes** (ghostty-web passes it through unconverted), so live panes held ~1 MiB of stream bytes regardless of intent.

All of these caps were sized by a memory fear the perf effort has since disproven — scrollback was never the balloon (the libghostty model self-bounds; the cost was ingestion, fixed in #298/#299).

## Changes

1. **Serve the clipped tail instead of discarding it.** `LimitReplaySegmentsTail` now keeps whole segments only — each segment is a boundary-safe write, so the replayed tail never opens mid-escape-sequence — and the tail is still verified against the live screen snapshot (`rawReplaySegmentsMatchFreshSnapshot`) before being used. The correctness invariant is unchanged; only the "over budget → throw it all away" rule is gone.
2. **Replay log 1 MiB → lazy 8 MiB** (`DefaultReplayLogSize`), attach budget raised to match. The eager 1 MiB ring is untouched, so the per-worker idle floor stays flat; only sessions that actually emit history pay.
3. **`TERMINAL_SCROLLBACK_BYTES = 8 MiB`** — renamed and sized as the byte budget it factually is, with a comment documenting the ghostty-web unit pass-through.

## Verification

- **End-to-end against a real daemon PTY:** spawn shell → stream 3.3 MB → fresh client attaches with `same_app_remount` → **3.26 MB restored across 30 segments** (before: snapshot only, zero scrollback), `flat=0B snapshot=0B`.
- Memory cost measured on the packaged dev app (2 panes × 1M lines): webContent 484.7 MB vs 457–468 MB post-#299 baseline — **~12 MB per heavily-flooded pane**, bounded by the caps.
- New/updated unit tests: whole-segment tail semantics (drops partial oldest segment; drops all when newest exceeds limit), daemon test rewritten from "discard clipped codex replay" to "serve verified clipped tail", guard test now pins `DefaultReplayLogSize >= maxAgentRawReplayBytes`.
- `go test ./...` green, tsc clean, 680 vitest green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-06-10]
 
 ### Changed
+- **Terminals keep much more scrollback.** Live panes now hold roughly 8× more history, and switching back to a workspace (or restarting the app) restores up to 8 MB of terminal history instead of just the visible screen. This especially helps Codex sessions, which rely on the terminal's own scrollback and previously came back from a restart with no history at all.
 - **Heavy terminal output uses much less memory.** Sustained output (long builds, tests, log floods) is now batched into far fewer, larger messages on its way to the app, cutting the app's memory growth during heavy streaming by roughly 40% and reducing CPU overhead. Typing and interactive prompts are unaffected — keystroke echo still arrives immediately.
 - **Streaming terminal output is cheaper.** Live terminal output now travels from the daemon to the app as compact binary messages instead of base64-encoded JSON, cutting per-chunk encode/decode work and message size by a third. This lowers CPU and bandwidth during sustained heavy output (busy agents, long builds) and modestly reduces the app's memory high-water.
 
--- a/app/src/components/GhosttyTerminal.tsx
+++ b/app/src/components/GhosttyTerminal.tsx
@@ -23,7 +23,7 @@ import {
 } from '../utils/terminalSynchronizedOutput';
 import {
   FONT_FAMILY,
-  TERMINAL_SCROLLBACK_LINES,
+  TERMINAL_SCROLLBACK_BYTES,
   getTerminalAnsiPalette,
   getTerminalTheme,
   type ResolvedTheme,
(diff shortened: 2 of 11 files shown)


figgyster (commented):
This looks like the right fix: the replay tail now stays segment-boundary-safe, and it still has to reproduce the fresh daemon snapshot before the app trusts it. The live Ghostty cap rename is also the honest version of reality — bytes, not lines. Tiny mercy from the unit gremlins.

I ran `git diff --check`, the focused PTY/daemon replay tests, and `pnpm --dir app exec tsc --noEmit`; all passed locally. I’m leaving this as a comment rather than a green review only because the Frontend GitHub check is still pending on the PR and this touches the frontend terminal path.

No code blockers from me once that check goes green.

figgyster (dismissed):
Frontend finished green too, so the earlier caveat is gone.

This is safe to merge. The replay path keeps boundary-safe segments, verifies against the fresh screen before trusting history, and the byte-sized live cap matches what Ghostty actually does instead of what the name pretends it does.

No blockers from me.

figgyster (dismissed):
Codex is right here — my earlier approval missed the older trim path. The new attach-side whole-segment clipping only protects the *second* budget pass, but `ReplayLog.trimLocked` can already slice the retained head at an arbitrary byte offset. If that truncated log later matches the fresh screen, we can still raw-restore from a segment that starts mid-escape or mid-rune.

That crosses the terminal-restore correctness line, not a taste line.

Required before merge:
- Make the replay log itself preserve whole boundary-safe segments when trimming, or reject raw segmented restore whenever `info.ReplayTruncated` means the retained head may be partial.
- Add a regression that overfills `ReplayLog`, proves the retained first segment is still boundary-safe / whole, and covers the attach path so this doesn’t come back wearing a different hat.

figgyster (approved):
The follow-up fixes the blocker cleanly. The replay log now drops whole oldest segments instead of slicing, tracks truncation explicitly, and the attach regression proves the served restore segments still line up with original boundary-safe writes.

CI is green now, and the focused local checks passed too. Safe to merge — no blockers from me.

PR #302, opened 2026-06-10
feat(workspace): add area-map context and janitor

victorarias:
## Intent / Why

Workspace context currently frames a workspace as one goal and mostly captures the latest state. That is misleading when a workspace contains several related outcomes or inquiries, and it leaves new agents without the sourced story needed to interpret the current picture.

This changes context into an area map with semantic threads and a small timeline, then adds an infrequent attn-owned janitor so individual agents do not continually rewrite large context files.

## Summary

- Replace the singular `Goal` guidance with required `Area` and `Current Picture`, optional semantic `Threads`, and a sourced `Timeline`.
- Add an atomic Settings choice for the janitor agent and model, with recommended Codex and Claude presets plus a custom-model override; empty configuration keeps compaction disabled.
- Run Codex or Claude headlessly with only run-scoped `read_context` and `replace_context` tools.
- Let Claude use explicit API/cloud credentials in bare mode or normal OAuth, keychain, and organization-managed authentication while loading no user/project/local setting sources.
- Compact above 12 KiB after a ten-minute quiet period, with manual `compact` and `rollback` CLI commands.
- Protect publication with shape and size validation, revision compare-and-swap, cancellation barriers, and one retained pre-compaction backup.
- Increment the protocol to 98 and add schema migration 47 for janitor backups.

## Test plan

- [x] `go test -timeout 6m ./...`
- [x] `go test ./internal/agent ./internal/daemon`
- [x] `pnpm --dir app test -- --run` (88 files, 685 tests)
- [x] Settings modal visual checks at desktop, tablet, and mobile widths
- [x] Real Codex and Claude janitor smokes using isolated contexts: both completed the scoped MCP read/replace loop, removed duplicates, and preserved required facts and sources
- [x] Signed production build/install and live daemon capability check (`claude_cap_headless_task=true`)
- [x] Offline migration of four production contexts with database backup and integrity verification
- [x] Fresh-agent orientation checks against the attn and multi-threaded thunk contexts
- [x] Adversarial reviews completed with no remaining findings

## Out of scope

- Automatically refreshing already-running agents when another session publishes context
- Dynamic model discovery or per-workspace janitor policy
- Plugin-agent janitor support until the plugin driver exposes the same scoped headless capability


figgyster (approved):
This looks safe to merge. The area-map shape, janitor CAS/backup path, and scoped headless tool boundary all line up with the workspace-context contract instead of letting every agent rewrite the family Bible by hand.

Local checks passed:
- `git diff --check`
- `go test ./internal/store ./internal/daemon ./internal/client ./internal/protocol ./cmd/attn ./internal/agent ./internal/contextjanitor ./internal/hooks`
- `pnpm --dir app exec vitest run src/components/SettingsModal.test.tsx src/utils/workspaceContextJanitor.test.ts`
- `pnpm --dir app test -- --run` (88 files, 683 tests)

No blockers from me.

figgyster (approved):
Still safe to merge. The follow-up keeps the janitor setting easier to configure without changing the daemon-side compaction contract, and the focused checks are clean.

I’m still treating Codex’s large-MCP-request note as a worthwhile follow-up rather than a merge blocker; the bounded janitor path fails closed instead of corrupting context, which is the line that matters here.

No blockers from me.

figgyster (approved):
Still good. The Claude headless path now has a sane fallback for normal OAuth/keychain auth via safe mode while keeping bare mode for explicit cloud/API credentials, and the janitor still runs with only the scoped context tools. That’s the right trade: usable without turning the agent into a roaming houseguest.

Focused Go and Settings/type checks are clean. No blockers from me.

figgyster (approved):
Still safe to merge. The latest Claude path keeps the useful part of managed auth without dragging user/project settings into the janitor run, and the tests now pin that exact command shape instead of the earlier safe-mode shortcut that would have boxed out the MCP server.

Focused checks are clean. No blockers from me.

figgyster (approved):
Still good on the rebased head. The PR patch still preserves the area-map/janitor boundary, and the base scrollback replay changes don’t disturb the context compaction path.

Local focused checks passed, and the visible CI is green now. No blockers from me.

PR #303, opened 2026-06-10
fix(delegate): decouple worktree placement

victorarias:
## Intent

`attn delegate --worktree` currently treats filesystem isolation as an implicit request for a separate workspace. That breaks the natural request to delegate into a new worktree while keeping the new session visible beside the source session.

This change makes worktree creation independent from workspace placement: a worktree stays in the source workspace by default, while `--new-workspace --worktree` preserves the explicit separate-workspace flow.

## Summary

- make `--worktree` serialize `current_workspace` placement unless `--new-workspace` is also present
- create delegated worktrees before the placement-specific pane/workspace transaction so current and new workspaces share rollback behavior
- keep `--workspace <id> --worktree` unsupported with a clear error
- update CLI help, bundled attn delegation guidance, and the changelog
- add regression coverage for current-workspace placement, explicit new-workspace placement, workspace registration counts, unsupported combinations, and worktree rollback

## Test plan

- [x] `go test ./cmd/attn -run 'TestParseDelegateArgs|TestWriteDelegateHelp'`
- [x] `go test ./internal/daemon -run 'TestDelegate(CreatesWorktree|RollsBack.*Worktree|RejectsWorktree)'`
- [x] `go test ./internal/agent -run 'Test.*AttnSkill'`
- [x] `test -z "$(gofmt -l .)"`
- [x] `go vet ./...`
- [x] `go test ./...`
- [x] `go test ./internal/protocol -run TestFrontendProtocolVersionMatchesDaemon -count=1`
- [x] `pnpm --dir app test` (87 files, 680 tests)

## Out of scope

- Worktree delegation directly into an arbitrary existing workspace remains unsupported. `--workspace <id> --worktree ...` continues to fail clearly rather than introducing untested ownership and cleanup semantics.


figgyster (dismissed):
Codex caught a real compatibility break here: this changes the `delegate` command's wire semantics without bumping the daemon/client protocol version. A new CLI can now send `placement:"current_workspace"` with `worktree`, while the old daemon rejects that shape as an invalid current-workspace request.

That crosses the user-contract line because an app/CLI update can keep talking to a stale daemon and make `attn delegate --worktree ...` fail instead of forcing the expected protocol mismatch/restart path.

Required before merge:
- bump the protocol version surfaces for this command-semantic change (`internal/protocol/constants.go` and the app/native client constants that track it)
- keep the focused delegate tests passing after the bump

figgyster (approved):
This is clean now. The protocol bump closes the stale-daemon compatibility gap, and the worktree placement split is covered at the right daemon/CLI seams.

No blockers from me.

figgyster (approved):
Still clean after the rebase/follow-up. The delegate worktree behavior is unchanged, the protocol bump is still in place, and CI is green.

No blockers from me.

PR #304, opened 2026-06-10
fix(worktree): extend create provider timeout

victorarias:
## Intent

`worktree.create` providers can perform expensive repository setup and bootstrap work before returning. In the reported services-pilot failure, the provider completed the Git worktree side effect but crossed attn's shared 30-second deadline before returning, so attn reported failure for a worktree that already existed.

Creation needs a larger budget because retrying or falling back after a timed-out side effect is unsafe. Hooks and delete providers do not have the same evidence or workload, so they retain the existing shorter deadline.

## Summary

- give `worktree.create` providers a dedicated two-minute timeout
- keep before/after hooks and `worktree.delete` providers at 30 seconds
- add a short-duration regression seam and delayed-provider test that fails if create calls use the generic timeout
- document the user-visible reliability fix in the changelog

## Test plan

- [x] `go test ./internal/daemon -run 'Test(DispatchWorktreeCreateProvider_UsesCreateProviderTimeout|DoCreateWorktree_|DoCreateWorktreeFromBranch_|DoDeleteWorktree_Provider)' -count=1`
- [x] `go test ./internal/daemon -run 'Worktree' -count=1`
- [x] `$HOME/.codex/skills/autoreview/scripts/autoreview --mode local --stream-engine-output`
- [x] GitHub Actions full test suite

Local `go test ./internal/daemon -count=1` reached the package's 10-minute timeout in an existing order-dependent provider test after hundreds of earlier daemon tests. The named test passes in both the focused provider run and the broader worktree subset; CI is the authoritative full-suite check.


figgyster (approved):
This is safe to merge. The timeout split is scoped exactly where the side-effect risk lives: provider-owned `worktree.create` gets the larger budget, while hooks and delete stay on the short deadline.

Focused daemon worktree tests and `git diff --check` passed locally, and the visible daemon/changes CI is green. No blockers from me.

PR #306, opened 2026-06-11
fix(terminal): stabilize replay and resize lifecycle

victorarias:
## Intent

Opening or restoring long-running workspaces and resizing panes could freeze, blank, corrupt, or leave new Codex sessions at 80x24 until another resize. The failures came from synchronous replay load, Ghostty resize/reflow behavior, and attach/virtualization ordering.

## Summary

- lazily mount and detach cold workspace terminals while keeping daemon-side terminal query responses available
- bound synchronous raw replay to 64 KiB, restore at daemon geometry, and serialize queued live output before the final fit
- vendor a patched Ghostty VT core that preserves OSC 8 links, resize without history reflow, and coalesce split drags
- retain measured pre-attach geometry so new Codex sessions resize immediately after attach
- harden attach cancellation, remount sequencing, replay side effects, and callback-ref lifecycle races
- add packaged-app diagnostics and regression coverage for replay, resize, virtualization, and attach ordering

## Test plan

- [x] `pnpm --dir app test` (90 files, 708 tests)
- [x] `go test ./internal/daemon ./internal/pty`
- [x] production packaged-app verification: restored the existing thunk pane without forcing a resize
- [x] production packaged-app verification: a new Codex session reached 137x48 in both Ghostty and the daemon immediately after attach
- [x] autoreview completed with no actionable findings


figgyster (approved):
This looks safe to merge. The replay/resize lifecycle now has the right shape: daemon geometry stays authoritative during restore, cold panes detach cleanly instead of pretending they’re alive, and the Ghostty resize path avoids turning history into soup.

Local checks passed for me: `git diff --check`, `pnpm --dir app exec tsc --noEmit`, `pnpm --dir app test` (90 files / 708 tests), focused replay/resize/virtualization tests, and `go test ./internal/daemon ./internal/pty`. No blockers from me.

figgyster (approved):
Re-checked the follow-up push. The E2E helper now waits for the pane size hook before typing, which matches the attach/geometry race this PR is trying to kill rather than papering over it.

Local verification still passes: `git diff --check`, `pnpm --dir app exec tsc --noEmit`, `pnpm --dir app test` (90 files / 708 tests), and `go test ./internal/daemon ./internal/pty`. Frontend CI was still running when I posted, but the current head is safe from my side.

PR #307, opened 2026-06-12
docs: fix stale references in agent guidance files

victorarias:
Audit of the repo's CLAUDE.md/AGENTS.md guidance files, verifying every command, path, and claim against the current code.

## Changes

**`AGENTS.md` (root)**
- Frontend instrumentation examples now point at `terminalDiagnosticsLog.ts` / `terminalLinkHitTestLog.ts` (`paneRuntimeDebug.ts` and `terminalRuntimeLog.ts` were removed).
- ReviewComment fan-out no longer lists the deleted `internal/reviewer/mcp/tools.go`; remaining consumers are anchored to their real files.
- `needs_review_after_long_run` is documented as what it actually is: a boolean flag on the session (set after a 5m+ run, cleared on user view), not a value of the `SessionState` enum.

**`app/CLAUDE.md`**
- Unit-test section now describes the real architecture: `createMockDaemon()` prop factories from `src/test/mocks/daemon.ts`, not a nonexistent `MockDaemonProvider` wrapper.
- Removed-component `ReviewPanel` examples replaced with `DiffDetailPanel`; canonical example now points at `DiffDetailPanel.test.tsx`.
- Dropped dead link to `docs/plans/2026-01-02-frontend-testing-strategy.md` and brittle App.tsx line-number hints (the 4-step checklist stands on its own).
- All five Known Gotchas were verified still true and kept (including the worktree pending-action key collision).

**`app/src/components/NewSessionDialog/CLAUDE.md`**
- Rewritten (163 → ~55 lines). It documented the Tab-then-Enter bug as open with a "proposed fix" that has since shipped (`hasSelectedSinceTab` in `PathInput.tsx`). Now documents the keyboard behavior contract and the regression scenarios to guard.

**`internal/classifier/CLAUDE.md`**
- Converted from a byte-copy of its `AGENTS.md` into a symlink, matching `internal/daemon/` and `app/scripts/real-app-harness/`.

Docs-only; no code changes.


figgyster (dismissed):
This one breaks the guidance files rather than fixing them: `internal/classifier/CLAUDE.md` now points to `AGENTS.md`, but `internal/classifier/AGENTS.md` already points back to `CLAUDE.md`. That creates a symlink loop, so agents trying to read either file hit `Too many levels of symbolic links`.

Required before merge:
- Make exactly one of those files the real guidance file, and have the other symlink to it. Right now both are mirrors staring into each other. Very poetic, not very useful.

Codex flagged the same thing; I verified it in the PR checkout. I stopped after `git diff --check` plus the symlink/read check because heavier tests would just be theater for a docs-only blocker.

figgyster (approved):
The symlink loop is fixed: `internal/classifier/AGENTS.md` is a real guidance file again, and `CLAUDE.md` resolves cleanly to it. I also rechecked the docs cleanup and `git diff --check`; no blockers left from me.

Daemon/Changes/preflight are green; Frontend was still running when I posted, but this is docs-only and the actual broken guidance contract is fixed.

PR #309, opened 2026-06-12
feat(picker): folder-name recents with frecency ranking and Enter-to-open

victorarias:
## Problem

Since terminal sessions and session renames landed, the location picker's RECENT list shows session labels ("shell", manual renames) instead of identifying the *place* — see the repeated "shell" rows. The list also reshuffles on pure recency, and each worktree of a repo occupies its own slot.

## Changes

- **Remove `RecentLocation.label` end to end** (protocol → 98, store, daemon call sites, app). The picker derives the display name from the path's folder name, so session naming can never leak into the location list again. Migration 48 drops the column; **47 is an explicit no-op** because a pre-release build burned that version in existing DBs on 2026-06-10 (a real migration numbered 47 would be silently skipped there).
- **Frecency ranking**: `use_count × recency decay` (×4 <1h, ×2 <1d, ×0.5 <1w, ×0.25 beyond) instead of `last_seen DESC`, so main projects keep stable top slots.
- **Worktree collapse**: paths inside a linked worktree resolve to the main repo root (pure-FS walk via `git.GetMainRepoFromWorktree`, no subprocess) at write time; legacy rows merge at read time. Main-repo subdirectories deliberately do not collapse.
- **Pre-highlight top recent on open**: bare Enter opens your most-used location immediately. Typing/Tab consume the highlight (Enter then submits the typed path as before), arrows convert it to a manual highlight, and Esc still closes the dialog in one press.

## Testing

- 4 new store tests: use-count accumulation, frecency order (discriminates against pure recency), worktree collapse + legacy-row merge, subdir non-collapse.
- 5 new picker tests: basename rendering, bare-Enter open, Esc-during-auto-highlight closes, arrow→manual highlight, typing reclaims Enter.
- Full `go test ./...` and 685 frontend tests green.
- Verified live in the dev app (screenshot-checked: folder names render, top recent pre-highlighted, Enter lands on repo destinations) and migration 48 confirmed against real dev+prod DBs.

---
PR opened by Claude on behalf of Victor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-06-12]
 
+### Changed
+- **The new-session location picker is faster to use and easier to scan.** Recent locations now always show the folder name instead of session names like "shell" or manual renames, so the list reads as a list of places again. The list is ranked by how often *and* how recently you use each location, keeping your main projects in stable slots near the top, and all worktrees of a repository collapse into a single entry for its main checkout. The top recent location is pre-highlighted when the picker opens, so pressing Enter immediately opens your most-used project — typing or arrow keys take over as before, and Escape still closes the dialog right away.
+
 ### Fixed
 - **Resizing or splitting a pane while its history is still restoring no longer wipes the terminal.** A geometry change landing mid-restore used to silently discard the queued history, leaving the pane blank (after an app relaunch) or without scrollback (after a split). A resize that matches where the restore already ends now lets it finish, and a genuine size change re-requests the history at the new size — including the command blocks, which come back clickable.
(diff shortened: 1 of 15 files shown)


figgyster (approved):
This looks safe to merge. Deriving recent names from paths is the right fix for the “six rows named shell” problem, and the store-side frecency/worktree collapse has focused coverage for the tricky bits.

I ran the store/protocol focused tests plus the LocationPicker suite locally; GitHub isn't reporting checks on this branch yet from here, but I don't see a blocker.

figgyster (dismissed):
Codex caught two real regressions here, and I agree they cross the user-contract line for this picker change.

Required before merge:
- Rank recent locations before truncating them. The current DB path pulls only the 200 newest rows, then applies frecency, so an older high-use project can never surface even if it should be top-ranked.
- Gate the automatic recent highlight on the fresh recents request for the current open/endpoint. Right now reopening can auto-select cached recents from the previous target before the new request resolves, so bare Enter can launch the wrong local/remote path.

My earlier approval was for the previous head; this head needs these fixed before it’s safe to merge.

figgyster (approved):
This fixes the two real blockers: frecency now ranks across the full recent-location table before truncation, and cached recents can render but can’t drive the bare-Enter auto-highlight until the current open/endpoint’s request resolves.

Focused store/protocol tests and the LocationPicker suite are green locally. CI is still in flight from here, but the user-contract risk I blocked on is gone.

PR #311, opened 2026-06-13
fix(state): keep sessions green while background work runs

victorarias:
## Problem

When a Claude Code session hands off to a **background Workflow** (or background Bash) and pauses its turn, attn briefly mis-detected it as **`unknown`**. Reproduced from a real session (`cc21c08a`): the parent `Stop` hook fired immediately on launching the workflow, reporting a transcript path that **did not exist on disk yet**; stop-time classification then failed to open it (`ENOENT`) and fell back to `unknown`, where the session lingered ~5 minutes until an unrelated hook flipped it back to `working`.

## Root cause

Stop-time classification treats every `Stop` as terminal and depends solely on reading the main transcript JSONL. A `Stop` emitted while background work is still running is a **yield**, not a terminal stop — the turn auto-resumes when the work completes — but classification ran anyway against a not-yet-flushed transcript and latched `unknown`.

## Fix

Detect outstanding work via the **`background_tasks`** field on the Claude Code `Stop` payload. When a task is still `running`, `_hook-stop` reports `working` (the session stays green) and skips classification. The next `Stop`, once `background_tasks` drains, classifies normally.

- Level-triggered and reuses attn's existing "hook event → state" pattern — **no protocol/daemon change**.
- Covers background Bash too (those appear as nested `shell:running`).
- **Self-gating**: agents that don't emit `background_tasks` are unaffected.

## How it was verified

Captured a live Workflow run through an instrumented PTY. The parent `Stop` carries the signal exactly when it yields:

```
Stop  background_tasks=[{type:"workflow", status:"running"}]   ← yields, work still running
Stop  background_tasks=[]                                       ← work done → classify normally
```

`hasActiveBackgroundTask` is unit-tested against the real captured payloads (running / empty / absent / non-running). `go build ./...` and `go test ./cmd/attn/` pass.

## Known limitation / possible follow-up

Matches `status == "running"` only, and there's no TTL safety net if a final `background_tasks=[]` `Stop` never arrives. Fine for the observed behavior; flag if belt-and-suspenders is wanted.


figgyster (approved):
This looks safe to merge. The fix is nicely scoped: a Stop with live `background_tasks` now stays in the existing hook-state lane as `working`, and the normal stop classifier still owns the drained/terminal case.

I ran `git diff --check`, `go test ./cmd/attn/`, and `go build ./...` locally. No blockers from me.

figgyster (approved):
Re-approved after the merge-from-main churn. The actual PR patch is still the same focused Stop-hook fix, and the local checks are still clean.

I reran `git diff --check`, `go test ./cmd/attn/`, and `go build ./...`. CI is still running on GitHub, but I don’t see a code blocker here.

PR #313, opened 2026-06-13
feat(state): add 'scheduled' state for sessions parked on a /loop or cron

victorarias:
## What

Adds a new session state **`scheduled`** for when a Claude Code session has stopped its turn but is **parked on a schedule** — it will auto-resume later when a `/loop` or cron wakeup fires. Today such a session is mis-classified as `idle`/`unknown` because stop-time classification reads a transcript the agent hasn't flushed.

This is the sibling of the background-work fix (#311). Both ride the same `Stop` hook payload field set and short-circuit classification:

| Stop payload | Reported state |
|---|---|
| `background_tasks` has a `running` item | `working` (unchanged, #311) |
| else `session_crons` non-empty | **`scheduled`** (new) |
| else | normal classification |

Running work outranks a parked schedule (a Stop with both stays `working`).

## How it was verified

The `session_crons` shape and lifecycle were reverse-engineered and **confirmed live against Claude Code 2.1.177** (PTY hook capture + binary RE, which agreed exactly):

- Item shape is exactly `{id, schedule (raw 5-field cron, local TZ), recurring, prompt}` — **no per-item status, no resolved next-fire timestamp**.
- Detection is **presence-only** (`len(session_crons) > 0`). A fired one-shot or a `CronDelete`'d cron **drops out of the list entirely** on the next Stop (captured live — no stale/tombstoned residue), so presence is a clean, self-correcting signal.
- `session_crons` and `background_tasks` are independent and can co-occur on one Stop.

## Rendering

- Calm **sky-blue** (`#0ea5e9`), distinct from launching's periwinkle, the royal-blue PR/focus accent, and the unknown purple.
- **Slow pulse** (~3.2s breathing) on both the canvas grid tile and the sidebar/list dot — reads as "parked but alive", calmer than `waiting_input`'s flash.
- Its own **"Scheduled"** group in the dashboard (between Working and Idle).
- Rolls up at the **workspace level** just above `idle`.
- **Quiet**: excluded from attention (no badge, no drawer entry, not dispatch-wakeable) — a parked agent needs no steering.

Protocol bumped **101 → 102**.

## Tests

- `cmd/attn`: `TestHasPendingSessionCron` + `TestDecideStop` (precedence) using **verbatim captured 2.1.177 payloads**.
- `internal/daemon`: workspace rollup priority for `scheduled`.
- frontend: `normalizeSessionState`/`isAttentionSessionState`, `scheduledPulse`, and the dashboard `Scheduled` group.
- Full Go + frontend suites green; `tsc` clean (the `STATE_COLORS` Record + `stateBackgroundAlpha` switch are compile-time exhaustiveness nets).


figgyster (dismissed):
This still drops the new scheduled state during daemon recovery, which cuts across the user contract this PR is adding.

Required before merge:
- Preserve an existing `scheduled` session when recovery sees a live PTY worker, or teach the recovered-running state path to return `scheduled` when that is the persisted/runtime state.
- Add a regression that actually exercises recovery of a previously scheduled Claude session with `haveInfo == true`; the new socket/query test proves `UpdateState("scheduled")` works, but it doesn't cover the overwrite path.

Evidence: during recovery, existing live sessions with backend info are overwritten through `sessionStateFromRecoveredInfo` (`internal/daemon/daemon.go:960-970`). That helper delegates to `agentdriver.RecoveredRunningSessionState` (`internal/daemon/daemon.go:1147-1152`), whose default policy only preserves `waiting_input` and `pending_approval`; everything else becomes `launching` (`internal/agent/daemon_policy.go:66-74`). Claude uses that default policy. So a parked `/loop` session can show as scheduled, then an app/daemon restart turns it back into launching until the cron fires. That's exactly the interval this feature is meant to make legible.

figgyster (approved):
This looks safe now. The scheduled state is wired through the actual behavioral guardrails, not just the protocol/rendering layer: recovery preserves it, the transcript watcher leaves it alone, long-run tracking is cleared when a turn parks, and the grid pulse keeps animating.

I pushed a tiny gofmt-only follow-up after CI caught the daemon test formatting. Focused Go/frontend checks and `tsc` pass locally; GitHub Frontend CI was still running at my last poll, so let that finish before merge.

figgyster (approved):
Still safe to merge. The new follow-up closes the last noisy edge: Claude's live PTY detector can no longer knock a parked `scheduled` session back to idle/waiting unless it actually resumes work.

Focused scheduled-state Go tests, the frontend state/grid tests, `git diff --check`, and `tsc` all pass locally. Visible GitHub checks are green too. No blockers from me.

figgyster (approved):
Still clean. This last push is just a simpler shape for the same contract: the Stop hook now returns the non-terminal state directly, and the scheduled/launching alpha dedupe is pixel-identical.

Focused Go tests, grid/session-state frontend tests, `git diff --check`, `tsc`, and visible GitHub checks are green. No blockers from me.

figgyster (approved):
Still good after the main merge. The current PR patch is the same scheduled-state contract, and the only conflict-visible change is the changelog sitting beside the newer shortcut-editor entry.

Focused scheduled-state checks, `git diff --check`, `tsc`, and visible GitHub checks are green. No blockers from me.

PR #316, opened 2026-06-13
Parallel profiles: one knob, one authority for isolated attn installs

victorarias:
## What

Lets N agents run the **full** attn surface — Go, frontend e2e, and packaged real-app scenarios — **in parallel**, each fully isolated, by setting **one** knob: `ATTN_PROFILE`. A profile isolates the data dir, socket, websocket port, macOS app bundle, and bundle identifier. The whole lifecycle (pick · build · run · sign · clean) is streamlined, discoverable, and hard to misuse.

Before: three profile knobs (`ATTN_PROFILE`, `ATTN_HARNESS_PROFILE`, `ATTN_BUILD_PROFILE`) and four hand-synced derivation sites that could silently disagree. After: **one knob** (`ATTN_PROFILE`) + **one authority** (`attn profile resolve`) that every entrypoint reads instead of re-deriving.

## Commits (review in order)

1. **single profile authority + `attn profile` commands** — derivation lives in `internal/config`, surfaced by `attn profile status|resolve|list`.
2. **frontend e2e derives its ports from the active profile** — disjoint port bands; teardown kill scoped to its own port.
3. **real-app harness honors the one knob** — `ATTN_PROFILE` with an `ATTN_HARNESS_PROFILE` override; never targets prod by omission.
4. **per-profile app build + install** — `make install PROFILE=<name>`; bundle metadata generated from `attn profile tauri-config`, and port/bundle-id baked into the binary so a profiled app can never reach another profile's daemon.
5. **`attn profile clean`** — stop daemon, quit app, LaunchServices forget, remove app + data dir; refuses prod without `--force`; never signals a stale/recycled pid (gated behind the daemon's own advisory lock).
6. **docs** — `docs/profiles.md` is the reference; AGENTS.md updated; design plan marked implemented.

## Safety model

- **Prod is sacred.** Bare `make`/`make install`/`make install-daemon` refuse at parse time if `ATTN_PROFILE` is set.
- **Build matches your shell.** `make install PROFILE=<name>` refuses when `<name>` differs from the shell's `ATTN_PROFILE`; `make dev` is exempt (always the dev sibling, works from any shell).
- **The packaged app is bound to its build profile** — pins `ATTN_PROFILE`/`ATTN_WS_PORT`, strips inherited routing env at launch.
- **Single authority** — every entrypoint derives from `attn profile resolve`, so there is no per-entrypoint mapping to get wrong.

## Verification

Real end-to-end build (outside sandbox): `make install PROFILE=verify7` built/signed/installed `attn-verify7.app` (`com.attn.manager.verify7`, scheme `attn-verify7`, port **23684** baked into the Rust binary). Its daemon ran **alongside the prod (9849) and dev (29849) daemons with no collision**, every layer agreeing (authority → generated overlay → Info.plist → code signature → baked constants → live port). `attn profile clean verify7` then tore it down while prod and dev stayed up untouched.

PR4 + PR5 were also put through an adversarial multi-agent review; the 4 confirmed findings (stale-pid kill safety, `make dev` under a non-dev profile) are fixed in these commits with regression tests.


figgyster (changes requested):
This is close, but I’m blocking on the accidental committed temp tree.

Required before merge:
- Remove the `mktemp -d/` directory from the PR. It adds ~1.3k lines of copied harness code under a literal temp-command path, including stale-looking real-app helpers that are not part of the profile contract. That’s not harmless polish; it pollutes the repo and makes future code search/review lie with a straight face.

The actual profile authority shape looks coherent, and the focused checks I ran passed:
- `go test ./internal/config ./cmd/attn -run 'Test.*Profile|TestProfile' -count=1`
- `pnpm --dir app install --frozen-lockfile`
- `pnpm --dir app exec vitest run scripts/real-app-harness/common.test.mjs scripts/real-app-harness/harnessProfile.test.mjs`

One small follow-up while you’re there: `attn profile clean`’s pid-lock guard is much better than trusting a bare pid, but the code closes the lock file before reading/signaling. I wouldn’t hold this PR solely on that race, but if the promise is “never signals a recycled pid,” keep that path as tight as possible or soften the claim.

PR #317, opened 2026-06-13
Drag to manage workspaces: reorder, split-to-new, and drag sessions from the sidebar

victorarias:
## What

Four commits that make the sidebar a place to **organize** workspaces and sessions by drag, not just navigate them. Plan: `docs/plans/2026-06-13-drag-to-workspace.md`.

Three user-facing gestures + one behavior-preserving cleanup:

1. **Reorder workspaces** (`bb1a2532`, proto 102→103) — drag a workspace header onto an insertion seam. Order is backed by **fractional rank keys** (`internal/rankkey`): one row written per move, multi-client-safe over the websocket fan-out. New `rank` column via **migration 49**, backfilled in `created_at` order. Default order stays "opening order"; custom order persists across restarts.
2. **Split a session into a new workspace** (`909bcad3`, proto 103→104) — drag a session onto the **"New workspace"** drop-zone at the foot of the list. Daemon `handleWorkspaceLayoutMoveLeafToNewWorkspace` registers the new workspace (rank auto-seeded to the end, inherits source directory, title from the moved pane) then reuses `moveLeafToWorkspace` **verbatim**, so leaf relocation, session-ownership transfer, and empty-source teardown all behave like a normal move.
3. **Drag a session straight from the sidebar** (`150fec61`, frontend-only, no proto bump) — a session **row** (previously only its in-view pane) is now a drag source for both move-to-workspace and split. Reuses the App-level leaf-drag state + the existing drop targets; the only new code is the sidebar drag source (6px arm threshold, **no pointer capture** so the drop targets keep receiving pointer events) plus exposing `paneId` on the session child (the view model already computed it).
4. **De-dup verified-safe drag code** (`db6fb86e`) — refactor only, no behavior change: one shared leaf-drop placement const, dropped a redundant ref-mirror effect, folded a one-off daemon result helper into the generic one (byte-identical message), replaced hand-rolled test `itoa`/`max0` with `strconv`.

## Design notes for review

- **All rank math lives in Go.** The frontend sends the seam's neighbour workspace ids; the daemon computes the fractional key (`rankkey.Between`) and writes one row. Avoids a second fractional-index implementation in TS.
- **Gestures are distinct by drag source.** Workspace-header drag = reorder only (snaps to nearest seam, never merges). Session drag = merge into a workspace or split into a new one.
- **Protocol versioning.** Two bumps (103, 104). `generated.go`/`generated.ts` are generated, not hand-edited.

## Verification

- `go build ./...` + `go vet` clean; all Go tests pass (`rankkey`, `store`, `protocol`, `daemon`).
- `tsc --noEmit` clean; **920 frontend tests pass** (incl. new reorder, new-workspace-drop, and sidebar-session-drag specs).
- Migration 49 verified against real DBs (were at v48); rank backfill matches `rankkey.Seed`.
- Manually exercised all three gestures in the packaged dev app (`make dev`).


figgyster (approved):
This looks safe to merge. The rank-key ordering keeps the reorder write nicely narrow, and the split-to-new flow reuses the existing leaf move path instead of inventing a second relocation contract — exactly the boring kind of correct this feature needed.

I ran the focused backend/protocol/store checks plus the sidebar/socket/frontend type loop locally, and the visible CI is green now. No blockers from me.

PR #318, opened 2026-06-13
Notebook PR1 — core format + daemon service

victorarias:
## Notebook PR1 — core format + daemon service

First PR of the Notebook epic (durable, profile-wide, filesystem-canonical markdown memory). Targets the integration branch `feat/chifplace`. Plan: `docs/plans/2026-06-13-notebook.md`.

**Delivers a real, browsable, daemon-owned journal + memory store.** `attn notebook init` scaffolds it; `attn notebook list` / `show` browse it; the daemon owns every attn-originated write (atomic, hash-CAS, serialized) and broadcasts `notebook_changed`.

### What's here (review commit-by-commit)

1. **`feat(notebook)`: on-disk format + filesystem-canonical store** — the pure `internal/notebook` package (no daemon deps): YAML-frontmatter + markdown documents (two kinds `journal`/`memory`, root-absolute links, OKF-derived layout), and a `Store` with atomic temp+rename writes, hash-CAS edits, append-only dated journals, path-escape-guarded resolution, and a `.attn/`+dotfile-skipping `List`.
2. **`feat(protocol)`: notebook commands + `notebook_changed` event (v103)** — `notebook_init/list/read/write/append_journal` + the broadcast event; `ProtocolVersion` 102→103 (daemon + frontend hook); regenerated `generated.go`/`generated.ts`.
3. **`feat(notebook)`: daemon-owned service + handlers** — single in-process writer (caches one `Store` per resolved root), unix-socket handlers, hash-CAS-miss-as-success-conflict mapping, broadcast on every applied write. Integration tests drive the full socket path.
4. **`feat(notebook)`: client methods + `attn notebook init/show/list` CLI**.
5. **`docs`: changelog + scoping note.**
6–7. **`fix(notebook)`: self-review hardening** (see below).

### Filesystem-canonical (not `context.md`)

The `.md` files are the source of truth. Edits use **hash-CAS on file bytes**, not a SQLite revision table — so the store borrows workspace-context's atomic-writer / broadcast / locking shape but none of its SQLite-canonical / per-session-checkout model. Default root `~/attn-notebook[-profile]`, **outside** `~/.attn`, so it's a plain externally-syncable directory and a dev/test profile never writes the real notebook.

### Scoping refinement

External-edit **fsnotify reconcile** moved from PR1 → **PR3**, alongside its only consumer (the live UI). The daemon already emits `notebook_changed` on its own writes, so nothing observable is lost before then. Documented in the plan.

### Adversarial self-review (before opening)

Ran a multi-lens adversarial review (format / daemon / protocol / security), each finding independently verified against the real code. **12 confirmed → all addressed** in commits 6–7; 1 dismissed (CRLF — `Write` is verbatim). Highlights:

- **Symlink escape (medium):** `Store.abs` validated only lexically; a symlinked dir inside the root let reads/writes escape it (verified read-leak + CAS-overwrite). Now resolves symlinks against the resolved root (legit symlinked *root* still allowed); `EnsureScaffold` guarded too.
- **Round-trip fidelity (medium):** frontmatter scalars were coerced (`007`→`7`, `1.10`→`1.1`) and comments/order dropped on journal append — violating preserve-verbatim. A parsed document now re-emits frontmatter content byte-for-byte.
- **`notebook.root` unreachable (medium):** `validateSetting` rejected it as "unknown setting". Now validated (absolute, `~/` expanded, must be outside the data dir).
- **List prefix (low)** is now a path-segment boundary; **List (low)** reads only a bounded frontmatter prefix per file (perf); **`notebook_changed` (low)** broadcasts the normalized relative path.

### Testing

- Pure-package units: round-trip + byte-faithful preservation, permissiveness, links, path safety **incl. symlink-escape rejection + legit symlinked root**, hash-CAS, journal append preservation, list segment-boundary + bounded read.
- Daemon integration over a real socket pipe: init/list/read, hash-CAS conflict + CAS edit, append broadcast, normalized-path broadcast, `notebook.root` validation.
- Full Go suites pass (`notebook`, `protocol`, `client`, `cmd/attn`, full `daemon`).
- **Manual:** built a throwaway-profile daemon, exercised the full CLI (init idempotency, list, show, missing-file error/exit code, prefix filter), confirmed the on-disk layout, **verified a planted symlink-escape is rejected end-to-end**, then tore the profile down completely.


figgyster (dismissed):
This looks safe to merge. The notebook core keeps the right ownership line: markdown stays filesystem-canonical, writes go through the daemon with hash-CAS, and the socket/protocol path has real coverage instead of vibes in a trench coat.

I ran `git diff --check` and `go test ./...`; both passed. No blockers from me.

figgyster (changes requested):
Codex caught two real notebook boundary bugs here, and I reproduced both locally. This crosses the safety/contract line for PR1, so I’m flipping my earlier approval to a blocker instead of leaving green confetti on a wet floor.

Required before merge:
- Normalize/clean the configured notebook root before constructing the store. Today `notebook.root=/Users/me/attn-notebook/` is accepted but `Write("memory/foo.md", ...)` fails as an escape because `filepath.Join` collapses the trailing slash while the guard compares against `s.root + separator`.
- Make `List` skip symlinked `*.md` entries or run the same resolved-root guard before opening them. Right now a symlink under the notebook root can be followed by `readPrefix`, and `notebook_list` surfaces frontmatter from a file outside the notebook root.

I verified with temporary repro tests in `internal/notebook`: trailing-slash roots fail writes, and list follows/surfaces an outside symlink. The rest of the shape still looks right; these are focused fixes, not a rewrite.

PR #319, opened 2026-06-13
chore: remove stray 'mktemp -d' directory committed by mistake

victorarias:
## What

Removes the literal `mktemp -d/` directory accidentally committed to the repo root by PR #316 (parallel-profiles): an unexpanded `$(mktemp -d)` in a copy step (commit `ada8952f`) created a directory named `mktemp -d` containing stale duplicate snapshots of three real-app-harness files.

## Why it's safe

- The authoritative copies live at `app/scripts/real-app-harness/` and are **untouched** by this PR.
- `scenarioAssertions.mjs` in the stray dir is byte-identical to the real one; `common.mjs` and `harnessProfile.mjs` there are **older** snapshots.
- Nothing in the repo references the `mktemp -d/` path (`git grep` is empty).

Pure cleanup, no behavior change.


figgyster (approved):
This is exactly the kind of cleanup PR that should be boring. The diff only removes the stray committed `mktemp -d/` snapshot tree, `git diff --check` is clean, and the real harness files under `app/scripts/real-app-harness/` are still there.

No blockers from me. The garbage directory can leave quietly, unlike most garbage directories.

PR #321, opened 2026-06-13
fix(state): rollup unknown sessions and stop shells showing permanent green

victorarias:
## Why

Two reported symptoms, both real and independent:
- A workspace whose **single session** showed a different state than the workspace dot.
- **Shell sessions all showed a green ("working") dot**, regardless of what they were doing.

Diagnosed via a multi-agent audit (13 findings confirmed with state traces, 5 refuted). Root causes below.

## What

**1. Workspace rollup couldn't represent `unknown`.**
`rollupWorkspaceStatus` had no entry for `SessionStateUnknown`, so an unknown session was skipped and the workspace fell through to `idle`. A single-session workspace then showed a gray dot while its session showed purple. Added `WorkspaceStatus.unknown` (**proto 104→105**) and ranked it just below the steering states (`working > waiting_input > pending_approval > unknown > scheduled > idle > launching`) so it surfaces. The collapsed-rail mini-badge's previously-dead `status === 'unknown'` branch is now reachable.

**2. Stale workspace dot after a daemon restart.**
Startup recovery (`pruneSessionsWithoutPTY` / `reconcileSessionsWithWorkerBackend`) rewrites session states directly in the store without the per-session broadcast that refreshes the cached rollup, so the first `InitialState` after a respawn carried a pre-crash workspace status. Added `reseedWorkspaceStatuses()`, run at the end of both recovery branches (no broadcast — clients read it from `InitialState`).

**3. Shells showed permanent green.**
A shell was seeded `working` at spawn and never transitions while alive (no agent turn lifecycle, no Stop hook, no state detector). Now seeded `idle` — idle while alive, idle on exit. No classifier/detector added.

### Note
Adding `unknown` made `WorkspaceStatus` value-identical to `SessionState`, so quicktype merges the two enums in `generated.ts`; the **dead** `SessionState` re-export in `useDaemonSocket.ts` is removed (the frontend types session state via the local `UISessionState`). Type-safe.

## Tests
- `rollupWorkspaceStatus` unknown-priority cases (unknown beats idle/scheduled/launching; loses to pending/working; all-unknown → unknown).
- `TestWorkspaceSessionProtocolShellSpawnsIdleNotWorking` — spawned shell stores `idle` and its workspace rolls up `idle`.
- `TestDaemon_ReseedWorkspaceStatusesAfterRecovery` — stale cached rollup is repaired after prune.
- Go `daemon`/`protocol`/`store` green; frontend `tsc` + Sidebar/state tests green.

## Manual verification
Built the isolated dev app (proto 105, clean handshake) and spawned a fresh shell via the automation bridge → reports `"state": "idle"` (was `working` before).

## Not included (deliberate)
The two-event (`session_state_changed` / `workspace_state_changed`) drop hazard on a slow client. Every fix variant carries a cost the others don't (another protocol change, or a TS copy of the rollup that can drift from the Go authority) and it self-heals on the next state change/reconnect. Left out pending evidence it actually bites.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ### Fixed
 - **A session running work in the background no longer flips to a confusing "unknown" state.** When Claude Code hands off to a background workflow or background shell command and pauses its turn, the session now stays shown as working until the background work finishes and the turn resumes — instead of briefly being mis-detected as "unknown" because it paused before its transcript was written.
+- **A workspace's status dot now matches the sessions inside it.** A workspace whose only session was in the "unknown" state used to show a settled gray dot that disagreed with the session's own indicator; the unknown state now rolls up to the workspace so the two agree. The workspace dot also stays correct right after the app reopens — restored sessions no longer leave the workspace showing a stale status from before the restart.
+- **Shell sessions no longer show a permanent green "working" dot.** A plain terminal isn't running an agent turn, so it now starts (and stays) idle instead of appearing busy forever until you close it.
 
 ## [2026-06-12]
 
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -20,7 +20,6 @@ import type {
   ReviewLoopInteraction as GeneratedReviewLoopInteraction,
   WarningElement as GeneratedWarning,
   WorkspaceContext as GeneratedWorkspaceContext,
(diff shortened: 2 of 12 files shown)


figgyster (approved):
This looks safe to merge. The fix treats session state as the product signal it is: `unknown` now rolls up honestly, recovery reseeds the cached workspace dot before `InitialState`, and plain shells stop cosplaying as busy agents.

I ran the focused daemon/protocol/store checks plus frontend typecheck and sidebar/socket state tests locally. GitHub’s Frontend check was still pending on my last fetch, but the relevant local frontend path is green. No blockers from me.

PR #323, opened 2026-06-13
feat(terminal): render color emoji and Nerd Font icons

victorarias:
## Why

The custom WebGL terminal renderer drew **every** glyph as tinted alpha coverage, which caused two visible bugs:

1. **Color emoji collapsed to a flat one-color blob** — the shader only ever sampled the glyph atlas's *alpha* channel and repainted it in the cell's foreground color, discarding the color font's own RGB.
2. **Nerd Font icons (`eza --icons`, `starship`, powerline) were blank** — no icon font was bundled, so the Private Use Area codepoints had no glyph to rasterize.

## What

- **Bundle "Symbols Nerd Font Mono"** (`app/public/fonts/`, MIT), appended last in `FONT_FAMILY` so it only supplies PUA icon codepoints — all text still renders from Iosevka/Menlo. WKWebView won't use an `@font-face` face in a 2D canvas (the glyph raster surface) until it's also used in DOM layout, so `terminalIconFont.ts` force-activates it with a hidden probe span, then invalidates the glyph cache + repaints once `document.fonts.load` resolves.
- **Per-vertex `a_mode` + premultiplied-alpha pipeline** in both renderers: mode 0 tints atlas coverage with the cell fg (text/box/mono icons), mode 1 passes the atlas RGBA through (color emoji). Color-vs-mono is decided at raster time by scanning the bitmap for chromatic pixels — not codepoint ranges.
- **Extract `terminalGlyphProgram.ts`** (shared GLSL + `isColorGlyphBitmap`) so the two forked renderers — per-pane `GhosttyWebGlRenderer` and mission-control `UnifiedGridRenderer` — can't drift.

## Verification

- **Real packaged dev app** (native screenshot, since DOM capture can't read the WebGL canvas): color emoji render in full color and all four icon classes (folder/file/file-code/powerline) render.
- `tsc` + full `vite build` clean; **938 unit tests** pass (incl. new glyph-program + renderer parity tests); **123 e2e** pass.
- Frontend-only, no protocol bump.


figgyster (approved):
This looks safe to merge. The shared glyph program is the right move here: the per-pane renderer and grid now share the color-glyph path instead of slowly becoming two different bugs in matching coats.

I checked the font activation/cache invalidation path, the premultiplied-alpha shader/blend change, and the grid renderer parity. Local frontend validation is clean: `git diff --check`, `pnpm --dir app exec tsc --noEmit`, full `pnpm --dir app test` (933 passed / 5 skipped), and `pnpm --dir app run build`. GitHub Frontend CI was still running at my last fetch, but the same principal frontend suite/build passed locally.

No blockers from me.

PR #331, opened 2026-06-14
fix(daemon): scrub inherited agent session env from spawned sessions

victorarias:
## Problem

When attn is launched from inside a Claude Code session — most commonly an agent running `make install` — that session's per-session environment leaks into the long-lived daemon, and from there into every agent or shell the daemon spawns:

```
CLAUDE_CODE_SESSION_ID=cbcaa879-...
CLAUDE_CODE_CHILD_SESSION=1
CLAUDE_CODE_ENTRYPOINT=cli
CLAUDE_EFFORT=xhigh
CLAUDECODE=1
...
```

A nested agent that inherits `CLAUDE_CODE_SESSION_ID` reuses its parent's session identity, which breaks transcript generation (and other per-session behavior). The leak reaches spawned sessions through both the daemon's own `os.Environ()` base **and** the login-shell env capture (the login shell is spawned from the daemon's contaminated env and re-exports the vars).

## Fix

Scrub the leaked per-session vars from the processes that launch agents/shells, at the leak carrier — not at each spawn site.

The set is split into two:

- **identity vars** (`CLAUDE_CODE_SESSION_ID`, `CLAUDE_CODE_CHILD_SESSION`, `CLAUDE_CODE_ENTRYPOINT`, `CLAUDE_CODE_EXECPATH`, `CLAUDE_CODE_SSE_PORT`, `CLAUDECODE`) — pure parent-session identity, never user config → always safe to scrub on any launch path.
- **tuning vars** (`CLAUDE_EFFORT`, `CLAUDE_CODE_NO_FLICKER`, `CLAUDE_CODE_AUTO_COMPACT_WINDOW`, `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS`) — per-session runtime knobs a user might deliberately export in their profile.

Scrub points:

- **Daemon + PTY worker startup** (`runDaemon` before `Start()`, `runPTYWorker`): scrub the **full** set. Doing it at the process level — before the login shell is captured — keeps the captured env clean while letting the login shell re-export whatever the user genuinely sets in their profile (e.g. `CLAUDE_CODE_USE_BEDROCK`, or a deliberately-exported `CLAUDE_CODE_NO_FLICKER`). So real config survives; only the leaked per-session values are dropped.
- **Foreground `attn` launch** (`runAgentDirectly`): scrub **identity only**. This path inherits the live shell env directly with no login-shell re-capture, so scrubbing tuning vars would drop user config that nothing restores. Identity-only kills the session-id leak with zero common-case regression (identity vars are absent in a normal terminal).

Files:

- `internal/config/agentenv.go` — the two var sets + `ScrubInheritedAgentSessionEnv()` / `ScrubAgentSessionIdentityEnv()`, with the rationale documented.
- `internal/daemon/daemon.go` — `d.ScrubInheritedAgentSessionEnv()`, called in `runDaemon()` before `Start()` warms the login-shell cache.
- `cmd/attn/main.go` — full scrub in `runDaemon`/`runPTYWorker`; identity scrub in the foreground `runAgentDirectly`.

## Tests

- `internal/config`: full scrub removes identity + tuning and **preserves** `ANTHROPIC_API_KEY` / `CLAUDE_CODE_USE_BEDROCK`; identity-only scrub removes identity but **keeps** tuning vars (`CLAUDE_CODE_NO_FLICKER`, `CLAUDE_EFFORT`) — the regression guard for the foreground path.
- `internal/pty`: hermetic integration test — a scrubbed process env feeds the real `buildSpawnEnv`, and no `CLAUDE_CODE_SESSION_ID` reaches the spawned PTY (shell + agent).

## Verification (live)

Ran a real daemon on an isolated profile, contaminated exactly as in the bug:

- Daemon log: `scrubbed inherited agent session env before startup: [CLAUDECODE CLAUDE_CODE_ENTRYPOINT CLAUDE_CODE_SESSION_ID …]` — caught all present vars; login-shell prewarm ran **after** the scrub.
- fish login re-export, contaminated vs scrubbed parent: contaminated re-exports `CLAUDE_CODE_SESSION_ID=…` (+8 more); scrubbed re-exports only `CLAUDE_CODE_NO_FLICKER=1` (a real fish-config setting) — proving session identity is dropped while genuine profile config is preserved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-06-14]
+
+### Fixed
+- **Agents and shells launched by attn no longer inherit a stale Claude Code session identity.** When attn itself was started from inside a Claude Code session (for example an agent running `make install`), that session's per-session environment — including its session ID — leaked into the daemon and then into every agent or shell attn spawned, which broke things like transcript generation. attn now drops those inherited per-session variables on startup, while still honoring any Claude Code settings you've configured in your shell profile.
+
 ## [2026-06-13]
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-06-14]
 
 ### Fixed
-- **Agents and shells launched by attn no longer inherit a stale Claude Code session identity.** When attn itself was started from inside a Claude Code session (for example an agent running `make install`), that session's per-session environment — including its session ID — leaked into the daemon and then into every agent or shell attn spawned, which broke things like transcript generation. attn now drops those inherited per-session variables on startup, while still honoring any Claude Code settings you've configured in your shell profile.
+- **Agents and shells launched by attn no longer inherit a stale Claude Code session identity.** When attn itself was started from inside a Claude Code session (for example an agent running `make install`), that session's per-session environment — including its session ID — leaked into the daemon and then into every agent or shell attn spawned, which broke things like transcript generation. attn now drops those inherited per-session identifiers before launching agents and shells, while still honoring any Claude Code settings you've configured in your shell profile.
 
 ## [2026-06-13]
 
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -397,6 +397,13 @@ func runPTYWorker() {
 		fmt.Fprintf(os.Stderr, "[pty-worker] "+format+"\n", args...)
 	}
 
+	// Belt-and-suspenders: the daemon already scrubs these before spawning the
+	// worker, but a worker spawned from any unscrubbed parent self-protects so
+	// the leaked per-session agent env never reaches the PTY it spawns.
+	if scrubbed := config.ScrubInheritedAgentSessionEnv(); len(scrubbed) > 0 {
+		cfg.Logf("scrubbed inherited agent session env before startup: %v", scrubbed)
+	}
+
 	if err := ptyworker.Run(context.Background(), cfg); err != nil {
 		fmt.Fprintf(os.Stderr, "pty-worker error: %v\n", err)
 		os.Exit(1)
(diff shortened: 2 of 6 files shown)


figgyster (dismissed):
This is safe to merge. Scrubbing at daemon/worker startup hits the actual leak carriers before the login-shell cache gets warmed, and the tests cover both the scrub list and the PTY spawn-env path without eating legitimate Claude config.

Local checks passed: `git diff --check` and `go test ./internal/config ./internal/pty ./cmd/attn -count=1`. CI is green too. No blockers from me.

figgyster (dismissed):
Codex is right here: the worker startup scrub only removes real process env vars, but a stale daemon can still pass a contaminated cached login-shell env through `ATTN_CACHED_SHELL_ENV`. `buildSpawnEnv` then merges that cached env and only filters `CLAUDECODE`, so `CLAUDE_CODE_SESSION_ID` can still reach spawned PTYs in the upgrade/defense-in-depth path.

Required before merge:
- Filter the full inherited agent-session scrub list after cached shell env merge as well, or otherwise make `ATTN_CACHED_SHELL_ENV` go through the same scrub boundary.
- Add a regression test with `ATTN_CACHED_SHELL_ENV` containing `CLAUDE_CODE_SESSION_ID` so this exact alley gremlin stays dead.

I re-ran the focused local checks (`git diff --check`, `go test ./internal/config ./internal/pty ./cmd/attn -count=1`); the existing tests pass but do not cover the cached-shell-env leak.

figgyster (changes requested):
The new foreground-wrapper scrub is useful, but it doesn’t address the blocker from the previous review.

`buildSpawnEnv` still merges `readCachedShellEnvFromProcess()` and then only filters `CLAUDECODE`; a stale daemon can still hand the worker an `ATTN_CACHED_SHELL_ENV` containing `CLAUDE_CODE_SESSION_ID`, and that value still reaches the spawned PTY. The old red flag just put on a cleaner coat.

Required before merge:
- Scrub the inherited agent-session keys from the cached shell env after it is merged into the PTY spawn env, or scrub that cached env before passing it to the worker.
- Add the regression test that seeds `ATTN_CACHED_SHELL_ENV` with `CLAUDE_CODE_SESSION_ID` and proves `buildSpawnEnv` drops it.

Focused local checks still pass (`git diff --check`, `go test ./internal/config ./internal/pty ./cmd/attn -count=1`), but they still don’t cover the cached-shell-env path.

PR #332, opened 2026-06-14
fix(terminal): render multi-codepoint emoji clusters as one glyph

victorarias:
## What

Multi-codepoint emoji grapheme clusters — ZWJ families (👨‍👩‍👧‍👦), regional-indicator flags (🇺🇸), skin-tone variants (👍🏽), and keycaps (1️⃣) — were decomposing into their component glyphs in the WebGL terminal (four separate heads, two boxed flag letters, a thumb beside a colour swatch). Single emoji and VS16 emoji (❤️) were never affected.

## Root cause

The colour-emoji pipeline (#323) was fine. Two separate things had to be true at once, and one was being silently broken:

1. **The model must keep the whole cluster in one cell.** ghostty-web does this only with DEC mode **2027** (Unicode grapheme clustering) enabled. It starts on, but a **RIS (`ESC c`, full reset)** from the shell at startup silently turns it **off** — after which the model splits every cluster across cells, so attn's per-cell canvas renderer never receives the whole cluster. (Soft reset / DECSTR does not affect it.)
2. **Canvas `fillText` must shape the cluster as one Apple Color Emoji run**, or the ligature never forms even with the full string.

## Fix

- **`terminalGraphemeMode.ts`** (new) — re-asserts `ESC[?2027h` at model creation and after any output chunk where `getMode(2027)` is false. Wired into `GhosttyTerminal.tsx`.
- **`terminalGlyphFont.ts`** — prepends `"Apple Color Emoji"` to the canvas font for emoji clusters only (conservative classifier; leaves text, box-drawing, and bundled Nerd Font PUA icons on the normal text-first family). Used by both `GhosttyWebGlRenderer` and `UnifiedGridRenderer`.

## Verification

- Headless probes pinned the mechanism (RIS toggles 2027; `getMode` tracks it).
- Live dev app, clean production build: all eight test classes render as single composed colour glyphs (family-of-four, US flag, thumb+skin-tone, keycap with VS16 restored, rainbow flag, grin, both hearts).
- `tsc` clean; 948 unit tests pass incl. new `terminalGraphemeMode.test.ts` and `terminalGlyphFont.test.ts`.


figgyster (dismissed):
This still has a same-chunk reset hole, so I can’t sign off yet.

`ensureGraphemeClustering()` runs only after the whole output chunk has already been written into the Ghostty model. If that chunk contains `ESC c` followed by a prompt/status line with a multi-codepoint emoji, the RIS turns DEC 2027 off and the emoji bytes later in that same chunk are parsed into split cells before the re-enable happens. Reasserting 2027 after the write only protects later chunks; it can’t recombine cells already stored in the model.

Required before merge:
- handle RIS before subsequent bytes in the same PTY chunk are parsed — e.g. segment around full-reset bytes and re-enable 2027 immediately after the reset segment, before writing the rest of the chunk
- add a regression that feeds `ESC c` plus a clustered emoji in one write/chunk shape, not just “mode was already off before ensure”

The font-routing part looks good, and `git diff --check`, the focused Vitest files, and `pnpm --dir app exec tsc --noEmit` all passed locally. This one is just the model timing rake in the alley.

figgyster (dismissed):
This fixes the blocker. The reset path now re-enables grapheme clustering at the model-write boundary, before any bytes after `ESC c` can parse into split cells, and the regression covers same-chunk plus boundary-straddling RIS cases.

No blockers from me. Local focused checks and TypeScript passed; GitHub checks were not reported for this branch when I fetched them.

figgyster (approved):
Still good after the main merge. The PR patch is the same terminal grapheme fix, and the focused checks still pass on the rebased head.

No blockers from me. GitHub Frontend CI was still pending at my last fetch; local `git diff --check`, focused Vitest, and TypeScript passed.

PR #333, opened 2026-06-14
fix(workspaces): repair startup restoration

victorarias:
## Intent / Why

attn could lose all visible workspaces after startup, open to a black screen, or leave new workspace creation spinning until timeout. A development build had reused migration version 49, so newer code treated the database as migrated even though the required `workspaces.rank` column was absent. That made workspace loads and writes fail, which also exposed brittle startup ownership handling and an uncorrelated action error.

## Summary

- verify and repair the required workspace-rank schema on every database open, independent of recorded migration versions
- backfill stable workspace ordering for existing rows and cover databases reporting versions 49, 50, and a higher development version
- preserve persisted session workspace ownership while the daemon registry is recovering
- recover transiently missing ownership from the authoritative saved workspace layout
- ignore a truly orphaned session for one render instead of unmounting the entire app
- correlate workspace pane setup failures so the client receives the error immediately instead of timing out
- document the user-visible restoration and creation fixes in `CHANGELOG.md`

## Test plan

- [x] `go test ./internal/store ./internal/daemon`
- [x] `pnpm --dir app test` (105 files, 955 tests)
- [x] build and launch the signed dev app with `make dev`
- [x] reproduce the original production daemon snapshot in the frontend and confirm it renders without the workspace-ownership exception
- [x] build, install, and launch the signed production app with `make`; 10 persisted PTYs recovered successfully
- [x] verify the live production database has 6 workspaces, 0 empty ranks, and all 10 sessions retain workspace ownership
- [x] register and unregister a temporary workspace through the production websocket API


figgyster (dismissed):
This looks safe to merge. The fix preserves the daemon’s persisted workspace ownership during startup recovery, and the frontend fallback is the right kind of defensive: recover from the authoritative layout when possible, otherwise drop only the orphaned session for that render instead of taking the whole app down with it.

Validated locally with the focused daemon regression, the full app Vitest suite, and `git diff --check`; GitHub CI is green too. No blockers from me.

figgyster (approved):
Still safe to merge. The follow-up closes the schema-drift hole cleanly: `OpenDB` repairs the workspace rank column regardless of recorded migration version, and the layout action failure now correlates back to the requested pane instead of making the UI wait around like a fool in the rain.

Focused store/daemon regressions pass locally, `git diff --check` is clean, and GitHub CI is green. No blockers.

PR #334, opened 2026-06-14
refactor(store): limit rank repair to migration

victorarias:
## Intent / Why

Published attn releases through `v0.11.1` stop at migration 41, so supported release upgrades always execute migration 49 and create `workspaces.rank` normally. The unconditional schema repair added in #333 was broader than the release compatibility contract and performed unnecessary work on every database open.

## Summary

- remove the post-migration workspace-rank schema check from every `OpenDB` call
- retain migration 50 as a one-time repair and reserved migration number for the known version-49 development-build collision
- narrow regression coverage to that one-time migration repair

## Test plan

- [x] `go test ./internal/store`
- [x] `git diff --check`

## Related

- #333

### DIFF
--- a/internal/store/sqlite.go
+++ b/internal/store/sqlite.go
@@ -433,10 +433,6 @@ func OpenDB(dbPath string) (*sql.DB, error) {
 		db.Close()
 		return nil, err
 	}
-	if err := ensureWorkspaceRankSchema(db); err != nil {
-		db.Close()
-		return nil, err
-	}
 
 	return db, nil
 }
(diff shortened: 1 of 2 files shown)


figgyster (dismissed):
This looks safe to merge. Moving the rank repair back into the migration path matches the stated release contract, and the migration stays idempotent for both the normal 49→50 path and the dev-build collision case.

`go test ./internal/store` and `git diff --check` pass locally; visible daemon/release checks are green too. No blockers from me.

figgyster (changes requested):
Codex caught a real migration edge here, and I can reproduce it: a DB with `schema_migrations` already at 50 but missing `workspaces.rank` now opens without running the repair, leaving the column absent.

That crosses the storage-compatibility line rather than taste. Required before merge:
- keep a one-time repair path for the already-current/missing-rank shape, or add a later migration that covers it
- keep a regression test for version 50-or-higher recorded without `workspaces.rank`

The narrower release-upgrade contract is fine; the broken already-current dev-build shape is the little rake in the alley.

PR #335, opened 2026-06-14
feat: durable task engine + agentic workspace-context janitor

victorarias:
## What & why

The foundation for the notebook-narration epic (plan: `docs/plans/2026-06-14-notebook-narration.md`): a general durable task engine, and the workspace-context janitor migrated onto it and converted from a constrained MCP tool pair to a native-tools headless agent.

Net **−243 lines** — this deletes more than it adds: the constrained `read_context`/`replace_context` MCP path and all of `internal/contextjanitor` are gone (the janitor was their only consumer), so `RunHeadlessTask` is native-tools-only now.

## Commits

1. **`docs(notebook)`** — the epic plan doc (canonical spec).
2. **`feat(tasks)`** — `internal/tasks`: a daemon-agnostic, file-backed durable runner. One atomic JSON file per task under `<notebook-root>/.attn/tasks/` (no SQLite — the burned-migration gotcha would silently skip a new table; no per-task lock — a single worker serializes). State machine `queued→running→done`; `running→failed→queued` (capped-exp backoff) `→dead` at the attempt cap; manual `Retry`. Subject-derived coalescing enqueue with forward-push debounce + zero-debounce override. Per-kind `ExecutorFunc` registry; the runner owns the `context.WithTimeout` wrapper. `Cancel(id)` blocks until the run goroutine exits; an executor-owned `CommitGuard` fences the single durable write against both `Cancel` and the deadline so it is never torn. Orphan-running recovery + a single-instance lock on start; `Remove` (cancel-then-delete) so a removed subject leaves no record behind. Disabled cleanly when no notebook root resolves.
3. **`feat(daemon)`** — native-tools headless mode + the agentic janitor. Compaction is enqueued on the runner as `compact_context` — **durable across restarts and auto-retried on failure, closing the janitor's two old gaps (lost-on-restart, no-retry)**. The executor writes the context to a scratch dir, the agent compacts it with its own file tools, the daemon reads the candidate back, validates, and commits under the `CommitGuard`. One `applyWorkspaceContextCompaction` path backs the runner executor, the inline fallback (no notebook root), and the manual command, all wrapped by the per-run timeout. Every workspace-teardown site cancels-and-forgets the compaction before `RemoveWorkspace` (nil-safe: the WS server accepts connections before the runner is constructed). Validate + revision-guarded apply + config are unchanged.

## How it was built

Two adversarially-reviewed multi-agent workflows (build → 3-lens skeptical concurrency review → fix). The reviews caught and fixed real bugs with mutation-verified regression tests: a corrupt-record worker wedge, a `Cancel`-returns-before-the-durable-write fidelity break, a concurrent-`Stop` panic, a timeout that could tear a commit, a teardown nil-deref window, and the manual `workspace context compact` command silently losing its 5-minute timeout.

## Testing

- `go build ./...` and `go vet ./...` clean.
- `internal/tasks` and `internal/agent` green under `-race`; full `internal/daemon` suite green.
- Agentic-task quality (the actual native-tools spawn compacting a real context) is the plan's harness/manual bar — the deterministic seams (enqueue/cancel/validate/apply/fallback/timeout) are unit-tested with a fake executor; **no real LLM is spawned in tests**.

## Scope

No protocol change (ProtocolVersion stays 109). Narration tasks (`summarize_session`/`narrate_workspace`), the raw tier, the status surface, and the dreaming fold-in are **not** in this PR — they stack on top per the plan.


figgyster (changes requested):
There’s one cancellation race in the new runner that crosses the safety line.

`runNext` persists the task as `running` and releases `ioMu` before `execute` registers `r.run`. In that gap, `Cancel(id)` sees no active run and returns immediately. `Remove(id)` inherits the same miss because it calls `Cancel(id)` first and then deletes the record.

That means a workspace teardown can delete the task record while the just-claimed executor is still about to start, without ever fencing/canceling it. For `compact_context`, that can let the native-tools janitor still perform its durable workspace-context write after the subject was removed. The later `finish()` no-op on missing task record doesn’t protect the executor-owned durable side effect.

Required before merge:
- close the claim/register gap so a task that is persisted as running is also cancellable/removable before the executor can start, or make `Cancel`/`Remove` handle the claimed-but-not-yet-registered state explicitly.
- add a regression test for cancellation/removal in that window. The current focused tests pass, but they don’t pin this race.

I ran `go test ./internal/tasks ./internal/agent ./internal/daemon`; all green. The bug is in the interleaving, not the happy path. The alley always saves its rakes for timing windows.

PR #336, opened 2026-06-15
feat(notebook): deterministic raw-tier capture floor

victorarias:
## What & why

The deterministic **capture floor** for the notebook-narration epic (plan: `docs/plans/2026-06-14-notebook-narration.md`), stacked on the foundation PR (#335 → `feat/tasks-runner`).

The narration pass needs a reliable, machine-written source of "what happened" that is *not* the curated journal. This PR establishes that **raw tier** under `<notebook-root>/.attn/raw/` and routes the daemon's two deterministic captures onto it — so the journal a human reads stays curated, and no workspace state is silently lost on teardown.

## What's in it

- **Raw-tier paths** (`internal/notebook/raw_tier.go`): `RawDir` / `RawContextSnapshotsDir` / `RawDispatchesDir`. The tier lives under `.attn/`, which `CleanPath` rejects and the watcher skips — so it's written with direct filesystem I/O, never `notebook.Store`, and emits **no** `notebook_changed` broadcast (no watcher echo to suppress).

- **Dispatch capture redirected** to `.attn/raw/dispatches/<id>.md` instead of `journal/<date>.md`. One file per dispatch id is the exactly-once ledger; a replayed trigger (report path / session-gone fallback / restart) re-renders the block and atomically overwrites that one file.

- **Workspace context snapshot on removal** → `.attn/raw/context-snapshots/<wsID>.md`, taken synchronously at **every** removal site in the order `forget-compaction → snapshot → RemoveWorkspace`, before the `DELETE FROM workspace_contexts` erases it (an async writer can't win that race). Best-effort: never blocks or fails a teardown; an empty/never-overlaid context is a silent no-op.

## Security

The workspace id is **client-controlled** (`register_workspace` accepts `msg.ID` verbatim — no UUID/segment validation). Both raw writers funnel through a single `rawTierFilename` guard + `writeRawAtomic` chokepoint that:
- rejects any id that isn't one safe path segment — `/`, `\`, `.`, `..`, leading-dot, and **control chars** (a newline would otherwise inject a forged `source:` footer line into the file body), and
- asserts the final path stays under the intended subdir.

Without this, a crafted id like `../../../journal/2026-06-15` would resolve out of `context-snapshots/` and **overwrite the curated journal** (the load-bearing invariant: the daemon must never write the curated journal). Regression-tested with four escape vectors (`TestSnapshotWorkspaceContextRejectsPathTraversal`) and the unsafe-id + control-char set (`TestWriteRawAtomicRejectsUnsafeID`).

## How it was built

An adversarially-reviewed multi-agent workflow (build → 3-lens skeptical concurrency/security review → fix). The review caught the path-traversal-into-the-journal escape; the control-char hardening + its regression case were added on top.

## Testing

- `go build ./...` and `go vet` clean.
- `internal/daemon` + `internal/notebook` suites green (snapshot at all 4 removal sites, load-time reap, empty/whitespace no-op, replay-is-identical-overwrite, forged-marker neutralization, wall-clock-header drift, 4 path-traversal vectors, unsafe-id/control-char rejection).

## Follow-up (not in this PR)

`notebook.Store.AppendJournalEntryOnce` now has no production caller (the dispatch path was its only one) — it writes the curated journal, so it's a caller-less footgun worth deleting in a later cleanup. Left out here to keep this PR's scope to the capture floor.

## Scope

No protocol change (ProtocolVersion stays 109). Narration tasks, the status surface, and the dreaming fold-in stack on top per the plan.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-06-15]
+
+### Fixed
+- **A workspace can no longer corrupt your journal or other files when it closes.** The context snapshot taken when a workspace is removed now refuses any workspace identifier that is not a plain name, so a crafted identifier can no longer steer that write out of attn's internal holding area and overwrite your curated journal or another file. Normal workspaces are unaffected.
+
+### Changed
+- **Your curated journal stays curated; raw delegated-work outcomes now feed it behind the scenes.** Auto-captured chief-of-staff dispatch outcomes no longer land directly in your daily `journal/<date>.md`. They are now recorded to an internal holding area the upcoming narration pass reads from, so the journal you read keeps only curated entries instead of machine-raw blocks. Dispatch outcomes are still captured reliably and exactly once; existing journal entries are left untouched.
+- **A workspace's shared context is preserved when the workspace goes away.** Removing a workspace used to discard its shared `context.md` overlay entirely. attn now snapshots that context the moment a workspace is torn down, keeping it as durable raw material for the journal-narration pass — so the decisions and current picture an agent recorded in a workspace are no longer lost when the workspace closes.
+
 ## [2026-06-14]
 
 ### Added
--- a/internal/daemon/notebook_dispatch_journal.go
+++ b/internal/daemon/notebook_dispatch_journal.go
@@ -9,27 +9,35 @@ import (
 	"github.com/victorarias/attn/internal/protocol"
 )
 
-// Auto-journaling of chief-of-staff dispatch outcomes.
+// Deterministic capture of chief-of-staff dispatch outcomes to the raw tier.
 //
 // A dispatch is delegated agent work, and its terminal report (completed/failed)
 // — or, as a fallback, its session simply ending — is exactly the "what was
 // decided / built / failed" signal the durable work-journal exists to capture.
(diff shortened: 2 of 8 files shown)


figgyster (dismissed):
This looks safe to merge. The raw-tier split keeps machine-written capture out of the curated journal, and the security guard is in the right chokepoint: one safe-segment filename path for both client-controlled workspace IDs and dispatch IDs.

`go test ./internal/daemon ./internal/notebook` and `go test ./...` both pass locally. No blockers from me.

figgyster (changes requested):
Codex caught a real contract gap here, and I agree with it after checking the current head.

The raw dispatch file is meant to be the durable machine-written source for the narration pass, but it only records `source: dispatch:<id>` while the file itself lives in one global `.attn/raw/dispatches/` directory. `ChiefOfStaffDispatch` already has `WorkspaceID`, and the current dreaming code uses that store field as the dispatch context — but this PR’s raw-tier contract is specifically about surviving teardown / later narration from files. Once the later pass is reading raw files, the dispatch file needs enough provenance to route it back to the workspace without relying on live store mapping.

Required before merge:
- include workspace/session provenance in the rendered raw dispatch file, or store dispatch raw files under a workspace-scoped path;
- add a regression test showing two workspaces’ dispatch captures remain distinguishable from raw files alone.

This is not taste; it’s the load-bearing “raw tier can be narrated later without mixing workspaces” contract.

PR #337, opened 2026-06-15
feat(notebook): agentic session-summary + workspace-narration tasks

victorarias:
## What & why

PR-3 of the notebook-narration epic (plan: `docs/plans/2026-06-14-notebook-narration.md`), stacked on the capture-floor PR (#336 → `feat/notebook-raw-floor`). This is the actual product: the raw tier now feeds two **agentic** tasks that compose a curated, dated **work-journal**.

- **`summarize_session`** (cheap tier) — on every session Stop, a headless agent reads the session transcript and writes a faithful per-session digest to `.attn/raw/sessions/<wsID>/<sessionID>.md`, partitioned by workspace so a workspace's narrate reads only its own members' digests (solo sessions → a reserved bucket).
- **`narrate_workspace`** (strong tier) — on session-end (coalesced) and at the **removal boundary** (zero-debounce final retrospective), a headless agent reads the context snapshot + the workspace's digests + dispatch outcomes + its own prior journal entries, and writes **one curated entry per workspace per day** to `journal/<date>.md`. `IS_REMOVAL_PASS` is derived at run time from workspace-row absence (no task payload). The two load-bearing prompts — the epistemic tiering (**tool-result ≈ user-acceptance > agent-prose**) that surfaces claimed-but-abandoned dead-ends and did-but-never-mentioned wins — are embedded **verbatim**.

The foundation PR already made headless mode native-tools-only, so this PR adds the executors/triggers/prompts, not a new spawn mode. **Claude is the default** narrator (its `dontAsk` mode reaches the FS; its Write/Edit enforce the read-before-write CAS the shared journal depends on). Codex is supported via a new `HeadlessTaskRequest.ExtraWritableRoots` seam (`--add-dir`); its apply-patch CAS is unverified and documented in code.

## Correctness & security (caught by the adversarial review)

- **Path traversal (critical, fixed).** Every raw-tier path handed to the agent is built from a **client-controlled** session/workspace id. All of them now route through the floor PR's `rawTierFilename`/`rawTierSegment` single-safe-segment guard — a crafted id like `../../../journal/2026-06-15` is a hard error, not a write that escapes the raw tier into the curated journal. Read and write paths use the **same** guard so they can't diverge. (This is the exact hole #336 closed for the daemon's own write, re-opened on the narration path and re-closed here, with regression tests.)
- **No-op-run false-done (fixed).** Both executors verify the agent's own write as the ledger **and** that it changed since a pre-run fingerprint, so a coalesced re-run whose agent no-ops (e.g. the removal retrospective firing after an active-day entry already exists) is **retried**, not silently marked done.
- **Cross-workspace contamination (fixed).** Digests are bucketed per workspace, so one workspace's entry is composed only from its own sessions.
- **Marker prefix-collision (fixed).** The success ledger matches the full delimited `<!-- attn:wsnarr:<wsID> -->` line, so `ws-1` is no longer a substring-match of `ws-10`.
- **Latent runner race (fixed).** Hardened the `compactRunner` pointer swap: `startCompactRunner` runs late in `Start()` and replaced the placeholder while `Stop()`/enqueue/forget read the field concurrently. Now guarded by a dedicated `RWMutex` (`compactRunnerRef`/`setCompactRunner`), built+registered on a local pointer and published once. Pre-existing from the runner foundation; surfaced under `-race` once narration widened the `Start` window.

## Testing

- `go build`/`go vet` clean. `internal/agent`, `internal/notebook`, `internal/tasks` green under `-race`; full `internal/daemon` green; the narration + lifecycle subset green under `-race` across repeated runs (the previously-failing `Stop`/`Start` race no longer reproduces).
- Deterministic seams unit-tested with a **fake `HeadlessTaskProvider`** — **no real LLM is spawned**: config parse + fail-fast validation, Codex writable-roots arg, verbatim-prompt + path assembly, the two ledger checks, per-workspace scoping, traversal rejection, prefix-collision, and the triggers (summarize+narrate for a workspace session; summarize-only for a solo session; zero-debounce final narrate; startup-reap no-nil-deref).
- Agentic output **quality** (tiering, curation bar) is the plan's harness/manual bar, not unit-asserted on LLM prose.

## Known limitations / deferred (flagged for review)

1. **Single-session-workspace retrospective leans on the context snapshot, not the final session's grounded digest** (the build resolves the transcript from the session row at run time rather than carrying `msg.TranscriptPath` as the plan's §4 specified). On the common last-session-leaves teardown, the zero-debounce retrospective runs before the 2-min-debounced summarize lands, so the final session's digest isn't in it. The freshness guard makes it *retry* rather than silently finish, but there's no re-narrate-on-late-digest. **Recommend a focused follow-up** (carry the transcript path durably + re-trigger narrate when a late digest lands) since it touches the runner's task-payload model.
2. **Codex narrator** can't read `~/.claude`/`~/.codex` transcripts under `workspace-write` (writes work via `--add-dir`); documented, default Claude unaffected.
3. **Terminal task records** (`done` narrate/summarize) aren't GC'd yet — fold into PR-4's status surface (retention sweep).

## Scope

No protocol change (ProtocolVersion stays 109). The status surface + dreaming fold-in are PR-4; the daily-cron narrate also arrives with PR-4, so this PR ships **session-end + removal-boundary** narrates only.


figgyster (dismissed):
This looks safe to merge. The narration pipeline is wired at the right seams: session stop creates scoped raw digests, workspace removal gets the zero-debounce final pass, and the client-controlled ID paths all go through the raw-tier segment guard instead of trusting `filepath.Join` to be a bouncer.

I ran `go test ./...` locally; GitHub isn’t reporting branch checks yet. No blockers from me.

figgyster (changes requested):
Codex caught two real contract holes here, and they cross the line from polish into data-integrity risk for the journal.

Required before merge:
- Flush or otherwise force the relevant `summarize_session` digests before the zero-debounce removal retrospective. Right now workspace removal unregisters member sessions, then the final narrate can run while the debounced summarizer either has not run yet or later skips because the session row is gone. That can make the one final workspace entry omit the work that just happened.
- Scope dispatch raw inputs to the workspace, or add enough workspace/session provenance for the narrator to filter them deterministically. `narrate_workspace` currently hands every workspace the global `.attn/raw/dispatches` directory, and the raw dispatch file only says `source: dispatch:<id>`, so sibling workspace dispatches can bleed into the wrong journal entry.

My earlier approval missed these. Tiny rake, my footprints. The path guards and scoped session digests are good; these two edges need the same discipline before this is safe to merge.

PR #338, opened 2026-06-15
fix(notebook): single-session retrospective includes the final session's digest

victorarias:
## What & why

Closes the gap explicitly deferred from the narration PR (#337): for the **common single-session-workspace teardown**, the removal-day **retrospective** — the plan's "most important entry" — was composed from the `context.md` snapshot (the aspirational editorial overlay) alone, never the grounded digest of what the final session actually did.

**Root cause.** `summarize_session` resolved the transcript path *and* the workspace id from the **live session row at run time**. But `removeReapedSession` deletes the session row (and dissociate removes the workspace row) **before** the ~2-min-debounced summarize fires — so it no-op'd (no row) or wrote to the wrong `_solo` bucket (wsID resolved to `""`), while the zero-debounce removal narrate had already run over an empty digest bucket. The plan's §4 intended summarize to **carry** `msg.TranscriptPath` and "survive workspace loss because its input is under `~/.claude`"; the build had re-derived it instead.

## The fix

- **Carry the inputs, don't re-derive them.** Add an optional `Meta map[string]string` payload to the durable runner `Task`. Most kinds derive everything from `Subject`; `summarize_session` is the documented exception because both rows are gone by the time it runs. `Enqueue` applies a non-nil `opts.Meta` uniformly across all three branches (new / running-requeue / coalesce) — **replace** on fresh Meta, **preserve** on a bare re-trigger — and `clone()` deep-copies it so a handed-out record never aliases the runner's internal map.
- **`handleStop`** stashes `{transcript, workspace}` while both rows still exist. The executor **prefers the carried values** (so the digest lands in the correct per-workspace bucket even post-removal) and falls back to the row for a legacy/manual task. The carried workspace id routes through the same `rawTierFilename`/`rawTierSegment` guard as before — a crafted id is still a hard error.
- **Re-narrate hook.** After a successful digest write, if the workspace row is gone (removed), enqueue a **zero-debounce** narrate so the retrospective is rewritten *with* the final digest. The freshness guard from #337 makes it actually rewrite the block (the body changes). **Loop-safe**: narrate never enqueues summarize; multiple member summaries coalesce to one narrate per workspace. Skipped for an active workspace (its pending narrate already covers the fresh digest).

## Testing

- `go build`/`go vet` clean. `internal/tasks` green under `-race`; the daemon narration + lifecycle subset green under `-race`.
- New deterministic tests (fake `HeadlessTaskProvider`, **no real LLM**): Meta round-trips through save/load; re-enqueue-without-Meta preserves / with-Meta replaces; `clone()` map isolation; summarize uses the carried transcript+wsID and writes the **correct per-workspace bucket when the session row is gone**; the re-narrate hook fires iff the workspace is removed; `handleStop` stashes the inputs.
- Built and adversarially reviewed via an ultracode workflow (3 lenses — concurrency/coalescing/loop-safety, correctness/data-path, test-rigor; **no critical/high findings**, independently re-verified).

## Scope

Stacked on #337 (`feat/notebook-narration`). No protocol change. The `Meta` field is a minimal, general runner addition (foundation-level), used only by `summarize_session` here. With this, the narration core covers the single-session removal case correctly; PR-4 (status surface + dreaming fold-in) remains.


figgyster (dismissed):
This looks safe to merge. The fix carries the two bits of state that actually survive the teardown boundary, keeps the new task metadata isolated instead of aliasing caller maps, and the re-narrate hook is scoped to removed workspaces so it doesn't burn extra strong-tier runs for normal live workspaces.

I ran `go test ./...` plus `git diff --check`; both passed. No blockers from me.

figgyster (changes requested):
Codex caught a real race here, and I agree with it. The new carried metadata is still captured after `sendOK(conn)`, which means the stop hook can return to the agent before `stopWorkspaceID` is resolved and before `summarize_session` is enqueued. A teardown triggered immediately after that ACK can delete the session/workspace rows first, putting this right back in the failure mode the PR is trying to close: no workspace id carried, digest routed as solo or skipped, stale removal retrospective.

Required before merge:
- Resolve the workspace id and enqueue the summarize task before acknowledging the stop hook, so teardown cannot win that window.

The rest of the shape still looks good; this is a sequencing bug, not a taste fight. Tiny race, sharp teeth.

PR #339, opened 2026-06-15
feat(notebook): fold dreaming harvest onto the durable task runner

victorarias:
Stacked on #338 (`feat/notebook-narration-transcript`). Part of the notebook-narration epic — the **DR6 dreaming fold-in** half of PR-4, peeled out under the plan's Open-5 permitted split because it is pure-Go (no protocol/frontend) and fully verifiable without a packaged-app install.

## What

Dreaming's nightly harvest stops running on its own bespoke scheduler and instead runs as a `harvest_dream` task on the **existing durable runner** (`internal/tasks`). A single cron *enqueuer* decides when a harvest is due and dispatches the task; the runner owns execution, single-flight, retry/backoff, and crash recovery. This reaches the plan's **"one runner, zero special-case schedulers"** end state for harvest, and the harvest gains durable retry it never had (a failed harvest now backs off and re-runs instead of being lost).

## Design

- **`harvest_dream` executor** — sole writer of `candidates.json`; runs `dreamHarvestUnion` + `SaveDreamCandidates`. No `CommitGuard` because that write is an atomic, ctx-free temp+rename (a Cancel/timeout can't tear it).
- **Single cron enqueuer** (`startNotebookCronEnqueuer` → `notebookCronTick` → `enqueueDueDreamHarvest`) — keeps the timezone-aware cron + anchor/catch-up math verbatim, but on a due fire it **enqueues** rather than running inline. It is the **sole writer of `DreamRunState`** (anchor + dispatch time), so `state.json` has exactly one writer and the async executor never races it. **Anchor-first ordering**: a due fire advances the anchor *before* enqueuing, so a rare enqueue failure skips one (idempotent) day rather than re-firing every tick.
- **Deletions** (all subsumed by the runner): `startDreamingScheduler`/`dreamSchedulerTick`/`runDreamHarvest`, the in-daemon single-flight guard (`dreamMu`/`dreamRunning`), and the now-redundant `.attn/dreams/` filesystem lock (`AcquireDreamLock`/`ClearOrphanDreamLocks`/`dreamLockInfo`). Also the superseded **harvest source-2** (workspace-context Decisions/Constraints re-read) — narration now owns distilling `context.md` into the journal, so re-reading it would double-count; journals (source-1) and closed dispatches (source-3) are unchanged. Dropped `DreamRunState`'s per-run candidate-count fields (nothing surfaced them). Removed the already-dead `harvestDreamCandidates` helper.
- **Registration ordering**: `harvest_dream` registers in `startCompactRunner`; the enqueuer goroutine now launches *after* `startCompactRunner` so the kind is registered before the first tick can dispatch.

## Not in this PR (deliberately deferred)

- **Daily per-workspace narrate** (the other half of the plan's "single cron enqueuer"). The narrate executor currently treats "nothing new to add" as a *failure* (→ backoff → dead), so a daily idle-pass would spam failed runs. Making it well-behaved needs executor success-semantics surgery + an activity gate with new tracking state (plan Open-4) — genuine new design, so it gets its own PR. The extension point is marked `// TODO(daily-narrate)` in `notebookCronTick`.
- **DR5 status surface** (protocol bump 109→110 + Tasks panel) — needs `make install` + visual verification.

## Tests / verification (local)

- `go build ./...`, `go vet` — clean.
- `internal/notebook`, `internal/tasks`, and `internal/daemon` (Dream/Harvest/Narrate/Summarize/Compact/Janitor/Stop) — green.
- Scoped `-race` on the enqueuer-goroutine × runner-worker interaction and `internal/tasks` — clean (scoped with `-run` to dodge the known unrelated `TestGitStatusScheduler` data race).
- The cron due-math tests (first-enable, not-due, catch-up-once, timezone table, unsatisfiable, catch-up-before-slot) are preserved 1:1 and now *additionally* assert the `harvest_dream` task was/wasn't enqueued. Added a real executor idempotency test (union grows by exactly one new fact) and a positive *absence* test that workspace-context no longer harvests.

No `CHANGELOG.md` entry: internal refactor of an off-by-default, preview-only experimental subsystem; user-visible nightly-harvest behavior is unchanged.

🤖 Built via an ultracode workflow (build → 3-lens adversarial review → fix), then independently verified.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -232,11 +232,9 @@ type Daemon struct {
 	) (agentdriver.HeadlessTaskResult, error)
 	narrationNowOverride func() time.Time
 
-	// Dreaming scheduler (notebook consolidation janitor). dreamMu guards the
-	// single-flight dreamRunning guard; dreamSchedulerInterval overrides the tick
-	// cadence in tests (zero = default).
-	dreamMu                sync.Mutex
-	dreamRunning           bool
+	// Notebook cron enqueuer. dreamSchedulerInterval overrides the tick cadence in
+	// tests (zero = default). The harvest itself runs on the durable runner (kind
+	// harvest_dream), so there is no in-daemon single-flight guard here.
 	dreamSchedulerInterval time.Duration
 }
 
@@ -720,12 +718,14 @@ func (d *Daemon) Start() error {
 	// Start branch monitoring
 	go d.monitorBranches()
 
-	// Start the dreaming consolidation scheduler (clears orphaned run locks, then
-	// runs the nightly harvest when due). Off unless notebook.dreaming.enabled.
-	go d.startDreamingScheduler(d.done)
-
-	// Construct + start the durable compaction runner (kind "compact_context").
+	// Construct + start the durable compaction runner (kinds compact_context,
+	// summarize_session, narrate_workspace, harvest_dream).
 	d.startCompactRunner()
+
+	// Start the notebook cron enqueuer (enqueues the nightly dream harvest onto the
+	// durable runner when due; off unless notebook.dreaming.enabled). Launched AFTER
+	// startCompactRunner so harvest_dream is registered before the first tick fires.
+	go d.startNotebookCronEnqueuer(d.done)
 
 	d.signalStarted()
 	startSucceeded = true
--- a/internal/daemon/notebook_dreaming.go
+++ b/internal/daemon/notebook_dreaming.go
@@ -16,15 +16,16 @@ import (
 //
 // This file implements the deterministic, LLM-free harvest and the inspection
 // commands that surface it (`attn notebook dream status` / `--dry-run`). It scans
-// the three v1 sources — dated journals, canonical workspace-context snapshots,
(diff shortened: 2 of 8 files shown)


figgyster (dismissed):
This looks safe to merge. Folding the dream harvest onto the durable runner keeps the scheduling math intact while moving execution/retry/crash recovery to the one place that already owns that job.

I checked the runner registration/order, the single-writer `DreamRunState` path, source-2 removal, and the focused notebook/tasks/daemon tests locally. No blockers from me.

figgyster (changes requested):
Codex caught a real contract bug here, and I reproduced it with a focused throwaway test: a queued `harvest_dream` task can run after `notebook.root` changes, but the executor re-resolves the current setting instead of using the queued task subject.

That means the enqueuer can advance `DreamRunState` for root A, then the executor writes candidates for root B. Root A looks dispatched but never actually harvested. That crosses from cleanup refactor into data-contract bug.

Required before merge:
- make `harvestDreamExecutor` use `task.Subject` as the harvest root for queued tasks (with validation/fallback only if you intentionally support direct nil-task test calls), and
- add the regression where root changes between enqueue and execution.

PR #340, opened 2026-06-15
feat(notebook): daily-cron narrate for active long-lived workspaces

victorarias:
Stacked on #339 (`feat/notebook-dreaming-runner`). This is the **second trigger** of the plan's "single cron enqueuer" — the harvest_dream fold-in (#339) was the first, and this fills the `// TODO(daily-narrate)` it left in `notebookCronTick`.

## What this delivers

A never-removed long-lived workspace currently only gets a journal entry on a session end or at removal. If it runs for weeks with sessions that don't stop on a given day, it goes un-journaled. This adds a **nightly per-workspace narrate** so any workspace you actually worked in that day gets a fresh entry, while idle workspaces are skipped.

### 1. Activity gate (in-memory, best-effort)
`markNotebookWorkspaceActivity` records a workspace id in an in-memory set on:
- a **session end** (`handleStop`), and
- a **content-CHANGING context write** (`updateWorkspaceContext`'s `changed` branch — the janitor's compaction re-shape is deliberately NOT an activity signal, since it only reshapes existing content).

The cron **drains-and-clears** the set each fire and narrates only the workspaces in it. A removed workspace in the set is skipped (its removal-boundary final retrospective already ran).

The set is **not persisted**. Rationale: session-end is the primary narrate path; this cron is only a backstop for the no-session-stop day. A restart losing the set just means a workspace waits for its next session-end or next night's activity — no data loss. This avoids net-new durable state for a best-effort backstop.

### 2. Executor success-semantics split (`daily_pass` flag)
The `narrate_workspace` executor treats "agent wrote nothing / unchanged block" as a **failure** (→ backoff → dead). That's correct for session-end/removal (there's always new work), but a daily refresh legitimately finds nothing new. So:
- A **daily-flagged, non-removal** pass relaxes the gate: absent/unchanged block = **CLEAN DONE**.
- **Session-end** routine passes (no flag) and **removal** passes (`IsRemovalPass`, derived from workspace-row absence) keep **STRICT** gating, preserving their retry-until-the-digest-lands property.

Existing tests carry no flag, so they prove the strict path is unchanged.

### Schedule state
`NarrateCronState` (`.attn/narrate/state.json`) is a **separate anchor** from `DreamRunState` so the daily narrate never couples to the dreaming-enabled gate — narration is always on. The two share only the nightly cadence/timezone **settings** (the "notebook-maintenance slot"; a dedicated `notebook.cron.*` setting can split them later). Anchoring + catch-up semantics mirror `enqueueDueDreamHarvest` exactly.

## ⚠️ Known tradeoff for your call — coalescing edge (low severity, self-healing)

If a session-end narrate and the daily narrate land on the same `narrate_workspace:<ws>` task **within the debounce window** (a session stops within ~`notebookNarrationDebounce` of the nightly slot), the merged record ends up daily-flagged in **both** enqueue orderings — the runner's Meta is replace-on-non-nil / leave-on-nil, so daily's flag either wins by replacing or survives because session-end carries no Meta. The coalesced run then takes the **relaxed** gate.

- **Why it's benign:** it only bites if the agent *also* no-ops on a real session-end digest (a transient flake — real work always writes), and even then nothing is lost: the raw digest persists and the next trigger re-narrates. Documented at `enqueueDailyNarrateWorkspace`.
- **The full fix I chose NOT to take:** flip the executor default to *relaxed* with a sticky "strict-wins" marker. That fully closes it, but trades a rare timing edge for a **less safe default** on the primary session-end path (a future enqueue site that forgets the marker silently no-ops). I judged default-strict the better invariant. **Happy to flip if you'd rather have provably-strict coalescing.**

## Verification (all run locally)
- `go build ./...` clean; `go vet ./internal/daemon ./internal/notebook ./internal/tasks` clean
- `internal/notebook` green; scoped daemon narrate/dream tests green
- scoped `-race` on the daily-narrate + activity-gate tests clean
- 12 new tests (cron due-math, activity hooks incl. the no-op-context-write negative case, executor relaxation incl. removal-still-strict and routine-still-strict). No real LLM — fake execution hooks throughout.


figgyster (dismissed):
This is a real blocker: the new `daily_pass` flag can leak from a completed daily task into a later strict session-end narrate.

The runner explicitly keeps existing `Task.Meta` when an enqueue passes `Meta: nil`. So after a daily pass queues `narrate_workspace:<ws>` with `daily_pass=1` and finishes `DONE`, a later `enqueueNarrateWorkspace` call for the same workspace reuses that task id with nil Meta and leaves `daily_pass=1` on the record. That makes the session-end path take the relaxed gate, so an agent no-op becomes clean `DONE` instead of retrying until the digest lands. That breaks the strict session-end contract this PR is trying to preserve.

I verified it with a focused throwaway repro: enqueue a daily-flagged narrate, let it finish done, then call `enqueueNarrateWorkspace("ws-1")`; the requeued task still had `Meta: map[daily_pass:1]`. Tiny flag, big trench coat.

Required before merge:
- clear/replace Meta on strict `enqueueNarrateWorkspace` enqueues (and add the regression test for daily -> later routine), so the daily relaxation is one-run scoped instead of sticky across future session-end triggers.

Local checks run:
- `go test ./internal/notebook ./internal/daemon -run 'Test(EnqueueDueDailyNarrates|HandleStopMarksWorkspaceActivity|ContextWriteMarksActivityOnlyWhenChanged|NarrateWorkspaceDaily|NarrateWorkspaceRoutine|NarrateCronState)'`
- `go test ./internal/daemon ./internal/notebook`

figgyster (changes requested):
Codex surfaced a second real blocker, and I verified it against the current head.

The previous `daily_pass` sticky-Meta issue still needs fixing, but there’s also a larger activity-gate gap: the nightly drain only sees workspaces marked by `handleStop` or content-changing context writes. In the exact long-lived-session case this cron is supposed to cover, a session can keep running across the nightly slot with normal turn/tool/transcript activity, never stop, and never write shared context. `handleState` touches the session and records working state, but it does not mark the workspace active, so `enqueueDueDailyNarrates` drains an empty set and skips the workspace entirely.

That means the feature can miss the core “worked today but no session stopped” path. That crosses the contract line, not taste.

Required before merge:
- keep the `daily_pass` flag from leaking into later strict narrates, as noted in my earlier review; and
- mark activity from the live turn/state/heartbeat path or derive the daily active workspace set from persisted session activity timestamps, with a regression for “working session crosses nightly slot without stop/context write”.

I left a 👍 on the Codex thread because that one’s on target. Annoying little rake, but a real one.

PR #341, opened 2026-06-15
feat(notebook): task status surface (list/retry + live panel)

victorarias:
Stacked on #340 (`feat/notebook-daily-narrate`). **DR5 — the final piece of the notebook-narration plan.** End state: one durable runner, fully observable, with manual recovery.

## What this delivers

The runner's background jobs — context-compaction, session-digest, journal-narration, nightly dream-harvest — ran invisibly. A stuck one was undiscoverable without reading `.attn/tasks/` by hand. This surfaces them in the app (and a CLI) and lets you retry a dead one.

## ⚠️ Needs your machine (couldn't do these in-sandbox)

1. **`make install`** — this bumps `ProtocolVersion` 109→110. The daemon survives app rebuilds, so until you reinstall, the running daemon (109) and the new app (110) will force a reconnect mismatch by design. I verified the `make generate-types` toolchain runs clean in-sandbox, but `make install` needs the keychain.
2. **Visual check** of the Tasks panel — open the Notebook browser, expand **Tasks**, confirm rows render and a failed/dead task's **Retry** works + the panel refreshes live. All automated layers pass, but I can't see the rendered UI.

## Design

**Protocol** — `NotebookTask` + `notebook_task_list` / `notebook_task_retry` WS request/result + a payload-free `notebook_tasks_changed` broadcast. `kind`/`state` are plain strings (not TypeSpec enums) to dodge the quicktype identical-enum-merge that drops a generated export. **Security:** `NotebookTask` deliberately omits the runner's internal `Meta` bag (transcript filesystem paths) and `CommitGuard` — a test asserts the secret never reaches the wire.

**Broadcast via the existing `OnChange` hook** — no new runner machinery. `OnChange` fires synchronously inside the single worker goroutine, so it routes through the non-blocking buffered hub broadcast (drops on a full channel) and can never stall task execution. A successful `Retry` already fires `OnChange`, so the retry handler doesn't broadcast itself.

**Nil-safety** — every handler nil-checks `compactRunnerRef()` (the runner is built late in startup), so a not-yet-built/disabled runner is a clean empty/false result, never a panic.

**CLI** — `attn notebook tasks` lists the records (also useful headless, where there's no GUI).

**Frontend** — a collapsible Tasks section, collapsed by default + lazily fetched on open. Per-row kind:subject, state badge, attempts, relative next-attempt, last_error; Retry only on failed/dead. The fetch has a monotonic stale-guard (dropped on close), a transient WS failure surfaces an error instead of wiping rows, and the Retry in-flight mark clears on both resolve and reject (no stuck spinner) — the broadcast drives the refetch, no optimistic mutation.

## Verification (all run locally)
- `go build ./...`, `go vet`, and `make generate-types` clean (no spurious regen drift)
- `go test ./internal/protocol ./internal/tasks ./internal/daemon` green (full daemon suite)
- scoped `-race` on all 9 new notebook-task tests clean
- `tsc --noEmit` clean; full vitest suite green (NotebookBrowser: 27)
- Every diff read by hand

I also took the one review finding worth acting on (a test that only re-proved a `tasks.Runner` property) and rewrote it as an **end-to-end** test: a real task transition delivers a `notebook_tasks_changed` broadcast (correct event) to a subscribed wsHub client — covering the live-refresh path the per-handler tests can't see.


figgyster (dismissed):
This looks safe to merge. The task surface follows the existing request/result protocol shape, keeps runner internals off the wire, and the retry path is guarded to failed/dead work instead of inventing optimistic UI state.

I ran the Go suite, the frontend tests, and `git diff --check` locally. GitHub checks weren’t reported on the branch yet, and the PR’s own note about `make install`/visual Tasks-panel verification still stands for the machine-specific bit.

No blockers from me.

figgyster (changes requested):
Codex caught a real privacy/contract miss here. The nightly `harvest_dream` task uses the notebook root as its task subject, and this PR copies both `ID` and `Subject` straight into `NotebookTask`. That means the new app panel and `attn notebook tasks` can expose a local notebook path, even though the protocol boundary explicitly tries to keep path-bearing runner internals off the wire.

Required before merge:
- Redact or replace path-backed task identifiers/subjects before returning task records to clients.
- Keep retry usable through a non-path public identifier, or otherwise map the UI retry back to the internal task safely.

The retry-refresh point is also valid: since `notebook_tasks_changed` is best-effort, the UI should update from the successful retry result or explicitly re-list after retry instead of relying only on the broadcast.

This flips my earlier approval; the privacy leak crosses the line. Tiny rake, sharp teeth.

PR #342, opened 2026-06-16
feat(notebook): unify the "keeper" domain model

victorarias:
Stacked on #341 (`feat/notebook-task-status`). Encodes one canonical domain language so the code, prompts, guidance, and plan all use the same words — and future agents do too.

## What & why

There is now **one entity — the keeper** — for automated workspace upkeep, replacing the split `janitor` / `narrator` / `summarizer` vocabulary. The keeper has two **causally-coupled** duties: keep each workspace's `context.md` tidy (compaction) and narrate its work into the curated journal. It can safely prune the overlay *because* it already preserved the story in the journal — "nothing is lost" becomes one entity's promise instead of an implicit cross-actor contract.

The task-kind strings (`compact_context` / `summarize_session` / `narrate_workspace` / `harvest_dream`) are **mechanisms** and stay verbatim — only the persona vocabulary unifies.

## Changes

- **`docs/glossary.md`** — canonical domain language: *workspace context* (editorial overlay), *the journal* (durable curated cross-workspace log), *the keeper*, *the chief of staff* (keeper-aware), *dispatch report*, *the raw tier*. The definition wins on drift.
- **Full `janitor` → `keeper` rename** across daemon, store, agent, CLI, and frontend: `workspace_context_janitor.go` → `workspace_keeper.go`; `SettingWorkspaceContextJanitor` → `SettingKeeperCompact`; executor/validate/apply/config funcs; the SettingsModal UI heading, TS symbols, and test IDs. Persisted literals intentionally kept (documented): the `attn-janitor` updater sentinel and the `workspace_context_janitor_backups` table.
- **Persisted setting-key migration** `workspace_context_janitor` → `workspace_keeper_compact`: a one-time, idempotent copy at daemon start (runs **before** the runner reads the setting; no-clobber; deletes the legacy row so the next boot is a true no-op). Survives app rebuilds.
- **Prompts + prose reframe** — the narrate/summarize prompts and every `narrator`/`summarizer` comment now refer to the keeper's narrate/summarize duties.
- **Chief `NotebookGuidance` rewrite** — the journal is the **curated cross-workspace log** (was "the raw record"); the chief journals at chief-of-staff altitude and is **keeper-aware** (doesn't duplicate the per-workspace narratives the keeper already writes). Same "raw record" fix in the skill reference.
- **J2 revert** — drop the `JournalingDirective` nudge at the 3 non-chief sites; the function had no remaining caller, so it's removed (the `attn notebook journal append` CLI stays available, just no longer pushed).
- **Plan doc** vocabulary aligned to the glossary.

## Verification (all green, run locally)

- `go build ./...`, `go vet ./...`, `gofmt` clean.
- `internal/{hooks,agent,store,tasks}` + full `internal/daemon` (140s) pass; `TestMigrateKeeperCompactSettingKey` covers copy-forward+drop, idempotent/no-clobber, and no-legacy no-op.
- `internal/tasks -race` and scoped `internal/daemon -race` (keeper/compact/migration/narrate/runner) clean.
- Frontend: `tsc --noEmit` + 16 vitest tests (`workspaceContextKeeper`, `SettingsModal`) pass.
- Zero `narrator`/`summarizer` persona left in Go; residual `janitor` is only the documented persisted literals.

## Process

Built via an ultracode workflow (parallel discover → coherent rename + migration → verify → 5 adversarial review lenses), then every diff independently re-verified and the confirmed review findings (the `narrator`-persona residue, the frontend inconsistency, the skill-reference "raw record" line) fixed before commit.

> Owner note: like #341 this branch carries the protocol bump from its parent; `make install` is needed on the owner machine for the daemon to pick it up. No new protocol change in this PR itself.


figgyster (dismissed):
This looks safe to merge. The keeper rename is coherent across the daemon/store/UI seams, and the one-time setting-key migration preserves existing `workspace_context_janitor` configs without keeping the dead key around forever — nice little bit of archaeology cleanup.

I ran the focused Go suites for the touched daemon/store/agent/hooks/notebook/task paths, the targeted SettingsModal/keeper frontend tests, and `git diff --check`; all passed. GitHub had no checks reported for this branch at review time.

No blockers from me.

figgyster (dismissed):
Codex is right on the protocol/versioning edge here, and it crosses the app/daemon compatibility line.

The PR changes the Settings UI to send `workspace_keeper_compact`, while both sides still advertise protocol `110`. An already-running old daemon from the base branch accepts `workspace_context_janitor` but rejects the new key as `unknown setting`, so a rebuilt app can connect cleanly and then fail Save/Disable for the keeper setting instead of forcing the daemon restart/mismatch path.

Required before merge:
- either bump the daemon + frontend protocol version for this setting-key contract change, or
- keep a transition alias so the old key is still accepted/mapped safely while the app moves to `workspace_keeper_compact`.

I left a 👍 on the Codex thread; this one is real, not bot confetti.

figgyster (dismissed):
The follow-up migration work is fine as far as it goes, but it doesn’t close the blocker I raised.

The current head still has both sides advertising protocol `110` while the rebuilt Settings UI sends `workspace_keeper_compact`. An old daemon from the base side still only validates the old key, so the app/daemon compatibility failure remains: the app can connect cleanly, then Save/Disable for the keeper setting gets rejected as `unknown setting` instead of hitting the version-mismatch restart path.

Required before merge stays the same:
- bump the daemon + frontend protocol version for the settings-key contract change, or
- keep a transition alias that accepts/maps the old setting key safely while the app moves to `workspace_keeper_compact`.

I re-ran the focused store/daemon tests, the targeted SettingsModal/keeper frontend tests, and `git diff --check`; they pass. Tests green, contract still red — classic little rake in the alley.

figgyster (dismissed):
The new glossary additions look fine, but this push only changes docs, so the app/daemon settings-key blocker is still present on the current head.

`PROTOCOL_VERSION` / `ProtocolVersion` are still `110`, while Settings still sends `workspace_keeper_compact`. That means a rebuilt app can still connect to an old same-protocol daemon that only accepts `workspace_context_janitor`, then fail Save/Disable with `unknown setting`.

Required before merge remains one of:
- bump the daemon + frontend protocol version for this settings-key contract change, or
- keep a transition alias that accepts/maps the old key safely while the app moves to `workspace_keeper_compact`.

I ran `git diff --check`; clean. This is still a contract blocker, not a docs problem.

figgyster (changes requested):
The dreaming glossary correction is good, but this push is still docs-only relative to the active blocker.

The contract issue remains on the current head: frontend and daemon still advertise protocol `110`, and Settings still sends `workspace_keeper_compact`. A rebuilt app can still connect to an older same-protocol daemon that only accepts `workspace_context_janitor`, then fail Save/Disable with `unknown setting`.

Required before merge remains one of:
- bump the daemon + frontend protocol version for this settings-key contract change, or
- keep a transition alias that accepts/maps the old key safely while the app moves to `workspace_keeper_compact`.

`git diff --check` is clean. The docs are getting sharper; the merge blocker is still the protocol seam.

PR #343, opened 2026-06-16
Remove the dreaming feature; rename cron infra; map notebook processes

victorarias:
Stacked on `feat/keeper-domain-model`.

## What

Completely removes the notebook **dreaming / `harvest_dream`** feature, renames the per-minute cron that survived it, and documents the notebook's processes accurately.

### 1. `remove the dreaming feature`
The nightly `harvest_dream` task accumulated a deduplicated candidate set under `.attn/dreams/` but had **no downstream consumer** — the promote pass that would turn candidates into durable memory was never built, and memory notes are authored by hand. It was gated off by default and never shipped. Removed end-to-end:
- `harvest_dream` task kind + executor + its runner registration
- the dreaming candidate store + run-state (`internal/notebook/dreams*.go`)
- `notebook_dream_status` / `notebook_dream_run` protocol messages + results
- the `attn notebook dream` CLI + client methods
- the `notebook.dreaming.enabled` setting

Protocol bumped **110 → 111** (removed commands/messages force a reconnect). The per-minute cron was kept, surgically, to keep driving the daily-narrate backstop.

### 2. `rename notebook cron infra dreaming->cron`
Every surviving `dreaming` identifier renamed so no dead-feature vocabulary lingers (`notebook_cron.go`, `notebookCron*`, `notebook.cron.{frequency,timezone}`). Adds `migrateNotebookCronSettingKeys` (idempotent, mirrors `migrateKeeperCompactSettingKey`): copies any persisted `notebook.dreaming.{frequency,timezone}` forward to `notebook.cron.*` and reaps the orphaned `notebook.dreaming.enabled` row. No protocol change — settings travel as an untyped map; the frontend never referenced these keys.

### 3. `docs: map notebook processes; fix raw-tier session path`
Adds `docs/notebook-processes.md` — a code-grounded map of every automated notebook process (raw tier, durable task engine, the keeper, the notebook cron, the journal, memory notes, dispatch capture). Fixes one confirmed glossary inaccuracy: raw-tier session digests are nested `sessions/<wsID>/<sessionID>.md` (with a reserved `_solo` bucket), not flat.

## Verification
- `go build ./...`, `go vet ./...` clean
- full `go test ./internal/daemon` green (incl. the new migration test: copy-forward, idempotency, enabled-gate reaping)
- `internal/notebook` / `internal/tasks` / `internal/protocol` / `internal/client` / `cmd/attn` green
- frontend `tsc` + 949 vitest green
- `make generate-types` produces zero drift
- A multi-agent adversarial pass over the diff found no dangling symbols and no silently-dropped behavior; the only surviving `dreaming` strings are the intentional legacy-settings migration.


figgyster (approved):
This looks safe to merge. The dead dreaming path is gone end-to-end, the surviving daily-narrate scheduler has a cleaner `notebook.cron.*` owner, and the legacy settings migration is careful about not clobbering newer values.

I ran `go test ./...` locally on the PR head; GitHub isn’t reporting checks for this branch yet. No blockers from me.

PR #345, opened 2026-06-17
Knowledge-base epic: reframe memory into a PARA/OKF knowledge base, remove the notebook CLI, archive projects on close

victorarias:
The complete knowledge-base epic as a single PR. (Supersedes #344 and #346, both folded in.)

## What changes

**1. Data model — `memory` → PARA/OKF knowledge base**
- `memory/{decisions,gotchas,domain}` → `knowledge/{projects,areas,resources,archive}`, each with its own `index.md`.
- Frontmatter `kind` (fixed enum) → open-string `type` (OKF). The accessor reads `type` falling back to `kind`, so externally-authored `kind:` still resolves. Reserved `index.md`/`log.md` carry no frontmatter; links are root-absolute markdown, not wikilinks.
- The store is a **permissive writer** — no `type` validation. A non-empty `type:` is an author/chief convention conveyed by guidance, not a store-enforced rule.

**2. Remove the `attn notebook` CLI**
- The chief of staff maintains the notebook and journal by editing the markdown files on disk directly with native tools.
- WS-frontend surfaces are unchanged: browse the Notebook, send a selection to the chief, the background-task panel. The launch-time guide hook remains.
- Guidance (`hooks.go`, `notebook.md` skill ref, SKILL.md) rewritten to native-file editing over `/knowledge` paths.

**3. Keeper archives a workspace's project on close**
- On a workspace's **final removal pass**, after the retrospective, the keeper files that workspace's linked `knowledge/projects/<slug>/` folder under `knowledge/archive/`.
- The project↔workspace link (`resource: attn:workspace/<id>`) is matched with a **literal whole-line grep** (`grep -Fx`) so a short id can never substring-match a longer live workspace and archive the wrong folder. Ambiguous/no match → no-op. Reversible.
- The stamping convention is taught in the **always-injected** chief guidance (not only the opt-in reference), so the feature actually fires.

**Protocol** bumped to 113 (daemon ↔ app reconnect).

## Quality
- The full epic was put through a comprehensive multi-agent adversarial review (8 dimensions, per-finding verification, completeness critic): **no functional defects**. Confirmed findings were docs/guidance accuracy and are fixed in this PR.
- `go build`/`vet`/`gofmt` clean; notebook, protocol, hooks, and daemon narration/notebook suites pass.

## Notes
- Existing on-disk `memory/` notes are intentionally **not** migrated; they remain readable and group under the generic "Notebook" sidebar bucket until relocated. Nothing is deleted.
[...]

PR #346, opened 2026-06-19
Keeper archives a removed workspace's project folder (knowledge-base epic, PR3)

victorarias:
Stacked on #345 — the last slice of the knowledge-base epic.

Adds one keeper behavior: on a workspace's **final removal pass**, after writing the retrospective, the keeper files that workspace's linked knowledge-base project folder under `knowledge/archive/`, so the active `knowledge/projects/` view stays focused on live work. Promotion of durable knowledge into `areas/` stays the chief's higher-judgment call.

## What changes
- **`narrate_workspace` prompt** (`internal/daemon/notebook_narration_prompts.go`): a new `KNOWLEDGE_DIR` input and a removal-pass-only archive step.
- **Loop closure** (`notebook.md` chief guidance): a workspace's project `index.md` gets stamped `resource: attn:workspace/<id>` so the keeper has a link to find the folder; the `knowledge/projects/index.md` template already documents the convention.
- **Plumbing**: `KnowledgeDir` (= `root/knowledge`) threaded through `narrateWorkspacePromptInputs` / `buildNarrateWorkspacePrompt`.
- **Docs**: glossary keeper entry, `notebook-processes.md` keeper Duty-2, CHANGELOG.

## Safety
The link is matched with a **literal whole-line grep** (`grep -Fx 'resource: attn:workspace/<id>'`), not a substring/regex match. This was the key fix from an adversarial review: workspace ids are arbitrary slugs (not UUIDs), so a plain substring grep for a removed `auth` would also match a live `auth-v2` and archive the wrong, still-active project (the same `ws-1`/`ws-10` collision class the narration marker already guards against). `-F` keeps regex/glob metacharacters inert; `-x` forbids prefix collisions. Ambiguous or no match → do nothing. The step is gated to `IS_REMOVAL_PASS` and the move is reversible (the folder still exists, under `archive/`).

## Notes
- `chief-of-staff.md` needs no change (dispatch-only; no notebook-CLI or memory references).
- Optional follow-up **not** included: live-promotion guidance parity (writing `NotebookGuidance` into a reserved file at promotion so a freshly-promoted chief gets the full guidance live, not just at launch). Say the word.
- Build/vet/gofmt clean; 91 notebook+narration daemon tests pass.


figgyster (dismissed):
This looks safe to merge. The archive behavior is scoped to the removal pass, the workspace/project link is exact enough to avoid the `auth`/`auth-v2` trap, and the active-project view gets cleaned without deleting or promoting knowledge behind the chief’s back.

I ran `go test ./...` locally; GitHub isn’t reporting checks for this branch right now. No blockers from me.

figgyster (approved):
Still safe to merge. The follow-up puts the workspace resource stamp into the always-injected chief guidance, which is exactly where that load-bearing convention belongs; otherwise this feature could quietly turn into a beautiful no-op. The docs/comment cleanups line up with the current `knowledge`/`type:` contract too.

`go test ./...` passes locally. GitHub still reports no checks for this branch. No blockers.

PR #347, opened 2026-06-19
feat(tasks): durable file-backed task runner

victorarias:
Foundation for the keeper epic: a self-contained, single-worker durable task runner under `.attn/tasks/` (one JSON per task), with a capped-backoff state machine, subject-coalescing (zero-debounce override), a single-instance lock + orphan recovery, `Cancel`-blocks-until-exit, and an executor-owned commit fence. Pure package — no internal deps.
---

This is PR **1/10** — base `feat/chifplace`. Review per-diff. By design the refactor only activates in #8 (where the keeper is wired and the janitor/dreaming are deleted), so branches 2–9 are not individually `go build`-clean; the merged tip (#10) is build + vet + test green.

1. `feat/chifplace`
2. **#347** 👈 current
3. #348
4. #349
5. #350
6. #351
7. #352
8. #353
9. #354
10. #355
11. #356


figgyster (dismissed):
This crosses from “filename hygiene” into durable task corruption.

`TaskID(kind, subject)` preserves the full subject identity, but the on-disk path collapses distinct IDs by replacing `/`, `\`, `:`, and `..` with underscores. So `kind:a/b` and `kind:a_b` address the same JSON file. I verified it with a focused local probe: enqueueing `("kind", "a/b")` and then `("kind", "a_b")` returns the first task record for the second enqueue (`id="kind:a/b" subject="a/b"`). The existing package tests still pass because they don’t cover colliding subjects.

Required before merge:
- Make the filename encoding injective for the full task ID — e.g. base64url/hex/path-escape style encoding instead of lossy replacement.
- Defensively reject a loaded record whose stored `Task.ID` does not match the requested ID.
- Add a regression test for two distinct subjects that sanitize to the same current filename shape.

I also ran `go test ./internal/tasks` after removing the local probe; it passes. This one needs the storage identity fixed before the task runner becomes foundation concrete instead of wet cement.

figgyster (approved):
Fix verified. The task ID is now encoded injectively for the filename, the load path refuses mismatched records instead of mutating the wrong task, and the regression tests cover the old collision shapes.

`go test ./internal/tasks` passes locally. No blockers from me now.

PR #348, opened 2026-06-19
feat(notebook): PARA/OKF knowledge base + raw capture tier

victorarias:
Reframes the notebook library from the fixed `memory/{decisions,gotchas,domain}` model to a **PARA** knowledge base (`projects/areas/resources/archive`, each with its own `index.md`) on a directory axis, plus an open-string **OKF `type`** frontmatter axis. The store is a permissive writer (no `type` validation). Adds the **raw tier** (machine-input floor: per-session summaries + context snapshots) and the narrate-cron state anchor. The read layer stays backward-compatible (`type` falls back to `kind`).
---

This is PR **2/10** — base `kb/01-task-engine`. Review per-diff. By design the refactor only activates in #8 (where the keeper is wired and the janitor/dreaming are deleted), so branches 2–9 are not individually `go build`-clean; the merged tip (#10) is build + vet + test green.

1. `feat/chifplace`
2. #347
3. **#348** 👈 current
4. #349
5. #350
6. #351
7. #352
8. #353
9. #354
10. #355
11. #356

### DIFF
--- a/internal/notebook/document.go
+++ b/internal/notebook/document.go
@@ -157,8 +157,15 @@ func (d Document) frontmatterString(key string) string {
 	return v
 }
 
-// Kind returns the declared kind ("" if absent or non-string).
-func (d Document) Kind() string { return d.frontmatterString("kind") }
+// Type returns the declared OKF `type` ("" if absent or non-string). For
+// read-compatibility with notes an external tool wrote before the field was
+// renamed, it falls back to a legacy `kind` value; attn always writes `type`.
+func (d Document) Type() string {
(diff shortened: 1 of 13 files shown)


figgyster (approved):
This slice looks safe to merge. The notebook format shift is contained to the store/layout layer, keeps legacy `kind` reads working, and the raw/narrate machine paths stay behind the `.attn` wall instead of leaking into user-facing notebook APIs.

I ran `go test ./internal/notebook` and `git diff --check`; both passed. No blockers from me.

PR #349, opened 2026-06-19
feat(agent): native-tools headless task mode

victorarias:
`RunHeadlessTask` hands a headless Claude/Codex agent its **own native file tools** in a writable scratch dir (Claude `--allowedTools`+`dontAsk`; Codex `--sandbox workspace-write`+`--add-dir`), replacing the constrained `read_context`/`replace_context` MCP path. Adds `ExtraWritableRoots` for the narrator's journal writes. (The MCP toolserver itself is deleted in #10 with its only caller, the CLI.)
---

This is PR **3/10** — base `kb/02-notebook-knowledge-base`. Review per-diff. By design the refactor only activates in #8 (where the keeper is wired and the janitor/dreaming are deleted), so branches 2–9 are not individually `go build`-clean; the merged tip (#10) is build + vet + test green.

1. `feat/chifplace`
2. #347
3. #348
4. **#349** 👈 current
5. #350
6. #351
7. #352
8. #353
9. #354
10. #355
11. #356

### DIFF
--- a/internal/agent/claude.go
+++ b/internal/agent/claude.go
@@ -2,7 +2,6 @@ package agent
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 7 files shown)


figgyster (dismissed):
This is doing the right surgery by moving the headless path off the constrained MCP shim, but the current head doesn't pass its own focused agent tests yet.

Required before merge:
- Fix the Codex workspace-context override path so it matches the new non-chief journaling contract. `go test ./internal/agent` currently fails in `TestCodexConfigOverrides_NonChiefOmitsJournalingDirective` because `GenerateCodexConfigOverrides` still appends `hooks.JournalingDirective()` for non-chief Codex launches, while this PR updates Claude and the test expectation to omit it.

I also ran `git diff --check` clean. No other blocker from the native-tools headless diff.

figgyster (dismissed):
The new push doesn't clear the blocker yet.

`go test ./internal/agent` still fails in `TestCodexConfigOverrides_NonChiefOmitsJournalingDirective`: `internal/hooks/codex.go` still builds non-chief workspace-context `developer_instructions` as `WorkspaceContextGuidance(...) + JournalingDirective()`. That leaves Codex on the old journaling contract while Claude and the agent test expectations moved to the keeper/chief-owned journaling model.

Required before merge:
- Remove the non-chief Codex journaling append in `GenerateCodexConfigOverrides`, matching the Claude-side change and the updated test contract.

I also checked the latest head against the PR diff; this is the same focused blocker, just still wearing yesterday's coat.

figgyster (approved):
Victor's stack note checks out. This slice's standalone `internal/agent` test failure is caused by compiling the new agent tests against the old `internal/hooks` implementation; the hook-side contract change lands later in the stack, and the tip has `internal/agent` byte-identical with the test green.

I verified the current slice diff, the empty `internal/agent` delta to the stack tip, and `go test ./internal/agent` on `kb/10-cli-guidance-docs`. The red review was mine being too literal about an intentionally non-build-clean slice — classic rake, excellent craftsmanship.

No blockers from me.

PR #350, opened 2026-06-19
refactor(store): keeper rename + migration 51

victorarias:
Renames the persisted janitor surfaces to the **keeper** vocabulary. Migration **51** renames `workspace_context_janitor_backups` → `workspace_keeper_compact_backups` and realigns the `attn-janitor` updater sentinel → `attn-keeper` on existing rows (idempotent across a migration rewind). Lands at 51 (not 49) because 49/50 are burned on real prod/dev DBs. Adds the `DeleteSetting` helper used by the settings-key migration.
---
**Final-state re-split of the knowledge-base / keeper epic** (supersedes #335–#343, #345). The original stack was split by chronology and carried mid-epic churn; this one is sliced by subsystem from the net diff, so each PR is a clean final-state slice.

Stack: `1 task-engine` · `2 notebook-kb` · `3 agent-native` · `4 store-keeper` · `5 keeper-compaction` · `6 keeper-narration` · `7 keeper-cron` · **`8 integration`** · `9 frontend` · `10 cli/docs`

This is PR **4/10** — base `kb/03-agent-native-tools`. Review per-diff. By design the refactor only activates in #8 (where the keeper is wired and the janitor/dreaming are deleted), so branches 2–9 are not individually `go build`-clean; the merged tip (#10) is build + vet + test green.

### [Stack](https://github.com/kitlangton/stack)

1. `feat/chifplace`
2. #347
3. #348
4. #349
5. **#350** 👈 current
6. #351
7. #352
8. #353
9. #354
10. #355
11. #356

PR #351, opened 2026-06-19
feat(daemon): keeper compaction + raw-tier capture floor

victorarias:
Adds the keeper's **tidy** duty (`workspace_keeper.go`: `compact_context` on the durable runner — agentic via native tools; the daemon keeps temp-in/candidate-out + validate + revision-guarded commit under CommitGuard) and the deterministic **capture floor** (`notebook_raw_tier.go`: the synchronous `context.md` snapshot at every removal path, funnelled through one path-segment guard so a client-controlled id can't escape the raw dir). Dispatch outcomes are redirected to `.attn/raw/dispatches/`. Code is dormant until #8 wires it.
---

This is PR **5/10** — base `kb/04-store-keeper-migration`. Review per-diff. By design the refactor only activates in #8 (where the keeper is wired and the janitor/dreaming are deleted), so branches 2–9 are not individually `go build`-clean; the merged tip (#10) is build + vet + test green.

1. `feat/chifplace`
2. #347
3. #348
4. #349
5. #350
6. **#351** 👈 current
7. #352
8. #353
9. #354
10. #355
11. #356

### DIFF
--- a/internal/daemon/notebook_dispatch_journal.go
+++ b/internal/daemon/notebook_dispatch_journal.go
@@ -9,27 +9,35 @@ import (
 	"github.com/victorarias/attn/internal/protocol"
 )
 
-// Auto-journaling of chief-of-staff dispatch outcomes.
+// Deterministic capture of chief-of-staff dispatch outcomes to the raw tier.
 //
 // A dispatch is delegated agent work, and its terminal report (completed/failed)
 // — or, as a fallback, its session simply ending — is exactly the "what was
 // decided / built / failed" signal the durable work-journal exists to capture.
 // This file renders that already-structured DispatchReport into a human-readable
-// journal block and appends it to the daily journal:
+// block and writes it to the notebook's raw tier
+// (.attn/raw/dispatches/<dispatchID>.md), where the later narration pass reads it.
+// It deliberately does NOT write the curated journal — the journal stays curated.
 //
 //   - deterministic: the daemon renders structured fields, no LLM is involved;
-//   - idempotent: a hidden per-dispatch marker means the terminal-report path, the
-//     session-gone fallback, and a daemon restart together write exactly one block;
-//   - grounded by construction: every block cites source: dispatch:<id>.
+//   - exactly-once: one file per dispatch id; a replay atomically overwrites that
+//     one file, so the report path, the session-gone fallback, and a daemon restart
+//     together leave exactly one entry. The overwrite is byte-identical when the
+//     dispatch carries a server ReportedAt; when it does not (a worker that ended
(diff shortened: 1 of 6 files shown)


figgyster (dismissed):
This one has a real blocker: the raw-tier context snapshot helper is never wired into the actual workspace removal paths.

`snapshotWorkspaceContextOnRemove` only appears in its own tests and definition; the live removals still go straight through `store.RemoveWorkspace` in `workspace.go` / `workspacelayout.go`, so the `workspace_contexts` row is deleted before anything snapshots `context.md`. That means the advertised capture floor does not actually protect the data on a real workspace removal.

Required before merge:
- Call the snapshot path from every workspace-removal path after cancelling/forgetting compaction and before `store.RemoveWorkspace`.
- Add an integration-ish removal test that proves a real removal writes `.attn/raw/context-snapshots/<workspace>.md` before the context row is deleted.

I also tried the focused daemon test slice, but this stack branch is expected not to build standalone yet; the local failure is duplicate janitor/keeper symbols from the pre-integration split, not separate evidence beyond the wiring gap above.

figgyster (dismissed):
The new head still leaves the blocker intact.

The follow-up commit changed the durable task store identity handling, but the removal paths still call `store.RemoveWorkspace` directly (`workspace.go` / `workspacelayout.go`) and `snapshotWorkspaceContextOnRemove` is still only used by its helper tests. So real workspace removal can still delete `workspace_contexts` without writing the promised raw-tier `context.md` snapshot.

- Wire `snapshotWorkspaceContextOnRemove` into every live workspace-removal path after compaction cancellation/forgetting and before `store.RemoveWorkspace`.
- Cover a real removal path with a regression test that proves `.attn/raw/context-snapshots/<workspace>.md` is written before the context row disappears.

I did run `git diff --check` and `go test ./internal/tasks`; those pass. The blocker is the unfixed removal-path contract, not the task-store cleanup.

figgyster (approved):
That context checks out. I verified this slice really does not own `workspace.go`/`workspacelayout.go`, and the stack tip has the removal ordering wired as `forgetWorkspaceContextCompaction` → `snapshotWorkspaceContextOnRemove` → `store.RemoveWorkspace` at the removal sites. I also ran the focused stack-tip removal tests:

`go test ./internal/daemon -run 'TestSnapshotWorkspaceContext(AtRemovalSites|AtLoadTimeReap|SwallowsWriteFailure)'`

They pass. So my earlier blocker was judging an integration-slice seam as if it had to be self-contained in `kb/05`. That was the wrong cut line.

No blockers from me on this slice. The raw-tier helper/guard work is fine here; the callsite wiring belongs in the later integration slice, where it exists and is covered.

PR #352, opened 2026-06-19
feat(daemon): keeper journal narration

victorarias:
The two agentic narration tasks: `summarize_session` (cheap tier → per-session raw digest) and `narrate_workspace` (strong tier → curated `journal/<date>.md`), both carrying verbatim mechanics+expectations prompts. The removal pass writes a full retrospective and then **archives the workspace's `knowledge/projects/<slug>/` folder** to `archive/`, matched by a literal whole-line grep (`grep -Fx`) on the `resource: attn:workspace/<id>` marker so a short id can't substring-match a longer live workspace. Dormant until #8.
---

This is PR **6/10** — base `kb/05-keeper-compaction-capture`. Review per-diff. By design the refactor only activates in #8 (where the keeper is wired and the janitor/dreaming are deleted), so branches 2–9 are not individually `go build`-clean; the merged tip (#10) is build + vet + test green.

1. `feat/chifplace`
2. #347
3. #348
4. #349
5. #350
6. #351
7. **#352** 👈 current
8. #353
9. #354
10. #355
11. #356

### DIFF
--- a/internal/daemon/notebook_narration.go
+++ b/internal/daemon/notebook_narration.go
@@ -1,0 +1,701 @@
+package daemon
+
+import (
+	"context"
+	"crypto/sha256"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	agentdriver "github.com/victorarias/attn/internal/agent"
+	"github.com/victorarias/attn/internal/notebook"
+	"github.com/victorarias/attn/internal/tasks"
+)
+
+// Notebook narration: two headless agent tasks that turn raw session work into a
+// curated daily work-journal.
+//
+//   - summarize_session (cheap tier): per-session. Reads ONE session transcript and
+//     writes a faithful digest to the raw tier, partitioned by workspace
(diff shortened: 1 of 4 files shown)


figgyster (dismissed):
This slice is safe under the stack rules. The narration duties keep the risky parts in the right place: raw/session digests are workspace-scoped, removal narration gets the final retrospective path, traversal/prefix-collision cases are covered, and the file-as-ledger success checks avoid the classic “agent said done” trap.

I could only run proportionate checks here: `git diff --check`, `gofmt` verification, and the task-runner Meta tests passed. `go test ./internal/daemon` is still blocked by the known mid-stack duplicate-symbol churn on these keeper branches, so I’m not treating that as new evidence against this PR.

No blockers from me.

figgyster (approved):
Re-approving the restacked head. The PR patch is byte-for-byte the same as the slice I already reviewed once you ignore Git’s index/base weather; the new task-store identity fix lives in the updated base, not in this PR’s current diff.

I re-ran the cheap checks that matter here: `git diff --check`, `gofmt` verification, and the focused task-store/Meta regression tests all pass. Still no blockers from me.

PR #353, opened 2026-06-19
feat(daemon): daily-cron narrate backstop

victorarias:
The nightly per-workspace narrate backstop on the durable runner: a single cron enqueuer with a **relaxed daily-pass gate** (absent or unchanged journal block = clean done) plus an in-memory **activity gate** so only drained, still-live workspaces are narrated. Session-end and removal passes stay strict. Separate `NarrateCronState` anchor (shares only cadence/timezone settings). Dormant until #8.
---

This is PR **7/10** — base `kb/06-keeper-narration`. Review per-diff. By design the refactor only activates in #8 (where the keeper is wired and the janitor/dreaming are deleted), so branches 2–9 are not individually `go build`-clean; the merged tip (#10) is build + vet + test green.

1. `feat/chifplace`
2. #347
3. #348
4. #349
5. #350
6. #351
7. #352
8. **#353** 👈 current
9. #354
10. #355
11. #356

### DIFF
--- a/internal/daemon/notebook_cron.go
+++ b/internal/daemon/notebook_cron.go
@@ -1,0 +1,270 @@
+package daemon
+
+import (
+	"strings"
+	"time"
+
+	"github.com/robfig/cron/v3"
+	"github.com/victorarias/attn/internal/notebook"
+)
+
+// Notebook cron enqueuer.
+//
+// A single per-minute, timezone-aware cron tick drives the notebook's scheduled
+// background work. The tick is a thin schedule-and-dispatch loop: it decides what
+// is due and enqueues onto the durable runner (internal/tasks), which owns
+// execution, single-flight, retry/backoff, and crash recovery. Today it dispatches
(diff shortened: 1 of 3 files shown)


figgyster (dismissed):
This looks safe to merge for this stack slice. The cron backstop is nicely boxed: separate narrate anchor, activity-gated fan-out, and the relaxed daily-pass semantics don’t weaken the stricter session-end/removal paths.

I couldn’t run the daemon package tests end-to-end because this mid-stack branch is still intentionally not build-clean (`workspace_keeper`/janitor and dreaming/dispatch duplicate symbols from adjacent slices), but I did run the available notebook state test and `git diff --check`; both passed. No blockers from me.

figgyster (approved):
Re-approved for the rebased head. The PR patch is identical to the slice I already reviewed; the head only moved with the stack/base weather.

`git diff --check` and the notebook state test still pass. The daemon package still has the expected mid-stack duplicate-symbol build failure, so I’m not treating that as a blocker here.

PR #354, opened 2026-06-19
feat(daemon): wire the keeper; retire the janitor and dreaming

victorarias:
**Integration PR — this is where the refactor activates.** Bumps the protocol (task-status commands in; dreaming + notebook-CLI commands out), wires the keeper compaction/narration/cron onto the runner in `daemon.go`, and adds the task-status WS surface (`notebook_task_list`/`notebook_task_retry` + a payload-free `notebook_tasks_changed` broadcast). Deletes the superseded subsystems end-to-end: the in-daemon dreaming harvest/scheduler and the old `workspace_context_janitor` (both Go sides), plus the notebook `dreams` harvest core they fed. Mostly deletions; the merged tip is build + vet + test green.

> Owner action after this lands: `make install` (protocol bump → daemon reconnect; needs keychain).
---

This is PR **8/10** — base `kb/07-keeper-cron-archive`. Review per-diff. By design the refactor only activates in #8 (where the keeper is wired and the janitor/dreaming are deleted), so branches 2–9 are not individually `go build`-clean; the merged tip (#10) is build + vet + test green.

1. `feat/chifplace`
2. #347
3. #348
4. #349
5. #350
6. #351
7. #352
8. #353
9. **#354** 👈 current
10. #355
11. #356


figgyster (dismissed):
This one crosses the boring-but-real line: the PR as sliced doesn't compile against its own protocol removal.

`go test ./internal/daemon ./internal/protocol ./internal/client ./cmd/attn` fails because `internal/client/client.go` and `cmd/attn` still reference the removed notebook Unix-socket API (`NotebookInit`, `NotebookAppendJournal`, `NotebookDream*`, and the old command constants). The protocol version test also says the frontend hook is still on the previous protocol version.

Required before merge:
- remove/update the leftover client + CLI callsites for the deleted notebook commands in this PR, or keep the protocol types until the CLI/docs slice removes them
- regenerate/update the frontend protocol version/types with the daemon protocol bump

The keeper/task wiring itself looks like the right direction, but a stacked slice that leaves the package graph red is still a rake in the merge path.

figgyster (approved):
Thanks for spelling out the stack boundary — with that context, my earlier build blocker was too literal for this subsystem-sliced series.

I re-checked the new head and the follow-up task-store identity fix is solid: hex filenames avoid the lossy task-id collision, and `go test ./internal/tasks` passes. The current slice still intentionally isn't standalone-buildable, but I verified the claimed tip contract: the removed notebook symbols are gone on `kb/10-cli-guidance-docs`, protocol `113` matches frontend `113`, and `go test ./internal/protocol ./internal/client ./cmd/attn` is green there.

No blockers from me on this slice. The stack is doing weird stack things, but at least it brought receipts instead of vibes.

PR #355, opened 2026-06-19
feat(app): Notebook knowledge-base browser, task panel, keeper settings

victorarias:
Frontend for the epic: the PARA-grouped Notebook browser, the background **task panel** (lazily-fetched list + per-row retry, monotonic stale-guard), keeper settings, regenerated protocol types, and the `workspaceContextJanitor` → `workspaceContextKeeper` rename.
---
**Final-state re-split of the knowledge-base / keeper epic** (supersedes #335–#343, #345). The original stack was split by chronology and carried mid-epic churn; this one is sliced by subsystem from the net diff, so each PR is a clean final-state slice.

Stack: `1 task-engine` · `2 notebook-kb` · `3 agent-native` · `4 store-keeper` · `5 keeper-compaction` · `6 keeper-narration` · `7 keeper-cron` · **`8 integration`** · `9 frontend` · `10 cli/docs`

This is PR **9/10** — base `kb/08-protocol-daemon-integration`. Review per-diff. By design the refactor only activates in #8 (where the keeper is wired and the janitor/dreaming are deleted), so branches 2–9 are not individually `go build`-clean; the merged tip (#10) is build + vet + test green.

### [Stack](https://github.com/kitlangton/stack)

1. `feat/chifplace`
2. #347
3. #348
4. #349
5. #350
6. #351
7. #352
8. #353
9. #354
10. **#355** 👈 current
11. #356

PR #356, opened 2026-06-19
feat: remove the notebook CLI; native-file chief guidance; docs

victorarias:
The chief of staff now maintains the notebook and journal by **editing the markdown on disk directly** with native tools, so the `attn notebook` CLI verbs, their client methods, and the `contextjanitor` MCP subcommand are removed. Rewrites the chief guidance (always-injected `hooks.go` + the `notebook.md` skill ref) to native-file editing over `/knowledge` paths, including the workspace→project **archive-stamping** convention (`resource: attn:workspace/<id>`) in the *always-injected* guidance so the archive-on-close feature actually fires. Adds `docs/glossary.md`, the code-grounded `docs/notebook-processes.md` process map, and the plan docs.

Existing on-disk `memory/` notes are intentionally **not** migrated — they remain readable and group under the generic "Notebook" bucket until relocated. Nothing is deleted.
---

This is PR **10/10** — base `kb/09-frontend`. Review per-diff. By design the refactor only activates in #8 (where the keeper is wired and the janitor/dreaming are deleted), so branches 2–9 are not individually `go build`-clean; the merged tip (#10) is build + vet + test green.

1. `feat/chifplace`
2. #347
3. #348
4. #349
5. #350
6. #351
7. #352
8. #353
9. #354
10. #355
11. **#356** 👈 current


figgyster (dismissed):
This looks safe to merge. Removing the notebook CLI lines up with the new filesystem-canonical Notebook model, and the chief guidance now points at the actual `/knowledge` paths instead of a dead command surface.

`go test ./...` is green here. No blockers from me.

figgyster (dismissed):
I can’t re-approve this head yet. The PR patch itself is unchanged from the head I approved, but the restack moved the base and the current merge head now fails the principal Go suite.

`go test ./...` fails in `internal/daemon` because websocket tests read a `notebook_tasks_changed` broadcast where they expect their own event:

- `TestDaemon_MuteRepo_ViaWebSocket`
- `TestDaemon_InitialState_IncludesRepoStates`
- `TestDaemon_StateChange_BroadcastsToWebSocket`

The focused packages touched by this PR still pass (`./cmd/attn ./internal/client ./internal/hooks ./internal/agent ./internal/tasks`), so this smells like base/stack weather rather than the CLI/docs diff wearing a fake mustache. But the current head is what GitHub will merge, and right now that head is not test-green.

Required before merge:
- fix or quarantine the `notebook_tasks_changed` websocket test interleaving so `go test ./...` passes on this head.

figgyster (approved):
This fixes the blocker cleanly. The websocket tests now wait for the event they actually care about instead of assuming the next frame belongs to them — exactly the right treatment for a broadcast bus with background task noise.

The targeted daemon tests pass, and `go test ./...` is green on this head. No blockers from me.

PR #357, opened 2026-06-19
feat(workflow): portable durable multi-agent workflow engine (gated, off by default)

victorarias:
## What

Adds a portable, durable multi-agent **workflow engine** to attn — Claude Code's Workflow capability re-implemented to run *inside* attn. A workflow is a deterministic JavaScript script that orchestrates one or more headless sub-agents (codex/claude); attn durably journals every step so a run can be observed live, retrieved later, and resumed.

**Gated OFF by default** behind the new `workflows_enabled` setting (Settings → Agents → Workflows). With it off, attn won't start a workflow and agents aren't told the capability exists — so this lands dormant.

## User-visible surface

- `attn workflow run|show|list|result|cancel` CLI, with an IPC journal to the daemon.
- A read-only **Workflow run view** docked in the app, showing the in-flight call as live progress.
- Opt-in model: an agent only starts a workflow when the user asks per task (`attn workflow …`) or per session (`hypercode`). Workflow sub-agents never receive the guidance, so workflows can't nest.

## Shape (the commits read as a clean sequence)

- **Design** — portable engine design doc.
- **Engine (E1–E6)** — goja realm + determinism bans + watchdog; real headless `agent()`; writable headless exec + MCP attach; fan-out concurrency cap + ordinal stability; `isolation:'worktree'` over attn's git-worktree infra.
- **Plumbing (S-\*)** — protocol `WorkflowRun*` models + **ProtocolVersion bump**; store durable journal + **migration 50**; daemon IPC + coalesced broadcast + attention; CLI verbs; read-only UI mount.
- **Guidance + gate** — settings-gated opt-in trigger guidance (developer-instructions / `--append-system-prompt`, no SessionStart hook); the `workflows_enabled` **master switch** (daemon refuses new runs at the socket entry when off; in-flight runs unaffected).

## Notes for review

- **ProtocolVersion bumped** — the daemon survives app rebuilds, so the bump forces a clean reconnect instead of a silent old-daemon/new-app mismatch.
- **DB migration 50** — adds the durable workflow journal. Independent of the keeper stack's migration 51.
- ~14k lines, but additive and cohesive: most of it is the new engine plus its tests, with a thin, gated touch on existing launch/settings paths.

## Verification

- `go build ./...`, `go vet`, full `go test ./...` (1427 tests), gofmt — all green.
- Frontend `tsc` + vitest — green.
- **Live dev-daemon** end-to-end: the master switch refuses `attn workflow run` when off (exit 1) and completes a run when on (exit 0); a real multi-call run journals and reports through `workflow show`.

CHANGELOG updated.


figgyster (dismissed):
This is a strong foundation, but cancel is currently a paper button for any run that is sitting inside `agent()`.

Required before merge:
- Thread the workflow run context into live agent dispatch, and make the event loop wake/return when that context is canceled while it is blocked waiting for worker results.

Evidence: `startCancelWatcher` cancels the root context when the daemon marks a run canceled, but `driverAgent.Run` immediately replaces it with `context.Background()`, so real headless Codex/Claude subprocesses never see cancellation. Worse, while the top-level workflow is awaiting `agent()`, `eventLoop.pump` is parked on `<-el.jobs`; `vm.Interrupt` from the cancel watcher only interrupts JS execution, not that mailbox wait. I verified this with a throwaway focused probe: `await agent("x")` with a blocking stub still failed to return 500ms after `cancel()`.

The normal checks I could run are otherwise healthy: `git diff --check`, targeted Go packages (`internal/workflow`, `internal/daemon`, `internal/store`, `internal/agent`, `cmd/attn`, `internal/workflowresult`), and the focused frontend workflow/socket/view tests all pass. So this is not a taste thing; it is the advertised durable workflow control surface not being able to stop live subagents. Tiny lever, load-bearing wall.

figgyster (dismissed):
The latest push doesn’t touch the cancellation path, so the blocker is still live on this head.

The only workflow-related changes since my previous review are formatting/alignment in `cmd/attn/workflow.go`, `internal/daemon/daemon.go`, and `internal/workflow/driveragent.go`. I reran the focused cancellation probe against `4dd49cb8`: `await agent("x")` with a blocking stub still does not return after the root context is canceled.

- Thread the workflow run context into live agent dispatch.
- Make the event loop wake/return when cancellation happens while it is waiting for an `agent()` result.

`git diff --check` and the targeted Go packages still pass, but the advertised `attn workflow cancel` contract remains broken. Same rake, new puddle.

figgyster (dismissed):
This fixes the blocker cleanly. The run context now reaches live `agent()` dispatch, the event loop wakes while parked on the agent await, and the new regression test pins the exact deadlock case instead of waving incense at it.

I re-ran the focused workflow/daemon/store/agent/CLI packages plus the workflow UI/socket tests locally; all passed. No blockers from me.

figgyster (approved):
Still looks good on the latest head. This follow-up mostly tightens the workflow boundary: engine mutation stays on the Unix-socket engine path, the UI is read/cancel only, the sink registry is bounded at terminal/cancel edges, and the old workflow test helper duplication got pulled into the real run path. That is the right direction.

I reran the focused workflow/daemon/store/agent/CLI packages and the workflow UI/socket tests locally; all passed. No blockers from me.

PR #358, opened 2026-06-19
refactor: post-review cleanup — dedup, dead code, comment fixes

victorarias:
## Post-review cleanup (kb/11)

Applies the maintainability-review recommendations over the notebook/keeper epic's net diff. **Net −33 lines, no protocol/version change.** Stacked on `kb/10`; builds standalone on the tip, so it's fully Figgy-reviewable. The 10 approved slices (#347–#356) are left byte-identical.

### Plan-check on dead code (the load-bearing step)

Before deleting anything, each dead/no-op item was cross-referenced against `docs/plans/2026-06-14-notebook-narration.md` and `docs/plans/2026-06-18-knowledge-base.md`. Four were genuine cruft / already-planned removal; **one was a planned behavior left unwired** and is now completed rather than deleted:

- **`notebookCronInterval`** — a test-only seam never assigned anywhere; tests inject `now` into the tick directly, and the plan never wanted a tunable production cadence. Removed.
- **`refreshCleanWorkspaceContextCheckouts`** — an empty no-op documenting the *deliberate* decision not to rewrite agent-owned checkouts (confirmed by both plans). Removed; rationale relocated to the call sites.
- **startup-reaper final narrate (finding #3)** — plan §4 wants every removal path, including startup reconciliation, to enqueue a final retrospective, but the call was a guaranteed nil-runner no-op (load runs ~150 lines before `startCompactRunner`). **Now wired:** `loadWorkspacesFromStore` returns the reaped ids; `Start` enqueues them after the runner exists, matching the live removal paths (which already narrate post-removal from snapshot + transcripts).

### Dedup
- `tearDownRemovedWorkspace` — the byte-identical 6-statement teardown at the three live removal sites (`workspace.go` ×2, `workspacelayout.go`). The startup reaper deliberately stays separate (omits prune + broadcast).
- `Runner.fenceAndWait` — shared by `Cancel` and `cancelActive`, in the caller-holds-`mu` form (no `mu` drop/re-take, so the commit-fence / blocks-until-exit contract is preserved).
- `renameSettingKey` — shared by the keeper-compact and notebook-cron settings renames (exact copy-forward + log format preserved).
- `notebookSessionDigestPath` now reuses `notebookWorkspaceSessionsDir` for the per-workspace bucket.

### Comments / micro-simplify
runner tuning-defaults comment · CommitGuard `committed` ref · `narrationTierDefault` redundant case · `formatNextAttempt` dead `now` param · NotebookBrowser two-effects → one (drops eslint-disable) · headless_test helper reuse · `assertJournalUntouched` doc.

### Deferred (not in this PR)
- `Response.notebook_tasks` field removal — leftover from the removed `attn notebook tasks` CLI. Needs a `ProtocolVersion` bump + `make generate-types` + `make install`; kept out so this PR stays protocol-neutral. Batch it with the post-merge `make install`.

### Verification
`go build ./...` · `go vet ./...` · `go test ./internal/{tasks,agent,daemon}` (full daemon suite, 123s) · app vitest (950) · `tsc --noEmit` — all green.

— on behalf of Victor

### DIFF
--- a/app/src/components/NotebookBrowser.tsx
+++ b/app/src/components/NotebookBrowser.tsx
@@ -378,20 +378,13 @@ export function NotebookBrowser({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [changeSignal]);
 
-  // Fetch tasks when the Tasks section is opened (and the browser is open). A bump
-  // of taskChangeSignal while it's open refetches via the effect below.
+  // Fetch tasks while the Tasks section is open (and the browser is open): once on
+  // open, and again whenever a notebook_tasks_changed broadcast bumps taskChangeSignal
+  // so runner transitions show without reopening the section.
   useEffect(() => {
     if (!isOpen || !tasksOpen) return;
     void refreshTasks();
-  }, [isOpen, tasksOpen, refreshTasks]);
-
-  // Live refresh: a notebook_tasks_changed broadcast refetches the task list while
-  // the section is open so runner transitions show without reopening it.
-  useEffect(() => {
-    if (!isOpen || !tasksOpen || taskChangeSignal === 0) return;
-    void refreshTasks();
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [taskChangeSignal]);
+  }, [isOpen, tasksOpen, refreshTasks, taskChangeSignal]);
 
   // Drop the staleness token when the browser closes so an in-flight tasks fetch
   // can't stamp rows onto a reopened panel.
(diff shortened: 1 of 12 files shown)


figgyster (approved):
This looks safe to merge. The cleanup keeps the important boundaries intact: startup-reaped workspaces now wait until the runner exists before final narration is queued, while the live removal teardown stays shared without dragging startup-only behavior into it.

I also ran the local proof: `git diff --check`, `go test ./...`, targeted daemon/tasks/agent tests, app Vitest (950 tests), and `tsc --noEmit` all passed. No blockers from me.

PR #359, opened 2026-06-19
refactor(protocol): drop the unused Response.notebook_tasks field

victorarias:
## Drop the unused `Response.notebook_tasks` protocol field (kb/12)

The deferred item from #358. The `Response` envelope carried `notebook_tasks?: NotebookTask[]` — a leftover from the removed `attn notebook tasks` CLI. The task surface is **websocket-only** (`NotebookTaskListResultMessage`), so the field has **no producer or consumer**: no Go or TS code reads `Response.NotebookTasks` (verified by grep across the tree, incl. tests).

Isolated in its own PR (stacked on `kb/11`) because it touches generated code + the version gate — kept out of #358 so that one stayed protocol-neutral and standalone-buildable.

### Change
- Remove `notebook_tasks` from `Response` in `main.tsp`. **Keep** the `NotebookTask` model (used by the WS task panel) and the `notebook_tasks_changed` broadcast.
- `ProtocolVersion` **113 → 114** + frontend `PROTOCOL_VERSION` mirror.
- Regenerated `generated.go` / `generated.ts` (`rm -rf tsp-output` first to avoid the orphaned-json-schema glob trap). Net −3 Go / −2 TS lines, no other schema churn.

### Owner step
`make install` (needs keychain) so the daemon adopts protocol 114 — fold this into the same install the epic already needs for the 113 bump.

### Verification
`go build ./...` · `go vet ./...` · `go test ./internal/protocol` + scoped daemon · full `go test ./internal/daemon` · `tsc --noEmit` · app vitest (950) — all green.

— on behalf of Victor

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -170,7 +170,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '113';
+export const PROTOCOL_VERSION = '114';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 interface PRActionResult {
--- a/app/src/types/generated.ts
+++ b/app/src/types/generated.ts
@@ -2753,7 +2753,6 @@ export interface Response {
     notebook_entries?:                     NotebookEntryElement[];
     notebook_guide?:                       NotebookGuide;
     notebook_read?:                        NotebookReadObject;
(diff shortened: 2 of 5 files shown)


figgyster (approved):
This is safe to merge. The removed field was only on the generic response envelope; the live notebook task surface still has its dedicated websocket result + change broadcast, and the protocol bump is mirrored on both sides.

I ran the focused protocol/daemon checks locally (`go test ./internal/protocol`, targeted notebook daemon tests, `go build ./internal/protocol ./internal/daemon`, `git diff --check`). GitHub didn’t report checks for this branch from here, but I don’t see a blocker.

PR #360, opened 2026-06-20
feat(notebook): Notebook knowledge-base + keeper (collapsed epic)

victorarias:
## What

Collapses the fully-reviewed 12-PR Notebook epic (#347–#359, `kb/01`–`kb/12`) into a single commit and lands the whole **Notebook knowledge-base + keeper** feature against `main`, rebased onto the gated workflow-engine work (#357).

The Notebook is a filesystem-canonical, profile-wide markdown memory layer for attn (dated journals + a PARA/OKF knowledge base), maintained by **the keeper** — one persona that owns automated workspace upkeep: context compaction (`compact_context`) and journal narration (`summarize_session` → raw tier, `narrate_workspace` → curated journal), all on a durable file-backed task runner with retry/backoff/coalesce/recovery.

## Conflict reconciliation (workflow engine × keeper)

`main`'s workflow engine and the keeper's headless-agent refactor both touched the same core. They now **coexist**: `RunHeadlessTask` dispatches keeper = native-tools path (no MCP) vs workflow = MCP-config path via `usesNativeToolsPath()`.

- Protocol bumped to **115**.
- Keeper-rename migration moved **51 → 52** (main owns 49–51 incl. its workflow tables); `applyMigration52` is `tableExists`-guarded — clean rename on a janitor-table DB, safe no-op on an already-renamed one.

## Verification

Driven to an **honest all-pass UAT** on a fresh dev build (protocol 115 / schema 52) with **two independent adversarial review rounds agreeing**. Real artifacts for scaffold/guidance/dispatch-capture (exactly-once across restart)/snapshot-floor/task-engine/migration; grounded summarize→narrate (Claude **and** Codex); keeper compaction + byte-exact rollback; cross-day / one-per-day / relaxed-vs-strict gate / path-traversal / cross-workspace guards (Go tests). Runbook + evidence in `docs/notebook-uat.md`.

`go build ./...`, `go vet`, full `go test ./...`, `tsc`, and the frontend suite are green.

## Two honest residual limitations (disclosed, not defects)

1. **Secret scrubbing is best-effort prompt-only** — no deterministic scrubber before the durable write, and the journal root is externally synced. A canary showed literal example tokens can land verbatim. The CHANGELOG's absolute *"never include secrets"* is softened to best-effort intent here; a token-shaped-string scrubber is the recommended follow-up hardening.
2. **Narration success gate is content-blind** — verifies the journal block changed, not that it is grounded. The agent-session env-leak fix (inherited from `main`'s `agentenv.go`) removes the live trigger; asserting non-sentinel / re-resolving the transcript path is the recommended follow-up.

## Notes

- The reviewed `kb/01`–`kb/12` PRs (#347–#359) targeted `feat/chifplace` and are superseded by this single PR; they can be closed.


figgyster (dismissed):
I found a few Notebook root boundary issues that cross the line from “rough edge” into “this can break or leak the filesystem contract.” The feature is solidly exercised otherwise, but these need fixing before merge.

Required before merge:

- Normalize the configured notebook root before constructing the store. A root entered with a normal trailing slash (for example `/Users/me/Notebook/`) makes regular paths like `index.md` fail the containment check because `Store.abs` compares against the raw root string and expects the prefix `/Notebook//`. That turns a perfectly reasonable settings value into “escapes the notebook root.” Clean the root in `notebookRoot()` or `notebook.NewStore`, and ideally use `filepath.Rel`-style containment rather than raw string prefixes.

- Apply the symlink containment guard in `Store.List` before reading frontmatter. `Read`/`Write` defend against symlink escapes with `checkWithinResolvedRoot`, but `List` walks the root and calls `readPrefix(p, ...)` directly. A synced/user-editable notebook root can contain `linked.md -> /outside/private.md`; the list endpoint will read and expose that file’s title/summary metadata over the websocket.

- Add symlink-aware containment to raw-tier writes. `writeRawAtomic` validates the leaf id and lexical parent, but it writes directly under `.attn/raw/...` without resolving ancestors. If a user/sync client turns `.attn/raw/dispatches` or another raw subdir into a symlink, daemon-generated raw captures can be written outside the notebook root. The comments promise containment; the current check only proves the string join stayed under `dir`.

Validation I ran: `go test ./internal/daemon ./internal/notebook ./internal/tasks ./internal/store ./internal/protocol`, `git diff --check`, and the targeted Notebook/settings frontend tests. Those passed; the blockers above came from focused path-boundary probes and code review. Tiny rake, but it’s sitting exactly where the filesystem contract walks.

figgyster (commented):
The filesystem-boundary fixes address the blockers I raised: configured roots are normalized, `List` now uses the same resolved-root guard before reading frontmatter, and raw-tier writes reuse that symlink-aware containment before writing.

I’m not flipping this green yet because the visible **Daemon** CI check is failing on the current head, and GitHub won’t expose the job log until the workflow finishes. Local focused validation is clean here:

- `go test ./internal/notebook ./internal/daemon ./internal/tasks ./internal/store ./internal/protocol`
- `git diff --check`
- targeted Notebook/settings frontend tests

So: the old Figgy blocker is fixed, but merge should wait for the failing Daemon check to resolve or show its real failure. GitHub weather, wearing steel-toed boots.

figgyster (approved):
This looks safe to merge now. The Notebook boundary fixes held, and the dispatch journal change closes the CI race the right way: durable-before-ack, with the raw file written before the socket response returns.

Local focused daemon/notebook validation is green, and the visible Daemon/Linux checks are green now. Frontend was still pending at my last fetch, but this last push only touched daemon dispatch journaling/tests, so I’m not holding the PR hostage for unrelated queue weather.

No blockers from me.

PR #361, opened 2026-06-20
feat(notebook): Notebook folder picker in Settings

victorarias:
## What

Adds a **Notebook Folder** picker under Settings → General so you can relocate attn's durable Notebook (dated journals + knowledge base) from the app.

## Why

PR #360 introduced the `notebook.root` setting with full backend support — `validateNotebookRoot` (absolute path, `~/` expansion, refuses paths inside `~/.attn`), and live re-pointing (`notebookStoreFor()` rebuilds the store and `ensureNotebookWatcher()` follows the new root). But the slice shipped **backend-only**: no UI to set it and no changelog line, so an app user couldn't actually change the folder. This closes that gap.

## Changes

- **SettingsModal**: a Notebook Folder block under General, mirroring the existing Projects Directory picker (text input + Browse via Tauri `open`), wired to `notebook.root`. Blank commits an empty override, which the daemon resolves back to the per-profile default.
- **Daemon**: surfaces a read-only `notebook.root.effective` in the settings payload (computed, never stored, rejected by `set_setting`) so the picker shows where the Notebook currently resolves to even when the override is blank.
- **CHANGELOG**: entry for the relocatable Notebook folder.

## Behavior decision

Changing the folder **re-points** attn at the new location (matches the daemon's existing lazy-rebuild behavior); existing notes are **not** moved. The UI says so — move/sync the folder yourself if you want the current contents to come along. (Move/copy-on-change could be a follow-up.)

## Verification

- `go build ./...`, `go vet`, targeted daemon tests green (`TestNotebookRootEffectiveSurfacedReadOnly`, `TestValidateNotebookRoot`, `TestNotebookRootResolution`, `TestNotebookWatcherFollowsRootChange`).
- Frontend: 17 `SettingsModal` tests pass incl. 3 new ones — render shows override + effective folder, placeholder falls back to the effective default, persist-on-blur and clear-to-default both call `onSetSetting('notebook.root', …)`.
- `tsc` clean. Dev app (`make dev`) rebuilt + reinstalled.

Verified via the rendering tests (which exercise the exact DOM path nav→General→input→effective→persist/clear) rather than a pixel screenshot — the block reuses the Projects Directory CSS classes verbatim. No protocol bump: `notebook.root.effective` is an additive key on the open settings map and degrades gracefully against an old daemon/app.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-06-20]
+
+### Added
+- **Choose where your Notebook folder lives.** Settings → General now has a Notebook Folder picker, so you can point attn's durable Notebook — your dated journals and knowledge base — at any folder you own instead of the default `~/attn-notebook` (separate per profile). Browse to a folder or type a path, and the picker shows where the Notebook currently resolves to; leave it blank to fall back to the default. attn starts using the new folder right away — your existing notes aren't moved, so move or sync the folder yourself if you want the current contents to come along.
+
 ## [2026-06-19]
 
 ### Added
--- a/app/src/components/SettingsModal.test.tsx
+++ b/app/src/components/SettingsModal.test.tsx
@@ -662,3 +662,81 @@ describe('SettingsModal review loop prompts', () => {
     expect(screen.getByTestId('settings-context-keeper-model-custom')).toHaveValue('gpt-custom');
   });
 });
+
+describe('SettingsModal notebook folder', () => {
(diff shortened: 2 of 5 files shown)


figgyster (approved):
This looks safe to merge. The backend keeps `notebook.root.effective` computed/read-only, the Settings UI mirrors the existing folder-picker pattern, and the copy is honest about re-pointing rather than moving notes — exactly the kind of small sharp edge that otherwise becomes user-data folklore.

I verified the focused daemon tests, the SettingsModal suite, `tsc`, `go vet`, `go build ./...`, and `git diff --check` locally. GitHub CI had the frontend job still running on my last fetch, but the principal frontend checks passed here.

PR #363, opened 2026-06-20
feat(keeper): configurable background duties + master on/off toggle

victorarias:
## What

Makes the keeper's three async background duties configurable and adds one master switch to turn all of them on/off together.

- **`notebook.tasks_enabled` master switch** (default ON; blank/unset => enabled, only explicit `false` disables). Gates every *background* enqueue and executor chokepoint for all three duties — `summarize_session`, `narrate_workspace` (incl. daily + removal-final paths), and `compact_context`. Flipping it off halts new keeper work and retires any already-queued run as a no-op success. The user-triggered inline/manual `attn workspace context compact` command stays **ungated** (explicit action, not background work).
- **Per-duty agent/model pickers** for Session summaries (default Claude Haiku) and Journal narration (default Claude Sonnet). The daemon already accepted `notebook.summarize_session` / `notebook.narrate_workspace`; they just had no UI.
- **Unified Settings → Agents "Keeper" section**: master toggle + one row per duty. Compaction keeps its opt-in *Disabled* affordance (blank = off); the always-on duties offer *Use default* to clear an override back to the tier default.

The daemon normalizes the master switch to its effective value in the settings payload, so the UI toggle reflects default-ON rather than reading a missing key as off.

## Why

The keeper ran off hardcoded defaults with the only UI being compaction's picker. There was no way to pick cheaper/stronger models per duty, and no single way to pause background keeper spend.

## Behavior preserved

No change to compaction's opt-in semantics or its presets/defaults (claude→opus, codex→gpt-5.4) — its row reuses the same setting key + testids, so the existing tests pass unchanged.

## Tests

- Go: `internal/daemon/notebook_tasks_enabled_test.go` — default-ON spellings, enqueue-skip when off, executor no-op for an already-queued run when off. Full keeper/narration/settings suite (168 tests) green; `go vet` clean.
- Frontend: 5 new `SettingsModal` keeper tests (master toggle on/off, always-on tier defaults + save override, no Disabled option on always-on rows vs. compaction, Use-default clears). Full suite 1017 green; `tsc` clean.
- Built the dev app (real bundle compiles).

🤖 Authored by Claude on behalf of Victor.


figgyster (approved):
This looks safe to merge. The keeper controls now line up across the daemon and Settings UI: default-on master switch, per-duty model overrides, and executor-side no-op retirement for work queued before the switch flips off.

I ran the targeted daemon keeper tests, the SettingsModal test slice, `git diff --check`, and frontend `tsc --noEmit`; all passed. GitHub’s Frontend check was still pending on my last poll, but the local frontend path is clean.

No blockers from me.

PR #364, opened 2026-06-20
feat(notebook): single live-preview editor (read + type in one surface)

victorarias:
## Stage 1 of the notebook UI rebuild

Toward the panel-based prototype (`docs/prototypes/kb-markdown-ui.html`); gap map in `docs/plans/2026-06-20-notebook-ui-gap.md`. Each stage is a small PR that's manually verifiable on its own.

**This stage: one mode — read and type in the same surface.** The notebook's View/Edit toggle and the rendered-markdown/textarea split are gone. A note now opens into a single CodeMirror 6 **live-preview** editor:

- Markdown renders inline as you read — headings sized, **bold**/*italic*/`code`/~~strike~~ styled, links shown as text — with the raw syntax markers hidden.
- The line your cursor is on reveals its raw markdown so you can edit it (Obsidian behavior). An **unfocused** editor renders fully clean.
- ⌘/Ctrl-click a wiki link to follow it; selecting text floats the existing **Send to chief** action.
- Edits **autosave** (debounced, plus a flush when you navigate away or close) over the same hash-CAS write + on-disk **conflict reconcile** as before. No Save button, no Cancel.
- Notes stay plain canonical `.md` (the source of truth is raw markdown), honoring the external-sync invariant.

Layout is otherwise unchanged (sidebar + document pane); backlinks moved to a foot strip for now (stage 3 relocates them to a right rail).

### Verify
Open a note → it renders **and** you can click anywhere and type → edits persist; there is no mode toggle.

### Testing
CodeMirror can't mount under happy-dom, so coverage is layered:
- **Headless unit** (`liveMarkdownPreview.test.ts`): decoration logic over an `EditorState` (heading sizing, marker hide/reveal by active line + focus, link href).
- **Unit, editor mocked** (`NotebookBrowser.test.tsx`): autosave/flush/conflict/navigation/selection orchestration.
- **Playwright harnesses** (real browser): `live-markdown-editor.spec.ts` (live preview render, typing, mod-click follow, selection) and `notebook-browser.spec.ts` (full modal, no toggle, autosave). Screenshots attached in `test-results/`.

All 1025 unit tests + the new e2e specs pass; `tsc` clean.

### Deferred (stage 4 polish)
List bullets/checkboxes and fenced-code styling still show raw (`- `, ```` ``` ````).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-06-20]
 
+### Changed
+- **Read and edit Notebook notes in one place — the Edit mode is gone.** Opening a note now shows it as rendered markdown you can type straight into, the way Obsidian's live preview works: headings, **bold**, *italic*, `inline code`, and links render as you read, and the raw markdown of the line your cursor is on reveals itself so you can edit it. There's no more View/Edit toggle, and your changes save automatically (no Save button) — you still get a heads-up if the note changed on disk underneath you, and notes stay plain `.md` files.
+
 ### Added
 - **Choose where your Notebook folder lives.** Settings → General now has a Notebook Folder picker, so you can point attn's durable Notebook — your dated journals and knowledge base — at any folder you own instead of the default `~/attn-notebook` (separate per profile). Browse to a folder or type a path, and the picker shows where the Notebook currently resolves to; leave it blank to fall back to the default. attn starts using the new folder right away — your existing notes aren't moved, so move or sync the folder yourself if you want the current contents to come along.
 - **Configure the keeper's background work — and switch it off with one toggle.** Settings → Agents now has a single Keeper section that gathers all of the keeper's background duties: session summaries, journal narration, and workspace-context compaction. Each duty has its own agent and model picker (with recommended defaults — Haiku for summaries, Sonnet for narration), so you can dial cost and quality per duty. A new **Background tasks** master switch at the top turns every keeper duty on or off together; while it's off the keeper queues and runs no background work, though your picks stay saved, and turning it off won't interrupt a run already in flight.
--- a/app/e2e/live-markdown-editor.spec.ts
+++ b/app/e2e/live-markdown-editor.spec.ts
@@ -1,0 +1,91 @@
+import { test, expect } from '@playwright/test';
+
+// The notebook's single read-and-type surface (CodeMirror live preview), rendered
+// in isolation by the component harness in a real browser — CM cannot mount under
+// happy-dom, so this is where its rendering and interactions are verified.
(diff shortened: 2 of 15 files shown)


figgyster (dismissed):
This is the right product direction — single live surface, plain markdown source, and the CodeMirror decoration tests are doing real work. But there’s one data-loss edge in the new autosave/navigation path that crosses the line.

Required before merge:

- Navigation/close flushes must preserve the same conflict-reconcile guarantee as the debounced autosave. Right now `flushSaveRef.current()` writes the outgoing note and drops every result; if the CAS write conflicts, the UI navigates away and replaces the buffer, so the user’s last edits can disappear without the conflict banner ever being shown. The PR explicitly promises the same hash-CAS + on-disk conflict reconcile as before, and this path is exactly where that promise matters most.
- While you’re in that area, put the same stale-navigation guard on the conflict “Reload from disk” path that `loadNote`/`saveDraft` already use. It currently awaits `readNotebook(path)` and then stamps `note`/`draft` unconditionally, so a slow reload from note A can overwrite the editor after the user has moved to note B.

I couldn’t rerun the frontend tests locally because the runner blocked my ephemeral pnpm command shape, but a parallel inspection did validate the live-preview CodeMirror path (`liveMarkdownPreview`, `tsc`, and the two new Playwright specs) cleanly. The blocker is narrower: don’t let conflict handling fall out of the car during the no-save-button navigation flow.

figgyster (dismissed):
The fix does what it needed to do. Navigation and close now route through the same conflict-aware persist path, conflict/error outcomes keep the editor open/on the current note, and `reloadFromDisk` has the same stale-note guard as the other async paths.

Frontend CI is green now, and `git diff --check` is clean. I still couldn’t run the local pnpm suite from this webhook runner because the command guard is being weirdly overprotective, but the code and new regression tests line up with the contract.

No blockers from me.

figgyster (dismissed):
Still safe to merge. This follow-up is just the CodeMirror theme escape hatch: `theme="none"` lets the custom transparent editor theme own the surface instead of @uiw's default light background painting the document pane white.

I rechecked the current head: TypeScript passes, `git diff --check` is clean, and the app Vitest run passed locally (1022 passing, 5 skipped). Frontend CI was still pending at my last fetch, but the changed surface is covered by the local principal checks.

No blockers from me.

figgyster (approved):
Still good. This follow-up tightens the live-preview theme instead of changing the notebook persistence contract: the editor now owns the CM cursor/selection colors and the harness forces real app tokens, so the dark-pane rendering regression has a proper browser-level guard.

I rechecked the current head: TypeScript passes, `git diff --check` is clean, the app Vitest run passed locally (1022 passing, 5 skipped), and the focused Playwright live-editor spec passed (5/5). GitHub Frontend CI was still pending at my last fetch; Changes is green.

No blockers from me.

PR #365, opened 2026-06-20
fix(notebook): render a clicked note instantly, decoupled from backlinks

victorarias:
## What

Clicking a note in the Notebook updated the tree selection immediately but left the editor showing the **previous** note's content until the read finished — selection and content visibly out of sync.

## Root cause

`loadNote` awaited `Promise.allSettled([readNotebook, backlinksNotebook])` and applied the content only after **both** resolved. Backlinks is far slower than a file read — the daemon's `Backlinks` walks every note in the notebook and reads each body to scan for links (O(all-notes)). So the fast content render was hostage to the slow backlinks walk.

## Fix

Decouple them: apply the note content the moment its read resolves; load backlinks independently and non-blocking. Both keep the load-token staleness guard so a superseded navigation is dropped.

## Tests

- New regression test: the clicked note's content renders while its backlinks fetch is still pending (deferred mock), and backlinks fill in afterward.
- Full suite green: `tsc` clean, 1028 unit tests pass.

Manually verified in the dev app — switching notes now feels instant.

_Authored by Claude on behalf of Victor._


figgyster (dismissed):
This fixes the editor lag, but it leaves the same stale-context bug in the backlinks panel.

When a previously loaded note already has backlinks, `loadNote()` now starts the next `backlinksNotebook(path)` request without clearing the old `backlinks` array. Since the backlinks walk is intentionally slow, the newly selected note can render with the previous note’s “Linked from” list until the new backlink request resolves. That’s a smaller panel than the editor, sure, but it is still selected-note metadata lying under the new selection — same family of bug, different hat.

Required before merge:
- Clear backlinks when a new note load starts (or show an explicit backlinks-loading state), so the visible backlinks never belong to the previous selected note.
- Add the regression case where note A’s backlinks have already rendered, then note B is clicked while B’s backlink fetch is pending.

I verified the current patch locally with `pnpm exec vitest run src/components/NotebookBrowser.test.tsx` and `pnpm run build`; both pass. CI’s Frontend job was still pending at my last fetch.

figgyster (approved):
This fixes the blocker cleanly. The backlinks panel now has its own loading state, clears the outgoing note’s relationships immediately on navigation, and the regression covers the exact old-links-visible-while-new-walk-pends case.

Local checks passed: `pnpm exec vitest run src/components/NotebookBrowser.test.tsx` and `pnpm run build`. GitHub checks weren’t reported for the branch at my last fetch, but the frontend path is covered locally.

No blockers from me.

PR #366, opened 2026-06-20
feat(fs): generic filesystem daemon surface (fs_list/read/write + fs_changed)

victorarias:
## What

Slice A of the **fs-based notebook pivot**: a generic filesystem daemon surface that the redesigned notebook UI will render a real folder tree over. This PR is **daemon-side only** (Go + protocol) — no UI yet.

Adds a raw filesystem view over the **same configurable root** as the notebook (the already-live `notebook.root` setting), exposed on the websocket:

- `fs_list` — shallow, one directory's immediate children (files + subdirs), directories-first. A tree UI expands lazily, one `readdir` per node.
- `fs_read` — any file's bytes + content hash.
- `fs_write` — create / hash-CAS edit, the **same conflict-reconcile contract as `notebook_write`**.
- `fs_changed` — broadcast on changes (origin `agent|ui|external`).

Unlike `notebook_*`, this surface is **structure-blind**: no `.md`/PARA/frontmatter awareness — it surfaces every file and directory exactly as it sits on disk. The curated notebook view stays; `fs_*` is the raw layer beneath it.

## Design notes

- **`internal/fsdoc`** reuses the notebook's vetted `EnsureWithinResolvedRoot` symlink-containment guard and `Hash` — no duplicated security logic. It defines its own generic path cleaner (no `.md` restriction).
- **One filesystem, one watcher.** The single shared root watcher (started lazily by the notebook path) now feeds *both* `notebook_changed` and `fs_changed`. `fs_write` records a self-write (so its own edit isn't echoed as external) and broadcasts `fs_changed(ui)` — mirroring `notebook_write`. It also normalizes the echoed/broadcast path so result/event paths match what `fs_list` returns.
- **Known limit (documented in code):** the watcher only surfaces `.md` paths today (`notebook.CleanPath` gates it), so a non-`.md` external edit doesn't yet emit `fs_changed`. Generalizing the watcher to all files lands with the non-text editor work.
- `ProtocolVersion` 115 → 116 (+ matching frontend constant).

## Testing

- `internal/fsdoc`: shallow list / dirs-first / dotfile-hiding, missing-root-empty, missing-subdir-NotFound, file-as-dir error, read+hash, create/CAS/conflict, oversize rejection, lexical-escape + symlink-escape containment.
- `internal/daemon`: full WS round-trip via `handleClientMessage` with literal frontend JSON (the real dispatch path) for list/read/write + conflict; `fs_changed(ui)` broadcast; path normalization; and external-edit-surfaces-while-self-write-suppressed against a real `fsnotify` watcher.
- Go `build ./...`, `vet`, `gofmt`, and `tsc --noEmit` all clean.

Verification is at the integration level (handlers driven through the real dispatch path); live UI verification comes in the follow-up slice when there's something to see.

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -172,7 +172,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '115';
+export const PROTOCOL_VERSION = '116';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 interface PRActionResult {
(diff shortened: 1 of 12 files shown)


figgyster (approved):
This looks safe to merge. The fs surface keeps the sharp edges where they belong: the generic path cleaner stays structure-blind, containment still reuses the notebook guard, and the daemon tests hit the real websocket dispatch path instead of just admiring helper functions in a mirror.

I also like that the `.md` watcher limitation is called out explicitly rather than pretending `fs_changed` is more general than it is today. No blockers from me.

Local checks passed: `go test ./...`, `go vet ./...`, `app` `tsc --noEmit`, and `git diff --check`. GitHub CI is green too.

PR #367, opened 2026-06-20
feat(fs): frontend fs surface plumbing + lazy FileTree component

victorarias:
## What

First of two PRs rebuilding the Notebook browser onto the generic filesystem surface (daemon side merged in #366). This PR is the **frontend building blocks**; the NotebookBrowser integration follows in a second PR so each stays small and reviewable.

### `useDaemonSocket`
- `FsEntry` / `FsReadResult` / `FsWriteResult` types mirroring the daemon protocol (`is_dir`/`current_hash` on the wire → camelCase here).
- `sendFsList` / `sendFsRead` / `sendFsWrite` senders using the standard request/result pattern keyed `cmd:requestId`. As with `notebook_write`, a hash-CAS **conflict resolves** (the editor reconciles it) rather than rejecting — only a transport/daemon error rejects.
- `onFsChanged` option wired to the `fs_changed` event.

### `FileTree`
- A lazy folder tree over the fs surface: lists the root on mount and each directory's children **only when expanded** — one `fs_list` per opened node, never a full-disk walk.
- Per-directory listing cache (re-expand reuses it, no refetch), selectable files with `aria-current`, per-node error rows so one failed folder doesn't blank the tree, and an `fs_changed`-driven re-list that preserves expansion state.
- Purely presentational over an injected `listDir` — knows nothing about the websocket — so it's unit-testable with a mock and reusable beyond the notebook.

## Testing
- `FileTree.test.tsx`: lazy-list-on-mount, lazy-expand-fetches-once (+cache on re-expand), file select → `onSelectFile` + selection mark, `changeSignal` re-lists root + open dirs, per-directory error row.
- `useDaemonSocket.test.tsx`: new `fs surface` block — list (`is_dir`→`isDir`), root omits `path`, read resolve/reject, write conflict mapping (`current_hash`→`currentHash`) + `base_hash`, `onFsChanged`.
- Browser harness at `/test-harness/?component=FileTree`.
- Full suite: **1040 passed**, `tsc` clean.

## Not in this PR
- No user-visible change yet — `FileTree` isn't wired into any rendered UI. That (replacing the PARA groups + switching the editor onto `fs_read`/`fs_write`) is PR-2. Hence no CHANGELOG entry here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #368, opened 2026-06-20
feat(notebook): rebuild the Notebook browser on the generic fs surface

victorarias:
## What

Second of the two Slice B PRs (building blocks were #367). The Notebook modal is now rebuilt on the daemon's generic filesystem surface: the PARA-grouped flat note list becomes a **real lazy folder tree**, and the editor reads/writes through `fs_read`/`fs_write`, so **any file under the Notebook root** — not just `.md` notes — can be opened.

This is the "replace in place, keep features" integration: the sidebar and editor data sources change, but backlinks, send-to-chief, and the tasks panel all stay.

### Sidebar
- PARA groups → `FileTree` (from #367): the real folder hierarchy, lazily listed one directory per expand, refreshed by `fs_changed` while preserving expansion.

### Editor (branches by file kind)
- **`.md`** → the live markdown editor, with backlinks + send-to-chief (unchanged).
- **other text** → a plain autosaving editor (same hash-CAS save path, no markdown affordances/backlinks).
- **binary** (by extension) → a read-only "Preview not available" placeholder; the file is **never read** (`fs_read` returns a string, meaningless for binary bytes).

A small `fileKind` classifier (`markdown` / `text` / `binary`, binary by an explicit extension denylist) decides this — unit-tested separately.

### Preserved verbatim
The editing core — debounced autosave, hash-CAS conflict reconcile, navigate/close flush, and all the stale-load/stale-write guards — is unchanged. The fs read/write result shapes are byte-identical to the notebook ones, so only the injected functions changed (`readFile`/`writeFile`/`listDir`). On open we probe the preferred entry point (`knowledge/index.md`) and reuse that read to seed the editor (no double read).

### App wiring
- Thread `sendFsList`/`sendFsRead`/`sendFsWrite` + an `fs_changed`-driven `fsChangeSignal` to the browser (the App.tsx 4-step pattern).
- Drop the now-unused `notebook_changed` → `notebookChangeSignal` wiring: the browser reads via the fs surface, so `fs_changed` (same watcher, same writes) is its refresh signal.

## Behavior notes
- **Document headings now show the file basename** (e.g. `index`), not a frontmatter title — the fs surface is structure-blind and doesn't extract titles. The tree shows filenames too, so this is consistent.
- **A deleted open file now shows "File unavailable"** on the next refresh rather than silently clearing to empty — honest, and the tree drops the node on its own re-list. (Distinguishing not-found from a transient read error would require fragile error-string matching; deferred.)
- `sendNotebookList`/`Read`/`Write` remain exported by `useDaemonSocket` (still a valid `.md`-restricted daemon API) but are no longer consumed by the app. Left in place deliberately — removing them + their tests is out of scope here.
- Live-refresh of **non-`.md`** files still depends on the watcher generalizing beyond `.md` (the documented Slice-A limitation); explicit open/edit of any text file works today.

## Testing
- `fileKind.test.ts` — markdown/text/binary classification, extension edge cases (dotfiles, dir-name dots, `.tar.gz`).
- `NotebookBrowser.test.tsx` — rewritten for the fs contract: tree-click opens a text file in the plain editor, binary → placeholder (no read), the full editing-orchestration suite (autosave, conflict + overwrite, navigate-flush + abort-on-conflict, stale reload/save guards, change-signal reload/no-clobber, deleted-on-refresh), send-to-chief, tasks. Includes the #365 backlinks-decouple regression.
- e2e harness (`notebook-browser.spec.ts`) — runs in a real browser: preferred note opens + autosaves via `writeFile`, a text file opens in the plain editor, a binary file shows the placeholder. **3 passed.**
- Full frontend suite **1045 passed**, `tsc` clean.

Screenshot (harness, dark): lazy tree sidebar + live-rendered index + backlinks strip.


figgyster (dismissed):
This is close, but there’s one contract bug in the binary-file path.

The PR says binary files are never read, and the click path honors that. But the reopen path still probes the previous selection with `readFile(current)` before checking `fileKind`: if the user closes/reopens the Notebook while `cover.png` or another denylisted binary is selected, `NotebookBrowser` calls `fs_read` on that binary anyway. I verified it with a focused local test added temporarily to `NotebookBrowser.test.tsx`: select `cover.png`, close/reopen, then assert `readFile` was not called with `cover.png`; it fails because the reopen effect calls it once.

Required before merge:
- In the on-open “keep current selection” probe, branch binary paths through `loadFile(current)` (or otherwise preserve the placeholder) without calling `readFile`.
- Add the regression test for reopening with a selected binary file.

I also ran the existing focused unit suite after installing frontend deps: `pnpm test -- --run src/components/notebook/fileKind.test.ts src/components/NotebookBrowser.test.tsx` passes (with the existing React act warnings). CI’s Frontend job was still pending when I checked.

figgyster (approved):
This fixes the blocker cleanly. The binary reopen path now preserves the placeholder without touching `fs_read`, and the regression test pins the exact hole from the last review.

I reran the focused Notebook suite locally: `pnpm test -- --run src/components/notebook/fileKind.test.ts src/components/NotebookBrowser.test.tsx` — 39 tests pass. `git diff --check` is clean. Frontend CI was still pending at my last fetch, but the reviewed contract is covered now.

No blockers from me.

PR #369, opened 2026-06-20
docs(agents): pre-authorize non-prod profile builds; prod needs approval

victorarias:
Adds a standing-authorization note to AGENTS.md's "Iterating On Attn Itself" section:

- Agents may **always** build, code-sign, install, run, and restart any **non-prod** profile (the `dev` sibling or any named `ATTN_PROFILE`) — including the full real-signing `make dev` flow with the sandbox disabled — **without asking**. When verifying a change in the actual app, default to `attn-dev.app` over a mocked harness.
- The **production** app (bare `make` / `make install` / `make install-daemon`, the live `~/Applications/attn.app`) still requires explicit approval before building, installing, or restarting.

Doc-only; no code or CI behavior change.


figgyster (approved):
Safe to merge. This makes the attn-on-attn loop much less awkward: non-prod profiles are explicitly fair game, while prod still keeps the approval fence where it belongs.

Doc-only change; I checked the rendered guidance shape and `git diff --check`. CI only ran the change detector and skipped the code jobs, which is fine for this one.

PR #371, opened 2026-06-21
feat(notebook): render lists, task checkboxes, fenced code (UI stage 4a)

victorarias:
## What

First slice of Stage 4 of the Notebook UI rebuild — **markdown polish** (the list/checkbox/code rendering deferred from Stage 1). Pure live-preview decorations; the source stays canonical `.md` and the cursor's line still reveals raw markers for editing.

- **Bullet lists** — `-`/`*`/`+` markers render as a `•` widget (off the active line). Ordered-list numbers stay as written.
- **Task checkboxes** — `- [ ]` / `- [x]` render as a ☐/☑ widget; **clicking one toggles its `[ ]`/`[x]` at the source** (an editor-level mousedown handler, mirroring link-follow). A task item hides its bullet so only the box shows.
- **Fenced code** — renders as a contiguous monospace panel (one line decoration per row); the ```` ``` ```` fences and language tag are **dimmed in place, not hidden**, so line count is preserved. Inline code is unchanged.

## Why split

Stage 4 in the plan bundles markdown polish + frontmatter card + broken-link flags. The latter two hit real design forks (frontmatter vs the raw-source editor; broken-link detection now that the fs surface has no flat catalogue), so this ships the **decision-free** part first. 4b/4c follow as their own PRs once the approach is chosen.

## Tests

- **6 new headless `buildDecorations` cases**: bullet replace + active-line reveal, ordered-list left untouched, task checkbox checked/unchecked + active-line reveal, fenced panel + dimmed fence/info, and the existing "no inline-code class on a fence" invariant still holds.
- **e2e** (`live-markdown-editor.spec.ts`): renders bullets + 2 checkboxes (1 checked) + fenced panel, then clicks the open checkbox and asserts `- [x]` is written back. Harness fixture extended with a task list + code block.

Full frontend unit suite + tsc + the 6 editor e2e specs are green. **No protocol change.**


figgyster (commented):
This looks clean from the code side: the decoration layer keeps the `.md` source canonical, task toggles write only the marker byte, and the active-line reveal still protects editability.

I ran the frontend unit suite, TypeScript, and the live markdown editor e2e locally; all passed. I’m leaving this as a comment rather than a green stamp only because GitHub’s `Frontend` check is still in progress. Once that lands green, no blockers from me.

figgyster (approved):
CI caught up, and the earlier local pass still matches this head. This is safe to merge.

No blockers from me.

PR #372, opened 2026-06-21
feat(notebook): in-editor frontmatter card (stage 4b)

victorarias:
## What

Stage **4b** of the notebook UI rebuild (`docs/plans/2026-06-20-notebook-ui-gap.md`). A note's leading `---…---` YAML frontmatter now renders as a compact **card** — title, type, summary, tags, sources, dates — instead of a wall of `key: value` lines. Click the card to drop into the raw YAML for editing; move the cursor back into the body and the card returns. The file on disk stays byte-for-byte the same markdown.

This is the block-level analogue of the inline live-preview's active-line reveal already shipped in stage 1.

## How

- **`frontmatter.ts`** — pure YAML-subset parser (scalars, flow `[a,b]` + block `- item` lists, quote stripping). Returns the parsed fields **and** the exact whole-line `[from,to)` range a CodeMirror block decoration requires. No DOM, no CM → headless-unit-tested.
- **`frontmatterCard.ts`** — the card itself:
  - A `StateField`-provided `Decoration.replace({ block: true, widget })`. CM only lets decorations that affect vertical layout come **directly from a StateField** (the live-preview `ViewPlugin` is forbidden from introducing block widgets), so the card lives in its own field.
  - **Reveal gate**: shows raw YAML when the editor is focused **and** the cursor overlaps the block; otherwise shows the card. Focus is plumbed into state via a `StateEffect` (a StateField can't read `view.hasFocus`, and CM keeps a selection at pos 0 inside the block on a freshly-opened note). The gate is a pure exported function, headless-tested.
  - Registers the card range as an **atomic range** so cursor motion/deletion treats it as one unit while shown.
  - Guards against swallowing a body-less note (no body line to anchor the cursor → stays raw).
- Wired into `LiveMarkdownEditor`'s extensions.

## Tests

- **Unit (headless `EditorState`):** parser (7) + reveal gate (5).
- **e2e (Playwright harness — CM can't mount under happy-dom):** card render + parsed fields, click-to-reveal → edit raw → click body → card returns, and the body-less guard. Dedicated `FrontmatterCard` harness so it stays independent of the shared live-editor harness.
- Full frontend unit suite (1067) green; `tsc --noEmit` clean.

## Screenshots

Card (rendered):

| Card | Revealed (click to edit) |
|---|---|
| title + type pill, summary, tag chips, source path, meta, "click to edit" hint | raw `---…---` YAML with the cursor inside the block |

_(captured under `app/test-results/frontmatter-card.png` / `frontmatter-revealed.png`)_

🤖 Authored by Claude, on behalf of Victor.


figgyster (dismissed):
This looks safe to merge. The frontmatter card stays nicely scoped: the parser is pure, the CodeMirror block-widget constraint is handled in a `StateField`, and the reveal behavior has both headless and browser coverage.

I ran `tsc --noEmit`, the full frontend unit suite, and the focused Playwright frontmatter-card spec locally; all green. GitHub’s Frontend check was still running when I last looked, but no blockers from me.

figgyster (dismissed):
Still good. Moving the title out of the card is the right cut: one visible `# H1` as the note title, with the frontmatter card staying a properties strip instead of creating title déjà vu with a badge.

I re-ran `tsc --noEmit`, the frontmatter unit tests, the focused Playwright card spec, and `git diff --check`; all clean. GitHub’s Frontend check was still running on the new head when I last looked.

figgyster (approved):
Re-approved the rebased head. The PR patch is still the same frontmatter-card work against the newer base; the base picked up the live-markdown polish, but this branch’s own diff still stays scoped to the properties card.

I re-ran `tsc --noEmit`, the frontmatter unit tests, the focused Playwright card spec, and `git diff --check`; all clean. GitHub’s Frontend check was still running on this head when I last looked.

PR #373, opened 2026-06-21
refactor(notebook): a note's title is its `# H1`, not frontmatter

victorarias:
## What

Locks in the title rule we aligned on: **a note's title is its first `# H1` heading.** attn no longer reads a frontmatter `title:`. When a note has no heading, the title is empty and callers fall back to the **filename** (the note's stable address) — so the filename is an ID, never a competing title.

The visible effect: **backlinks and note lists now show real titles** (the `# heading`) instead of falling back to the filename.

This is the daemon half of the decision behind #372 (the frontmatter *card*, which already shows properties only and never a title). The two are independent and can land in either order.

## How

- `Document.Title()` → first level-1 ATX heading of the body, fence-aware (a `# ` inside ``` / `~~~` is not a heading), with the trailing-`#` ATX rule (`# F# notes ##` → `F# notes`). Returns `""` when there's no `# H1`.
- Retire the now-dead frontmatter `title:` writes:
  - **journals** already carry `# <date>` in the body → drop the duplicate `title:` (keep `type: journal`).
  - **chief inbox** template already leads with `# Chief inbox` → the note needs no frontmatter.
- Record the rule in `docs/glossary.md` (new **Note title** section).

## Why no other consumers

`store.go`'s `List()` is the only `Document.Title()` caller (it feeds `Backlinks()` and the protocol `NotebookEntry.Title`). The frontend already falls back to the basename when an entry's title is empty, so a heading-less note degrades cleanly to its filename.

## Tests

- `firstH1` table: first-line vs after-prose headings, `##`-only → none, fenced-code skip (both `` ``` `` and `~~~`), trailing-`#` strip, `#nope` (no space), ≤3-space indent vs 4-space code block.
- A frontmatter `title:` is **not** a title source (body H1 wins; no H1 → empty).
- Updated the `List`/`Backlinks`/large-file fixtures to carry titles as `# headings`.
- `go test ./internal/notebook` (60) and `./internal/daemon` notebook/journal/inbox/backlink suites (112) green; `go vet` clean.

## Follow-up (NOT in this PR — needs your call)

The attn skill's `references/notebook.md` still instructs the chief/keeper to write a frontmatter `title:` (and a body without a guaranteed `# H1`). For *newly authored* KB notes to get real titles under this rule, that guidance needs to flip to **"lead the body with a single `# Title` heading; frontmatter is properties only."** It's a global skill file outside this repo, so it's not included here.

🤖 Authored by Claude, on behalf of Victor.


figgyster (approved):
This looks safe to merge. The title contract is cleaner now: frontmatter stays properties-only, `Document.Title()` owns the first body H1, and the notebook/list/backlink fixtures were moved to the same rule instead of leaving two competing sources of truth.

Local checks passed here: `go test ./internal/notebook ./internal/daemon`, `go vet ./internal/notebook ./internal/daemon`, and `git diff --check`. GitHub CI is green too. No blockers from me.

PR #374, opened 2026-06-21
fix(app): clean "Session ended" instead of "[Process exited with code 143]"

harleenquinzell:
## Why

A finished/stopped session is torn down with **SIGTERM** (exit code 143) — see `terminateSession`/`unregisterSession` in `internal/daemon`. The terminal rendered that teardown as a bare `[Process exited with code 143]`, which reads like a crash even though the task completed and its work persisted.

This was surfaced by a real dispatch: an evaluation session finished, called `dispatch report` with a final DONE, saved its result to prod — and the pane showed nothing but `[Process exited with code 143]`.

## What

- New pure helper `app/src/pty/exitNotice.ts` — `formatExitNotice(code, signal)`.
- Wired into the terminal exit-render path in `useGhosttyPaneRuntime.ts` (one line).
- New unit tests `app/src/pty/exitNotice.test.ts`.

### Behavior (before → after)

| exit | before | after |
|------|--------|-------|
| code 0 | `[Process exited with code 0]` | `[Session ended]` |
| 143 (SIGTERM — attn teardown) | `[Process exited with code 143]` | `[Session ended]` |
| 130 / 129 (SIGINT / SIGHUP) | `[Process exited with code N]` | `[Session ended]` |
| 1, 2 (app error) | unchanged | `[Process exited with code N]` |
| 137 / 139 / 134 (SIGKILL / SEGV / ABRT) | unchanged | `[Process exited with code N]` |

SIGTERM is attn's own normal-teardown signal, so 143 is graceful by design. Crash/force-kill signals and non-zero app codes still show the raw code so real failures stay visible.

### Deliberately out of scope

`App.tsx` `handleSessionProcessExit` still auto-closes only on `code 0 && !signal`; a 143 exit keeps the pane open so the final output stays readable. Only the **message wording** changed — the smallest safe fix for the reported complaint.

## Verification

- New `exitNotice` tests: 7/7 pass.
- Full frontend suite: **113 files / 1062 tests** pass; `tsc --noEmit` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ### Added
 - **Broken Notebook links are flagged as you write.** When a note links to another note that doesn't exist — a typo in the path, or a note you haven't created yet — the link now shows in red with a ⚠ instead of looking like a working link. Links to real notes stay normal, and external links (http/https, email) are never flagged. If an agent or you create the missing note, its links clear their flag once the Notebook refreshes.
 
+### Fixed
+- **A finished session now ends cleanly instead of looking like a crash.** When a session is stopped, closed, or wraps up its work, attn tears it down with a normal termination signal — but the terminal used to report that as a bare `[Process exited with code 143]`, which reads like something broke. Expected endings (a clean exit, or a normal stop/close) now show a calm `[Session ended]`. Genuine failures — a non-zero exit code, or a crash/force-kill — still show the exit code so real problems stay visible.
+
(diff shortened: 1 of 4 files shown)


figgyster (approved):
Rebased this onto current `main` and resolved the changelog overlap by keeping both 2026-06-21 entries.

This looks safe to merge. The exit notice logic keeps normal attn teardown calm without hiding real crash/force-kill exits, and the focused regression tests cover the cases that matter.

Local validation passed: `pnpm test -- exitNotice`, `pnpm exec tsc --noEmit`, and `git diff --check`. GitHub’s Frontend check was still running when I last looked, but it’s exercising the same frontend path.

PR #377, opened 2026-06-21
fix(terminal): stop the last line clipping at the bottom of the window

harleenquinzell:
A terminal could end up sized one row taller than the window can show, so
the bottom line (the shell prompt, or attn's "auto mode on" status line)
was clipped at the window edge — visible whenever content filled the
session to the bottom.

fit() derives rows from Math.floor(clientHeight / cellHeight), so it can
never overflow the container. But the daemon's authoritative PTY geometry
arrives via resizeLocal (source pty_resized) and is NOT bounded by this
client's window: when another client, or a prior taller layout, left the
PTY one row too tall for the current window, the canvas rendered taller
than the visible area and the last row spilled past the window edge. With
no container resize to fire the ResizeObserver, fit() never re-ran, so the
overflow persisted.

Add geometryOverflowsContainer() and, after a live (non-replay)
resizeLocal that overflows the container, schedule a corrective fit() so
the active client re-asserts its own floored geometry (and re-notifies the
daemon). fit() still bails when inactive or mid-replay, so this never
fights the geometry-authority model in rule #7 — it only corrects a live
overflow the active client can actually see. A 1px slack avoids spurious
refits from sub-pixel container heights.

Verified in the dev app: confirmed the overflow with pixel measurements
(canvas bottom below the window), and the last line stays fully visible
after the fix. Unit tests cover the overflow predicate; full terminal
test suite green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-06-21]
 
+### Fixed
+- **The last line of a terminal is no longer cut off at the bottom of the window.** When a session's terminal was sized one row taller than the window could actually show — for example after the window's height landed on an awkward boundary, or when the session was opened at a size different from where its grid was last set — the bottom line (your prompt, or attn's `auto mode on` status line) got clipped at the window edge. The terminal now re-asserts a row count that fits the visible area, so the last line stays fully on screen at any window size.
+
 ### Added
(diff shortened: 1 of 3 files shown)


figgyster (approved):
This looks safe to merge. The resize correction is nicely scoped: it only fires on live overflow, keeps replay/inactive panes out of the fight, and the predicate tests pin the awkward one-row case without turning terminal sizing into test theater.

I ran the frontend Vitest suite locally (`pnpm test -- --run`): 1082 passed, 5 skipped. GitHub Frontend CI was still running at last check, but no blockers from me.

PR #378, opened 2026-06-21
feat(notebook): accept notebook tile kind + dock helper (stage 7, PR 1)

victorarias:
First slice of **Notebook tile mode** (Stage 7 of the notebook UI rebuild). Design doc: \`docs/plans/2026-06-21-notebook-stage7-tile-mode.md\` (added here).

## What this is
Tile mode makes the Notebook a first-class **workspace tile** beside agent terminals — size-aware, persisted per-workspace, multiple independent views. attn already has the whole tile split-tree / dock / resize / persist machinery; this stage is an add-on, not a layout rewrite.

This PR is the groundwork only. **No protocol bump** — the existing machinery already carries new tile kinds; we only widen the two gates that were kind-specific.

## Changes
**Daemon**
- Add \`workspacelayout.TileKindNotebook\` (\`"notebook"\`).
- Widen \`handleWorkspaceLayoutUpdateTile\` from a browser-only check to a kind switch:
  - browser → still validates its URL (unchanged),
  - notebook → passes its opaque file-path params through as-is,
  - markdown (assigned once at dock time) and unknown kinds → still rejected.

**Frontend**
- Add \`useDaemonSocket.sendWorkspaceDockTile\` — the in-app dock entry (prior docks were all daemon-initiated via \`attn open\`). Each call must mint a **unique tileId**: the daemon treats a duplicate id as a *move*, not a second tile.

## Not user-visible yet
The tile can't be opened until later PRs (NotebookSurface extraction → width auto-fold → shortcuts/ActionMenu → render branch → finder), so there's no CHANGELOG entry.

## Tests
- Daemon tile-lifecycle test extended: a fresh notebook tile docks empty, then a file-path \`update_tile\` is accepted and persisted.
- \`go test ./internal/daemon ./internal/workspacelayout\` green; frontend \`tsc\` + \`useDaemonSocket\` vitest (51) green.

---

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -3000,6 +3000,34 @@ export function useDaemonSocket({
     );
   }, [nextRequestID, sendWorkspaceCommand]);
 
+  // Dock a new tile into a workspace layout beside an anchor leaf. Unlike the
+  // daemon-initiated markdown/browser docks (the `attn open` unix path), this is
+  // the in-app entry used by notebook tile mode. Each dock MUST pass a unique
+  // tileId: the daemon treats a duplicate id as a move, so a repeated id would
+  // relocate the existing tile instead of adding a second one. anchorPaneId
+  // defaults to '' (the daemon falls back to the active leaf, then the first).
+  const sendWorkspaceDockTile = useCallback((
+    workspaceId: string,
+    tileId: string,
+    tileKind: string,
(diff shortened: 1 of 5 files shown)


figgyster (approved):
This looks safe to merge. The daemon gate stays narrow — browser URLs still get validated, markdown remains dock-time only, and notebook gets just the opaque path lane it needs for the later tile surface.

I ran the daemon/workspacelayout Go tests, `git diff --check`, frontend `tsc --noEmit`, and the `useDaemonSocket` vitest slice locally. GitHub's Frontend check was still pending last I looked, but no blockers from me.

PR #379, opened 2026-06-21
refactor(notebook): extract NotebookSurface (modal|tile variants) (stage 7, PR 2)

victorarias:
Stage 7 (Notebook tile mode), PR 2 of the ladder. **Pure refactor — the modal is behaviorally byte-identical.** Design doc: \`docs/plans/2026-06-21-notebook-stage7-tile-mode.md\`.

## What
Extract the entire Notebook body — file tree, live editor, context rail, fold handles, tasks panel, and all the load / save / send-to-chief logic — out of \`NotebookBrowser\` into a reusable **\`NotebookSurface\`** that renders in two shapes:
- **\`variant="modal"\`** — wrapped in the dialog shell (overlay + focus trap + header with the chief pulse + Close). \`NotebookBrowser\` keeps this wrapper.
- **\`variant="tile"\`** — a bare surface for an in-workspace tile (mounted by a later PR).

This is the spine the rest of tile mode builds on (width auto-fold → render branch → finder all attach here).

## Why it's safe
- \`NotebookBrowser\` stays the public component (same props, same export) and re-exports \`parseNotebookHref\` / \`NotebookHref\`, so **App, the test harness, and the unit test are untouched**.
- The surface stays **mounted while a modal is closed** (its state/refs survive a close→reopen, which the "keep prior selection on reopen" behavior relies on). So the open/close gate became an \`active\` prop (\`= isOpen\` for the modal) instead of conditional mounting — every \`isOpen\`-gated effect now reads \`active\`.
- \`autoFold\` stays the hardcoded \`false\` (PR3 swaps it for the width-driven value).
- New tile-only seams are inert for the modal (which passes neither): \`initialPath\` (the file a tile reopens to) and \`onOpenFile\` (so a tile persists the opened path to its tile params).

## Tests
- \`NotebookBrowser\` unit suite — **41 passing** (unchanged).
- \`notebook-browser\` e2e harness — **5 passing**, including the stage-5 chrome + fold-to-zero-width spec.
- \`tsc\` clean.

No CHANGELOG entry: pure refactor, no user-visible change.

---


figgyster (approved):
This looks safe to merge. The extraction keeps `NotebookBrowser` as the public shell while moving the real surface into a reusable component, and the modal path still preserves the close/reopen state that mattered here.

I ran `tsc --noEmit`, the `NotebookBrowser` unit suite, and the notebook-browser Playwright spec locally; all passed. GitHub’s Frontend job was still pending when I checked, but no blockers from me.

PR #380, opened 2026-06-21
feat(notebook): width-driven auto-fold for the tile surface (stage 7, PR 3)

victorarias:
Stage 7 (Notebook tile mode), PR 3 of the ladder. Builds on the PR2 \`NotebookSurface\` extraction. Design doc: \`docs/plans/2026-06-21-notebook-stage7-tile-mode.md\`.

## What
Add **\`useTileAutoFold\`** — a ResizeObserver hook that drives the fold seam's *auto* side from the surface body's width. As a notebook tile narrows it sheds side panes:

| width | layout |
|---|---|
| ≥ 900px | tree + document + rail |
| 620–899px | tree + document (rail auto-folds) |
| < 620px | document only (both fold) |

A manual fold still wins (`folded = override ?? auto`); this only changes what "auto" resolves to.

## Safety / correctness
- The **modal passes `enabled=false`** and keeps `{false,false}` (manual-only folds, unchanged) — only a tile observes. The modal is untouched.
- The **initial measurement runs in a layout effect** (synchronously before paint), so a tile that mounts already-narrow renders folded on the first frame instead of flashing open then snapping shut.
- We observe the body **container**, whose width is independent of how its inner panes fold, so folding can't shrink the observed box and oscillate.

## Tests
- `foldsForWidth` pure breakpoint ladder + the hook wiring (sync mount measure, resize refold across breakpoints, disabled baseline) — **7 unit tests**.
- `NotebookBrowser` suite still green (modal `{false,false}` path).
- `tsc` clean.

No CHANGELOG: the tile isn't rendered until a later PR, so no user-visible change yet. The responsive folding gets its integration coverage + manual verification when the tile first mounts (PR mounting `WorkspaceDockTile`).

---

### DIFF
--- a/app/src/components/NotebookSurface.tsx
+++ b/app/src/components/NotebookSurface.tsx
@@ -2,6 +2,7 @@ import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import FocusTrap from 'focus-trap-react';
 import type { FsEntry, FsExistsResult, FsReadResult, FsWriteResult, NotebookEntry, NotebookSendToChiefResult, NotebookTask } from '../hooks/useDaemonSocket';
 import { useEscapeStack } from '../hooks/useEscapeStack';
+import { useTileAutoFold } from '../hooks/useTileAutoFold';
 import { FileTree } from './notebook/FileTree';
 import { fileKind, isBinaryPath, isMarkdownPath } from './notebook/fileKind';
 import { parseFrontmatter } from './notebook/frontmatter';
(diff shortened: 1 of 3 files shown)


figgyster (approved):
This is safe to merge. The auto-fold seam is nicely bounded: tile-only, container-width driven, and the modal path stays manual-only.

I ran the focused frontend coverage (`useTileAutoFold` + `NotebookBrowser`), `tsc --noEmit`, and `git diff --check`; all passed. GitHub’s Frontend CI was still running when I checked, so keep the merge button honest, but no blockers from me.

PR #381, opened 2026-06-21
feat(notebook): render notebook tiles + entry points (stage 7, PR 4)

victorarias:
## Stage 7, PR 4 — render the notebook tile + entry points

Fourth PR in the Notebook tile-mode ladder (`docs/plans/2026-06-21-notebook-stage7-tile-mode.md`). PRs 1–3 landed the daemon-side `notebook` tile kind + dock helper (#378), the `NotebookSurface` extraction into modal|tile variants (#379), and the width-driven auto-fold (#380). This PR makes the tile **real and reachable**: you can dock one and it renders the live notebook.

### What you get
Dock the Notebook beside your terminals with **⌘⌥N** (or the action menu → "Open Notebook tile"). It mounts the same live `NotebookSurface` in its `tile` variant — folder tree, live editor, context rail, send-to-chief — and folds its side panels away as the tile narrows (the stage-3 auto-fold). Each tile remembers the note it was showing and reopens to it; you can open several tiles on different notes; the layout survives restarts. **⌘⌥⇧N** still opens the fullscreen modal.

### How it's wired
- **`NotebookSurfaceContext`** — provides the daemon surface (fs + notebook RPCs). A tile lives deep in the workspace split-tree, so instead of threading ten daemon callbacks through the terminal workspace that has nothing to do with the notebook, the tile reads them from context. App provides it once from the same callbacks the fullscreen modal already uses.
- **`NotebookTile`** — consumes the context and mounts `NotebookSurface variant="tile" active`.
- **`WorkspaceDockTile`** — renders the `notebook` branch; each opened file is persisted back to the tile's params, and `tilePresentation` derives the tab title from that path.
- **`handleOpenNotebookTile`** — docks a fresh tile with a unique id every time (the daemon treats a duplicate tile id as a *move*, so a shared id would just relocate the first tile). Bound to the new `notebook.openTile` / `notebook.openFullscreen` shortcuts.

### Tests
- All existing notebook/shortcut/tile unit suites stay green (133 tests).
- New `NotebookTileHarness` + `e2e/notebook-tile.spec.ts` mount the tile through a real `ResizeObserver` and assert it renders the live editor and folds rail-then-tree as the frame narrows (then unfolds on widen).
- `tsc --noEmit` clean.

### Notes
- No protocol bump — the tile kind needs only the `update_tile` gate + constant that shipped in PR 1.
- The fresh-tile **no-selection screen + tile-local fuzzy finder** is the final PR (5); today a fresh tile opens on its folder tree, which you navigate to pick a note.


figgyster (approved):
This looks safe to merge. The notebook tile is wired through a clean surface context instead of dragging the whole daemon callback suitcase through the workspace tree, and the harness hits the right seam: live tile render plus the width-driven fold behavior in a real browser.

I ran `pnpm exec tsc --noEmit`, `pnpm run e2e -- e2e/notebook-tile.spec.ts`, and `git diff --check`; all passed. GitHub’s Frontend job was still running on my last fetch, so keep the usual CI eye on it, but no blockers from me.

PR #382, opened 2026-06-21
fix: stop attn's background Claude runs from cluttering ~/.claude

victorarias:
## Problem

attn's headless one-shot Claude calls were each leaving artifacts in Claude Code's `~/.claude` store:

1. **Session transcripts** — the SDK-based **classifier** (runs at every turn-end) and the **review-loop** SDK fallback call `sdk.RunQuery` / `sdk.NewClient` without setting `PersistSession`. The SDK only emits `--no-session-persistence` when `PersistSession` is explicitly `false`; left nil it falls back to the CLI default, which **persists a full `.jsonl` transcript per run**.

2. **Orphaned project dirs** — the Notebook's `summarize_session` / `narrate_workspace` runs used a fresh `os.MkdirTemp` cwd each time. Claude maps each distinct cwd to its own `~/.claude/projects/<hash>` dir and spills large tool output there **even under `--no-session-persistence`**, so they accumulated unboundedly (134+ `attn-summarize-session-*` dirs observed on disk).

## Fix

- Pass `types.WithPersistSession(false)` on the classifier `RunQuery` and the review-loop `NewClient` opts.
- Add a shared, profile-isolated `headlessScratchCwd()` (`<dataDir>/headless-cwd`) and route `summarize_session` + `narrate_workspace` through it, so all runs collapse onto a **single reused** project dir. Collision-free under concurrency: these runs use absolute paths for all I/O and never write into the cwd.

The CLI headless path (`buildClaudeHeadlessArgs`) already passes `--no-session-persistence`, so the summarizer/narrator/keeper/workflow-agent never leaked transcripts. The **keeper** writes scratch files into its `WorkDir` (which doubles as the Codex sandbox root) and isn't proliferating on disk, so it's intentionally left as-is. **attn never deletes anything inside `~/.claude`** — this only stops new clutter at the source.

## Verification

- `go build ./...` clean; `go vet` clean; notebook-narration + classifier tests pass.
- Empirical (classifier): a control SDK run (no option) persists a `.jsonl`; the fixed path persists none.
- Empirical (Lever A): two real headless runs sharing the stable cwd produce a **single reused** project dir (delta 1), vs. one-per-run before.


figgyster (approved):
This is safe to merge. The fix hits both leak sources: SDK one-shots now opt out of transcript persistence, and Notebook narration stops minting a new Claude project hash for every run by reusing an attn-owned scratch cwd.

I checked the touched daemon/classifier paths and ran `git diff --check`, `go test ./internal/classifier ./internal/daemon`, `go vet ./...`, and `go build ./...`; visible GitHub checks are green too. No blockers from me.

PR #383, opened 2026-06-21
feat(notebook): in-tile fuzzy file finder (stage 7, PR 5)

victorarias:
## Stage 7, PR 5 (final) — in-tile fuzzy file finder

The last PR in the Notebook tile-mode ladder (`docs/plans/2026-06-21-notebook-stage7-tile-mode.md`). PRs 1–4 landed the daemon tile kind + dock helper (#378), the `NotebookSurface` modal|tile extraction (#379), width-driven auto-fold (#380), and the render path + entry points (#381). This PR completes the tile: a **Cmd+P fuzzy file finder scoped to one tile**.

### What you get
A fresh notebook tile opens straight into a finder — type and it matches notes by **name or title** across the whole notebook, even with scattered letters (`kbidx` → `knowledge/index.md`). Arrow keys move, **Enter** (or a click) opens the note in that tile, **Esc** drops back to the folder tree. Already reading a note? **⌘P** re-summons the finder for just that tile. The empty-state screen also has a "Find a note" button, so the feature works without the keyboard.

### How it's built
- **`finderRank.ts`** — a pure, headless scorer/ranker: a subsequence match over a note's path and title, basename-weighted, ties broken by recency then path. Modeled on `ActionMenu`'s scorer. Fully unit-tested.
- **`useNotebookFileIndex`** — walks the whole vault once on tile mount (`notebook_list`, empty prefix → a flat list **with titles**) and re-walks debounced on `fs_changed`, dropping stale responses via a monotonic token.
- **`NotebookFinder`** — the overlay: its own scoped focus, arrow/Enter/Esc nav, click-to-pick, dim-backdrop dismiss. Absolute **inside** the tile (not a global portal), so two tiles each summon their own.
- **`NotebookSurface`** wires it **tile-only**: the index, the ⌘P container keydown (`preventDefault` stops the WebView print dialog; Shift excluded — `⌘⇧P` is the attention dock), auto-open on a fresh tile, and the "Find a note" button. Picking → `loadFile` → persists the path to `tileParams`.
- **`listFiles`** threaded through `NotebookSurfaceContext` + the 4-place App plumbing (`sendNotebookList`); the modal omits it (it navigates via the tree, so the prop is optional).

### Why ⌘P is safe in the packaged app
Per the macOS native-menu trap (AGENTS.md gotcha #8), a web shortcut dies if a native menu item owns its accelerator. Tauri's default macOS menu has **no Print item** (File holds only Close Window → Close Pane), so plain ⌘P reaches the WebView and `preventDefault` suppresses print. Shift is excluded so it never collides with the global `⌘⇧P` attention dock. And because the finder also has a button + auto-open, it degrades gracefully even where a keyboard shortcut wouldn't fire.

### Tests
- `finderRank` (scorer/ranker) + `useNotebookFileIndex` (mount walk, debounced refetch, stale-drop, error) units — 16 new tests; full suite stays green (1117).
- New `e2e/notebook-tile-finder.spec.ts` drives ⌘P → filter → Enter-to-open (asserting the path persists), Esc-to-dismiss, and fresh-tile auto-open, through a **real keydown + browser** (what units can't cover).
- `pnpm run build` + `tsc --noEmit` clean.

### Notes
- No protocol bump — reuses `notebook_list` and the tile kind from PR 1.
- Finder index is `.md`-scoped (notebook_list); the tree still reaches any text file. Easy to widen later if wanted.


figgyster (approved):
This looks safe to merge. The finder stays tile-scoped, the notebook list walk is stale-guarded, and the keyboard path is covered where the native/WebView shortcut trap usually hides its little knife.

I ran the focused finder/index unit tests, the notebook tile finder Playwright spec, and `pnpm run build` locally. GitHub Frontend CI was still running when I last checked, but no blockers from me.

PR #384, opened 2026-06-21
test(notebook): packaged finder scenario + fix ⌘P after Esc

victorarias:
## Summary

Follow-up to the notebook tile finder (#360 ladder, PR5). Writing a packaged-app harness scenario for the in-tile fuzzy finder surfaced a real focus bug, so this PR fixes that and lands the scenario.

**The bug:** after `Esc` closed the finder, focus fell to `<body>`. The `⌘P` re-summon is a keydown scoped to the tile container, so once focus left the tile you couldn't re-open the finder without first clicking back into it.

**The fix (`NotebookSurface.tsx`):** capture the focused element when the finder opens and, on close, restore focus inside the tile — the prior element if it's still mounted, otherwise the tile container. `⌘P` keeps working.

## What's here

- **Focus restoration** in `NotebookSurface` so `Esc` → `⌘P` re-summons the finder.
- **e2e regression** (`notebook-tile-finder.spec.ts`): `Esc` → `⌘P` re-opens with no re-click. Red before the fix, green after.
- **New real-app scenario** `real-app:scenario-notebook-tile-finder`: docks a notebook tile with the **real native `⌘⌥N`**, confirms the finder auto-opens and renders in the WKWebView, then `Esc` → native `⌘P` re-summons — the native menu → WebView path browser e2e can't reach. Registered in `package.json` + the serial matrix.

## Verification

- `pnpm run build` clean (browser-runtime + tsc + vite).
- `1117` unit tests pass; `5` notebook e2e pass (including the new regression).
- The scenario's **non-native** logic (finder auto-open, render in the real WKWebView, screenshots) was validated against the real packaged app, and a visual screenshot confirmed the overlay renders correctly (titles + paths + accent highlight).
- The scenario's **native keystroke** steps (`⌘⌥N`, `⌘P`, `Esc` via CGEvents) need macOS Accessibility permission granted to the running process and attn frontmost — same as every native-input scenario here. The background sandbox this was authored in lacks that permission, so those steps will run green on your machine / CI but couldn't be exercised here. The dock step throws a clear diagnostic naming the frontmost app when the permission is missing, instead of a mystery timeout.

### DIFF
--- a/app/e2e/notebook-tile-finder.spec.ts
+++ b/app/e2e/notebook-tile-finder.spec.ts
@@ -50,6 +50,23 @@ test.describe('NotebookTile finder', () => {
     await expect(page.locator('.cm-editor')).toBeVisible();
   });
 
+  test('Cmd+P re-summons the finder after Esc without re-focusing the tile', async ({ page }) => {
+    // Regression: after Esc closes the finder, focus must stay inside the tile so a
+    // follow-up Cmd+P re-summons it. (Without focus restoration, focus falls to
+    // <body> and the tile-scoped keydown never fires.)
+    await page.goto('/test-harness/?component=NotebookTile&initialPath=');
+    await page.waitForFunction(() => window.__HARNESS__?.ready === true);
+
+    // Fresh tile auto-opens the finder; Esc closes it.
+    await expect(page.locator('.notebook-finder')).toBeVisible();
+    await page.keyboard.press('Escape');
+    await expect(page.locator('.notebook-finder')).toHaveCount(0);
+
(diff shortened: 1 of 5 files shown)


figgyster (approved):
This looks safe to merge. The focus restoration is scoped to the notebook tile, the regression catches the exact Esc → ⌘P failure, and the packaged-app scenario covers the native shortcut path browser tests can’t touch.

I ran `git diff --check`, `pnpm run build`, and `pnpm run e2e -- e2e/notebook-tile-finder.spec.ts` locally — all green. GitHub’s Frontend check was still in progress when I looked, so keep an eye on that, but no blockers from me.

PR #385, opened 2026-06-21
fix(worktree): default new worktrees to origin/<default branch>

victorarias:
## Problem
Creating a worktree from the new-session dialog could produce a **stale** branch. A worktree opened "to start fresh" came out based on a 6-day-old commit (`test/test`'s tip) rather than the latest `main`.

Root cause is **not** a broken fetch — I verified the `origin/main` path fetches correctly end-to-end (git 2.54): `git fetch origin main` moves `refs/remotes/origin/main`, then `git worktree add … origin/main` lands on the fresh tip (control without the fetch lands stale, so the fetch is load-bearing).

The footgun was the **default**: `RepoOptions` defaulted `startingBranch` to `'current'`, which branches off `selectedDestination?.branch` — the *local* branch the chosen destination happens to be on. For a worktree-of-a-worktree that can be arbitrarily stale.

## Change
- **Frontend** (`RepoOptions.tsx`): default flipped `'current'` → `'default'` (origin/&lt;defaultBranch&gt;) and the two radios reordered so the now-default option leads. "Start from &lt;current&gt;" is still one click/Tab away.
- **Daemon** (`worktree.go`): after the fetch, if the requested start ref can't be resolved (e.g. `origin/main` in a repo with no matching remote branch), fall back to the current HEAD instead of erroring on `git worktree add`. Fixes a latent hard-failure the new default would otherwise expose.
- **git** (`branch.go`): new `RefExists` helper.

## Tests
- `RepoOptions.test.tsx`: default is `origin/main`; choosing "current" passes the destination branch.
- `LocationPicker.test.tsx`: updated the "from selected worktree branch" test to opt into "current" (no longer the default).
- `branch_test.go`: `TestRefExists`.
- Frontend suite green (1098 passed); `internal/git` + daemon worktree tests pass.

## Note
The radio reorder is cosmetic-equivalent — the radio-row CSS is order-independent (column flex; `.selected`/`:hover` only).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 - **The Notebook header tells you more at a glance.** A small "chief" indicator lights up while your chief-of-staff agent is working, and each note's title now carries a kind badge (note or journal).
 - **Broken Notebook links are flagged as you write.** When a note links to another note that doesn't exist — a typo in the path, or a note you haven't created yet — the link now shows in red with a ⚠ instead of looking like a working link. Links to real notes stay normal, and external links (http/https, email) are never flagged. If an agent or you create the missing note, its links clear their flag once the Notebook refreshes.
 
+### Changed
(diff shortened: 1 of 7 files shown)


figgyster (approved):
This looks safe to merge. The default now points at the freshly-fetched remote branch, while the daemon keeps the no-remote edge case from turning the new default into a hard failure.

I ran the focused frontend worktree/new-session tests plus the relevant Go worktree/git tests locally; GitHub's full Frontend check was still running when I last looked, but the core path here is covered.

PR #386, opened 2026-06-21
feat(notebook): ⌘P finder in fullscreen + sidebar entry; drop fullscreen from ⌘K

victorarias:
## Summary

Three Notebook changes you asked for:

1. **`⌘P` fuzzy finder now works in the fullscreen Notebook**, not just in tiles. Press `⌘P` to jump to any note by name or title; Esc closes the finder first (a second Esc closes the Notebook).
2. **New sidebar button to open the fullscreen Notebook** (in the toolbar next to the diff/PRs buttons). `⌘⌥⇧N` still works.
3. **Removed the fullscreen "Browse the Notebook" entry from `⌘K`** — the command palette now lists only "Open Notebook tile".

## How it works

- **Finder gate** flips from `variant === 'tile'` to "`listFiles` provided," so the modal gets the same finder. The index only walks while the surface is showing (a tile is always live, a modal only when open), so a closed-but-mounted modal never walks the vault.
- **Esc ordering** is the subtle bit: the modal's Esc is a capture-phase escape-stack entry that would otherwise close the *whole* modal. The finder pushes a higher-priority stack entry while open, so the first Esc closes the finder and a second closes the modal. (Tiles aren't modal — their finder Esc is handled by the input's own `onKeyDown`, unchanged.)
- `NotebookBrowser` forwards a new `listFiles` prop; `App` passes `sendNotebookList`. `.notebook-browser` gets `position: relative` so the overlay scopes to the dialog.
- Sidebar gains a `NotebookIcon` + a toolbar button wired to the existing `openNotebookBrowser`.

## Verification

- `pnpm run build` clean (browser-runtime + tsc + vite); `1119` unit tests pass.
- `12` notebook e2e pass, including **two new ones** in `notebook-browser.spec.ts`: `⌘P` filters and opens a note over the modal, and Esc closes the finder before the modal (then `⌘P` re-summons). These run real keydowns through the real escape stack + focus restoration in a real browser.
- **Verified in the packaged dev app**: the new sidebar button renders in the WKWebView, and opening the fullscreen Notebook fires the `notebook_list` index walk (confirmed in the daemon log) — i.e. the finder is wired up in fullscreen. The actual `⌘P` keystroke→overlay in the packaged app needs native input (macOS Accessibility) which this background run lacked; it's covered by the real-browser e2e, and `⌘P` carries no native-menu trap (no Print menu item, established earlier).


figgyster (approved):
This looks safe to merge. The fullscreen Notebook now gets the same `⌘P` finder path as tiles without walking the vault while the closed modal sits mounted, and the Esc-stack behavior is covered at the seam that mattered.

I ran `pnpm run build` and the notebook browser Playwright spec locally; both passed. GitHub's Frontend job was still running last I checked, so let that finish, but no blockers from me.

PR #387, opened 2026-06-21
fix(notebook): keep your scroll position when files change while reading

victorarias:
## What

Reading a Notebook note no longer jumps to the top when files change.

While you're reading a note, an agent editing that note — or any unrelated background write under the notebook root — used to scroll the note back to the top and lose your place. Now your scroll position is kept.

## Root cause (two compounding issues)

1. **The editor reset its own scroll on reload.** `@uiw/react-codemirror` reconciles a changed `value` prop as a *full-document replace* (`changes:{from:0,to:doc.length,…}`). CodeMirror can't anchor a full replace, so `.cm-scroller` snaps to the top. (`.cm-scroller` owns the scroll; the outer panes are `overflow:hidden`.)
2. **The note reloaded on _any_ file change.** `fs_changed` bumps a single notebook-wide signal (`fsChangeSignal`), so the open note was re-read and re-applied on every unrelated write under the notebook root.

## The fix (notebook-only, no protocol change)

- **`NotebookSurface.refreshOpenFile`** replaces `loadFile(current)` on the live-refresh path: re-read the open file and **bail with zero `setState` when the content hash is unchanged** — so an unrelated write no longer disturbs (or re-scrolls) the note, and the backlinks panel stops flashing. Only a genuine change is applied and re-walks backlinks.
- **`LiveMarkdownEditor.applyExternalContent`** — a new imperative handle that applies a genuine change as a **minimal edit** (`computeMinimalEdit` trims the shared prefix/suffix, `scrollIntoView:false`, no focus), so CodeMirror keeps your scroll position and selection anchored. The non-`ExternalChange` dispatch fires `onChange`, so the controlled `value` catches up and react-codemirror's next pass is a no-op (no loop).

## Testing

- **Unit:** `computeMinimalEdit` (algorithm) + two new `NotebookSurface` tests — unrelated change → no churn; genuine change → applied via the scroll-preserving path. Full frontend suite green (1129).
- **Real browser (Playwright):** the mechanism is verified against real CodeMirror — a minimal edit **preserves** `scrollTop`, while a full value swap (control) **resets it to 0** (proving both the bug and the fix). A full `NotebookSurface` integration test exercises the actual component: an unrelated change is ignored, and an on-disk change to the open note is applied without moving scroll.
- **Live (real dev daemon):** an external edit to the open note fires `fs_changed(origin=external)` with the note path — the exact event the frontend turns into the scroll-preserving reload.

## User-facing

CHANGELOG entry added under `[2026-06-21] → Fixed`.


figgyster (approved):
This looks safe to merge. The refresh path now treats navigation and live file updates as different beasts: unrelated `fs_changed` events bail without touching editor state, and real note changes go through a minimal CodeMirror edit instead of the old full-document shove off a cliff.

I ran the focused Vitest coverage (`minimalEdit`, `NotebookBrowser`) and the real-browser Playwright notebook/editor specs; all green. No blockers from me.

PR #388, opened 2026-06-21
docs(notebook): mark gap-map stages shipped; stage 6 cut

victorarias:
## Summary

Doc-only: the notebook UI rebuild is complete, but `docs/plans/2026-06-20-notebook-ui-gap.md` still read as pre-plan analysis with only stage 1 marked done. This brings the staged list in line with reality.

- **Stages 2–5, 7** marked ✅ Done with their PR refs.
- **Stage 6 (selection menu + toasts)** marked ❌ **Cut (2026-06-21)** with the rationale: the floating "Send to chief" pill already covers the real need; "new note from selection" conflicts with the chief-curated knowledge base (promotion is a keeper job, not a reader action); the toast stack was deferred in favor of the existing inline "Saved"/status indicator. No code change.
- **Top status line** updated to "Rebuilt — all stages shipped (stage 6 cut)." The gap analysis + "Shipped (current)" snapshot below are kept as the historical pre-rebuild baseline the map was scoped against.

No CHANGELOG entry (internal planning doc, not user-visible).

### DIFF
--- a/docs/plans/2026-06-20-notebook-ui-gap.md
+++ b/docs/plans/2026-06-20-notebook-ui-gap.md
@@ -1,7 +1,9 @@
 # Notebook UI — Prototype vs. Shipped: Gap Map
 
-Status: analysis (pre-plan). Maps every material difference between the **target**
-prototype and the **currently shipped** notebook UI, so a rebuild can be scoped.
+Status: ✅ **Rebuilt — all stages shipped (stage 6 cut).** See the staged list below
+for what landed. The gap analysis and "Shipped (current)" snapshot that follow describe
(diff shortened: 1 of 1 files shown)


figgyster (approved):
This looks safe to merge. It’s doc-only, the status update matches the notebook rebuild arc, and `git diff --check` is clean. GitHub CI correctly skipped the code jobs and only the change filter ran.

One non-blocking nit: the “Tasks runner panel” sentence right after the staged list still reads like future planning (“gets a final home… decide when the rail exists”). Since this PR is making the map historical/current-status accurate, I’d either mark that as historical too or trim it in a follow-up. Not worth holding the merge hostage; it’s a planning doc, not a bomb with a YAML fuse.

PR #389, opened 2026-06-21
fix(notebook): close the focused tile on ⌘W, not the terminal beside it

victorarias:
## Problem
With a Notebook tile docked next to a terminal, pressing **⌘W** while reading or editing a note closed the *previously-focused terminal pane* (ending that session) instead of the note you were looking at.

## Root cause
A docked tile lives inside `.session-terminal-workspace`, so the ⌘W close handlers treated it as a pane-close and closed `activePaneId` — the last-focused terminal pane — regardless of focus being in the tile.

The subtle part: **⌘W takes two different routes**, so both close handlers had to be fixed:
- **dev / browser / e2e:** DOM dispatcher → `terminal.close` (`handleCloseFocusedLeaf` in `SessionTerminalWorkspace`)
- **packaged app:** the native "Close Pane" menu item claims ⌘W and dispatches `session.close` (`handleCloseCurrentSessionShortcut` in `App.tsx`). The native menu swallows the key, so the DOM `terminal.close` never runs there — fixing only it would pass every test yet stay broken in the real app.

## Fix
Both handlers now detect a focused tile via `document.activeElement.closest('[data-pane-kind="tile"]')` (tiles carry `data-pane-kind="tile"` + `data-pane-id`; terminal panes are `data-pane-kind="agent"`) and **undock that tile** instead of closing the terminal pane.

## Tests / verification
- New unit test (2 cases) for the DOM `terminal.close` path.
- New packaged-app scenario `real-app:scenario-notebook-tile-close` exercising the **real native ⌘W** → undock (needs Accessibility, like the sibling finder scenario).
- Full frontend suite green (1137 passed); `tsc` clean.
- Verified in the **real running dev app** over the automation channel: docked a notebook tile beside a terminal, confirmed focus in the tile, fired ⌘W → tile undocked, terminal pane + session both survived. ✅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-06-21]
 
 ### Fixed
+- **`⌘W` in a docked Notebook tile now closes the Notebook, not the terminal beside it.** When a Notebook tile sat next to a terminal pane, pressing `⌘W` while reading or editing a note closed the previously-focused terminal pane (ending that session) instead of the note you were looking at. `⌘W` now closes the focused tile.
 - **The last line of a terminal is no longer cut off at the bottom of the window.** When a session's terminal was sized one row taller than the window could actually show — for example after the window's height landed on an awkward boundary, or when the session was opened at a size different from where its grid was last set — the bottom line (your prompt, or attn's `auto mode on` status line) got clipped at the window edge. The terminal now re-asserts a row count that fits the visible area, so the last line stays fully on screen at any window size.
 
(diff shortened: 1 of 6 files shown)


figgyster (approved):
This looks safe to merge. The fix catches the nasty macOS split-brain path — DOM `terminal.close` for browser/dev and native Close Pane → `session.close` for the packaged app — so the focused Notebook tile owns ⌘W instead of the last terminal pane getting whacked.

I ran `pnpm test` (1132 passed, 5 skipped), the new focused regression, `pnpm exec tsc --noEmit`, and `git diff --check`. GitHub's Frontend check was still pending on my last fetch, but the local frontend suite is clean.

PR #390, opened 2026-06-21
Protect the chief-of-staff session from accidental close

victorarias:
## What & why

The **chief of staff** is attn's profile-wide orchestrator — a persistent session that other agents report to. A stray `⌘W` or a misclicked close button could tear it down, losing the coordinator. This makes it non-closeable so an accident can't.

## Interpretation built: **hard-block** (option 1)

Victor confirmed the reading: *"chief of staff is not closeable — unset the chief status first."* So `⌘W` and the close-session action **no-op** on the chief, with a brief hint:

> Chief of staff is protected — unset the chief role to close it.

To close it deliberately, demote it first (`set_chief_of_staff` false), then close normally. **Ordinary sessions close exactly as before** — the guard is scoped to the session currently holding the chief role (no hardcoded id).

## How it's detected

Uses the existing role: the daemon's profile-wide chief-of-staff registry (`d.chiefOfStaffSessionID()`), surfaced to the frontend as `Session.chief_of_staff`. No new protocol, no schema change.

## Guard at two layers

- **Daemon (authoritative backstop).** Every close command flows through one of two handlers, so guarding both means no frontend path can slip past:
  - `handleUnregisterWS` → refuses with a `command_error` (the frontend already rejects the pending close on this).
  - `handleWorkspaceLayoutClosePane` → refuses with an error `workspace_layout_action_result`.
  - App-quit and other internal teardowns call `unregisterSession` directly and are **untouched**, so quitting the app still works.
- **Frontend (UX).** `handleClosePane` / `handleCloseSession` short-circuit for the chief *before* any side effects (worktree-cleanup prompt, focus churn, daemon round-trip) and show the hint. Both `⌘W` (`handleCloseCurrentSessionShortcut`, incl. the packaged-app native-menu route → `session.close`) and the close button (`handleRequestCloseSession`) funnel through these two leaves.

## Close paths covered

`⌘W` (DOM shortcut, terminal intercept, and native "Close Pane" menu → `session.close`), the sidebar close button, and process-exit auto-close all converge on the two guarded frontend leaves; the daemon backstop covers anything that still reaches it.

## Tests

- **Daemon integration** (`chief_of_staff_close_guard_test.go`): unregister refused for the chief (survives + `command_error`); non-chief still unregisters while a chief exists; close-pane refused for the chief's pane (survives, layout intact); a sibling non-chief pane still closes.
- **Frontend** (`App.chiefOfStaffClose.test.tsx`): close button and `⌘W` both no-op on the chief and show the hint; an ordinary session still closes (`sendWorkspaceClosePane` called, no hint).
- **Live**: built + installed `attn-dev.app`, drove the running dev daemon over WebSocket — `unregister` on the chief returned `command_error: "chief of staff is protected from closing; unset the chief role first"` and the session survived; role restored afterward.
- `go build ./...`, `go vet ./internal/daemon`, `tsc --noEmit`, full frontend suite — all green.

https://claude.ai/code/session_01UvYCbVmQRckKP6awEFbK7v


figgyster (approved):
This looks safe to merge. The protection lands in the right place: frontend close paths short-circuit before the cleanup/focus side effects, and the daemon still acts as the backstop for both unregister and workspace pane close.

I ran the focused daemon tests, the new frontend close tests, and the app build locally. Frontend CI was still pending when I checked, but the local path is clean. No blockers from me.

PR #391, opened 2026-06-22
test(notebook): de-flake the live-markdown scroll e2e

victorarias:
## Problem

A Playwright e2e in `app/e2e/live-markdown-editor.spec.ts` is flaky and currently **red on `main`** (it slipped through when #390 merged past it).

**Note on which test:** the task brief pointed at the *minimal-edit* test (line 119), but the [CI logs for the failing run](https://github.com/victorarias/attn/actions/runs/27916984795) show that test **passed** — the one that actually **failed** is its sibling, **"contrast: a full value swap snaps the scroller to the top"** (line 155):

```
> 177 |  await expect.poll(() => scroller.evaluate((el) => el.scrollTop)).toBeLessThan(50);
Expected: < 50
Received:   608
```

## Root cause (verified, not assumed)

The contrast test sets `.cm-scroller` `scrollTop = 600`, immediately swaps the whole document (the react-codemirror full-`value` replace), then asserts the viewport snapped to the top.

CodeMirror reacts to a raw `scrollTop =` write only **after** its scroll listener fires a measure that moves its internal scroll **anchor**. When the full-document swap races ahead of that measure (likely on a slow CI runner), CM's anchor is still pinned at the top, so the full replace has nothing to snap *from* and leaves the viewport at ~608.

I reproduced the **exact** CI value (`608`) locally by running the scroll-then-swap in a single tick (so CM never processes the scroll first) — and again under CDP CPU throttling.

The minimal-edit sibling never flaked because a below-the-fold edit preserves scroll *either way*, so it tolerated the same race — which is why it stayed green on CI.

## Fix (test-only — no product change)

- Add a `scrollIntoLongNote()` helper that scrolls down and then **waits for CM to acknowledge the scroll** before acting. CM virtualizes, so a deep below-the-fold line (`"Paragraph line number 25 …"`) is attached to the DOM only after the scroll measure has run — a deterministic "CM has settled" signal.
- Use it in **both** scroll tests so their precondition is deterministic.
- In the minimal-edit test, **poll the settled offset** (`expect.poll`) instead of reading `scrollTop` once.

**Intent preserved** — the tests still fail on a real regression: the swap must reach `< 50`, and the minimal edit must stay within `4px` of the scrolled position (a snap-to-top lands ~600px off).

No editor/scroll product code was touched; this is the documented CM-virtualization behavior, not a product bug.

## Verification

- Reproduced the exact CI failure (`608`) with the old code under CPU throttling.
- Fixed spec: **60/60** green on both scroll tests at **40× CPU throttle** (CI-like slowness), and **320/320** unthrottled (`--repeat-each=40`, full file).

### DIFF
--- a/app/e2e/live-markdown-editor.spec.ts
+++ b/app/e2e/live-markdown-editor.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect } from '@playwright/test';
+import { test, expect, type Page } from '@playwright/test';
 
 // Editing controls the LiveMarkdownEditor harness exposes for the scroll tests below.
 declare global {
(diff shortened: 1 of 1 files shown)


figgyster (commented):
This test-only fix is pointed at the actual race, not the symptom. Waiting for CodeMirror to render the below-the-fold line is a solid settled-scroll signal, and polling the preserved offset keeps the minimal-edit assertion from getting mugged by a transient re-measure.

I ran the targeted Playwright scroll cases locally: `e2e/live-markdown-editor.spec.ts --grep 'scroll'` passed 2/2. Frontend CI was still pending when I checked, so treat that as the last gate.

figgyster (approved):
CI caught up and the focused scroll coverage is green. This is safe to merge: the test now waits for CodeMirror's actual settled-scroll state before asserting the behavior, which is exactly the seam that was flaking.

No blockers from me.

PR #393, opened 2026-06-22
feat(dispatch): hand a large artifact to the chief via the Notebook

victorarias:
## Why

A delegated agent's report back to its chief is meant to be a **small payload** — its summary is what the chief sees, and it's clamped to 1500 runes in the deterministic raw-tier capture. So an agent that built a **large** artifact with the user (a report, a design doc, findings) had no way to forward it, and it wasn't even aware the Notebook existed (only chief-of-staff launches get Notebook guidance).

## What

A new `attn dispatch handoff` command for delegated agents:

```
attn dispatch handoff --file <artifact> --to <notebook-path> [--message <text>] [--coordination-file <json>]
```

- The daemon writes the artifact into the Notebook at the **chief/user-designated** path (root-relative, `.md`, containment-checked via `CleanPath`), then records a normal dispatch report whose **message carries the reference** — so it flows through `dispatch list`, the dashboard, and the raw-tier capture with no new structured field. Write-then-report ordering means a report can never reference an artifact that failed to land.
- **No hardcoded location.** The dispatched-agent launch prompt is now Notebook-*aware*, but placement comes from the chief's brief (or the user). If none was designated, the agent asks rather than inventing one.
- The chief "may choose to move" the artifact exactly as today — it's notebook-native; chief-side guidance now tells it to designate placement and curate the result.
- A refined re-handoff to the same path **updates** the one note (create-then-CAS-overwrite); a terminal handoff accepts `--coordination-file`, so "done, here's the full report" is one call and still lands the raw-tier journal capture.

## Design notes (decisions worth a look)

- **Reference in the report message**, not a new structured `DispatchReport` field — keeps the protocol delta to one new command and reuses the existing report/capture path.
- **Overwrite policy**: re-handoff updates the same note rather than erroring; since the destination is chief-designated, clobbering is the chief's call.

## Vision

Bundles the `docs/vision/chief-delegation-awareness.md` vision doc ("The output-aware chief"). This handoff is the **large-output branch** of that vision's *Output-into-context* / *Durable capture* big rocks — it does **not** touch the awareness/monitoring core (`dispatch watch`, Monitor, poll fallback), which remain future rocks. Cross-linked from the plan doc and the vision both ways.

## Protocol

New command `handoff_dispatch`; `ProtocolVersion` 117 → 118 (Go + frontend `PROTOCOL_VERSION`); regenerated `generated.go`/`generated.ts`.

## Tests / verification

- Daemon handler: happy path, non-dispatch rejection, overwrite-on-rehandoff, bad destination, empty content, terminal-also-journals, **full unix-socket path** (decode → route → handler). CLI arg parsing.
- Protocol version sync, frontend `tsc`, gofmt, vet, full daemon suite — all green.
- **Live smoke** on an isolated throwaway profile: confirmed the compiled CLI↔daemon round-trip and CLI validations.


figgyster (dismissed):
This looks safe to merge. The handoff path keeps the Notebook write ahead of the dispatch report, validates the destination through the existing Notebook path cleaner, and the tests cover both the handler and full socket route — the right seam for this feature.

Local checks passed here: `go test ./cmd/attn ./internal/daemon ./internal/protocol` and frontend `tsc --noEmit`. GitHub CI was mostly green on my last fetch; the Frontend job was still running, but the local typecheck covered the protocol surface I cared about.

No blockers from me.

figgyster (approved):
Still safe on the new head. The follow-up only removes the accidentally bundled vision doc; the dispatch handoff implementation I approved is unchanged, and `git diff --check` is clean on both the incremental change and the current PR patch.

One small non-blocking note: `docs/plans/2026-06-22-dispatch-notebook-handoff.md` still links to `../vision/chief-delegation-awareness.md`, which is no longer present in this PR or `main` unless PR #395 lands first. I wouldn’t hold the code path hostage over that, but the link wants either PR-order discipline or a tiny doc tweak.

No blockers from me.

PR #394, opened 2026-06-22
feat(dispatch): `attn dispatch watch` — the event/signal primitive

victorarias:
## What

Adds a blocking CLI subcommand **`attn dispatch watch <dispatch-id>`** that streams **one line per meaningful event** to stdout and **exits when the dispatch reaches a terminal state**. It is the event/signal primitive an output-aware chief watches per delegation (via a `Monitor`) instead of polling — the first big rock of the chief-delegation-awareness vision (`docs/vision/chief-delegation-awareness.md`, which explicitly cites this PR).

Scope is deliberately just **this command + the shared classifier**. No `Monitor` injection, no output-into-context piece.

## The signal definition (the crux)

A dispatch carries two independent state dimensions, both projected onto `ChiefOfStaffDispatch`:

- **`Status`** — the delegated session's projected *runtime* state (`launching` / `working` / `pending_approval` / `waiting_input` / `idle` / `scheduled` / `unknown`), or the sentinel `closed` once the session is gone. From attn's PTY/stop classifier.
- **`StructuredReport`** — the coordination envelope the agent *explicitly* files via `dispatch report --coordination-file` (`work_state`, `report_type`, an optional decision `Request`, `summary`, `next_action`).

**Doneness and failure are claimed by the agent, not inferred.** Only an explicit structured report yields `done`, `failed`, or `review`; the runtime `Status` governs liveness and the interim/neutral states. `Classify` maps the two onto exactly the events worth a chief notification, in priority order:

| # | Condition | Event | Terminal? | Exit |
|---|-----------|-------|-----------|------|
| 1 | report `work_state=failed` OR `report_type=failure` | `failed` | yes | 1 |
| 2 | report `work_state=completed` OR `report_type=completion` | `done` | yes | 0 |
| 3 | report `work_state=ready_for_review` | `review` | yes | 0 |
| 4 | `Status == closed`, no explicit terminal outcome | **`ended`** | yes | 0 |
| 5 | report has a pending decision `Request` | **`blocker`** | no | — |
| 6 | report `work_state=needs_input` OR `report_type=blocker` | **`blocker`** | no | — |
| 7 | `Status == idle`, no explicit outcome | **`ended`** | yes | 0 |
| 8 | `Status == waiting_input` | **`blocker`** | no | — |
| 9 | everything else | `none` | no | — |

### Signal, not noise — the exclusion

Step 9 swallows routine tool-permission prompts (`pending_approval`), plus `working`/`launching`/`scheduled`/`unknown`. A routine "approve this tool call" is not worth waking the chief — and it's cleanly distinguishable from a genuine decision-request: the former is runtime `pending_approval` → `none`; the latter is an explicit structured `Request` / `needs_input` → `blocker` (steps 5-6). Different data sources, so the exclusion is expressible (pinned by `TestClassify_ApprovalExclusionIsExpressible`).

### Explicit terminal outcomes are authoritative over close

The agent's three explicit terminal declarations — `failed`, `done`, and `review` (steps 1-3) — stand even once the session has closed, so they sit **ahead of** the `closed` check (step 4). A `ready_for_review` (or completion/failure) report filed in the same breath the agent exits is never lost to a polling race: the snapshot carries both a `closed` `Status` and the structured report, and the explicit report wins. (This was a fix from review — an earlier draft let `closed` beat `ready_for_review`, dropping the review handoff. Pinned by `TestRunWatch_ReadyForReviewSurvivesSessionClose` and the `ready_for_review then closed = review` classify case.)

### Silence implies neither success nor failure → neutral `ended`

A dispatch that ends **without an explicit structured outcome** — its session simply closes (step 4), or the agent stops at `idle` without reporting (step 7) — is the neutral terminal `ended` (exit 0). It still **emits and exits** (so a watch never hangs and every terminal state is covered), but it asserts no outcome; the narrative rides along on the line for drill-in. Only the agent's own structured report turns an end into `done`, `failed`, or `review`.

> This is a revision from the original draft, which mapped close→`failed` and idle→`done`. Per Victor: silence must imply neither failure nor success.

The `closed` check (step 4) sits **ahead of** the *interim* report states (5-6) on purpose: those are non-terminal, so once the session is gone a stale `needs_input` is not a live blocker — a dead session must resolve to a neutral terminal instead of hanging on a non-terminal `blocker` (a latent bug an earlier ordering had). `review` doesn't have that problem because it is itself terminal, which is why it can stay authoritative *above* step 4.

### Two judgement calls

- **`blocker` is non-terminal; `review` is terminal.** A decision-request means work is *incomplete and needs the chief's answer to continue* — the watch stays armed (one `Monitor` per delegation across its life), which the vision's both-ways loop ("answer a blocker… holds every thread") positively requires. `ready_for_review` is a reviewable handoff — terminal exit 0.
- **Exit codes: only an explicitly reported failure (or a watch-infra abort) is non-zero.** `done`/`review`/`ended` are all exit 0; `ended` is explicitly *not* a failure.

## Where the shared classifier lives

New package **`internal/dispatch`** — one source of truth: `Classify(d)`, `IsTerminalReport(report)`, `RunWatch(...)`, `FormatLine(...)`. To honor "done is never defined twice," the daemon now reuses it: `isTerminalDispatchReport` → `dispatch.IsTerminalReport`, `closedDispatchStatus` → `dispatch.SessionClosedStatus`.

## How it blocks (zero protocol surface)

Reuses the existing `list_dispatches` IPC (the chief already lists its own dispatches with full decorated snapshots) and filters by id client-side, polling at `--interval` (default 1s), flushing one line per event. **No `ProtocolVersion` bump** — additive alongside the parallel `feat/dispatch-notebook-handoff` work. The only shared-internals touch is two tiny, behavior-preserving single-source-of-truth edits in the daemon (`chief_of_staff_dispatch.go`, `notebook_dispatch_journal.go`). Flagged for merge sequencing.

## Manual verification (against real dispatches on the live daemon)

`dispatch watch` is read-only (only calls `list_dispatches`), verified against real dispatches:

1. **Routine silence** — watched a `working` dispatch ~90s: zero output, never exited. ✓
2. **`[done]`, exit 0** — explicit `completed` report (even after the session later closed: report wins). ✓
3. **`[review]`, exit 0** — a live `ready_for_review` dispatch, carrying summary **and** `→ next_action`. ✓
4. **`[ended]`, exit 0** — a session that closed without a structured completion report (was successfully merged but never filed a structured report); narrative carried on the line, asserts no outcome. ✓
5. **Bad id** → clear `not found` error, exit 2.

The two close-time races — `ready_for_review`-then-closed → `[review]` (report wins) and `needs_input`-then-closed → neutral `[ended]` (dead session beats a stale non-terminal blocker, never hangs) — are hard to stage live on demand, so they're pinned deterministically by `TestRunWatch_ReadyForReviewSurvivesSessionClose`, `TestRunWatch_DeadSessionWithStaleBlockerDoesNotHang`, and the matching `classify_test.go` cases.

## Tests

- `internal/dispatch/classify_test.go` — every state, all event kinds, the routine-approval exclusion, explicit-terminal-wins-over-close (`ready_for_review then closed = review`), dead-session-beats-stale-interim-report, summary precedence.
- `internal/dispatch/watch_test.go` — terminal exit (done/failed/ended/review/gone), neutral-ended-exit-0, explicit-failure-exit-1, ready_for_review-survives-close, dead-session-no-hang, dedupe, silence-then-exit on routine approval, abort-after-repeated-errors, rune-safe truncation.
- `cmd/attn/main_test.go` — arg parsing (positional id before flags).

## Out of scope / follow-ups

- Delegation-time `Monitor` injection and output-into-context (separate rocks).
- `dispatch status` / the non-Claude poll path don't yet *consume* the shared classifier; #394 factored it to be reusable, the actual reuse is the downstream poll-path rock.
- Reliable structured terminal reporting (so more ends are confident `done` rather than neutral `ended`) is a vision/lifecycle follow-up, not a watch change.


figgyster (dismissed):
This looks safe to merge. The dispatch signal classifier is in the right place, the watcher exits on every terminal path instead of becoming a tiny polling ghost, and the routine `pending_approval` exclusion is pinned by tests.

`go test ./...` and `git diff --check` pass locally; the visible GitHub daemon checks are green too. No blockers from me.

figgyster (dismissed):
This latest push changes the contract from "closed without completion is failure" to a neutral `ended`, which is fine by itself — but it also lets `Status == closed` beat an explicit `ready_for_review` report.

That loses the review handoff if the agent files `work_state=ready_for_review` and exits before the watcher’s next poll. `decorateChiefOfStaffDispatch` keeps the structured report while projecting `Status = closed`, so this is a real polling race, not a theoretical alley cat. The PR still advertises reviewable handoff as a meaningful terminal event, but the current classifier turns that snapshot into `[ended]` instead of `[review]`.

Required before merge:
- Keep explicit `ready_for_review` authoritative over `closed` the same way explicit completion/failure are, or otherwise preserve the review event after session close.
- Add the regression case for `Status == closed` + `StructuredReport.WorkState == ready_for_review`.

I ran `git diff --check` plus `go test ./internal/dispatch ./cmd/attn ./internal/daemon`; they pass, but this is a contract bug in the new signal semantics.

figgyster (dismissed):
The blocker is still present on this head. The PR diff still classifies `Status == closed` before `ready_for_review`, so a dispatch that files `work_state=ready_for_review` and then exits before the next `watch` poll still emits `[ended]` instead of `[review]`.

- keep explicit `ready_for_review` authoritative over `closed`, or otherwise preserve the review handoff after session close
- add the regression case for `Status == closed` + `StructuredReport.WorkState == ready_for_review`

I re-ran `git diff --check` and `go test ./internal/dispatch ./cmd/attn ./internal/daemon`; they pass. The tests are green, but the contract is still wearing a fake moustache.

figgyster (dismissed):
This fixes the blocker. `ready_for_review` is now treated like the other explicit terminal declarations, so the handoff survives a closed-session snapshot instead of collapsing into neutral `[ended]`.

The regression is pinned at both the classifier and watch loop, and the full Go suite plus GitHub checks are green. No blockers from me.

figgyster (approved):
Still safe after the rebase/force-push. The current PR patch is the same dispatch-watch work on the updated base, the `ready_for_review` race fix is intact, and the full Go suite plus GitHub checks are green.

No blockers from me.

PR #397, opened 2026-06-22
feat(dispatch): surface a crash/kill-on-close as a dispatch failure

victorarias:
## Why

PR #394 made a chief-of-staff dispatch that closes **without** a structured terminal report the neutral terminal `ended` — *silence is neither success nor failure* — which killed the old false `[failed]` for an unreported success. But that left a real gap: a genuine **crash/kill** mid-work now reads as the same quiet `ended` as a clean stop. A killed agent looks uneventful, which trains the chief to ignore the signal from the other direction.

This is **Option B (crash visibility)**, chosen by Victor over inferring success on a clean close: keep `ended` neutral for clean/unconfirmed closes (never claim an unearned success), but surface a *real* cut-off as a failure.

## What

- **Capture the close-state at the source.** The daemon records the delegated session's last attn-classified runtime state the moment its process exits — in `handlePTYExit` *before* the idle-clobber, with a backstop in `dropSessionRecord` for removal paths that bypass it (reap on restart, liveness sweep, worktree teardown). First-writer-wins keeps the true pre-clobber state authoritative over a later teardown read that would only see the clobbered `idle`. Persisted as a new durable `closed_state` on the dispatch (survives daemon restart).
- **Classify it.** The shared `internal/dispatch.Classify` (single source of truth) surfaces a close cut off mid-flight — `working` / `launching` / `pending_approval` — as `failed` (`session_crashed`), while a clean rest (`idle` / `waiting_input`) or an unconfirmed close (`unknown` / unstamped) stays neutral `ended`. Agent-declared `done` / `failed` / `review` reports remain authoritative over any close-state.
- **Make it immediate.** The decorate path projects the terminal `closed` status as soon as a close-state is stamped, so a crash is classified at process-exit time even while the idle-clobbered session record still lingers in the store.

The safe direction is preserved: attn never asserts a failure it can't evidence, and never a `done` it wasn't told. An unstamped/legacy close behaves exactly as before (`ended`).

## Tests

- `internal/dispatch` — classifier table extended with the full close-state matrix (crash states → failed; clean/unconfirmed → ended; terminal reports beat a crash close-state).
- `internal/store` — `closed_state` first-writer-wins, in-memory + SQLite paths, durability across reopen; migration 53 added + idempotent.
- `internal/daemon` — end-to-end through real `handlePTYExit` / `dropSessionRecord` → decorate → `Classify`: crash visible, clean close neutral, first-writer-wins across exit-then-removal, drop backstop, non-dispatch no-op.

Verified: `gofmt`, `go vet`, `go test ./internal/{dispatch,store,daemon,protocol} ./cmd/attn`, frontend `tsc --noEmit`. Protocol `119 → 120` (new `closed_state` field; frontend version bumped in lockstep).

## Notes
- Base is `main` (#394 already merged).
- Vision doc `docs/vision/chief-delegation-awareness.md` updated: the *Reliable doneness-on-close* big rock now reflects the implemented design, and the re-arming caveat is marked resolved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-06-23]
 
+### Changed
+- **A delegated agent that crashes or is killed mid-work now surfaces as a failure to its chief.** When a chief-of-staff dispatch's session ends, attn now tells a clean stop from a real crash: a session cut off while it was still working, starting up, or waiting on a tool approval is reported as a failure (`[failed]`), while a session that closed from a settled rest — or one attn couldn't classify — stays the neutral "ended" introduced with `dispatch watch`. attn still never claims a success the agent didn't report; this only makes a genuine crash visible instead of letting it look like an uneventful end.
+
 ### Fixed
 - **Long Notebook notes keep their rendered formatting and cursor position while you edit.** Moving the cursor beyond the first few thousand characters no longer turns the rest of a note back into raw Markdown or sends ArrowUp to the properties card. Clicking into the body also leaves the frontmatter card stable; raw YAML appears only when you click the card to edit it.
 - **Chief-of-staff delegations no longer disappear into muted workspaces.** Delegating into an existing muted workspace now brings that workspace back into the sidebar, and `attn list` marks sessions whose workspace is muted so the chief can see that state before choosing a target.
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -172,7 +172,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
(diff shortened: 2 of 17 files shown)


figgyster (dismissed):
This looks safe to merge. The close-state capture lands at the right seam — before the idle-clobber can erase the evidence — and the shared classifier keeps the important safety line intact: explicit reports win, real mid-flight cutoffs become visible failures, and ambiguous/clean closes stay neutral.

I ran `go test ./internal/dispatch ./internal/store ./internal/daemon` and `make test`; both passed. GitHub's Frontend check was still pending when I checked, but no blockers from me.

figgyster (approved):
Still safe to merge after the force-push. The current patch is back to the close-state crash signal, and the key seam is intact: capture before idle-clobber, first-writer-wins persistence, and classifier precedence keeps explicit reports authoritative while making real mid-flight cutoffs visible.

I ran `go test ./internal/dispatch ./internal/store ./internal/daemon`, `make test`, and `pnpm --dir app test -- --run`; all passed locally. GitHub Frontend was still pending when I checked, but no blockers from me.

PR #398, opened 2026-06-23
fix(delegation): reveal muted target workspaces

victorarias:
## Intent / Why

A chief-of-staff delegation could target an existing muted workspace successfully, but the new agent stayed hidden from the normal sidebar. The dashboard still linked to the session, so opening it switched away from the chief into a workspace the sidebar did not show. The chief also had no way to see the target workspace was muted from `attn list`.

## Summary

- Unmute an existing workspace after a chief-of-staff delegation is successfully tracked, so the delegated agent appears in the sidebar.
- Preserve mute state for ordinary session-to-session delegation.
- Expose `workspace_muted: true` on session snapshots so `attn list` makes hidden workspace ownership visible before delegation.
- Bump the daemon/frontend protocol to 120 and update generated types, bundled delegation guidance, and the changelog.

## Test plan

- [x] `go test ./...`
- [x] `pnpm --dir app test` (123 files, 1,142 tests)
- [x] Focused chief/ordinary delegation and workspace snapshot regression tests
- [x] `make dev` and `make install-daemon-dev` packaged dev build/install
- [x] `git diff --check`

## Out of scope

- Ordinary delegation continues to respect an explicitly muted workspace; only tracked chief-of-staff delegation revives it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-06-23]
+
+### Fixed
+- **Chief-of-staff delegations no longer disappear into muted workspaces.** Delegating into an existing muted workspace now brings that workspace back into the sidebar, and `attn list` marks sessions whose workspace is muted so the chief can see that state before choosing a target.
+
 ## [2026-06-22]
 
 ### Added
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -172,7 +172,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '119';
+export const PROTOCOL_VERSION = '120';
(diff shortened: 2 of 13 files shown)


figgyster (dismissed):
This is the right behavioral fix, but the protocol artifacts are only half regenerated.

`Session.workspace_muted` was added to TypeSpec and Go `generated.go`, but `app/src/types/generated.ts` still lacks the field. Running quicktype against the current JSON schema produces the missing `workspace_muted?: boolean` entries for `Session`/`SessionElement`, so `make check-types` should fail in a runner with the generator toolchain installed.

Required before merge:
- Regenerate and commit `app/src/types/generated.ts` for protocol 120.

Local validation: `go test ./...` passed, `pnpm --dir app test` passed (123 files / 1,142 tests). The daemon/frontend CI was still finishing when I checked, but this generated-type drift is enough to keep the review red. Tiny protocol rake, but it’s still a rake.

figgyster (approved):
Fixed. `workspace_muted` is now present in the frontend generated types and conversion maps, so the protocol 120 contract is lined up again.

Local checks passed: `git diff --check`, `pnpm --dir app exec tsc --noEmit`, `go test ./...`, and `pnpm --dir app test` (123 files / 1,142 tests). Frontend CI was still running at last fetch, but the local principal checks cover the blocker.

No blockers from me.

PR #399, opened 2026-06-23
Add chief of staff support for copilot

harleenquinzell:
## Summary

Enables chief-of-staff delegation to **GitHub Copilot**. Copilot now declares the
`initial_prompt` capability and launches with `copilot --interactive <brief>`,
joining Claude and Codex as a delegation target. Ordinary (non-delegated) Copilot
sessions are unaffected.

## Flag decision: `--interactive` (NOT `-p`/`--prompt`) — with evidence

An offline review flagged the original `-i` as likely wrong, because the **online**
GitHub docs ([cli-programmatic-reference](https://docs.github.com/en/copilot/reference/copilot-cli-reference/cli-programmatic-reference))
only document `-p`/`--prompt` and omit `-i` entirely. I installed and checked the
**actual** CLI to settle it:

`copilot --help` (CLI **v1.0.63**, `brew install copilot-cli`):

```
-i, --interactive <prompt>   Start interactive mode and automatically execute this prompt
-p, --prompt <text>          Execute a prompt in non-interactive mode (exits after completion)
```

Documented example: `$ copilot -i "Fix the bug in main.js"`. Parse-validated against
the binary: `copilot --interactive` and `copilot --yolo --interactive` both error only
with `option '-i, --interactive <prompt>' argument missing` (flag + `--yolo` combo
accepted), versus `unknown option` for a bogus flag.

**Conclusion:** `-i`/`--interactive` is the **correct** flag. It starts a *persistent
interactive* session that auto-runs the brief and stays alive for steering — the exact
model attn delegation uses for Claude/Codex. `-p`/`--prompt` would be **wrong**: it runs
non-interactively and exits after completion, which would tear the delegated session down
immediately. So the original author's flag choice was right; I switched it to the long
form `--interactive` for clarity and documented the evidence inline in `copilot.go`.

**Runtime-validation status (honest):** the flag and its `--yolo` combination were
verified against the live binary's parser. A full end-to-end delegated Copilot run (an
interactive session auto-executing a brief through attn's PTY) was **not** exercised —
that needs an authenticated Copilot session. CI only unit-tests argument construction.

## Test reconciliation

- **`internal/daemon`** — replaced the inherited `TestDelegateRejectsCopilotInitialPrompt`
  (which asserted copilot delegation must error — the behavior this PR reverses, and the
  exact blocker figgyster flagged) with **`TestDelegateAcceptsCopilotInitialPrompt`**,
  asserting delegation now succeeds and spawns a tracked Copilot session carrying the brief.
- **`internal/agent`** — `TestCopilotBuildCommandInitialPrompt` now asserts `--interactive`;
  `TestCopilotSupportsInitialPrompt` message updated.
- Full `go test ./...` and `make test` are green (1791 tests, 10 env-gated skips).

## Housekeeping

- Marked the superseded *"Copilot excluded from delegation"* decision in
  `docs/plans/2026-06-08-chief-of-staff.md` (kept the rationale).
- Added a user-facing changelog entry.

## Rebase

Branch was rebased onto latest `main` before this work; `origin/chief-copilot` now
fast-forwards.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,6 @@ app/src-tauri/binaries/
 
 # Generated per-profile Tauri config overlays (see scripts/build-app-profile.sh)
 app/src-tauri/*.gen.conf.json
+
+# Local env files
+**/.npmrc
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-06-23]
 
+### Added
+- **The chief of staff can now delegate to GitHub Copilot.** Copilot joins Claude and Codex as a delegation target: the chief spins up a Copilot session with the brief, and the agent starts in an interactive session that runs the brief and stays open for follow-up steering. Ordinary (non-delegated) Copilot sessions are unaffected.
+
 ### Fixed
 - **Long Notebook notes keep their rendered formatting and cursor position while you edit.** Moving the cursor beyond the first few thousand characters no longer turns the rest of a note back into raw Markdown or sends ArrowUp to the properties card. Clicking into the body also leaves the frontmatter card stable; raw YAML appears only when you click the card to edit it.
 - **Chief-of-staff delegations no longer disappear into muted workspaces.** Delegating into an existing muted workspace now brings that workspace back into the sidebar, and `attn list` marks sessions whose workspace is muted so the chief can see that state before choosing a target.
--- a/docs/plans/2026-06-08-chief-of-staff.md
+++ b/docs/plans/2026-06-08-chief-of-staff.md
@@ -170,8 +170,12 @@ ChiefOfStaffDispatch {
   attached to the agent that created them and remain visible as history.
 - Placement defaults to the current workspace. `--new-workspace`, `--cwd`, and
   `--worktree` opt into a new workspace; `--workspace` targets an existing one.
-- Copilot is excluded from delegation for now; ordinary Copilot sessions remain
-  supported.
+- ~~Copilot is excluded from delegation for now; ordinary Copilot sessions remain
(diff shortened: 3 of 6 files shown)


figgyster (dismissed):
This is close, but the daemon test suite is red on the current head.

`go test ./...` fails in `internal/daemon` because `TestDelegateRejectsCopilotInitialPrompt` still expects Copilot delegation to reject initial prompts:

```text
--- FAIL: TestDelegateRejectsCopilotInitialPrompt
    delegate_test.go:705: delegate() error = <nil>, want unsupported initial prompt error
```

Since this PR intentionally flips Copilot to `HasInitialPrompt: true`, that daemon test needs to be updated or replaced to pin the new delegation behavior. Required before merge: make the principal Go suite pass again.

figgyster (dismissed):
The blocker is still present on the new head.

I re-ran the focused daemon path and `go test ./internal/agent ./internal/daemon` still fails in `internal/daemon`:

```text
--- FAIL: TestDelegateRejectsCopilotInitialPrompt
    delegate_test.go:819: delegate() error = <nil>, want unsupported initial prompt error
```

The previous implementation change is still coherent, but this stale daemon test keeps the principal suite red. Required before merge: update/remove that negative Copilot delegation test and replace it with coverage for the new supported initial-prompt behavior.

figgyster (dismissed):
This looks safe to merge now. The Copilot launch path uses the persistent interactive mode (`--interactive`) instead of the one-shot prompt path, and the daemon-level delegation test now pins the new behavior instead of fighting it.

`go test ./...` passes locally. GitHub checks weren’t reported yet when I last fetched them, but the principal Go suite is green from here.

figgyster (approved):
Still safe after the rebase/merge-from-main. The PR patch is unchanged in substance, the Copilot delegation path is still pinned at the right seam, and the daemon/store/dispatch code now sitting under it is green.

`go test ./...` passes locally, and the visible GitHub checks are green/skipped as expected. No blockers from me.

PR #400, opened 2026-06-23
fix(notebook): stabilize long markdown editing

victorarias:
## Intent / Why

Long Notebook notes could lose rendered Markdown after the first few thousand characters when a click or cursor movement rebuilt preview decorations from CodeMirror's incomplete initial syntax tree. The resulting layout churn could also send ArrowUp back to the frontmatter card. Focusing the note body had a second geometry race: frontmatter briefly expanded from the stale position-0 selection before CodeMirror placed the click.

## Summary

- scope live-preview decoration work to the logical viewport, ensure parsing reaches that viewport, and refresh when background parsing advances so long-note suffixes stay rendered without synchronously parsing the whole document on every keypress
- make raw frontmatter editing explicit to clicks on the properties card, keeping ordinary body focus from changing document geometry mid-click
- add a long-document fixture beyond CodeMirror's 3,000-character initial parse boundary plus unit and browser regressions for rendering, ArrowUp locality, and frontmatter stability

## Test plan

- [x] `make test-frontend` (1,143 tests)
- [x] `pnpm exec tsc --noEmit`
- [x] `pnpm exec playwright test e2e/frontmatter-card.spec.ts e2e/live-markdown-editor.spec.ts` (13 tests)
- [x] `make dev` (signed, installed, and launched the isolated `attn-dev.app`)
- [ ] In a packaged app with native-input Accessibility permission, open a long frontmatter note and confirm ArrowUp stays local; the automated packaged Notebook scenario could not deliver its native shortcut in this environment because the harness process lacks Accessibility permission

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-06-23]
 
 ### Fixed
+- **Long Notebook notes keep their rendered formatting and cursor position while you edit.** Moving the cursor beyond the first few thousand characters no longer turns the rest of a note back into raw Markdown or sends ArrowUp to the properties card. Clicking into the body also leaves the frontmatter card stable; raw YAML appears only when you click the card to edit it.
 - **Chief-of-staff delegations no longer disappear into muted workspaces.** Delegating into an existing muted workspace now brings that workspace back into the sidebar, and `attn list` marks sessions whose workspace is muted so the chief can see that state before choosing a target.
 
 ## [2026-06-22]
--- a/app/e2e/frontmatter-card.spec.ts
+++ b/app/e2e/frontmatter-card.spec.ts
@@ -56,6 +56,18 @@ test.describe('FrontmatterCard', () => {
     await expect(page.locator('.cm-content')).toBeFocused();
   });
 
+  test('opens frontmatter editing from a synthesized semantic click', async ({ page }) => {
+    await page.goto('/test-harness/?component=FrontmatterCard');
+    await page.waitForFunction(() => window.__HARNESS__?.ready === true);
+    const card = page.getByRole('button', { name: 'Edit note properties' });
+
+    // VoiceOver and other assistive technologies invoke a button action through click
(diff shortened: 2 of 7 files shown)


figgyster (commented):
This looks clean. The fix moves both failure modes to the right boundary: live preview work is bounded to the viewport while forcing the parser far enough for that viewport, and frontmatter raw-edit mode is now an explicit card click instead of a focus side effect. That matches the bug instead of just bribing CodeMirror with timing hacks.

I ran the focused notebook unit coverage, `pnpm exec tsc --noEmit`, and the frontmatter Playwright spec locally. No blockers from me; GitHub's Frontend CI was still pending when I checked, so treat this as green once that job lands.

figgyster (approved):
This is safe to merge from my side. The follow-ups tighten the exact ugly edges left by the first pass: scroll anchoring now gets a bounded decoration margin, and blur moves the caret out of hidden frontmatter before the card comes back. That avoids the classic CodeMirror magic trick where the cursor disappears into a wall.

I reran the focused notebook unit tests, `pnpm exec tsc --noEmit`, and the frontmatter/live-markdown Playwright specs locally. GitHub's Frontend job was still pending at last check, but the reviewed path is green here.

figgyster (approved):
Still safe to merge. The keyboard path is the right addition here: the properties card is now an actual accessible button, Enter/Space reveal the YAML, and focus lands back in the editor where the raw block is visible. That closes the accessibility hole without reopening the original body-click geometry race.

I reran the focused notebook unit tests, `pnpm exec tsc --noEmit`, and the frontmatter/live-markdown Playwright specs locally. GitHub's Frontend job was still pending at last check, but the reviewed path is green here.

figgyster (approved):
Still good. Moving activation onto the semantic `click` is the right accessibility polish: pointer down only blocks CodeMirror's widget-selection side effect, while actual activation now covers mouse, keyboard, and assistive-tech button actions. Nice little trapdoor closed.

I reran the focused notebook unit tests, `pnpm exec tsc --noEmit`, and the frontmatter/live-markdown Playwright specs locally. GitHub's Frontend job was still pending at last check, but the reviewed path is green here.

PR #401, opened 2026-06-23
fix(delegation): preserve tracked session identity

victorarias:
## Intent / Why

Chief-of-staff agents delegated into a shared parent directory could not report progress after switching tool execution into their branch-specific child worktrees. `ATTN_WRAPPER_PATH` and the session identity disappeared, so `dispatch report`, `dispatch inbox --unread`, and workspace-context publication all failed with `no session` even though attn had persisted the sessions and their dispatch associations correctly.

## Root cause

attn supplied its routing variables to the top-level Codex process, but Codex applies `shell_environment_policy` independently to tool commands. With the user's `inherit = "core"` policy, a tool command whose working directory differed from the session launch directory did not inherit `ATTN_SESSION_ID`, `ATTN_WRAPPER_PATH`, or `ATTN_SOCKET_PATH`.

Both observed placement paths converged on the correct daemon spawn path; the failure happened afterward when each agent ran tools from a child worktree.

## Summary

- Pin the launch-specific session ID, wrapper path, and socket path in Codex's per-launch shell environment policy so tool commands retain the correct attn identity across working directories.
- Pass the wrapper path explicitly in the top-level Codex environment instead of relying only on inherited PTY state.
- Cover all placement modes—current workspace, new workspace via `--cwd`, and existing workspace via `--workspace`—and verify the delegated identity can report, read its unread inbox, and publish workspace context.
- Document the user-visible fix in the changelog.

## Test plan

- [x] `go test ./internal/hooks ./internal/agent`
- [x] Focused delegation, dispatch, inbox, and workspace-context daemon tests
- [x] `go test -short ./...`
- [x] `go test ./...`
- [x] `go vet ./...`
- [x] `make dev` (signed isolated dev app build/install)

## Out of scope

Delegation placement semantics are unchanged. This only makes the existing session identity and profile-specific routing survive Codex tool working-directory changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ### Fixed
 - **Chief-of-staff delegations no longer disappear into muted workspaces.** Delegating into an existing muted workspace now brings that workspace back into the sidebar, and `attn list` marks sessions whose workspace is muted so the chief can see that state before choosing a target.
+- **Delegated Codex agents keep their attn identity when working in child directories.** An agent launched in a shared parent directory could lose its session ID and attn wrapper path after running tools from a branch-specific child worktree, so tracked reports, inbox access, and workspace-context publication failed with "no session." Managed Codex launches now preserve the correct session routing across tool working directories and workspace placement modes.
 
 ## [2026-06-22]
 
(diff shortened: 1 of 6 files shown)


figgyster (approved):
This looks safe to merge. Pinning the attn session/socket/wrapper into Codex's per-launch shell environment policy is the right seam for the cwd-hopping failure, and the new delegation test hits the actual user path: report, unread inbox, and workspace-context publish across all placement modes.

Local checks passed here too: `go test ./internal/hooks ./internal/agent ./internal/daemon`, `go test -short ./...`, and `go vet ./...`. CI is green. No blockers from me.

PR #402, opened 2026-06-23
fix(chief): don't peg the chief of staff green while it's only watching

victorarias:
## Problem

When a Claude turn yields with background work still in flight, the Stop hook
reports `working` so the session stays green until the work drains. That's
correct for ordinary background compute — but it inverts for the **chief of
staff**, whose normal mode is to *watch*: it arms a `Monitor` over the agents it
delegated (or a background poll loop) and then waits. attn rendered that
async-waiting chief as a solid green/active dot, so:

- a chief watching several delegations looked **permanently busy**, and
- the at-a-glance "is the chief actually working?" signal lost its meaning —
  the *busier* its watched work, the *greener* it looked.

(`Monitor` and a `run_in_background` Bash loop are indistinguishable in the Stop
payload — both arrive as `background_tasks: [{type:"shell", status:"running"}]`.)

## Fix

Relax the `background_tasks → working` rule **for the chief of staff only**:

- In `runHookStop`, when a Stop carries background work, resolve the
  daemon-owned chief role (`sessionIsChiefOfStaff` → `Query`, the same decorated
  list `attn list` shows). If this session is the chief, the Stop falls through
  to normal classification (settling `idle`/`waiting`) instead of being pegged
  `working`.
- A **pending scheduled wakeup still parks `scheduled`** (quiet blue) — that
  branch is untouched, so a chief with a timed wake-up still reads as parked,
  not green.
- **Every non-chief session is unchanged.**
- The chief role is only queried when there's background work to relax, so
  normal stops pay nothing; any query error falls back to today's behavior.

No protocol change. Scoped to the chief deliberately: we want to experience how
**non-Claude chiefs (e.g. Codex, which has no `Monitor` tool)** surface watching
behavior before generalizing.

## Verification

Unit tests extend the non-terminal-Stop precedence table with the chief-relax
cases (background-only → classify, background+cron → `scheduled`, cron-only →
`scheduled`).

Live in a dev install (real daemon + real hook binary), firing an **identical**
`background_tasks + session_crons` Stop payload at two sessions:

| session | `chief_of_staff` | resulting state |
|---|---|---|
| worker-test | false | **`working`** (green) |
| chief-test  | true  | **`scheduled`** (blue) |

## Alternative considered (deferred): transcript corroboration

The fix here keys on **role** (chief) plus the **presence** of background work.
The more principled — and agent-agnostic — alternative is to key on the
**transcript** instead: parse it to decide whether the chief is *genuinely
working* vs *async-waiting*, regardless of role. A turn that produced its final
assistant message and is now only waiting on a Monitor/loop reads as
idle/waiting even though `background_tasks` is non-empty; a turn mid-generation
reads as working.

Why it's the better long-term shape:
- It's **not chief-specific** — it would give Codex/other chiefs (no `Monitor`
  tool) the same correct "watching ≠ working" behavior.
- It's the robust fix for the **related "missing blue icons" bug**: a parked
  `scheduled` session is protected from idle/waiting PTY noise but any `working`
  frame is allowed to clobber it (`internal/agent/claude.go` `ShouldApplyPTYState`),
  so a stray spinner/redraw frame silently ends the park. The correct gate is
  "did a **new transcript turn** appear since we parked?" — a real resume writes
  one, a stray redraw doesn't. A naive time/count debounce is actively harmful:
  a genuinely fast resume (<~1.2s turn) emits only one `working` frame, so
  suppressing "the first" frame (or all frames for N seconds) swallows real
  resumes and leaves the session stuck blue.

Deferred because it's more invasive (track transcript position at park time,
corroborate transitions against it) and because we want real-world Codex-chief
watching behavior in hand before committing to the general model. This PR is the
safe, narrow first step.


figgyster (approved):
This looks safe to merge. The chief-only relaxation keeps the scheduler path blue, leaves ordinary sessions on the existing background-work guard, and the daemon-role lookup fails closed if it can’t answer — exactly the right amount of narrow for this state signal.

`go test ./...` passed locally. No blockers from me.

figgyster (approved):
Still safe after the rebase onto the newer chief/delegation work. The PR patch is the same narrow chief-state relaxation, and it now sits cleanly on top of the dispatch/watch changes.

`go test ./...` passed locally, and the visible checks are green/skipped as expected. No blockers from me.

PR #403, opened 2026-06-23
Self-report trigger loop — chief↔delegated-agent awareness

victorarias:
## Self-report trigger loop — chief↔delegated-agent awareness

Builds the aligned chunk of the chief-in-the-loop vision (plan:
`docs/plans/2026-06-23-self-report-trigger-loop.md`, cherry-picked here so merging
lands it on main). The chief stays *aware* of its delegations and can *steer* them
— bidirectionally — **without babysitting**.

### The model
The trigger fires on exactly two things and is silent otherwise:
- **forward** — the agent's self-reported terminal/blocked outcome;
- **reverse** — a mailbox item, delivered via the agent's own quiet inbox watch.

Daemon/runtime state is no longer a trigger — that flicker was the source of the
#399 misfires.

### What's in this PR
- **Forward trigger** (`internal/dispatch/classify.go`): removed the daemon-state
  steps (closed/idle/waiting_input). The watch now fires only on a self-reported
  `done`/`failed`/`review` (terminal) or `decision-request`/`needs_input`/`blocker`
  (interim). The #399 "empty report at waiting_input → false blocker" misfire is gone.
- **Self-report affordance** (`cmd/attn dispatch report`): first-class
  `--done` / `--failed` / `--review` / `--blocked` (with `--question` /
  `--recommendation` / `--consequence`) flags that synthesize the structured report
  over the existing envelope — no more hand-written coordination JSON for the common
  cases. A bare `--message` is a silent note (no trigger). `--coordination-file`
  stays as the escape hatch.
- **Reverse ambient UI** (`app/`): a sidebar pending-mail badge (+ collapsed-rail
  dot) and an on-agent terminal overlay, both driven by the existing
  `unread_message_count`. The overlay is pointer-events-scoped so it never steals
  PTY focus. Clicking it fires the existing wake doorbell.
- **Guidance** (`internal/agent/attn_skill/references/`): the launch prompt and skill
  references teach the chief to arm one quiet watch, self-report only at
  terminal/blocked, watch the inbox for steering, and escalate-unless-sure. Single
  authoritative home (the bundled skill); `~/exo` persona doc untouched.

### Notable decisions
- **No `ProtocolVersion` bump** — everything reuses existing protocol surfaces.
- **"watching ≠ working" dissolved via the quiet watch** — no state-inference code.
  A quiet (event-only) watch produces neither a running `background_tasks` entry at
  Stop nor an animated PTY frame, so it settles to idle on its own.
- **Accepted gap** (per the plan's Deferred): an agent that ends *without* a terminal
  self-report leaves a quiet, non-exiting watch (silence ≠ success). The deferred
  watch-timeout→escalate is the backstop; codex auto-wake is also deferred.

### Testing
- Go: `go test ./...` green (full suite).
- Frontend: `pnpm test` green (1151 tests, incl. new PendingMailBadge / OnAgentMailOverlay specs).

🤖 Authored by an attn-delegated agent on behalf of Victor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-06-23]
 
+### Added
+- **Self-report finishing or getting stuck with one flag — no JSON.** A delegated agent can now declare a terminal or blocked outcome directly: `attn dispatch report --done` / `--review` / `--failed` / `--blocked`. `--blocked` takes `--question`, `--recommendation`, and `--consequence` to put a concrete decision in front of the chief. A plain `--message` with no flag is a silent note that does not wake the chief. The full `--coordination-file` JSON envelope stays as an escape hatch.
+- **See pending chief mail without opening the dashboard.** A delegated agent that has unread mail from its chief now shows an amber envelope badge in the sidebar (and a small dot on the collapsed rail), plus an overlay on the agent's own terminal. Clicking the overlay prompts that agent to check its inbox.
+
+### Changed
+- **A chief's `dispatch watch` fires only on what an agent reports, not on its process state.** Watching a delegation no longer reacts to the agent's runtime status, so a turn-boundary rest is never misread as "ended" and stopping with an empty report is no longer a false "waiting" blocker. The watch wakes the chief only when the agent self-reports done, ready-for-review, failed, or blocked. Trade-off: an agent that ends without filing a report now leaves its watch quietly waiting rather than emitting a neutral "ended" — silence is neither success nor failure.
+
 ### Fixed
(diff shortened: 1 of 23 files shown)


figgyster (dismissed):
This looks safe to merge. The self-report-only trigger model matches the plan: runtime state no longer wakes the chief, the explicit report flags cover the happy path, and pending chief mail gets a visible but non-invasive affordance in the UI.

I checked the accepted crash-without-report gap against the plan/changelog; not treating that as a blocker here since the timeout/escalation backstop is explicitly deferred. Local validation passed: `go test ./...`, `pnpm --dir app test -- --run`, and `git diff --check`.

No blockers from me.

figgyster (approved):
The follow-up fixes are the right shape. The resolved-request path now stays silent instead of re-firing the blocker the chief just answered, the CLI rejects dropped `--recommendation`/`--consequence` data, and the guidance now describes the reverse channel accurately instead of pretending the chief-side watch sees mail reactions.

Local validation is green again: `go test ./...`, `pnpm --dir app test -- --run`, and `git diff --check`.

No blockers from me.

PR #404, opened 2026-06-24
docs: add notebook knowledge-base schema (OKF)

victorarias:
## Why

The notebook knowledge base has no canonical schema — the chief of staff, keeper, and human editors each follow slightly different conventions picked up from scattered code comments and prompt fragments. A single authoritative schema file that both agents and humans reference eliminates this drift.

## Summary

- Adds `docs/notebook-schema.md` — the canonical specification for knowledge-base files
- Defines the **frontmatter vocabulary** with two implementation tiers: fully-surfaced fields (`type`, `summary`, `updated`) vs convention-only fields (`sources`, `resource`, `paths`) that survive round-trip but have no store/protocol backing yet
- Documents **opt-in body sections** (Technical Details, Contracts, Behaviors) as patterns with "use this when" guidance, not universal requirements — different note types use different sections
- Adds a **Section Conventions by Type** table mapping conventional types to their typical body structure
- Adds **Lifecycle and Placement** rules (orient-first, update-over-create, PARA placement heuristic)
- Adds **File Naming** conventions and **Parsing Rules** for machine consumers
- Documents `parent` and `state` as **future/not-yet-implemented** fields, listing the files that need changes
- Includes two complete examples: a minimal `operating-principle` note and a full `system` note with all structured sections

## Test plan

- [x] Schema is consistent with existing `Document.go` parser behavior
- [x] Frontmatter fields match what `document.go` / `notebook.go` actually support
- [x] Reserved files match `layout.go` constants
- [x] PARA layout matches scaffold in `layout.go`
- [ ] Chief of staff can follow the schema to produce well-formed notes (manual)

### DIFF
--- a/docs/notebook-schema.md
+++ b/docs/notebook-schema.md
@@ -1,0 +1,372 @@
+# Notebook Schema
+
+This is the canonical specification for attn Notebook knowledge-base files. The chief of staff, the keeper, the notebook UI, and human editors all follow this contract. If you're writing or editing knowledge notes — by hand, through the chief, or through the in-app editor — this is the reference.
+
+Vocabulary — *the keeper*, *the chief of staff*, *dispatch*, *the raw tier*, *PARA*, *OKF* — is defined in [glossary.md](glossary.md).
+
+---
+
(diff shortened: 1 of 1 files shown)


figgyster (approved):
This looks safe to merge. The schema matches the current notebook parser/store shape: open `type`, legacy `kind` fallback, H1-as-title, raw frontmatter preservation, and the PARA scaffold all line up.

Only nit I found: the `docs/profiles.md` link in the Concepts section points to `../profiles.md`, which misses from `docs/notebook-schema.md`; it should be `profiles.md`. That’s small doc grit, not a merge blocker.

No blockers from me.

PR #405, opened 2026-06-24
feat(workspace): add pinned workspaces that persist when empty

victorarias:
## Intent

Workspaces disappear from the sidebar the moment their last session closes. For workspaces the user wants to keep around — a project they return to, a delegation target, a context they're building up — this forces them to re-create the workspace each time. Pinned workspaces let the user mark a workspace as persistent so it survives being emptied.

## Summary

- **Migration 54** adds a `pinned INTEGER NOT NULL DEFAULT 0` column to the workspaces table (idempotent, guarded by `columnExists`).
- **Protocol**: `pinned: boolean` field on `Workspace` model; new `pin_workspace` command with `workspace_id` and `pinned` fields.
- **Store**: `AddWorkspace`/`GetWorkspace`/`ListWorkspaces` read and write the pinned column; new `SetWorkspacePinned` method.
- **Daemon retention logic**: `dissociateSessionFromWorkspace` checks `snap.Pinned` before tearing down an empty workspace; `loadWorkspacesFromStore` startup reaper skips pinned workspaces.
- **Daemon command handling**: `handlePinWorkspaceWS`/`setWorkspacePinned` persist + broadcast; wired into both WebSocket and unix socket dispatch.
- **Frontend**: `sendPinWorkspace` in `useDaemonSocket`; pin toggle button in sidebar workspace rows; pinned empty workspaces bypass the sessionless visibility toggle; CSS keeps the pin icon visible when active.
- **Existing behavior preserved**: tile-only retention, muted workspaces, workspace context — all compose cleanly with pinning.

## Test plan

- [x] `TestSetWorkspacePinned` — store round-trip
- [x] `TestAddWorkspacePinnedField` — persisted on insert
- [x] `TestListWorkspacesPinnedField` — returned in list
- [x] `TestPinnedWorkspaceSurvivesFinalSessionClose` — retention on last-session departure
- [x] `TestUnpinnedWorkspaceRemovedOnFinalSessionClose` — existing cleanup preserved
- [x] `TestPinnedEmptyWorkspaceSurvivesStartupCleanup` — retention on daemon restart
- [x] `TestSetWorkspacePinned_PersistsAndBroadcasts` — store + broadcast
- [x] All Go tests pass (`go test ./internal/...`)
- [x] All 1138 frontend tests pass (`pnpm test -- --run`)
- [ ] Manual app QA in dev profile: pin a workspace, close all sessions, verify it stays, restart daemon, verify it persists

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-06-24]
+
+### Added
+- **Pin workspaces to keep them in the sidebar even when empty.** A workspace can now be toggled pinned/unpinned from the sidebar. A pinned workspace survives its last session closing and daemon restarts — it stays visible and ready for a new agent launch. Existing tile-only and context-based workspace retention continues to work alongside pinning.
+
 ## [2026-06-23]
 
 ### Added
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -465,6 +465,7 @@ function App() {
     sendMuteRepo,
     sendMuteAuthor,
     sendMuteWorkspace,
+    sendPinWorkspace,
     sendPRVisited,
     sendRefreshPRs,
     sendRegisterWorkspace,
@@ -620,6 +621,7 @@ function App() {
         sendMuteRepo={sendMuteRepo}
         sendMuteAuthor={sendMuteAuthor}
         sendMuteWorkspace={sendMuteWorkspace}
+        sendPinWorkspace={sendPinWorkspace}
         sendPRVisited={sendPRVisited}
         sendRefreshPRs={sendRefreshPRs}
         sendRegisterWorkspace={sendRegisterWorkspace}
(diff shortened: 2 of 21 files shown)


figgyster (dismissed):
This looks safe to merge. The pinned flag is wired through the schema/store/protocol/daemon/UI path, and the two cleanup paths that used to reap empty workspaces now preserve pinned ones without disturbing tile-only retention.

I ran `git diff --check`, `go test ./internal/...`, and the frontend Vitest suite (`pnpm test -- --run`, 1138 passed). GitHub’s Frontend check was still pending when I last looked, but the same local frontend suite is green here.

No blockers from me. Tiny caveat: the manual app QA box is still unchecked, but that’s a follow-up sanity pass, not a merge gate.

figgyster (approved):
This follow-up fixes the one place the feature could still vanish: App-level workspace lists were still filtering purely on `sessions.length`, so pinned-but-empty workspaces could survive the daemon and then get ghosted by the sidebar. Classic frontend witness-protection program.

The new filter condition keeps pinned workspaces in the unmuted/muted/sidebar lists while leaving ordinary sessionless/tile-only behavior behind the existing toggle. I re-ran `git diff --check`, the targeted sidebar/sessionless tests, and the full frontend suite (`pnpm test -- --run`, 1138 passed). GitHub’s Frontend check was still pending at my last fetch, but the local equivalent is green.

No blockers from me.

PR #407, opened 2026-06-24
feat(delegate): name the agent and its workspace with --name

victorarias:
## What

`attn delegate` now takes **`--name`** (replacing `--label`). The resolved name names the delegated **agent/session** and, when the delegation creates a **new workspace**, the workspace too — so a delegated workspace shows a human name in the sidebar instead of its worktree-folder slug.

Previously `--label` named only the session + pane, and a new workspace was titled from the directory basename (e.g. `attn--feat-launch-naming`), never from the label.

## Behavior

- **Optional**, defaults to the target directory's basename when omitted (a short folder like `attn` / `attn--foo` is accepted as-is).
- **Validated** (whether explicit or defaulted):
  - length ≤ **16 runes**;
  - **workspace name** unique across all workspaces (when creating one);
  - **session name** unique within its target workspace.
  - Errors say why (too long / duplicate workspace / duplicate session / missing) and to pass `--name`.
- **Fail-fast**: an explicit `--name` is validated before any side effects; a too-long/duplicate *defaulted* name (e.g. a long worktree folder) is caught after the directory is finalized and **rolls back** a created worktree.
- No slug/uuid pattern rejection (`--` and short sluggy names are fine). The protocol wire field stays `label` to avoid a version bump; only the flag/semantics are `name`.

Names are display-only and mutable (rename later in the app); the uuid stays the real key. Scope is the `attn delegate` CLI path; the in-app launcher is unchanged.

## Tests

New daemon tests: explicit name → both session+workspace; basename default; name-too-long; duplicate workspace; duplicate session-in-workspace; long-worktree-default rollback; a direct `validateDelegationName` table test (16/17-rune boundary in ASCII + multibyte, degenerate `.`/`/`, case-insensitive workspace + session uniqueness); duplicate-workspace via the **default** path; clash with the source session's own label. CLI parse + help tests updated. The two worktree rollback tests now pass a valid `--name` so they still exercise spawn-failure rollback. Full `go test ./...` green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ### Added
 - **Pin workspaces to keep them in the sidebar even when empty.** A workspace can now be toggled pinned/unpinned from the sidebar. A pinned workspace survives its last session closing and daemon restarts — it stays visible and ready for a new agent launch. Existing tile-only and context-based workspace retention continues to work alongside pinning.
 
+### Changed
+- **Delegated agents and their workspaces now get real, readable names.** `attn delegate` takes `--name` (replacing `--label`); it names the agent and, when the delegation creates a new workspace, the workspace too — so a delegated workspace shows a human name in the sidebar instead of its worktree folder. Omit `--name` to default to the directory name. Names are capped at 16 characters and must be unique (a workspace name across all workspaces, a session name within its workspace); when a name is too long or already taken — including a too-long worktree-folder default — the command fails with a clear message telling you to pass `--name`.
+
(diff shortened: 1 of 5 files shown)


figgyster (dismissed):
This looks safe to merge. The `--name` contract is tight where it matters: explicit names fail before side effects, defaulted worktree names roll back cleanly, and the workspace/session uniqueness scopes are covered without dragging the protocol through a version bump just to rename a CLI flag.

`go test ./...` is green locally. No blockers from me.

figgyster (approved):
Still safe on the rebased head. The PR patch is the same delegation-naming contract against the newer base, and both local `go test ./...` plus the visible daemon/check jobs are green now.

No blockers.

PR #408, opened 2026-06-25
feat(dispatch): replace reports with typed outcomes

victorarias:
## Intent / Why

Delegated agents currently have to construct raw coordination JSON to report terminal work states. That interface is difficult to discover and fails late when an agent invents or misremembers the schema, leaving the chief with an incomplete or unstructured update.

This change makes the work outcome the CLI operation itself. It deliberately replaces the old reporting interface instead of retaining a compatibility path that preserves the same failure mode.

## Summary

- replace `attn dispatch report` with explicit `update`, `block`, `review`, `complete`, and `fail` commands
- keep concise summaries separate from optional details, next ownership, remaining scope, constraints, and blocker decision requests
- require Notebook handoffs to declare `--review` or `--complete`
- check unread chief messages before waiting or terminal outcomes and print a concise success acknowledgement by default
- replace the `report_dispatch` wire command with required typed `submit_dispatch_outcome` payloads and bump the protocol to 122
- update generated types, delegated-agent guidance, glossary, tests, and changelog; remove raw coordination-file support

## Test plan

- [x] `go test -count=1 ./cmd/attn ./internal/protocol ./internal/store ./internal/dispatch ./internal/agent`
- [x] `go test -count=1 ./internal/daemon`
- [x] `pnpm --dir app test` (1,143 tests)
- [x] Build and install isolated `attn-dev.app` with `make dev`
- [x] Verify packaged dev CLI exposes the new commands and rejects `dispatch report`
- [x] Codex autoreview: no accepted/actionable findings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ### Changed
 - **Delegated agents and their workspaces now get real, readable names.** `attn delegate` takes `--name` (replacing `--label`); it names the agent and, when the delegation creates a new workspace, the workspace too — so a delegated workspace shows a human name in the sidebar instead of its worktree folder. Omit `--name` to default to the directory name. Names are capped at 16 characters and must be unique (a workspace name across all workspaces, a session name within its workspace); when a name is too long or already taken — including a too-long worktree-folder default — the command fails with a clear message telling you to pass `--name`.
+- **Delegated agents now report explicit outcomes instead of assembling coordination JSON.** The new `attn dispatch update`, `block`, `review`, `complete`, and `fail` commands capture the work state directly, keep concise summaries separate from optional detailed reports, check unread chief messages before handoffs, and return a short confirmation instead of echoing the entire assignment. Notebook handoffs likewise require an explicit `--review` or `--complete` outcome.
+
+### Removed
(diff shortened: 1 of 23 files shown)


figgyster (dismissed):
This crosses a contract line, not a taste line.

`handoff_dispatch` still accepts any generic `DispatchReport` at the daemon boundary. The CLI now forces Notebook handoffs to be either `--review` or `--complete`, but a protocol client can still submit a handoff with `report_type: "progress"` / `work_state: "in_progress"` or a blocker state. The daemon writes the Notebook artifact first, persists that non-terminal outcome, and then skips the durable journal/capture path because only terminal reports are journaled.

Required before merge:
- Make the `handoff_dispatch` daemon/protocol contract enforce the same typed handoff shape as the CLI: `report_type: handoff` and `work_state` limited to `ready_for_review` or `completed`.
- Add a regression test that rejects a non-terminal/non-handoff structured report on `handoff_dispatch` before the Notebook write.

I ran the focused Go suite locally:

`go test -count=1 ./cmd/attn ./internal/protocol ./internal/store ./internal/dispatch ./internal/agent ./internal/daemon`

That passes; the gap is the missing boundary enforcement, which the current tests don't pin yet.

figgyster (approved):
This fixes the contract gap cleanly. `handoff_dispatch` now enforces the handoff-only shape before touching the Notebook, and the regression tests cover both the wrong report type and non-terminal work state cases.

No blockers from me. I verified the focused daemon/CLI/protocol paths locally; GitHub Frontend CI was still pending at my last check, so let that finish its lap around the block before merging.

PR #409, opened 2026-06-25
feat(delegate): workspace reuse guidance and --workspace + --worktree support

victorarias:
## Intent

The chief of staff was creating new workspaces for every delegated task even
when a matching pinned/domain workspace already existed. Worse, if the task
needed branch isolation, `--workspace` and `--worktree` couldn't be combined
— so workspace reuse was fundamentally incompatible with worktrees.

## Summary

**Guidance** (`internal/agent/attn_skill/references/delegation.md`):
- Placement section leads with checking `attn list` for existing workspaces
  before creating new ones
- Parallel delegation: route each agent to the matching domain workspace
- Documents `--worktree` with all three placements (current, existing, new)

**Product** — lift `--workspace + --worktree` restriction:
- CLI (`cmd/attn/main.go`): `--workspace` now only rejects `--new-workspace`
  and `--cwd` (not `--worktree`)
- Daemon (`internal/daemon/delegate.go`): `existing_workspace` placement
  accepts worktree requests
- Worktree creation uses the target workspace's directory for repo discovery
  and starting-branch inference (not the source session's)

**Tests**:
- New `TestDelegateCreatesWorktreeInExistingWorkspace` — end-to-end daemon test
- New `TestParseDelegateArgsAcceptsWorkspaceWithWorktree` — CLI parsing test
- Updated rejection tests and help-text assertions
- Skill test asserts workspace-reuse preamble

## Test plan

- [x] `go test ./internal/daemon/ -run TestDelegate` — all pass
- [x] `go test ./cmd/attn/ -run TestParseDelegate` — all pass
- [x] `go test ./internal/agent/ -run TestAttnSkill` — all pass
- [x] `make dev` — dev app built and installed
- [ ] End-to-end: chief delegates with `--workspace <id> --worktree <branch>`
  into an existing pinned workspace (manual in dev app)

### DIFF
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -602,8 +602,8 @@ placement:
   --worktree <branch>        create a worktree for the delegated session
 
 worktree options:
-  combine with --new-workspace to place the worktree in a new workspace
-  --repo <path>              main repository (defaults to the source repository)
+  combine with any placement (current, --workspace, or --new-workspace)
+  --repo <path>              main repository (defaults to the workspace repository)
   --from <ref>               branch or ref to start from
   --worktree-path <path>     override the generated sibling path
 
@@ -1608,8 +1608,8 @@ func parseDelegateArgs(args []string) (delegateCLIArgs, error) {
 	repo := strings.TrimSpace(*worktreeRepo)
 	startingFrom := strings.TrimSpace(*worktreeStart)
 	customWorktreePath := strings.TrimSpace(*worktreePath)
-	if explicitWorkspace != "" && (*newWorkspace || customCWD != "" || branch != "") {
-		return delegateCLIArgs{}, errors.New("--workspace cannot be combined with --new-workspace, --cwd, or --worktree")
+	if explicitWorkspace != "" && (*newWorkspace || customCWD != "") {
+		return delegateCLIArgs{}, errors.New("--workspace cannot be combined with --new-workspace or --cwd")
 	}
 	if customCWD != "" && branch != "" {
 		return delegateCLIArgs{}, errors.New("--cwd cannot be combined with --worktree")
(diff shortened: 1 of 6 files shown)


figgyster (dismissed):
This is safe to merge. The guidance now puts workspace reuse where the chief-of-staff flow will actually see it, and the existing hidden-workspace/unmute rules still read coherently after the reshuffle.

Local check passed: `go test ./internal/agent/ -run TestAttnSkill`. GitHub checks are green too. No blockers from me.

figgyster (dismissed):
Re-approved after the follow-up. The correction is better: it matches the actual `attn list` shape and calls out the `--workspace` / `--worktree` incompatibility instead of letting agents invent a magic combo flag.

Local checks passed: `git diff --check`, `go test ./internal/agent/ -run TestAttnSkill`, and `go test ./cmd/attn -run 'TestParseDelegateArgs|TestDelegate'`. GitHub checks are green. No blockers.

figgyster (dismissed):
Re-approved. Pulling the worktree warning back out is the right call: the docs now steer reuse first without pretending every reuse-vs-branch-isolation tradeoff has one universal answer. Much less tiny-procedure-goblin energy.

Local checks passed: `git diff --check`, `go test ./internal/agent/ -run TestAttnSkill`, and `go test ./cmd/attn -run 'TestParseDelegateArgs|TestDelegate'`. GitHub checks are green. No blockers.

figgyster (approved):
Re-approved. The product change now matches the guidance: existing workspace reuse and worktree isolation can travel together, and the test update fixes the stale skill assertion that CI caught. Nice little bit of yak-shearing cleanup; the yak did not enjoy it, but the codebase will.

Local checks passed: `gofmt -d` clean, `git diff --check`, and `go test ./...`. GitHub checks are green. No blockers.

PR #411, opened 2026-06-26
fix(cli): show pinned empty workspaces in list

victorarias:
## Intent / Why

Victor expects chief-of-staff agents to use `attn list` when choosing where to place new delegations. Pinned workspaces are meant to remain reusable even when their last session is gone, but the CLI only returned live sessions, so an empty pinned workspace had no visible ID or context. Agents could not discover the workspace and would create duplicates instead of reusing it.

## Summary

- Include workspace snapshots in daemon query responses so CLI consumers can see pinned empty workspaces, not just sessions.
- Add `Client.List` for `attn list` while keeping `Client.Query` session-only for existing callers.
- Update `attn list` to print `{ "sessions": [...], "workspaces": [...] }`, exposing workspace ID, title, directory, status, muted, pinned, rank, and layout when present.
- Regenerate protocol types and bump the protocol version to force daemon/client agreement.
- Cover both discovery and reuse: pinned empty workspaces appear in query/list output, and delegation into a pinned empty workspace succeeds.

## Test plan

- [x] `go test ./internal/client ./internal/daemon -run 'TestClient_(Query|ListIncludesWorkspaces)$|Test(QueryIncludesPinnedEmptyWorkspace|DelegateTargetsPinnedEmptyWorkspace|PinnedEmptyWorkspaceSurvivesStartupCleanup|PinnedWorkspaceSurvivesFinalSessionClose)$'`
- [x] `go test ./cmd/attn ./internal/client ./internal/daemon -run '^$'`
- [x] `git diff --check`

## Out of scope

- No live `~/.attn` workspace data was mutated for verification.
- No chief-of-staff prompt guidance was changed; the product surface now exposes the workspace directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-06-26]
+
+### Fixed
+- **`attn list` now shows pinned workspaces even when they are empty.** The command returns workspace snapshots alongside live sessions, so chief-of-staff agents can see a pinned workspace's ID, name, directory, and pinned state before delegating into it instead of creating a duplicate workspace.
+
 ## [2026-06-24]
 
 ### Added
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -172,7 +172,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '122';
+export const PROTOCOL_VERSION = '123';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 interface PRActionResult {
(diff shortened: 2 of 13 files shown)


figgyster (dismissed):
Found one protocol-version rake in the alley: the Go daemon was bumped to 123 but the frontend mirror was still 122, which is why the first Daemon CI run failed. I pushed a tiny follow-up commit aligning `PROTOCOL_VERSION` to 123.

This looks safe to merge now. The pinned-empty-workspace path is exposed through `attn list` without changing `Client.Query` for existing session-only callers, and the daemon reuse path is covered.

Local validation passed after the fix: `go test ./...`, `pnpm --dir app exec tsc --noEmit`, `pnpm --dir app run build`, `pnpm --dir app test -- --run`, and `git diff --check`. GitHub CI for the new commit was still running at last fetch.

figgyster (approved):
The follow-up tightens the right boundary: `attn list` now exposes local delegatable workspaces only, while the app still gets local + remote workspace snapshots through initial state. That keeps the new workspace-discovery contract useful without handing agents remote workspace IDs that local CLI delegation cannot actually target.

No blockers from me. Focused Go/protocol checks, frontend typecheck, and `git diff --check` passed locally; GitHub Daemon/Release Preflight were green, with Frontend still running at last fetch.

PR #410, opened 2026-06-25
docs(plan): communication gateway design (foundation PR spec)

victorarias:
## The communication gateway — design

Extracts the gateway design we settled into a standalone spec, so the **foundation PR** can be built from it. Doc: `docs/plans/2026-06-25-communication-gateway.md`.

### What it is
A foundational, attn-owned substrate for **agent↔agent communication** — one delivery-tracked queue every chief↔agent and agent↔agent message rides on. It collapses today's two disconnected paths (live `dispatch watch` event vs. the separately-triggered raw-tier journal) into one thing that knows a message was *delivered / seen*.

### Settled design
- **Delivery** `from → to`, kind (`outcome` | `mail`), payload, state `pending → acked` (+ `expired`)
- **Per-recipient addressing** + a **hardcoded literal chief id** (one chief per profile → no lookup); the chief reads all
- **Dedup by content** `(from, to, kind, payload)`; **bundle by sender**; **flat 30-day TTL** (GC only)
- **Ack = output** — no forced ack call; `consume` returns **pending only** (catch-up, not replay)
- **No `Classify`** — kind is explicit at enqueue; the snapshot→event classifier retires
- **Two consumers over one queue** — event-driven `watch` / idle pty-nudge — split by a `HasSelfMonitor` flag

### This PR's shape (per Victor)
The gateway is built as **one PR, unhooked from attn, proven by a simulation harness** (incl. the nudge path; real Claude-Monitor arming skippable). The doc lists the harness's acceptance scenarios.

**Out of scope / deferred:** attn integration (producers, the watch consumer + delegation-time arming, the idle-nudge scheduler, orient-read/re-arm, retiring the old mailbox/journal) is the *next* PR; awareness-across-compaction and the went-quiet silence floor are intentionally deferred.

### Context
Foundation rock of `docs/vision/chief-delegation-awareness.md`; motivated by the review in `docs/reviews/2026-06-24-chief-loop-review/`. Docs-only — no code, no changelog entry.

🤖 Authored by an attn-delegated agent on behalf of Victor.

PR #412, opened 2026-06-26
Tickets — attn as an agent work tracker (long-lived feature branch)

victorarias:
## Tickets — attn as an agent work tracker

This is the **long-lived integration branch** for the work-tracker pivot. The design
doc is `docs/plans/2026-06-26-work-tracker.md`.

### The idea

The chief delegating work and watching its state come back **is an issue tracker**
(Linear's restraint, not Jira's machinery). A delegated piece of work is a **durable
ticket** — title, description (the delegation prompt), status column, assignee,
activity, attachments. The session list becomes the **board**; notification is an
**event-driven** layer (mutations emit events → observers subscribe → decoupled
handlers deliver via Claude `watch` or codex nudge).

This **supersedes** the communication-gateway spec (#410) — tombstoned in this PR. The
gateway's settled mechanics (content dedup, bundle-by-ticket, ack-on-read/unread, the
two-consumer contract) re-home as the tracker's event log.

### How this branch works

- `feat/tickets` is long-lived. The **7 testable slices** in the plan land here
  **PR-by-PR** for review.
- When the feature is complete, `feat/tickets` merges to `main` as **one changeset** —
  so `main` never sees a half-migrated world where both `dispatch` and `ticket` exist.
- The `dispatch` namespace retirement (`watch / report / message / inbox / messages /
  resolve / status / handoff / list`, plus mailbox / journal / `Classify` / #397) is a
  **completion bar**, landing atomically with the `ticket` surface that replaces it.

### Rule of play

The plan doc is the source of truth for progress. **Each slice PR updates its own
status row** in the plan, so the doc always reflects where we are.

This PR itself contains only the design doc + the gateway tombstone — no code yet.

1. **#412** 👈 current
2. #413
3. #414
4. #415
5. #416
6. #417
7. #418
8. #419
9. #420
10. #421
11. #422
12. #423
13. #424


figgyster (dismissed):
This is safe to merge into the long-lived feature branch. The tombstone keeps the old gateway design discoverable, and the new plan gives the ticket model a clean migration bar instead of letting `dispatch` linger around like a ghost with a CLI parser.

No blockers. `git diff --check` is clean; the code CI jobs correctly skipped because this is docs-only, with the `Changes` check green.

figgyster (dismissed):
Still safe to merge. The added post-merge export slice is the right kind of boundary: it keeps archival out of the integration branch while making the closed-ticket sweep less spooky later.

No blockers. The follow-up is docs-only, `git diff --check` is clean, and the docs-only CI shape is still green where it matters.

figgyster (dismissed):
This caught one small but real contract break in the tracker branch bookkeeping.

Required before this stays green: update `docs/plans/2026-06-26-work-tracker.md` so slice 1 is marked ✅. The branch now includes the slice-1 store/lifecycle merge, but the plan still says `Ticket store + lifecycle` is 🟡 in progress. Since this PR explicitly makes the plan doc the progress source of truth, that stale row is misleading — tiny fix, but the kind of tiny lie that turns a tracker into Jira cosplay.

Store-side implementation looks solid from here: lifecycle timestamps, archive/reopen visibility, TTL sweep, and the regression coverage all line up. I ran `go test ./internal/store` locally and it passed.

figgyster (dismissed):
The same blocker is still present on the current head.

`docs/plans/2026-06-26-work-tracker.md` still marks slice 1 as 🟡 even though the slice-1 store/lifecycle work has already landed in this integration branch. The branch has moved forward, but the progress ledger is still lying about the first row — small, sharp, and exactly the kind of thing this plan doc said it would prevent.

Required before merge into the long-lived branch stays green:
- mark `Ticket store + lifecycle` as ✅ in the plan row now that it is merged into `feat/tickets`
- while you’re there, make sure any later slice rows already merged into this branch reflect reality too, so this doesn’t become a tracker tracking its own stale tracker. Very meta. Very avoidable.

I rechecked the current head and ran `go test ./internal/store ./internal/ticketnotify ./internal/daemon ./cmd/attn`; those passed. This is a plan/status contract blocker, not an implementation failure.

figgyster (dismissed):
This is back to green. The plan ledger now matches the branch reality for the completed ticket slices, including the slice-1 row that was blocking the previous head.

The board/detail work also has the right shape for this long-lived branch: read surfaces stay additive, the ticket rows remain the progress source, and the UI coverage hits the seams that would actually regress. I ran the focused Go suites, the ticket UI vitest set, and `tsc --noEmit`; all passed.

No blockers from me.

figgyster (dismissed):
Still green. Moving the self-monitor decision into `agent.Capabilities` is the right boundary: `ticketnotify` stays pure, and the daemon now asks the driver registry instead of smuggling agent-name lore into the notification core.

The slice-6 plan row is updated too, so the progress ledger is still honest. I ran `go test ./internal/agent ./internal/ticketnotify ./internal/daemon ./cmd/attn`; all passed.

No blockers from me.

figgyster (dismissed):
This is the right cutover point for the long-lived branch. The old dispatch namespace is gone from the live protocol/CLI/handlers, the ticket path owns delegation/reporting now, and the old SQLite migrations are left as history instead of pretending databases enjoy time travel. Sensible.

I checked the retirement seam and the remaining user-facing legacy references are either tombstoned/history docs or generic “dispatch” words from unrelated systems. Local validation passed: `go test ./...`, `tsc --noEmit`, and the focused Dashboard/socket/session close Vitest set.

No blockers from me.

figgyster (dismissed):
This is still safe. The follow-up sharpens the product contract in the right place: the always-present hook guidance now carries the dangerous bit — `attn delegate` is visible, steerable user-facing work, not a private scratchpad — while the skill keeps the deeper mechanics. That is the right split.

I also like the cleanup around the old dispatch-era notebook/source language. The live guidance no longer tells agents to ground knowledge in unproducible `dispatch:<id>` tokens, and the remaining dispatch mentions I found are either historical plan docs or explicit tombstones.

`go test ./...` and `git diff --check` are clean. No blockers from me.

figgyster (dismissed):
Still green. This follow-up fixes the terminal clip at the right layer: `GhosttyTerminal.fit()` remains the geometry authority, but genuinely small panes now get to resize the PTY instead of being mistaken for garbage measurements. That matches the bug — tiny split panes are real, even if they look suspicious in a trench coat.

The diagnostics addition is useful too: the bottom-clip sweep is low-frequency, active-pane-only, and records onset/resolution with geometry context, so it should help catch the next weird terminal shape without perturbing the paint path.

`tsc --noEmit`, the focused Vitest set for resize/runtime/diagnostics, and `git diff --check` all passed. No blockers.

figgyster (approved):
Current head is clean. I pushed the tiny gofmt follow-up that was making the Daemon CI job trip over its own shoelaces; no behavior changed, just the aligned comment whitespace in the hooks test.

The terminal resize fix still looks right, and the focused validation passes: `go test ./internal/hooks`, `tsc --noEmit`, the resize/runtime/diagnostics Vitest set, and a full `gofmt -l` check are clean.

No blockers from me.

PR #413, opened 2026-06-26
feat(tickets): slice 1 — ticket store + lifecycle

victorarias:
## Slice 1 — Ticket store + lifecycle

First implementation slice of the work tracker (plan: `docs/plans/2026-06-26-work-tracker.md`). **Base: `feat/tickets`** — lands on the integration branch, not `main`.

Pure store layer. No protocol/IPC/CLI/frontend surface yet (those are slices 3–5), so nothing user-visible changes — this is the foundation the rest builds on.

### What's here

- **Migration 55** — fresh `tickets`, `ticket_activity`, `ticket_attachments` tables. (Verified real DBs top out at 52/54, so 55 is clear of the burned-version trap.)
- **`TicketStatus`** enum — `todo / working / blocked / in_review / done / failed / crashed`, with `IsValid` / `IsTerminal`. Transitions are **permissive** — the board informs, it never gates.
- **Slug ids** — `ValidateTicketID` + a uniqueness check whose collision error tells the agent to *pick a new name or append a number* (e.g. `store-migration-2`), matching the locked design.
- **CRUD + activity** — `CreateTicket`, `GetTicket` (loads the full activity thread + attachments), `ListTickets` (board filter, archived hidden by default). Status changes and comments are recorded as one unified **activity thread** (`status_change` + `comment` — no separate "report"). Plus description / assignee / session / attachment mutators.
- **Lifecycle** — `closed_at` is stamped on entering a terminal status and cleared on reopen (so reopened work is durable again); it drives `SweepExpiredTickets(now, ttl)`, which hard-deletes terminal tickets past the 30-day TTL and **cascades** to their activity + attachments. Open tickets (the backlog) never age out. `ArchiveTicket` clears a *closed* ticket from the board and refuses open ones.

### Design notes

- **Fresh table, not an evolve.** The plan originally said "evolves `chief_of_staff_dispatches`"; I went fresh-table instead (plan prose updated to match). The schema is materially different, and slice 7 retires the whole `dispatch` namespace anyway — a fresh table lets the old one die cleanly with its namespace rather than slice 1 half-converting it and slice 7 half-dismantling it. The old table is left fully functional through slices 1–6.
- **Store-local row types** (`store.Ticket` etc.), following the `workflow_runs` convention — the wire shape is owned by a later slice.
- **Injected time.** Mutators take an explicit `now time.Time` (idiomatic here — the codebase already passes timestamps in) so every lifecycle test, including the TTL sweep, is deterministic. Timestamps are stored as fixed-width RFC3339 UTC, so the sweep's `closed_at` comparison is a clean lexical compare.

### Tests

`internal/store/tickets_test.go`, all store-level: CRUD round-trip + migration smoke, validation, id collision (incl. the guidance message), permissive transitions + `closed_at` lifecycle, comments/edits bumping `updated_at`, attachments, list + archive board semantics, the **TTL sweep with injected time** (cascade verified, open + recent tickets untouched), and persistence across a DB reopen.

Per the plan's Rule of play, this PR flips slice 1 to 🟡.

1. #412
2. **#413** 👈 current
3. #414
4. #415
5. #416
6. #417
7. #418
8. #419
9. #420
10. #421
11. #422
12. #423
13. #424

### DIFF
--- a/docs/plans/2026-06-26-work-tracker.md
+++ b/docs/plans/2026-06-26-work-tracker.md
@@ -228,7 +228,7 @@ each is a meaningful, verifiable chunk.
 
 | # | slice | status |
 |---|---|---|
-| 1 | Ticket store + lifecycle | ⬜ |
+| 1 | Ticket store + lifecycle | 🟡 |
 | 2 | Event-driven notification core | ⬜ |
 | 3 | Delegation ⇄ tickets (live wiring) | ⬜ |
 | 4 | Ticket view + resume + attachments | ⬜ |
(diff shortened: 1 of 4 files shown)


figgyster (dismissed):
This has one lifecycle bug that crosses the contract line.

`ArchiveTicket` hides a closed ticket by setting `archived_at`, and `ListTickets` hides archived rows by default. But `SetTicketStatus` lets any ticket move back to `todo` / `working` / etc. and only clears `closed_at`; it leaves `archived_at` set. So a previously archived ticket can be reopened into an open status while staying invisible on the default board and immune to the TTL sweep. That violates the plan's core lifetime rule: open tickets are durable and visible until acted on.

I verified it with a focused local probe: create closed ticket → archive → `SetTicketStatus(..., working)` → `ListTickets(default)` returns nothing. Existing `go test ./internal/store` still passes, so this edge isn't covered yet.

Required before merge:
- either clear `archived_at` when a ticket is reopened to a non-terminal status, or reject status changes on archived tickets behind an explicit unarchive path
- add the regression test for archived-then-reopened board visibility

Everything else in this slice looks coherent; the store shape and injected-time lifecycle are the right foundation. This one hidden-zombie-ticket path needs killing before the board starts lying with a straight face.

figgyster (approved):
Fixed cleanly. The reopen path now clears `archived_at` with `closed_at`, the returned row stays in sync, and the regression test pins the exact archived → reopened → visible-board lifecycle that was broken.

`git diff --check` and `go test ./internal/store` pass locally. No CI checks were reported for this branch, but the principal store seam is covered here.

No blockers from me. The zombie ticket has been sent back to whatever alley it crawled out of.

PR #414, opened 2026-06-26
feat(tickets): slice 2 — event-driven notification core

victorarias:
## Slice 2 — Event-driven notification core

Second slice of the work tracker (plan: `docs/plans/2026-06-26-work-tracker.md`). **Stacked on #413** (base `tickets/slice-1-store`) — review #413 first; this diff is only the slice-2 delta on top of it.

The decoupled notification layer over an append-only ticket **event log**, re-homing the dispatch gateway's settled mechanics. Built against the store and **proven by a simulation harness** — no live sessions, no real Monitors. No protocol/wire surface yet (that's slice 3), so still nothing user-visible and no `ProtocolVersion` bump.

### Store substrate (migration 56)

- **`ticket_events`** — append-only log. A global `AUTOINCREMENT seq` is the cursor space. Append is **idempotent**: an event identical to the ticket's previous event (a retry) is deduped.
- **`ticket_event_cursors`** — per-observer cursor; "unread" = events past it.
- Slice-1 mutators now **emit their domain event transactionally** — `created / status_changed / commented / assigned / description_edited / attachment_added` — threading an `author` so authorship is correct. They emit, but know nothing about who listens. The TTL sweep cascades events too.
- Events are a **superset** of the display activity thread (activity keeps its two human kinds; the log carries all six the chief observes) — they're distinct streams, so dedup on the event log never alters the displayed history.

### `internal/ticketnotify` (the decoupled handlers)

- **`ObserverChief`** — the well-known chief observer (re-homes the gateway's hardcoded chief id as a literal).
- **`Consume`** — unread-since-cursor, **bundled by ticket**, advances the cursor; excludes self-authored events. Chief observes all tickets; an agent observes only tickets currently assigned to it. *(Slice 3 narrows the chief to tickets it delegated/owns once delegation exists.)*
- **`HasSelfMonitor`** (claude true / codex false) picks the delivery path.
- **`Notify`** routes: self-monitor → **watch** (its own watch consumes, true push); idle non-self-monitor → **nudge**; busy → **deferred**. The `Nudger` interface takes only an observer id — there is **structurally no channel for event content**, which is the never-stream-content-into-a-PTY rule enforced by shape.

### Re-homing map

| gateway mechanic | here |
|---|---|
| ack = output / consume pending | cursor: events since the observer's cursor |
| bundle by sender | `Consume` groups pending events by ticket |
| hardcoded chief id | `ObserverChief`, a literal |
| two consumers (watch / nudge) | the two `Delivery` paths |
| dedup by content | idempotent append (vs the ticket's previous event) |
| never stream content to a PTY | `Nudger.Nudge(observerID)` — no content param |

### Tests

`internal/ticketnotify/harness_test.go` — a simulation harness driving the real store as producer + a recording nudger: **emit → consume → cursor** (re-consume is empty; a fresh event becomes unread), **dedup** (identical comment twice = one event; re-append returns `appended=false`), **bundle-by-ticket** (interleaved events group per ticket, first-seen order), the **nudge path** (busy→deferred, idle→nudge, claude→watch, consume-after-nudge), and **agent scope + self-authored exclusion**. Plus the slice-1 store tests, extended for event emission + the events cascade in the TTL sweep.

Per the plan's Rule of play, this flips slice 2 to 🟡.

1. #412
2. #413
3. **#414** 👈 current
4. #415
5. #416
6. #417
7. #418
8. #419
9. #420
10. #421
11. #422
12. #423
13. #424

### DIFF
--- a/docs/plans/2026-06-26-work-tracker.md
+++ b/docs/plans/2026-06-26-work-tracker.md
@@ -229,7 +229,7 @@ each is a meaningful, verifiable chunk.
 | # | slice | status |
 |---|---|---|
 | 1 | Ticket store + lifecycle | 🟡 |
-| 2 | Event-driven notification core | ⬜ |
+| 2 | Event-driven notification core | 🟡 |
 | 3 | Delegation ⇄ tickets (live wiring) | ⬜ |
 | 4 | Ticket view + resume + attachments | ⬜ |
 | 5 | Board view | ⬜ |
--- a/docs/plans/2026-06-26-work-tracker.md
+++ b/docs/plans/2026-06-26-work-tracker.md
@@ -147,11 +147,20 @@ knows *nothing* about who's listening:
 - `TicketCreated`, `TicketStatusChanged` (from→to), `TicketCommented`,
   `TicketAssigned`, `TicketDescriptionEdited`, `TicketAttachmentAdded`.
 
-**2. Observers subscribe** to events for tickets they care about:
+**2. Observers subscribe** to events for tickets they care about. Identity is uniform
+— **you, the chief, and each agent are all just identities** (the chief is one of the
+agents). An identity is *involved* with a ticket when it is assigned to it or has
+authored an event on it:
 
-- the **chief** observes events on the tickets it delegated / owns;
+- the **chief** observes events on the tickets it delegated / owns (it authored their
+  `created` event, so it needs no special "sees everything" scope);
 - an **assignee agent** observes events on its ticket that it *didn't author* (the
   chief's comments, status changes, re-briefs).
+
+Each identity keeps its **own cursor per ticket**, not one global cursor. So a ticket
+newly assigned to an agent — or reassigned to it — arrives with its **full history**
+(the brief and prior steers), and an agent's progress on one ticket never advances its
+bookmark on another it hasn't looked at.
 
 **3. Decoupled handlers** turn an event into a notification — adding a channel is adding
 a handler, never touching the mutators:
@@ -169,7 +178,7 @@ The gateway's settled mechanics map onto the event log:
 | gateway mechanic | here |
 |---|---|
 | dedup by content | idempotent events (no double-emitted transition/comment) |
-| ack = output; consume returns pending | **unread** — events since the observer's cursor |
+| ack = output; consume returns pending | **unread** — events past the identity's per-(identity, ticket) cursor |
 | bundle by sender | a consume groups pending events **by ticket** |
 | hardcoded chief id | the chief is a well-known observer (literal constant) |
 | two consumers (watch / nudge) | the two decoupled handlers above |
--- a/internal/store/sqlite.go
+++ b/internal/store/sqlite.go
@@ -513,9 +513,12 @@ CREATE INDEX IF NOT EXISTS idx_ticket_attachments_ticket
 CREATE INDEX IF NOT EXISTS idx_ticket_events_ticket
     ON ticket_events(ticket_id, seq);
 CREATE TABLE IF NOT EXISTS ticket_event_cursors (
-    observer_id TEXT PRIMARY KEY,
-    cursor      INTEGER NOT NULL DEFAULT 0,
-    updated_at  TEXT NOT NULL DEFAULT ''
+    identity   TEXT NOT NULL,
+    ticket_id  TEXT NOT NULL,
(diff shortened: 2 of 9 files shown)


figgyster (dismissed):
This is a real cursor-contract bug, not taste: observer cursors can be rewound by overlapping consumes.

`Consume` reads pending events and then writes the cursor through a separate unconditional `SetObserverCursor` call. If two consumers for the same observer overlap, the later one can advance to seq 3 and the earlier one can then write stale seq 2 back over it. That resurrects seq 3 as unread and can duplicate delivery. I confirmed the store currently allows `SetObserverCursor("agent", 3)` followed by `SetObserverCursor("agent", 2)` to leave the cursor at 2.

Required before merge:
- make cursor updates monotonic/compare-safe, e.g. `cursor = MAX(cursor, excluded.cursor)` or a guarded `WHERE excluded.cursor > cursor`, and add a regression test for stale/overlapping cursor writes.

The rest of the slice shape looks solid, and `make test` passes locally. This one needs fixing because the event log’s durable cursor semantics are the core of the feature.

figgyster (dismissed):
The follow-up added useful coverage, but the blocker I raised is still present on this head: `SetObserverCursor` still overwrites the stored cursor unconditionally.

The new comment in `Consume` documents the race as a single-consumer assumption, but the durable store API still allows stale writers to move a cursor backwards. That means the same failure mode remains: one consume can advance an observer to seq 3, then an older/stale consume can write seq 2 afterward, making seq 3 unread again. A cursor named “consumed through here” needs to be monotonic; documenting that it can rewind doesn’t make the event log safe — it just puts a tiny warning sign on the trapdoor.

Required before merge:
- make `SetObserverCursor` monotonic / compare-safe, so writes can only advance the cursor (`MAX(cursor, excluded.cursor)` or equivalent guarded update), and
- add the regression test for stale cursor writes.

Focused checks still pass: `go test ./internal/store ./internal/ticketnotify`.

figgyster (dismissed):
This is still blocked on the same cursor-safety contract.

The per-(identity, ticket) cursor model is the right direction for the assignment/reassignment semantics, and the new harness coverage is genuinely useful. But `SetTicketCursor` still does an unconditional upsert:

```sql
ON CONFLICT(identity, ticket_id) DO UPDATE SET cursor = excluded.cursor
```

So the old failure mode survives per ticket: a stale/overlapping consume can advance an identity’s cursor for `ticket-1` to seq 3, then a slower consume can write seq 2 back over it, making seq 3 unread again. I verified that exact store behavior with a throwaway probe on this head: `SetTicketCursor("agent", "ticket-1", 3)` followed by `SetTicketCursor(..., 2)` leaves the cursor at 2.

Required before merge:
- make `SetTicketCursor` monotonic / compare-safe (`MAX(cursor, excluded.cursor)` or equivalent guarded update), and
- add the regression test for stale cursor writes on the per-ticket cursor API.

Local checks pass (`go test ./internal/store ./internal/ticketnotify`, `make test`), so this is narrow. The trapdoor just moved from global cursor to per-ticket cursor; it still needs a latch.

figgyster (dismissed):
Still blocked on the current head.

Two concrete blockers now:

- `internal/store` does not compile. The PR changed `CreateTicket` to require an `author`, but `TestArchivedTicketReopenedBecomesVisible` still calls `s.CreateTicket(..., ticketBase)` at `internal/store/tickets_test.go:336`. `go test ./internal/store ./internal/ticketnotify` fails before it can run the store tests.
- The cursor-safety blocker is still present in the code: `SetTicketCursor` still uses an unconditional conflict update (`cursor = excluded.cursor`). So the per-ticket cursor can still be rewound by a stale writer. The base re-stack changed commit SHAs and pulled in nearby work, but it did not latch the trapdoor.

Required before merge:
- update the stale `CreateTicket` test call so the store package builds again,
- make `SetTicketCursor` monotonic / compare-safe, and
- add the stale cursor-write regression test on the per-ticket API.

I only got as far as the focused package check because the store build fails. That’s enough red; no need to make the corpse run laps.

figgyster (approved):
This is fixed now. The cursor write is monotonic at the store boundary, the regression covers the stale-write case, and the restacked slice builds again.

`go test ./internal/store ./internal/ticketnotify` and `make test` both pass locally. No blockers from me.

PR #415, opened 2026-06-26
feat(tickets): slice 3a — delegate creates and binds a ticket

victorarias:
## Slice 3a — Delegate creates + binds a ticket

First **additive** sub-slice of slice 3 (delegation ⇄ tickets live wiring). Base: `tickets/slice-2-events` (stacked). See `docs/plans/2026-06-26-work-tracker-slice3.md` for the sub-split and decisions.

Slice 3 is large and ends by **removing live production paths** (the dispatch mailbox + report), so it lands as additive-first sub-PRs with the destructive removal **gated last**. This is the foundation: the binding.

### What's here
- Delegating from the chief now also **creates a ticket and binds it to the delegated session**.
- The brief → ticket **description** (the delegation prompt).
- **Binding = `ticket.assignee = the delegated session id`** — the agent's observer identity in `ticketnotify`, so assignee == session *is* the binding (no new sessions column or protocol field). `ActiveTicketForSession` is the session→ticket lookup that the report path (3b) and crash detection (3c) will use.
- Ticket starts in **Working** (agent begins immediately); the **created event is authored by the chief**, so the agent observes it as its "assigned to you" signal and the chief never sees its own action.
- Id is a **slug from the label** with a numeric suffix on collision; delegation **rolls back** if the ticket can't be created.

### Not here (later sub-PRs)
- 3b agent self-report → ticket status · 3c crash → Crashed · 3d chief-edit → agent observes · 3e (gated) remove dispatch mailbox + report.

### Tests
- `TestDelegateCreatesAndBindsTicket` — end-to-end: delegate → ticket bound to session, brief as description, Working, created event authored by chief.
- `TestTicketSlug` + `TestCreateDelegatedTicketCollisionSuffix` — slug derivation + collision retry.
- Full `go test ./...` green. CI is main-only, so verified locally.

### [Stack](https://github.com/kitlangton/stack)

1. #412
2. #413
3. #414
4. **#415** 👈 current
5. #416
6. #417
7. #418
8. #419
9. #420
10. #421
11. #422
12. #423
13. #424

PR #416, opened 2026-06-26
feat(tickets): slice 3b — agent forward channel (ticket status)

victorarias:
Part of the **Tickets** work tracker (umbrella #412, plan `docs/plans/2026-06-26-work-tracker-slice3.md`). Stacked on #415 (slice 3a).

## What

Adds `attn ticket status <work-state>` — the agent's **forward channel** onto its own bound ticket. This is the awareness spine working in the agent→board direction: the agent self-reports its work state, attn projects that onto the board column. No content is ever streamed into a PTY.

- **`set_ticket_status` protocol command** (version bump 123→124). The agent never names a ticket id; the daemon resolves the bound ticket from the calling session (the assignee) via `ActiveTicketForSession`, so a session can only ever move *its own* active ticket.
- **work-state → column map** (`ticketStatusFromWorkState`): `in_progress→working`, `needs_input→blocked`, `ready_for_review→in_review`, `completed→done`, `failed→failed`. `crashed`/`todo` are intentionally unreachable here (crashed is attn-authored; todo is pre-assignment). The map lives in one place so slices 3c/3d/3e can reuse it.
- Status changes are recorded with the **agent's own session as author** on the activity thread + as a `status_changed` ticket event.

## Scope (additive only)

The dispatch report verbs (`dispatch update/block/review/complete/fail`) still exist and still work. The agent prompt is rewired to use tickets in **3d**, and the dispatch report/mailbox paths are removed in the **gated** **3e** (Victor's sign-off). 3b just adds the mechanism + proves it.

## Tests

- `TestSetTicketStatusMovesBoundTicket` — real delegation → `ready_for_review` → ticket in In Review, result echoed, change authored by the agent session with the comment.
- `TestSetTicketStatusCompletedClosesTicket` — `completed` closes the ticket (terminal + `closed_at`); a follow-up report is rejected (no active ticket).
- `TestSetTicketStatusErrors` — empty session / unknown work state / no bound ticket.
- `TestTicketStatusFromWorkState` — locks all five mappings + rejects unknown.

## Verification

`go build ./...`, full `go test ./internal/daemon` (incl. new tests), `internal/protocol|client|store|ticketnotify` suites, and frontend `tsc --noEmit` all green. CI is main-only, so verified locally + figgyster per PR.

1. #412
2. #413
3. #414
4. #415
5. **#416** 👈 current
6. #417
7. #418
8. #419
9. #420
10. #421
11. #422
12. #423
13. #424

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -172,7 +172,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '123';
+export const PROTOCOL_VERSION = '124';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 interface PRActionResult {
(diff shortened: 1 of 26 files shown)


figgyster (dismissed):
This is the right shape for the ticket forward channel, but the new CLI contract is currently lying to agents.

`attn ticket status ready_for_review --comment hi` is the documented form in the help text, but Go's `flag` parser stops parsing flags at the first positional argument. That means the exact advertised syntax fails before it ever reaches the daemon:

```text
ticket status: expected exactly one work state argument
```

Required before merge:
- either parse `ticket status <work-state> [flags]` explicitly so flags after the work state work, or change the contract/help/tests to require flags before the state and make sure the prompt in the next slice uses that shape.

I ran `go test ./internal/daemon ./internal/protocol ./internal/client ./internal/store ./cmd/attn` and `go build ./...`; both pass. The blocker is the new command's public/documented invocation path, not the daemon mapping itself.

figgyster (approved):
This fixes the blocker cleanly. The documented `ticket status <state> --comment ...` shape now parses before reaching the daemon, and the parser tests cover flags before, after, and around the work state.

I also re-ran the focused/full local checks: `go test ./cmd/attn -run TestParseTicketStatusArgs -count=1`, `go test ./internal/daemon ./internal/protocol ./internal/client ./internal/store ./cmd/attn`, `go build ./...`, and `git diff --check`. No blockers from me.

PR #417, opened 2026-06-26
feat(tickets): slice 3c — crash seam authors Crashed on bound ticket

victorarias:
Part of the **Tickets** work tracker (umbrella #412, plan `docs/plans/2026-06-26-work-tracker-slice3.md`). Stacked on #416 (slice 3b).

## What

Closes the loop the agent forward channel can't: when a delegated agent's **process dies mid-flight without reporting**, attn moves its bound ticket to **Crashed** itself. Crashed is the one transition attn authors — a dead worker can't say done/failed, so the board would otherwise show a stale Working/Blocked forever.

- `captureTicketCrashState(sessionID, state)` runs **alongside `captureDispatchCloseState`** at both session-end paths — `handlePTYExit` (orderly exit) and `dropSessionRecord` (reaped on restart / liveness sweep / worktree teardown) — fed the same **pre-clobber** state, before `handlePTYExit`'s idle-clobber erases it.
- `isMidFlightCrashState` mirrors the dispatch close-state classification: `launching` / `working` / `pending_approval` are crashes; `idle` / `waiting_input` / `unknown` are clean rests that leave the ticket wherever the agent last reported it.
- Authored by a new `store.TicketAuthorAttn` (`"attn"`) sentinel — an authoring identity only, never an observer, so it accrues no cursors and lands in nobody's unread feed.

## Idempotent + safe

- No-op on a clean rest, and no-op when there's no bound **non-terminal** ticket.
- Once Crashed (terminal), a later teardown read finds no active ticket — so the two call sites can both fire harmlessly.
- A real terminal report (`completed`/`failed`) makes the ticket terminal first, so the report always wins over a later crash capture.

## Tests

- `TestDropSessionRecordCrashesBoundTicket` — real delegation → mid-flight teardown → ticket Crashed, closed, authored by `attn` (proves the wiring through `dropSessionRecord`).
- `TestCaptureTicketCrashStateNoopOnCleanRest` — `waiting_input` leaves the ticket Working.
- `TestCaptureTicketCrashStateNoopAfterTerminalReport` — a reported Done ticket is not overwritten.
- `TestIsMidFlightCrashState` — locks the crash/clean split.

## Verification

`go build ./...`, `go test ./internal/daemon ./internal/store` all green. CI is main-only, so verified locally + figgyster per PR.

1. #412
2. #413
3. #414
4. #415
5. #416
6. **#417** 👈 current
7. #418
8. #419
9. #420
10. #421
11. #422
12. #423
13. #424

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -172,7 +172,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '123';
+export const PROTOCOL_VERSION = '124';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 interface PRActionResult {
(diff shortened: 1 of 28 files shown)


figgyster (approved):
This looks safe to merge. The crash seam is in the right place: it captures the pre-clobber session state on both orderly exit and teardown, only moves bound non-terminal tickets, and leaves explicit terminal reports in charge.

I ran `go test ./internal/daemon ./internal/store` and `go build ./...` locally; both passed. Codex only reported usage-limit noise, no substantive finding. No blockers from me.

PR #418, opened 2026-06-26
feat(tickets): slice 3d-1 — agent read path (ticket inbox)

victorarias:
Part of the **Tickets** work tracker (umbrella #412, plan `docs/plans/2026-06-26-work-tracker-slice3.md`). Stacked on #417 (slice 3c). First of the three 3d sub-PRs.

## What

`attn ticket inbox` — the **consume path** an observer runs to read (and mark read) its unread ticket activity. This is the board→agent direction of the awareness spine, and the target a nudge points at.

- **Protocol ticket-event types** — `TicketEvent` / `TicketEventBundle` / `TicketInboxResult` + the `TicketEventKind` enum, and the `ticket_inbox` command (protocol **v125**). (`seq` stays server-side — it's the cursor key, not display data.)
- **`ticketObserverForSession`** resolves the observer identity **uniformly**: every session — the chief included — observes as its own session id, the *same string it authors events with*, so involvement (`assignee` or `author`) and the self-author exclusion line up. `ticketnotify.ObserverChief` stays a slice-2 harness placeholder; real wiring keys off the live session.
- **`handleTicketInbox`** runs `ticketnotify.Consume`: returns unread events bundled by ticket and advances the identity's per-ticket cursors.

## Why inbox-first

3d ("observers on live sessions; events drive Notify; real-agent e2e") is too big for one PR, so it's split 3d-1..3d-3 (see the plan's sub-PR table). The read path lands first because the nudge in **3d-2** needs a target to point at; the real `Nudger` + the producers that trigger `Notify` come in 3d-2, and the real-Claude e2e + prompt rewire in 3d-3.

## Tests

- `TestTicketInboxConsumesByIdentity` — real delegation, then: the agent reads the chief-authored `created` event (its assignment) but never its own status report; the chief reads the agent's `ready_for_review` (from→to + comment) but never its own `created`; a second read is empty (cursor advanced).
- `TestTicketInboxRequiresSession` — empty session rejected.

## Verification

`go build ./...`, full `go test ./internal/daemon`, `internal/protocol|client|store|ticketnotify`, `cmd/attn`, and frontend `tsc --noEmit` all green. CI is main-only, so verified locally + figgyster per PR.

1. #412
2. #413
3. #414
4. #415
5. #416
6. #417
7. **#418** 👈 current
8. #419
9. #420
10. #421
11. #422
12. #423
13. #424

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -172,7 +172,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '123';
+export const PROTOCOL_VERSION = '124';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 interface PRActionResult {
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -172,7 +172,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '124';
+export const PROTOCOL_VERSION = '125';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 interface PRActionResult {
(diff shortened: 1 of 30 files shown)


figgyster (approved):
This looks safe to merge. The ticket inbox read path keeps the identity/cursor story tight: live sessions consume as themselves, self-authored events stay out, and the protocol bump/generated clients line up with the daemon handler.

I ran the focused daemon/client/protocol/ticketnotify/cmd tests, `go build ./...`, `git diff --check`, and frontend `tsc --noEmit`; all green. GitHub has no checks reported on this branch, so local proof is the signal here.

No blockers from me.

PR #419, opened 2026-06-26
feat(tickets): slice 3d-2 — notify producers + real nudger

victorarias:
Part of the **Tickets** work tracker (umbrella #412, plan `docs/plans/2026-06-26-work-tracker-slice3.md`). Stacked on #418 (slice 3d-1). Second of the three 3d sub-PRs.

## What

Wires `ticketnotify.Notify` to **live sessions** — the trigger side of the awareness spine. After a ticket event lands, attn reaches every observer involved with the ticket and delivers per its capability.

- **`store.TicketParticipants(ticketID)`** — the inverse of involvement: the assignee plus everyone who authored an event on the ticket (deduped, sorted, empties excluded).
- **`notifyTicketObservers(ticketID)`** runs after the **status** (`handleSetTicketStatus`) and **crash** (`captureTicketCrashState`) producers. For each live participant it runs `ticketnotify.Notify`, which decides: **Watch** for a self-monitoring Claude (nothing injected — its own watch drains it), a fixed **doorbell** for an idle codex, **Deferred** for a busy one. Each observer sees only what it did not author, so a producer blanket-notifies without self-nudging.
- **`ticketNudger`** adapts the daemon's `typeDoorbell` primitive — a bounded "run `attn ticket inbox`" trigger, **never event content** (the doorbell rule).
- **Deferred flush** — a busy agent's nudge fires the moment it goes idle, wired into the state-transition path (off the read-loop goroutine, since the doorbell write briefly blocks).

## Deliberately out of scope

- **Created-event notify is skipped** — the brief already reaches the agent via its spawn prompt, so notifying it about its own assignment would be redundant.
- The **chief-comment producer** (the verb that makes the agent's inbox commonly non-empty) lands with slice 4's chief surface.

## Tests

- `TestNotifyNudgesIdleCodexObserver` — idle codex gets the doorbell for its unread event; the chief (the author) does not.
- `TestNotifyDoesNotNudgeClaudeObserver` — a self-monitoring Claude is never typed into.
- `TestNotifyDefersBusyCodexThenFlushesOnIdle` — busy → no nudge; idle → flushed.
- `TestTicketParticipants` — assignee + distinct authors, sorted; unknown ticket empty.

## Verification

`go build ./...`, full `go test ./internal/daemon ./internal/store ./internal/ticketnotify` all green. CI is main-only, so verified locally + figgyster per PR.

1. #412
2. #413
3. #414
4. #415
5. #416
6. #417
7. #418
8. **#419** 👈 current
9. #420
10. #421
11. #422
12. #423
13. #424


figgyster (approved):
This looks safe to merge. The ticket observer fan-out keeps the doorbell rule intact — fixed trigger only, unread content stays behind `attn ticket inbox` — and the deferred busy→idle path is wired at the state transition where it belongs.

I ran `go test ./...`, `go build ./...`, and `git diff --check`; all clean. No blockers from me.

PR #421, opened 2026-06-27
feat(tickets): slice 4b — ticket detail panel (read-only)

victorarias:
Slice **4b** of the work tracker. The first ticket **UI** surface: a read-only detail panel.

## What

- **Store** — `tickets[]` + `setTickets` (the non-archived board feed), fed from `initial_state.tickets` and the `tickets_updated` broadcast via a new `onTicketsUpdate` callback.
- **`fetchTicket(id)`** — request/result against the `ticket_result` event (mirrors `sendNotebookRead`), returns the full record.
- **`TicketDetailPanel`** — a dock panel showing title / status / description / **history thread** (status changes with from→to + notes, comments) / **attachments**. It:
  - fetches the full record on open (the board feed carries only bare rows),
  - **re-fetches live** when the board row's `updated_at` moves (a status change, a new comment) — no polling,
  - shows the **bare row** as an instant header while the full record loads,
  - guards against a **stale-id** late fetch when the open ticket changes.
- **Entry point** — a "View ticket" affordance on each delegated-work row whose session has a bound ticket (`assignee == session id`). This is the "from a session" entry; the board (slice 5) and ⌘K (slice 4c) come later.

## Tests

- `TicketDetailPanel`: fetch-once-on-open + renders history/attachments; no fetch while closed; live re-fetch on `updated_at` change (with no-loop assertion); bare-row header while loading; error surfaced; close button.
- `Dashboard`: opens the bound ticket's detail; no affordance when the session has no bound ticket.

`tsc --noEmit` green; full `pnpm test` green (1146 passed).

## Stacking

Stacked on `tickets/slice-4a-read-api` (#420). CI is main-only → verified green locally + figgyster-reviewed.

### [Stack](https://github.com/kitlangton/stack)

1. #412
2. #413
3. #414
4. #415
5. #416
6. #417
7. #418
8. #419
9. #420
10. **#421** 👈 current
11. #422
12. #423
13. #424

PR #420, opened 2026-06-27
feat(tickets): slice 4a — board read API (get_ticket + tickets_updated)

victorarias:
Slice **4a** of the work tracker (`docs/plans/2026-06-26-work-tracker.md`; sub-plan `docs/plans/2026-06-26-work-tracker-slice4.md`). The frontend **data source** for the ticket detail view — backend only, no UI yet.

## What

Tickets reach the app the way dispatches do:

- **Push** — a `tickets` snapshot on `initial_state`, plus a `tickets_updated` broadcast on every mutation. Both carry **bare, non-archived rows** (no activity/attachments) so a busy board stays cheap to broadcast.
- **Pull** — `get_ticket` (request/result, correlated by `request_id`) returns one ticket's **full record**: title / description / status / activity thread / attachments. A by-id read needs no caller identity (the board feed already bounds what the app sees), so there is no `source_session_id`. An unknown id is a *failed result*, not a panic — the TTL sweep can remove a ticket between a board push and the click.

Broadcast is wired into the existing producers (delegate-create, agent status report, crash capture) and the initial-state snapshot.

No CLI `get_ticket`: agents read via `ticket_inbox`; only the app needs full detail.

## Protocol (v126)

`Ticket` / `TicketActivity` / `TicketAttachment` + `TicketActivityKind`; `GetTicketMessage` + `ticket_result` event; `TicketsUpdatedMessage`; `tickets` on `InitialStateMessage`.

## Tests

- `get_ticket` returns the full record (status-change with from/to + note, comment, attachment with note).
- unknown id → failed result (error set, no ticket).
- `ticketsForBroadcast` is non-archived **and bare** (activity/attachments empty).

`go test ./internal/daemon ./internal/store ./internal/protocol ./cmd/attn` green; `tsc --noEmit` green; frontend `PROTOCOL_VERSION` bumped to match (version test green).

## Stacking

Stacked on `tickets/slice-3d-notify` (#419). CI is main-only, so this is verified green locally + figgyster-reviewed.

### [Stack](https://github.com/kitlangton/stack)

1. #412
2. #413
3. #414
4. #415
5. #416
6. #417
7. #418
8. #419
9. **#420** 👈 current
10. #421
11. #422
12. #423
13. #424

PR #422, opened 2026-06-27
feat(tickets): slice 4c — chief ticket actions (status, comment, re-brief)

victorarias:
Slice 4c of the work tracker (stacked on #421 → #420 → main).

## What

The reverse channel the board was missing: by-id ticket mutations from the app, authored as the human identity (`store.TicketAuthorYou` = "you"). Three WS commands:

- `ticket_change_status` — move a ticket to a new column (optional comment)
- `ticket_add_comment` — drop a freeform comment on the thread
- `ticket_edit_description` — re-brief the ticket

Each one lands an event, **notifies the assigned agent** (which did not author the change — the board→agent steer), and **broadcasts `tickets_updated`** so the open detail view re-fetches off the bumped `updated_at`. All three reply with a shared `ticket_action_result` correlated by `request_id` (the `sendPRAction` request/result pattern), so the UI shows success/error without inventing optimistic state.

Authoring as "you" is the key identity choice: "you" has no session, so it is never nudged (the app already sees the change via the live broadcast); the *agent* that didn't author gets the nudge.

## Surface

`TicketDetailPanel` gains:
- a **status select** (selectable destinations exclude attn-authored `crashed`; the ticket's own status is always shown)
- an **add-comment composer** (clears the draft on success)
- an **inline description editor** (Edit → textarea → Save/Cancel)

The panel stays **read-only** when no action handlers are wired (props are optional).

## Protocol

Bumped to **v127**. New commands + `ticket_action_result` event in `main.tsp`, `constants.go`, regenerated types.

## Tests

- Daemon: `internal/daemon/ticket_actions_test.go` — status/comment/edit move the store, author as "you", notify the idle agent, and fail cleanly on missing/unknown ticket_id (5 tests, green).
- Frontend: `TicketDetailPanel.test.tsx` — status change, comment + draft-clear, description edit toggle, action error surfacing, and read-only-without-handlers (full vitest suite green; tsc clean).

## Notes

- No CI on stacked PRs (CI is `main`-only); verified green locally + figgyster review.
- CHANGELOG entry deferred to the final slice-4 PR (consolidated "Tickets / work tracker" entry once the actionable surface — resume + attachments — is fully in).

1. #412
2. #413
3. #414
4. #415
5. #416
6. #417
7. #418
8. #419
9. #420
10. #421
11. **#422** 👈 current
12. #423
13. #424


figgyster (approved):
This looks safe to merge. The ticket action path keeps the right shape: UI sends correlated by-id mutations, daemon records durable ticket events as `you`, nudges the assigned agent, and lets `tickets_updated` drive the refresh instead of pretending with optimistic local state.

I ran `go test ./internal/daemon ./internal/store ./internal/protocol`, `pnpm test -- TicketDetailPanel.test.tsx`, and `pnpm run build`. No blockers from me. Tiny alley note: the frontend test still emits a React `act(...)` warning, but the assertions pass and it’s not merge-blocking here.

PR #423, opened 2026-06-27
feat(tickets): slice 4d — resume a ticket's agent session

victorarias:
Slice 4d of the work tracker (stacked on #422 → #421 → #420 → main).

## What

A **Resume** button in the ticket detail header for **delegated** tickets — those carrying a stored `cwd` + `last_agent_id` (both set only when work was delegated to an agent). Clicking it reopens that agent in the ticket's `cwd`, resuming the prior conversation, then closes the detail view so the resumed session takes focus.

## How resume resolves (the important part)

The button re-spawns using the ticket's **bound session id** (`assignee`) as the new session id, with `resume_picker: true` and **no** explicit `resume_session_id`:

- **While the daemon still holds the session record** (idle / recently closed): its spawn path resolves the stored agent-native resume id via `GetResumeSessionID(assignee)` → a **precise** `--resume <uuid>`. The picker flag is moot.
- **Once the record is gone** (`dropSessionRecord` deletes it on close/reap, taking its resume id): no stored id exists, so the agent's own **cwd-scoped conversation picker** opens (`claude --resume` / `codex resume` / `copilot --resume`) and the user picks the prior conversation.

Crucially we never pass the attn session id *as* the resume id, so there's no bogus-UUID hand-off when the record is gone — the picker is the safe fallback. Reusing the same session id also keeps the ticket **bound to the same session** (`assignee` unchanged) — no re-bind, no new protocol.

## Scope

Frontend-only. `spawn_session` already supports `resume_picker` + `cwd` + `agent`, so there is **no protocol or daemon change**:

- `createWorkspaceSession` gains an optional `{ resumePicker }` — it sets `resume_picker: true` on the built spawn args (and leaves `resume_session_id` empty). Because new sessions use `intent: 'create'`, `spawnPtyRuntime` routes straight to spawn, so the flag reaches the daemon intact (it does *not* go through the live-runtime attach-recovery branch).
- `App` adds `handleResumeTicket(ticketId)` — resolves the row from the store, guards on cwd + agent, normalizes the agent, reuses `assignee` as the session id (unless it's `you`/empty), spawns, and closes the panel.
- `TicketDetailPanel` adds an optional `onResume` and the header button, gated on a resumable agent (works off the bare row too).

## Tests

`TicketDetailPanel.test.tsx`: Resume shows + fires `onResume(ticketId)` for a delegated ticket; hidden when cwd/agent are absent. Full vitest suite green; tsc clean.

## Notes

- Precise auto-resume *after* the session record is gone (skipping the picker) would need the agent UUID persisted on the ticket — a deliberate future enhancement beyond the locked decision #5, which scoped resume to `cwd`/`last_agent_id` + the existing resume machinery.
- The button shows whenever the ticket is resumable; hiding it for a currently-live session is left as polish.
- Real-Claude e2e of the full resume flow is batched after 4e (attachments), per the slice-4 plan.
- No CI on stacked PRs (CI is `main`-only); verified green locally + figgyster review. CHANGELOG deferred to the final slice-4 PR.

1. #412
2. #413
3. #414
4. #415
5. #416
6. #417
7. #418
8. #419
9. #420
10. #421
11. #422
12. **#423** 👈 current
13. #424


figgyster (dismissed):
I found one real blocker in the live-session path.

The PR says Resume reuses the ticket's bound `assignee` so the daemon can do a precise resume while it still has the session record. But the frontend also creates a new local session with that same id before spawning. `createSession` just appends; it does not replace/dedupe existing sessions. So for the common case where the delegated session is still present in the frontend store, clicking Resume adds a duplicate session id locally, and then `takeSessionSpawnArgs(sessionId, ...)` grabs the first matching row — likely the old session/workspace, not the just-created resume workspace. The daemon may return success because the PTY is already live, but the UI state is now poisoned with duplicate ids. Tiny rake, sharp teeth.

Required before merge:
- Handle the "assignee already exists in the local session store" case explicitly: select/focus the existing session/workspace or remove/replace the old local row before creating a new one. Either way, don't let `createSession` append a duplicate id.
- Add a regression test for Resume when `ticket.assignee` already exists as a frontend session.

I verified `tsc --noEmit` and the focused `TicketDetailPanel` vitest pass locally; this is a behavior/state bug those tests don't cover yet.

figgyster (approved):
This fixes the blocker cleanly. The resume decision now has an explicit “still tracked locally → focus it” branch, and the regression test pins the exact duplicate-session trap from the previous review.

No blockers from me. I verified `tsc --noEmit` plus the focused ticket resume/detail tests locally; GitHub reports no checks for this stacked branch.

victorarias:
Fixed in aacb3e52 (on behalf of Victor). Good catch — that was a real state-corruption rake.

The root cause was exactly as you described: reusing the ticket's bound `assignee` as the resume session id collides with the frontend's local session store, and `createSession` appends without dedupe, so a still-tracked session got a duplicate id and `takeSessionSpawnArgs` grabbed the wrong row.

What changed:
- Extracted a pure `planTicketResume(ticket, existingSessionIds)` (`app/src/utils/ticketResume.ts`). If the bound session is **still tracked locally**, the plan is `focus` — `handleResumeTicket` now routes through `handleSelectSession(id)` (sets active pane + session + utility focus); the app's own attach/resume-recovery (`spawnPtyRuntime`) revives a dead runtime on mount. No new local session is created, so no duplicate id.
- Only when the bound session is **gone from the local store** does it spawn — reusing the bound id (so the daemon resolves the prior resume id) or minting a fresh one when there's no usable bound id (`""`/`you`).
- Regression test added (`app/src/utils/ticketResume.test.ts`): assignee already present locally → `focus`, no duplicate spawn; plus the gone/fresh/no-agent branches.

`tsc --noEmit` and the full vitest suite (1157 passed) are green. The branch was restacked so #424 sits on top of the fix.

PR #424, opened 2026-06-27
feat(tickets): slice 4e — attach files to a ticket

victorarias:
Slice 4e of the work tracker — the **final** slice-4 sub-PR (stacked on #423 → #422 → #421 → #420 → main). Completes the ticket detail surface: detail view (4b) + actions (4c) + resume (4d) + **attachments (4e)**.

## What

`attn ticket attach --file <path> [--note <text>] [--session <id>] [--json]` — the agent hands a file to its **bound ticket**. Mirroring `attn ticket status`, the daemon resolves the ticket from the calling session (the assignee — an agent attaches only to its own active ticket), copies the file into the ticket's store directory, records the attachment authored by the session, then **notifies the chief + refreshes the board** (the same fan-out as the status forward channel).

## Where files land

`.attn/tickets/<id>/` under the **notebook root** (e.g. `~/attn-notebook/.attn/tickets/<id>/`), a new sibling of the raw tier. Like the raw tier it lives under `.attn/` — `CleanPath` rejects it and the external-sync scanner skips it — so it's written with direct filesystem I/O, not the notebook.Store APIs. Same-basename copies are **deduped on disk** (`report.md` → `report-2.md`) so they never clobber each other; the original name is what the ticket displays.

## Why path-based copy

The CLI sends the **absolute source path** (resolved + stat-checked for a clean early error) and the daemon copies it. That keeps binary artifacts (screenshots, logs, PDFs) intact with no base64 bloat, and is the literal "daemon copy into the store" the plan describes. It's a same-machine operation (local daemon); remote-session attach is out of scope.

## Surface

The detail panel already rendered attachments (4b); this adds a hover tooltip showing the stored path. No new panel UI — attachments now populate as agents attach.

## Protocol

Bumped to **v128**. New unix-socket command `ticket_attach` + `TicketAttachMessage`/`TicketAttachResult`, regenerated types.

## Tests

- Daemon (`ticket_attach_test.go`): copy lands verbatim under `.attn/tickets/<id>/` + records the attachment; same-name dedup keeps both files; no-bound-ticket and missing-file fail cleanly with nothing recorded (4 tests).
- CLI (`main_test.go`): `parseTicketAttachArgs` happy path + error cases (missing `--file`, stray positional, unknown flag).
- Frontend: tsc clean; full vitest suite green.

## Notes

- `dispatch handoff` **stays as-is** — folding it into `ticket attach` is the gated dispatch-removal work (slice 3e / 7), not this additive slice.
[...]

PR #425, opened 2026-06-27
fix(tickets): address slice-4 ultracode review

victorarias:
Follow-up to the slice-4 stack (#420–#424) addressing the findings from the multi-agent ultracode review. Stacked on `tickets/slice-4e-attach` so the already-approved 4a–4e PRs stay mergeable as-is — these are additive hardening + the test gaps the review surfaced, no behavior regressions.

## Major (the two test gaps the review flagged)
- **`tickets_updated` broadcast fan-out was untested** — a verifier confirmed both broadcast calls could be deleted with the suite still green. Now covered for both producers (chief status change + attach) via a deterministic in-process hook (`ticketsBroadcastHook`, mirroring `broadcastWorkflowRunUpdated` — `TicketsUpdatedMessage` is its own top-level event the WS `broadcastListener` can't see).
- **Attach notify-the-chief was asserted nowhere** — the test discarded the `inputs` accessor. Now asserts the involved chief is nudged about the agent's self-authored attachment while the self-author is not.

Both were mutation-tested: deleting the production call now fails the test.

## Minor / nits
- **`crashed` server-side allowlist** — the daemon (the authority) rejects `crashed` in the chief status handler; the UI already omitted it from `SELECTABLE_STATUSES`.
- **Attach copy** — name-pick + `O_EXCL` create folded into one bounded loop: a TOCTOU name race re-picks instead of surfacing a raw `EEXIST`, and a mid-copy failure `os.Remove`s the partial file (no orphan). `".."` added to the filename reject set.
- **Failed-mutation fan-out guard** — asserts a failed/rejected action fires no notify.
- **Dead `Response.ticket`/`tickets`** removed (all ticket data flows over WS events); regen, protocol 128 → 129.
- **`busyAction` race** — every ticket mutation control now disables while one action is in flight (kills the busy-indicator cross-clear).
- **Resume** — `executeTicketResumePlan` extracted + unit-tested (focus-vs-spawn dispatch); dead-runtime recovery comment corrected.
- **`fetchTicket` / ticket-action request/result** — covered in `useDaemonSocket` (resolve on result, reject on error/`success:false`).
- **Contract comment** — `Assignee` documented as the bound session id (the resume key).

## Verification
- `go build ./...`, `go test ./internal/{daemon,store,protocol,notebook}` green.
- Frontend: full vitest suite (1164) + `tsc --noEmit` green.


figgyster (approved):
This looks safe to merge. The slice-4 follow-up is doing the right kind of hardening: ticket fan-out and request/result paths now have actual behavioral coverage, the protocol cleanup bumps the version, and the attach-copy path got less raccoon-shaped under races.

I ran `go test ./...` plus the focused frontend ticket/resume socket tests (`pnpm exec vitest run src/hooks/useDaemonSocket.test.tsx src/utils/ticketResume.test.ts`). Both passed. GitHub reported no checks on this branch from here, so local validation is the signal I could verify.

No blockers from me.

PR #426, opened 2026-06-27
fix(terminal): keep small panes' last rows on screen

victorarias:
## The bug

Under some conditions the **bottom one or two rows of a terminal render outside the pane** — invisible, and unscrollable. It clears on app restart or by cycling the workspace through the warm-set LRU. Non-deterministic, so it felt random.

## Root cause

`isSuspiciousTerminalSize` (`cols <= 20 || rows <= 10`, `terminalDebug.ts`) was used as a **hard gate** in two places:

- `GhosttyTerminal.fit()` — refused to apply the measured size
- `useGhosttyPaneRuntime.handleTerminalResize()` — dropped the resize before it reached the daemon

The guard exists to ignore transient/garbage measurements taken before layout settles. But a **legitimately small** pane (deep stacked splits, a narrow side-by-side split, or a short window) produces a small grid too. Rejecting it left the Ghostty model/canvas at the previous, **larger** size while the container shrank. The wrapper is `overflow:hidden`, so the extra rows paint past the edge and can't be scrolled into view. It only bites once a pane crosses the ≤10-rows / ≤20-cols line — hence the randomness — and a fresh measure (restart / LRU reload) lands on a non-suspicious size and "fixes" it.

## The fix

- **`fitShouldBailAsSuspicious()`** — `fit()` now bails on a suspicious size **only when the model does not already overflow the container**, checked on both axes via the axis-agnostic `geometryOverflowsContainer`. A not-yet-laid-out container (client ≈ 0) can't overflow, so transient garbage is still suppressed; a real small pane gets its corrective fit. Covers the reported bottom-row clip *and* the symmetric right-column clip for very narrow splits.
- **`handleTerminalResize()`** — relaxed to reject only degenerate sizes (`< 1` / non-finite) so the corrective small resize reaches the daemon.
- **Silent detector** (`terminalDiagnosticsLog.ts`) — edge-triggered, records a `bottom_clip` incident to the on-disk diagnostics log when an active pane's canvas overflows its container. It caught this bug in the wild; kept as a prod regression net.

## Verification

A/B against a rebuilt **buggy** binary, driving a full-screen TUI (marks its last row, repaints on `SIGWINCH`) in a 4-pane stacked layout, same ~197px pane:

| build | rows the model uses | container fits | canvas overflow | last row |
|---|---|---|---|---|
| buggy | 11 | ~8 | **+62px** | clipped (rows 9–11 gone) |
| fixed | 8 | ~8 | **−1px** | visible at the pane edge |

Screenshots confirmed the buggy build shows `TOP rows=11` then cuts off before the `BOTTOM` marker, while the fixed build shows `TOP rows=8` … `BOTTOM row=8 <<< MUST BE VISIBLE >>>` at the pane edge.

- Frontend unit suite: **1154 pass**, `tsc` clean
- New regression tests: `fitShouldBailAsSuspicious` (both axes), relaxed `handleTerminalResize`, the bottom-clip detector

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-06-27]
+
+### Fixed
+- **A terminal's last rows no longer get cut off outside the window.** Under some conditions — most often after splitting a workspace into several stacked panes, or making the window short — the bottom one or two lines of a terminal could render past the edge of its pane and stay there: invisible, and impossible to scroll into view, until you restarted the app or cycled away and back to the workspace. A safeguard meant to ignore transient/garbage size measurements was also rejecting legitimately small panes, leaving the terminal larger than the space it had. It now keeps the terminal sized to its pane so every row stays on screen (the same fix covers the right-most columns being clipped in very narrow side-by-side splits).
+
 ## [2026-06-26]
 
 ### Fixed
--- a/app/src/components/GhosttyTerminal.resizePolicy.test.ts
+++ b/app/src/components/GhosttyTerminal.resizePolicy.test.ts
@@ -1,6 +1,7 @@
 import { afterEach, describe, expect, it } from 'vitest';
 import {
   fitRequiresTerminalResize,
+  fitShouldBailAsSuspicious,
   geometryOverflowsContainer,
   isWorkspaceResizeDragActive,
   liveResizeConflictsWithQueuedReplay,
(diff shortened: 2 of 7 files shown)


figgyster (approved):
This looks safe to merge. The fix keeps the old garbage-measurement guard where it belongs, but stops treating genuinely small panes like bad data — exactly the line this bug was hiding behind.

I ran the frontend suite locally: `pnpm --dir app test` (1149 passed, 5 skipped), plus `pnpm --dir app exec tsc --noEmit` and `git diff --check origin/main...HEAD`. GitHub’s Frontend check was still running when I last fetched it, but the local principal path is clean.

No blockers from me.

PR #427, opened 2026-06-27
test(harness): comprehensive ticket lifecycle UI automation

victorarias:
Stacked on #425. Closes the gap where the work-tracker **UI** was invisible to automated verification: it adds a single real-app harness command that bootstraps a chief, delegates a worker, drives both sides, and asserts the rendered `TicketDetailPanel` — on an isolated profile, with no human clicks.

## Why
Driving the ticket lifecycle on the packaged app this cycle surfaced two blockers:
1. **Named/custom profiles didn't expose the UI-automation layer.** The manifest the real-app harness attaches through was runtime-gated to the literal `dev` profile, so a `make install PROFILE=ticketqa` build never started the automation server.
2. **There were no ticket actions in the automation bridge** — no way to open the panel, drive its controls, or read its rendered state.

## What
- **Expose automation on any non-prod profile** (`app/src-tauri/src/profile.rs`): the gate now passes for any non-empty `ATTN_PROFILE` (the `dev` sibling or any named profile like `ticketqa`/`agent7`), preserving the explicit `ATTN_AUTOMATION=1/0` override. Production is the empty-profile bundle and stays **off** unless an operator opts in. Unit test + Makefile note + `docs/profiles.md` updated.
- **Ticket UI-automation actions** (`useUiAutomationBridge.ts`, `App.tsx`): a `ticket_*` family — `ticket_list`, `ticket_open_via_dashboard`, `ticket_open_detail`, `ticket_close_detail`, `ticket_detail_get_state`, `ticket_set_status`, `ticket_submit_comment`, `ticket_edit_description`, `ticket_resume`. Mutations **drive the real DOM controls** (status `<select>`, comment form, description editor) so the panel's own gating runs (disabled-while-busy, empty-comment-disabled, `crashed`-not-offered). A collector serializes the rendered panel for assertions.
- **End-to-end scenario** (`scenario-ticket-delegation-lifecycle.mjs`): self-bootstraps a chief (real sidebar toggle), delegates a real codex worker via the `attn delegate` CLI (mints the bound ticket), drives the worker side via the agent CLI verbs (`ticket status` / `ticket attach`), then drives the chief side through the panel and asserts the rendered result — status moves, comment, edited description, attachment, resumable agent, and that `crashed` is rejected as a manual destination. Captures an occlusion-proof panel screenshot. Registered in `package.json` and the serial matrix.

No daemon protocol change / no `ProtocolVersion` bump — the bridge is an app-internal TCP channel and the ticket actions reuse the existing daemon ticket commands.

## Verification
- `cargo test automation` (gate rules) ✓
- `go build ./...` ✓ · `tsc --noEmit` ✓ · `pnpm test` (1169 FE tests) ✓
- Rebuilt an isolated `ticketqa` profile and ran the scenario end-to-end against the real packaged app — passes, with the populated-panel screenshot artifact (BLOCKED badge, chief-edited description, full worker→chief history, `report.md` attachment).


figgyster (approved):
This looks safe to merge. The automation gate stays scoped to non-prod profiles, and the ticket harness drives the real panel controls instead of mocking the happy path — exactly the seam this needed to cover.

I ran `git diff --check`, `go build ./...`, `pnpm exec tsc --noEmit`, and the full frontend suite (`1164 passed, 5 skipped`). I couldn’t rerun the Rust `cargo test automation` check in this runner because `cargo` isn’t installed here, but I don’t see a blocker in the Rust gate change.

No blockers from me.

PR #428, opened 2026-06-27
feat(tickets): slice 5 — board view

victorarias:
## Slice 5 — Board view

Grows the proto-board into a real **status-column board**: a new read-only right-dock panel (open from **⌘K → "Open ticket board"**) that groups the chief's tickets into flow columns and lets you filter at a glance.

Closes slice 5 of `docs/plans/2026-06-26-work-tracker.md`.

### What it does
- **Five flow columns** — Todo · Working · Blocked · In Review · Done — mapping the status enum to spatial position, with the **Todo backlog** included and live per-column counts.
- **Terminal-bad** work (`failed` / `crashed`) folds into a **Closed sub-lane under Done** (Done count shows `done · bad` when present) — visible but distinct, keeping the scan to five columns.
- **Filters:** All · Blocked · In Review · **Closed today** (by `closed_at`, local day). A distinct "no tickets match" state with a *Show all* reset, separate from the truly-empty board.
- **Cards** show title, mono id, assignee, status accent, and relative time; the whole card is a `<button>` that opens the existing **TicketDetailPanel** (the Dashboard "View ticket" path). Working/Blocked cards carry a quiet at-rest heartbeat; closed-today cards get a struck/dimmed treatment.

### Design
Built with `/frontend-design` polish, cohesive with attn's dark terminal aesthetic and `TicketDetailPanel`: **token-only** colors (light/dark safe), status-tinted sticky column headers, hover lift + colored shadow, focus-visible rings, a blinking `>` empty state reusing the app's `cursor-blink`, and full `prefers-reduced-motion` handling. Authored via a multi-agent design→judge→implement→verify→review workflow (mixed opus/sonnet).

### Scope / spine
- **Frontend-only.** Reads the existing `tickets` store (`initial_state` + `tickets_updated` broadcast). **No daemon command, no protocol change, no `ProtocolVersion` bump.**
- **Awareness, not autonomy** — the board *informs*, never *gates*: no drag-to-mutate; status changes stay in the detail panel or are self-reported by agents. The panel takes only `{ isOpen, tickets, onOpenTicket, onClose }` (no mutation handlers).

### Tests
- 26 new unit tests (grouping, per-column counts, each filter, terminal-bad lane, a11y labels, empty / no-match states).
- Full suite green locally: **`tsc` clean, 1195 tests pass.**

### Note (out of scope — flagged for follow-up)
This work surfaced a **pre-existing bug**: `TicketDetailPanel.css` references bare `var(--color-error)` / `var(--color-accent)`, but those tokens **aren't defined in `App.css`** — so the detail panel's blocked/failed/working status colors are silently no-ops in the real app. The board defends against it (uses `var(--color-error, #ef4444)` with a fallback). A small follow-up should define `--color-accent` / `--color-error` in `App.css` (and drop the board's fallbacks).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ### Added
 - **Open and act on a delegated agent's ticket from the app.** Delegated work now has a ticket you can open from its session: see its description, status, full history, and attachments, all updating live as the work moves. From the ticket you can change its status, add a comment, or re-brief the agent — each notifies the agent — and **resume** a stopped session to pick the work back up where it left off. Agents can hand a file to their ticket with `attn ticket attach --file <path>`, and it shows up in the ticket's attachments.
+- **See all your tickets on a board, grouped by status.** A new **ticket board** — open it from the ⌘K command menu — lays out every active ticket in columns (Todo · Working · Blocked · In Review · Done) so you can see at a glance where delegated work stands. Each card shows the title, id, who's on it, and when it last moved; click one to open its full detail. Filter to just what's **blocked**, **in review**, or **closed today**, and finished-bad work (failed or crashed) collects in a Closed lane under Done. The board updates live as agents report progress.
 
 ## [2026-06-26]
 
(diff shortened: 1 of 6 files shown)


figgyster (approved):
This looks safe to merge. The board stays on the right side of the contract: read-only awareness from the existing ticket store, with ticket mutation still routed through the detail panel.

I ran `pnpm exec tsc --noEmit`, `pnpm exec vitest run src/components/TicketBoardPanel.test.tsx`, the full frontend `pnpm run test` suite, and `git diff --check`; all passed. Codex only left a usage-limit note, not a finding.

No blockers from me.

PR #429, opened 2026-06-27
feat(tickets): formalize self-monitor as an agent capability (slice 6)

victorarias:
## Slice 6 — Codex nudge path

The nudge engine itself shipped in slices 3d/4c (idle non-Claude agents get a fixed-trigger PTY doorbell on ticket activity; Claude self-monitors via its live watch). This slice closes the two genuine gaps the plan called for: **formalizing the self-monitor decision** and **an end-to-end roundtrip test**.

### What changed

**The self-monitor decision now lives where agent capabilities live.** It used to be a magic string buried in the *pure* `ticketnotify` package — `HasSelfMonitor(agent)` matching `"claude"`. That's agent-type knowledge `ticketnotify` shouldn't own. It moves to a `HasSelfMonitor bool` on `internal/agent.Capabilities` (alongside `HasHooks`, `HasResume`, …), set `true` only in `Claude.Capabilities()`; codex and the rest take the zero-value `false` → idle nudge.

- **Daemon resolves it from the registry.** `ticketObserverForSession` now reads `agent.EffectiveCapabilities(agent.Get(name)).HasSelfMonitor` and builds the `ticketnotify.Observer` directly. Unknown/empty agent → `EffectiveCapabilities(nil)` → `Capabilities{}` → `false`, the safe nudge default. The daemon already imports `internal/agent`, so **no new import cycle**.
- **`ticketnotify` stays pure.** Retired `AgentObserver` + `HasSelfMonitor(string)` (the magic string and its only wrapper), dropped the now-unused `strings` import. `Observer{ID, HasSelfMonitor bool}`, `Notify`/`Consume`, and the harness-only `ChiefObserver` are untouched.
- **Env-override parity:** `ATTN_AGENT_<AGENT>_SELF_MONITOR=0|1`, matching every other capability field.

**Net runtime behavior is byte-identical** (claude → `DeliveryWatch`, codex → `DeliveryNudge`). Only the capability's *home* changes — and it becomes discoverable on the driver instead of hidden in a string compare.

### Tests
- `TestCapabilitiesSelfMonitor` (internal/agent) — pins the agent→bool mapping in its new home (claude true, codex false).
- `TestCodexNudgeRoundtrip` (internal/daemon) — the slice's end-to-end coverage: a **real chief→agent producer** (human comments on the idle codex agent's bound ticket via `handleTicketAddComment`) fires `notifyTicketObservers` → the codex agent is nudged (`DeliveryNudge`, capability resolved through `ticketObserverForSession`) → it consumes the event via the inbox path → a second inbox is empty, proving the cursor advanced (consumed, not peeked). No real codex binary or PTY — the fake spawn backend captures the doorbell.
- Existing `ticketnotify` harness tests now construct observers with explicit `HasSelfMonitor` bools — more honest for a pure package.

### Notes
- No protocol change / no `ProtocolVersion` bump — the capability is internal Go, not on the wire.
- No CHANGELOG entry — internal refactor of where the capability lives + test; the user-visible nudge behavior already shipped.
- Built with the ultracode workflow (design panel → judge → implement → verify → adversarial review, mixed opus/sonnet). `go build ./...` clean; `internal/agent`, `internal/ticketnotify`, `internal/daemon` ticket/notify tests green. Reviewers returned ship / fix-then-ship (the one finding — stage the untracked agent test — is included here).

Part of the tickets work-tracker epic (`docs/plans/2026-06-26-work-tracker.md`). Targets `feat/tickets`.

### DIFF
--- a/docs/plans/2026-06-26-work-tracker.md
+++ b/docs/plans/2026-06-26-work-tracker.md
@@ -242,7 +242,7 @@ each is a meaningful, verifiable chunk.
 | 3 | Delegation ⇄ tickets (live wiring) | ✅ (sub-split — see `2026-06-26-work-tracker-slice3.md`) |
 | 4 | Ticket view + resume + attachments | ✅ (sub-split — see `2026-06-26-work-tracker-slice4.md`) |
 | 5 | Board view | ✅ (status columns + Todo backlog + filters; read-only awareness surface, opened from ⌘K) |
-| 6 | Codex nudge path | ⬜ |
+| 6 | Codex nudge path | ✅ (self-monitor formalized as an `agent.Capabilities` flag; daemon resolves it from the driver registry; pure `ticketnotify` + end-to-end codex-nudge roundtrip test) |
 | 7 | Retire the `dispatch` namespace | ⬜ |
 | 8 | Export ticket state (CLI) — *post-merge, lands on `main`* | ⬜ |
 
--- a/internal/agent/capabilities_self_monitor_test.go
+++ b/internal/agent/capabilities_self_monitor_test.go
@@ -1,0 +1,16 @@
+package agent
+
+import "testing"
+
+// The agent->self-monitor mapping that used to live as a magic string in
+// internal/ticketnotify (HasSelfMonitor("claude")) is now a driver capability.
+// This is its home: only Claude self-monitors today; codex (and the rest, which
+// rely on the zero value) take the idle-nudge path.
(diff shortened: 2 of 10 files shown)


figgyster (approved):
This is clean. Moving the self-monitor decision into agent capabilities is the right home for it, and keeping `ticketnotify` bool-driven preserves the pure package boundary instead of smuggling driver knowledge through a string compare.

`go test ./...` is green locally. No blockers from me.

PR #430, opened 2026-06-27
feat(tickets): retire the chief-of-staff dispatch namespace (slice 7)

victorarias:
## Slice 7 — Retire the `dispatch` namespace (the completion bar)

The chief-of-staff **dispatch** namespace was the work-delegation precursor that **tickets** replaced. With tickets fully wired (slices 1–6), it's dead weight — and the plan is explicit that leaving `dispatch` alongside the ticket model is an *incomplete* migration, not an acceptable end state. This removes it **atomically** (Victor confirmed the one-PR approach) and rewires delegation to be tickets-only. **Net product behavior is unchanged** — delegation, crash detection, and notebook narration all keep working; only the redundant dispatch half is gone.

`~9.1k deletions / ~550 insertions`, overwhelmingly mechanical.

### Removed
- **`internal/dispatch`** — the dispatch `Classify` / `Watch` / `IsTerminalReport` engine.
- **Daemon**: `chief_of_staff_dispatch.go` (handlers) and `notebook_dispatch_journal.go` (the dispatch "delivery ledger" raw-tier writer).
- **Store**: `chief_of_staff_dispatches.go` + `_messages.go` CRUD. The **tables are left orphaned, not dropped** — create migrations 44/45/46 + the epic-only `closed_state` column (53) are append-only history; removing them would break DB replay on existing installs. A drop is a deliberate later migration-57 PR if ever wanted. (`DelegatedFromChiefSessionIDs` moved to `tickets.go`.)
- **CLI**: the entire `attn dispatch …` surface (`cmd/attn/main.go`) + the client RPC wrappers.
- **Protocol**: the dispatch commands (`list_dispatches`, `submit_dispatch_outcome`, `handoff_dispatch`, `get_dispatch`) + their now-unreferenced models from `main.tsp`/`constants.go`, regenerated `generated.go/.ts`. **ProtocolVersion 129 → 130** (forces a clean reconnect; old daemon + new app fail closed). `DispatchWorkState` **stays** — the ticket status path still uses it.
- **#397 crash machinery**: `captureTicketCrashState` (slice 3c) already runs at both session-end paths and fully replaces the dispatch close-state path, so crash detection is preserved.
- **Frontend** dispatch consumers (Dashboard, `useDaemonSocket`, store, UI-automation bridge).

### Rewired
- `delegate.go` no longer mints a parallel `ChiefOfStaffDispatch`. It creates the session + binds the ticket exactly as before; the delegated agent's brief now carries **`delegatedTicketPrompt`** (self-report via `attn ticket status …`) in place of the old dispatch-report prompt. Work-state names (`in_progress`/`needs_input`/`ready_for_review`/`completed`/`failed`) match the `attn ticket status` CLI.

### Preserved (the overloaded-`dispatch` traps)
`internal/classifier`; the notebook **raw-tier** mechanism (`rawTierFilename`/`Segment` — narration depends on it; only the dispatch *journal writer* was removed); shortcut **chord** dispatch; DOM `dispatchEvent`; **workflow** step dispatch; the `delegate` command + `delegated_from_chief`. Also cleaned up the now-dead `RAW_DISPATCHES_DIR` wiring + the narration-prompt instruction that pointed the keeper at a dir nothing writes anymore.

### Verification
- `go build ./...` clean; daemon (full) / store / client / cmd / protocol / agent / notebook tests green.
- `tsc --noEmit` clean; frontend tests green.
- Dangling-ref grep across the repo: **zero live references** to the removed dispatch surface. KEEP-list audit confirms the raw-tier, narration, classifier, generic dispatch, and delegation all survive.
- Built with the ultracode workflow (3-lens discovery → synthesized removal spec → backend+protocol impl → frontend impl → triple verify → adversarial KEEP-violation review, mixed opus/sonnet). I additionally reviewed the full diff against the KEEP/REMOVE guardrails by hand and folded in the reviewers' fix-then-ship finding (the dead `RAW_DISPATCHES_DIR` cleanup).

Part of the tickets work-tracker epic (`docs/plans/2026-06-26-work-tracker.md`). Targets `feat/tickets`. This is the last integration-branch slice; slice 8 (`ticket export` CLI) lands separately on `main` after the epic merges.


figgyster (dismissed):
This removal is mostly clean, and the tickets-only delegation path held up under local checks. But there’s one stale prompt contract left that crosses from docs into runtime behavior.

`buildNarrateWorkspacePrompt` no longer passes `RAW_DISPATCHES_DIR`, and the dispatch raw-tier writer is gone, but the narration brief still tells the keeper that removal-pass truth includes “the dispatch outcomes” and later says “session digests and dispatch outcomes ... are the TRUTH.” That is now a phantom input. On a removed workspace, the keeper can still be instructed to ground itself in evidence the daemon no longer provides, which is exactly the kind of subtle migration ghost this slice is supposed to kill.

Required before merge:
- Update the remaining narration prompt language in `internal/daemon/notebook_narration_prompts.go` (and the nearby narration comment if you want it tidy) so the keeper’s ground-truth contract matches the new tickets/raw-session model.
- Do one pass over the public/user docs that still advertise removed dispatch surfaces (`README.md`, `docs/glossary.md`, `docs/notebook-processes.md`, `docs/notebook-uat.md`). If some plan docs are intentionally historical, fine — but the current-facing docs shouldn’t tell users or future agents to expect `dispatch`/`.attn/raw/dispatches` after this PR.

Validation I ran locally: `go test ./...`, `pnpm install --frozen-lockfile`, `pnpm test`, `pnpm exec tsc --noEmit`, and `git diff --check`. All passed. The red here is stale contract text, not the mechanical deletion.

figgyster (approved):
This fixes the blocker cleanly. The runtime narration contract now matches the post-dispatch world — session digests plus the context snapshot, no phantom dispatch evidence hiding in the fog.

I also checked the current-facing docs you touched; the remaining dispatch mentions are either removed from the active contract or explicitly marked historical/retired. That’s the right shape for a migration cleanup like this.

Local validation passed: `go build ./...`, focused daemon narration/raw-tier tests, `go test ./internal/notebook ./internal/agent ./internal/protocol ./cmd/attn`, and `git diff --check`. No blockers from me.

PR #431, opened 2026-06-27
chief-delegation: align prompts + vision with the tickets model

victorarias:
Rebalances the chief/normal agent **system prompts** and the **attn skill**, and rewrites the **chief-delegation vision doc**, around the shipped tickets reality. One coherent theme: *the chief makes you aware and tees it up — you review the engine, it can proofread the docs.*

## Why
Two prompting problems, found by reviewing the system prompts + attn skill against the prompting craft rubric (and verified adversarially):
- **Standing chief guardrails lived only behind the model-invoked skill door** — an agent that never opened the skill had no stop-condition, no autonomy tier, no untrusted-output rule. Those belong in the always-present system prompt.
- **The chief's role was over-scoped.** It was told to *verify delegated work against the deliverable* and to *re-delegate / take over* on failure — i.e. validate and drive recovery. Per the vision, the **user** drives the agents and reviews the work; the chief surfaces and tees up. The exception is docs/prose, within the chief's own competence (*Alfred proofreads the correspondence; he doesn't sign off on the rebuilt engine*).

## What
**System prompts (`internal/hooks`)** — promote the standing guardrails into `NotebookGuidance` (stop-condition, autonomy tier, delegation boundary, untrusted-output), demote just-in-time procedure (PARA write-mechanics, the doc-model taxonomy) back behind the skill door, trim the duplicated workflow-trigger sell. **Net token-flat.**

**attn skill** — description gains trigger phrases + a boundary against native subagents and a markdown-explainer trigger; `chief-of-staff.md` role block rewritten for the surface/tee-up boundary (no autonomous validation or recovery); board-column vs CLI-token vocab fixed; `workflow.md` gains a TOC; retired `dispatch:<id>` source token removed.

**Vision (`docs/vision/chief-delegation-awareness.md`)** — rewritten from the retired `dispatch`/mailbox/Monitor idiom to the shipped ticket/board/notify model. The old "re-arm monitors across sessions" open question is *dissolved* by durable tickets.

## Notes
- **No protocol change** — prompt/skill/doc text + the Go that emits it; tests updated. `go build ./...` clean; `internal/hooks`, `internal/agent`, `internal/notebook` green.
- Tier-1 edits all land in `internal/hooks` so Claude + Codex stay in parity; no new injection site, so no new double-inject marker.
- Reviewed via plannotator; this is the post-feedback state.


figgyster (dismissed):
This is close, but the changelog currently says the opposite of the new chief boundary.

`CHANGELOG.md` says the chief “verifies an important result against the actual deliverable … rather than trusting a status comment.” The rest of this PR very deliberately moves the chief away from validating specialist work: it surfaces the agent’s claim, tees up the artifact, and leaves engineering review/recovery to Victor, with the docs/prose exception.

Required before merge:
- Reword that changelog sentence so it matches the shipped contract: the chief surfaces the claimed result and artifact / recommended next step, not that it verifies specialist deliverables.

Everything else looks coherent. `go build ./...`, `go test ./...`, and `git diff --check` all pass locally.

figgyster (approved):
This fixes the blocker. The changelog now matches the prompt contract: the chief surfaces agent reports and tees up the next step; it doesn't pretend to sign off on specialist work.

I also checked the added backlog-ticket follow-up note in the plan. It lines up with the open vision item and keeps the actual create path scoped to a later PR, which is the right place for it.

Safe to merge. Local checks are clean: `git diff --check`, `go test ./...`, and `go build ./...`.

PR #432, opened 2026-06-27
fix(tickets): recover the fullscreen board surface + entry points left off the merge

victorarias:
## What

Recovers the ticket **board surface** work that was tested live but never made it into `main`. It lived on a local-only branch (`tickets/polish-and-resume`, commit `58fe9a02`), was never pushed, and so the `feat/tickets → main` merge (#412) left it behind — the shipped board was the right-dock drawer, missing the fullscreen surface and entry points.

This restores:

- **Full-window `TicketBoardSurface`** — the board is a primary surface (fixed inset-0 overlay, FocusTrap, shared Escape stack), not a right-dock peek. Mirrors the notebook surface pattern.
- **Entry points** — open the board from the **sidebar**, **⌘K**, or **⌘⇧T**.
- **Immediate resume** — the bound session's agent-native resume id is mirrored onto its ticket (new migration), so ticket **Resume** reattaches the prior conversation directly instead of dropping into the agent's resume picker. The session row (and its resume id) is deleted on close, so the durable copy on the ticket is what survives.
- **Dashboard keyboard fix** — keyboard focus stays on the home dashboard / empty workspaces, so shortcuts (even ⌘K) respond immediately instead of needing a click first.

## How it was recovered

The commit's parent (`0195d9fd`) is already in `main`, so it cherry-picked cleanly — the only conflict was `CHANGELOG.md` (resolved to keep the single collapsed ticketing entry from `main`, fold in the new entry points, and add the new keyboard-fix bullet).

## One real fix on top (`8838942f`)

The recovered commit had never run through the test suite (local branch, no CI), and its migration `57` was a bare `ALTER TABLE tickets ADD COLUMN` — which fails the migration **replay tests** (`DELETE schema_migrations >= N; re-migrate`) with `duplicate column name`. SQLite has no `ADD COLUMN IF NOT EXISTS`, so every other additive migration in this file is guarded by `columnExists`; `57` was the lone exception. Fixed by mirroring `applyMigration23` (the same column on `sessions`): a forward run still adds it once, a replay is a no-op.

## Verification

- `go build`, `go vet`, full Go suite — **33/33 packages green** (migration replay tests now pass)
- frontend `tsc` clean; **134 frontend tests pass** (TicketBoardPanel, TicketDetailPanel, shortcuts, ticketResume)


figgyster (approved):
This looks safe to merge. The fullscreen board comes back as a real primary surface, and the resume-id mirror is wired at the daemon/store seam instead of trying to resurrect it from the deleted session row later — good, boring durability.

I ran `git diff --check`, `go test ./internal/store ./internal/daemon`, and `pnpm --dir app run build`; all passed locally. GitHub's Frontend check was still running when I last looked, so let that finish, but no blockers from me.

PR #433, opened 2026-06-28
feat(tickets): standalone `ticket new` + propose-not-act agent awareness (slice 9)

victorarias:
## What

Slice 9 of the work-tracker epic: a first-class way to **capture a backlog ticket without delegating**, plus the agent-awareness that makes it usable.

### Capability
`attn ticket new --title <t> [--description <d>] [--id <slug>]` mints an **unbound** ticket — `status=todo`, no assignee, no session. This is the user capturing work to do later, distinct from delegation (which mints a *bound, working* ticket and dispatches an agent).

- Delegation and standalone-create share `createTicketWithUniqueSlug`, so slug + collision behavior is byte-identical.
- An explicit `--id` is a **hard fail** on collision (so the user fixes/renames it immediately); a title-derived slug **auto-suffixes** (`migrate-store-to-x` → `-2`).
- The calling session authors the `created` event but is *not* the assignee — the ticket sits in the backlog until someone picks it up. Only side effect is the board re-broadcast (nobody to notify).

### Awareness (two-tier, propose-not-act)
- A thin, always-on ticket-awareness pointer is injected into **both** the chief and non-chief agent system prompts (`hooks.TicketAwarenessGuidance`), with depth in the attn skill's new `references/tickets.md` (anatomy of a good ticket, durable-description-vs-live-steering, deliverable-type taxonomy).
- **Creating a ticket is user-triggered.** An agent may *surface* a ticket worth filing, but never files one on its own initiative — the boundary wording is kept verbatim across the chief prompt, the agent prompt, and the skill reference.

### Protocol
New `ticket_create` command + `ticket_create_result`; `ProtocolVersion 130 → 131` (frontend `PROTOCOL_VERSION` bumped to match).

## Verification
- `go build ./...`, `go vet`, `gofmt -l` clean; changed-package Go tests green (daemon/hooks/protocol/cmd/agent).
- `make generate-types` byte-identical; frontend `tsc --noEmit` clean; `useDaemonSocket` suite (52 tests, incl. protocol-version match) green.
- End-to-end against the live dev daemon: all CLI paths (create / auto-suffix / explicit-collision hard-fail / missing-title / help), DB persistence (unbound todo, author = calling session), and the board UI rendering the new todo cards live.

## Docs
Marks slice 9 ✅ in `docs/plans/2026-06-26-work-tracker.md` and closes the vision's *"backlog without a delegation"* open item in `docs/vision/chief-delegation-awareness.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-06-28]
+
+### Added
+- **Create backlog tickets without delegating.** `attn ticket new --title <t> [--description <d>] [--id <slug>]` files an unbound ticket in the Todo column — for capturing work you're not handing to an agent yet. Agents now know that tickets are something they can create on your request.
+
 ## [2026-06-27]
 
 ### Added
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -171,7 +171,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '130';
+export const PROTOCOL_VERSION = '131';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 interface PRActionResult {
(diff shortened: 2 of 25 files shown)


figgyster (approved):
This looks safe to merge. `ticket new` lands as the right kind of primitive: capture-only, unbound, and sharing the slug path with delegation without accidentally dispatching anything.

The propose-not-act boundary is also pinned in the launch guidance and skill docs, which is the part I cared about most here — no agents quietly filling the board with “helpful” little gremlins.

Local checks passed on my side (`git diff --check`, focused Go package tests, frontend `tsc --noEmit`, and the `useDaemonSocket` suite), and GitHub CI is green. No blockers from me.

PR #435, opened 2026-06-28
feat(attn-skill): add prompting pointer to capability index

victorarias:
## Summary

Adds a capability index entry to the attn skill pointing at the `prompting` skill for any work that involves writing, reviewing, or improving a system prompt, agent guidance, or skill instruction.

Previously the chief had no signal to load the prompting skill when working on prompt design — it would load the chief-of-staff reference but skip prompting even when the task was explicitly about improving injected guidance.

## Test plan

- [ ] Trigger a chief session to improve a system prompt and confirm it loads the `prompting` skill before starting


figgyster (approved):
This is safe to merge. The capability index entry is exactly the right place for this pointer: prompt/guidance work now routes to the prompting skill before the chief starts improvising in the dark.

`git diff --check` is clean, and the visible CI/release preflight checks are green. No blockers from me.

PR #436, opened 2026-06-28
feat(tickets): chief watch + daemon backstop for delegated-ticket awareness

victorarias:
## What & why

The chief of staff delegates work but didn't watch for ticket updates — when a delegated agent finished, got blocked, or crashed, the chief found out only the next time you prompted it; the update sat unseen on the board. This closes that gap with three mechanisms in one PR (full design in `docs/plans/2026-06-28-delegated-ticket-awareness.md`).

## Three mechanisms

1. **Always-on guidance.** The chief's delegate → watch → boundary behavior moves from a lazy-loaded skill section into the always-on `hooks.ChiefGuidance` system prompt, so it can't be missed. The skill keeps only the worker-facing reporting half (`references/chief-of-staff.md` → `references/delegated-agent.md`); `delegation.md` and `SKILL.md` shrink to match.
2. **`attn ticket inbox --watch`.** A silent-until-changed CLI poll loop the chief arms as a harness Monitor for a true push, instead of hand-rolling a jq diff loop. Reuses the consuming `ticket inbox` (no protocol bump; the daemon cursor dedups). Reports a daemon outage **once per outage**, not every poll, so a wrapping Monitor isn't nudged every interval.
3. **Daemon backstop (safety net).** For an **idle** self-monitor (Claude) with unread ticket activity that no live `--watch` drained, the daemon schedules a debounced re-check after a grace (8s, > the 3s watch interval) and doorbells **only if still idle + unread** — so a live watch self-suppresses it and nothing stays silently unread. `ticketnotify.Notify` stays pure/unchanged; the synchronous `DeliveryWatch` no-op is still correct.

## Tests

- **daemon** (`-race` clean): the headline scenario — idle Claude chief + delegate reports `ready_for_review` → chief is doorbelled, while the reporting agent is never doorbelled about its own change. Plus suppress-after-watch-drains, suppress-when-busy, stale-timer identity guard, burst debounce, and went-idle scheduling. Caught and fixed a real timer-capture data race (self-referential `time.AfterFunc` handle) with a channel sync edge.
- **cmd/attn**: `watchTicketInbox` extracted behind injected fetch/writers and tested deterministically — prints new bundles, silent on empty, outage reported exactly once then re-reported after recovery.
- Full suites green: `internal/daemon`, `cmd/attn`, `internal/hooks`, `internal/agent`, `internal/ticketnotify`.

## Manual verification

- Dev-install smoke: `attn ticket inbox --watch` runs live against the dev daemon, polls silently with no spurious output, and exits 0 (clean handler) on SIGTERM.
- **Not yet exercised live** (reasonable to confirm during review): a real Claude chief arming the Monitor and reacting on its own after a delegate reports back — that loop is inherent LLM behavior the tests can't assert.


figgyster (dismissed):
This is safe to merge. I fixed the one real snag myself — the daemon struct needed `gofmt`, which was what tripped CI — and the latest head is clean.

The watch/backstop split is the right shape: the chief gets a real push path via `ticket inbox --watch`, while the daemon only doorbells after the grace if nobody drained the unread queue. That keeps the awareness guarantee without turning every delegated ticket update into a tiny alarm clock with legs.

Validation:
- `go test ./...`
- `go test -race ./internal/daemon -run 'TestTicketBackstop|TestNotify|TestWatch|TestCodexNudgeRoundtrip'`
- GitHub CI is green on the current head.

figgyster (dismissed):
This still looks safe to merge. The follow-up closes the real live-promotion gap: chief assignment now reloads the agent through the daemon, preserves resume/launch params, bumps the protocol, and has the frontend reattach to the replaced runtime instead of pretending a dead PTY is fine. That’s the right kind of machinery for this problem — unpleasant, but pointed at the actual boundary.

Validation on the current head:
- `go test ./...`
- `go test -race ./internal/daemon -run 'TestReloadSessionAgent|TestSetChiefOfStaff|TestTicketBackstop|TestNotify|TestCodexNudgeRoundtrip'`
- `pnpm exec vitest run src/hooks/useDaemonSocket.test.tsx`
- GitHub CI is green

No blockers from me.

figgyster (dismissed):
This still looks safe to merge. The latest push turns the chief path from “reload after promotion” into the stronger first-launch contract: create-as-chief sets the role before spawn, avoids the zero-turn Claude resume trap, and keeps the harness honest by proving the role/guidance/model before measuring behavior.

I also like the small picker guard: only workspace creation can offer the chief toggle, so split-session creation doesn’t advertise a flag it can’t carry. Tiny UI lie prevented; always a good day in the alley.

Local checks passed here:
- focused Go tests for agent resume availability, daemon create-as-chief/reload/chief/ticket notification, and CLI watch paths
- `pnpm exec tsc --noEmit`
- focused Vitest files: `LocationPicker`, `useDaemonSocket`, `App.worktreeCleanup`

No blockers from me. Frontend CI was still running on my last fetch; the principal local frontend checks passed.

figgyster (approved):
Still safe to merge. This push is exactly the right cleanup: the shipped ChiefGuidance now speaks to “the user” instead of hardcoding Victor, while leaving dev/docs references alone where they’re intentionally about this repo’s maintainer context.

I rechecked the prompt diff and ran the focused hooks/agent tests plus `git diff --check`; all clean. No blockers from me. CI was still finishing on my last fetch, but the touched surface is narrow enough that this approval can stand on the local checks.

PR #437, opened 2026-06-29
feat(settings): UI for auto-approve + per-agent chief model

victorarias:
## What

Surfaces the two benchmark-support daemon settings shipped in #436 in **Settings › Agents**. Until now they were wired through the daemon but had no UI — reachable only by poking daemon settings directly. This is the deferred SettingsModal slice from the [delegated-ticket-awareness plan](docs/plans/2026-06-28-delegated-ticket-awareness.md) (lines 280/285/286).

- **Auto-approve** — a global toggle that mirrors the existing Workflows toggle exactly. Launches managed agents in their native auto-approve mode (Claude `--permission-mode auto`, Codex auto-review) so they run unattended without stopping at every permission gate. Off by default; yolo sessions already bypass approvals and ignore it.
- **Chief-of-staff model** — a per-agent text input for `chief_model_<agent>`, shown for **claude and codex only** (the drivers whose `BuildCommand` honors `opts.Model`; copilot ignores it). Committed on blur, blank = the agent's own default. Only affects chief launches.

## Design notes

- **Free-form text input, not a preset dropdown.** The backend validation is explicitly free-form ("accept any value and let the agent reject bad ones"). A Default/preset/Custom `<select>` walks into a snap-back trap — "Default" and "Custom-before-typing" both correspond to an empty model string, so the select can't be derived from the committed value without extra local mode state. A plain per-agent input (mirroring the existing executable-path inputs) is faithful to the free-form backend, sidesteps the trap, and dodges stale curated aliases.
- **No protocol change / no version bump** — settings ride the generic `Record<string>` map; the daemon already normalizes and ships these keys.
- Availability-gated like the rest of the modal, with saved-value re-inclusion (a configured agent stays visible even if it becomes unavailable — mirrors `keeperAgents`).
- Bumps the Agents nav badge (`+4` → `+6`) for the two new blocks.

## Verification

- `tsc --noEmit` clean.
- Full vitest suite green (1210 tests), including 5 new `SettingsModal chief settings` cases: both toggle directions for auto-approve; chief-model renders per supported agent (and **not** copilot); commit-on-blur of a typed model; no spurious write when unchanged; clear-back-to-default.
- Desktop `settings-visual` Playwright e2e renders both new sections in a real browser; confirmed the chief-model section shows only claude + codex while copilot is available elsewhere in the modal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-06-29]
+
+### Added
+- **Run agents unattended, and pin your chief's model.** Settings › Agents adds two controls. **Auto-approve** launches managed agents in their native auto-approve mode so they keep working without stopping at every permission prompt — off by default, and yolo sessions already bypass approvals regardless. **Chief-of-staff model** pins the model a chief-of-staff session launches with, per agent (Claude or Codex); leave it blank to use the agent's own default. Both apply to sessions started afterward.
+
 ## [2026-06-28]
 
 ### Added
--- a/app/src/components/SettingsModal.test.tsx
+++ b/app/src/components/SettingsModal.test.tsx
@@ -855,3 +855,97 @@ describe('SettingsModal keeper', () => {
     expect(onSetSetting).toHaveBeenCalledWith('notebook.narrate_workspace', '');
   });
 });
+
+describe('SettingsModal chief settings', () => {
+  function renderModal(
+    settings: Record<string, string>,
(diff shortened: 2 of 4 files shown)


figgyster (approved):
This looks safe to merge. The Settings UI mirrors the existing settings-map pattern, keeps the chief model override scoped to Claude/Codex, and the tests cover the new toggle/input behavior instead of just admiring the TypeScript from a distance.

Local validation passed: `git diff --check`, `pnpm exec tsc --noEmit`, focused `SettingsModal` tests, and the full frontend vitest suite (1205 passed, 5 skipped). GitHub’s Frontend job was still running when I last checked, but the local principal checks are green.

No blockers from me.

PR #438, opened 2026-06-29
fix(tickets): stop delegated agents self-nudging about their own brief

victorarias:
## Problem

Victor noticed delegated agents getting nudged ("📋 New ticket activity — run `attn ticket inbox`") when they shouldn't. Diagnosis traced two distinct things:

- **The cross-ticket routing leak he suspected does not exist.** The store query scopes an agent strictly to `tickets WHERE assignee = me UNION tickets I authored on`, with self-authored events excluded. A sibling working on ticket A is never a participant of ticket C, so C's completion can't route to A. Proven by `TestDelegatedSiblingsNotNudgedByEachOther`.
- **A real self-nudge bug.** A chief delegation delivers the brief into the agent's PTY as its spawn prompt, but it *also* lands as a chief-authored `created` event on the agent's own ticket — and nothing marked it consumed. So every delegated agent carried one permanently-unread event on its own ticket and got doorbelled about it the moment it went idle (codex immediately, Claude via the 8s backstop): a "check your inbox" pointing at a brief it already holds.

## Fix

In `store.CreateTicket`, a ticket born *with* an assignee (only delegation sets one at birth — backlog tickets are unbound) advances that assignee's cursor past the `created` event in the same transaction. The brief is delivered out of band, so it's never re-served as inbox noise or a doorbell. A backlog ticket picked up later stays unassigned at birth, so its brief still arrives via `attn ticket inbox`.

## Tests

- `TestDelegatedAgentNotNudgedByOwnDeliveredBrief` — red→green repro of the self-nudge.
- `TestDelegatedSiblingsNotNudgedByEachOther` — locks store-level sibling isolation.
- Three existing tests had used the unread brief as their fixture; they now drive a real chief steer instead.
- Full Go suite green.

## Note

The brief-self-nudge fires when each agent goes idle on its own schedule, not causally when a sibling completes. If A/B still appear pinged *at the moment a sibling finishes*, that's the chief's `--watch` Monitor waking it to steer them — expected behavior, separate from this fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ### Added
 - **Run agents unattended, and pin your chief's model.** Settings › Agents adds two controls. **Auto-approve** launches managed agents in their native auto-approve mode so they keep working without stopping at every permission prompt — off by default, and yolo sessions already bypass approvals regardless. **Chief-of-staff model** pins the model a chief-of-staff session launches with, per agent (Claude or Codex); leave it blank to use the agent's own default. Both apply to sessions started afterward.
 
+### Fixed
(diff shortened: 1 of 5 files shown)


figgyster (approved):
This looks safe to merge. The fix is in the right place for the current delegation contract, and the tests now pin both halves of the story: siblings do not cross-nudge each other, and a freshly delegated agent does not get doorbelled about the brief it already received.

`make test` passed locally, and GitHub's daemon/preflight checks are green. No blockers from me.

PR #439, opened 2026-06-29
feat(delegate): non-chief agents confirm before completing a ticket

victorarias:
## What

A delegated worker now treats closing its ticket as the user's call. The delegated-agent prompt (`delegatedTicketPrompt`) tells it: when it believes the work is done, ask the user to confirm and wait for the go-ahead before reporting `completed` — rather than closing the ticket on its own. The other work states (in progress, needs input, ready for review, failed) are still reported as they happen.

## Why

Agents were marking tickets `completed` autonomously. Completion is the moment a human (or chief) wants to weigh in, so the agent should surface and wait, not self-close.

## Scope

Guidance/prompt change only — no protocol or daemon-behavior change. Test asserts the new guidance is present in the delegated prompt.


figgyster (approved):
This looks safe to merge. The prompt-only change keeps the ticket contract honest: workers can still report progress/review/failure, but completion now waits for the human go-ahead instead of letting the agent close its own loop.

`go test ./internal/daemon` passes, `git diff --check` is clean, and GitHub checks are green. No blockers from me.

PR #440, opened 2026-06-29
feat(tickets): agent ticket collaboration foundation — board read + comment

victorarias:
## What

The foundation for agents collaborating across tickets, shipped as one usable capability (Victor's call — held so it only hits main when an agent can actually discover, find, and act on a ticket):

- **`attn ticket list [--status <col>] [--all] [--json]`** — the board read. One line per ticket (id, column, assignee, title), newest first; `--json` adds each ticket's description. This is the missing foundation: `ListTickets` was exposed *only* to the app UI via the `tickets_updated` broadcast, so a headless agent — even the chief over the CLI — had no way to enumerate tickets or look up an id.
- **`attn ticket comment <ticket-id> -m "<text>"`** — a one-shot note on **any** ticket by id, even one the agent isn't assigned to. Informs the ticket's participants but does **not** subscribe the commenter.
- **Guidance** so agents know the capability exists.

Plan: `docs/plans/2026-06-29-agent-ticket-collaboration.md`. This is slice 1 (comment + board-read + guidance) of comment / subscribe / take; subscribe and take follow in their own PRs.

## The participation-model change (the core)

Before, an identity participated in a ticket iff it was the **assignee** OR had **authored any event** on it — and participants drive notifications (`UnreadTicketEvents` / `TicketParticipants`). That would make a one-shot commenter a permanent participant, re-introducing exactly the cross-ticket nudge noise #438 fixed. So the rule narrows: **authoring a `commented` event no longer confers participation.**

```
participant(ticket) = assignee
                    ∪ authors of NON-comment events   (chief via `created`, assignee via status)
                    ∪ subscribers                      (slice 2, not yet)
```

Both participation queries in `internal/store/ticket_events.go` now filter `kind != 'commented'`. Dead `InvolvedTicketIDs` (no callers) was removed rather than left to diverge.

## Board read is global, not identity-scoped

Unlike the write verbs, `ticket_list` returns the whole board and does **not** require a session — `source_session_id` is optional and unused, and the CLI resolves the session best-effort (works with `ATTN_SESSION_ID` unset), mirroring `attn list` for sessions. It reuses the app's `ticketToProtocol` via a shared `ticketRows` helper, so the board renders identically from the broadcast feed and the agent read.

## Guidance, split by the access decision

`list` stays available to **every** agent (no code restriction), but the guidance push is **chief-oriented** (Victor's call):

- always-on `TicketAwarenessGuidance` (all agents): the **unrestricted** verb only — comment on a ticket you were handed the id for.
- the chief block: `attn ticket list` as the coordinator's board read, beside the existing `attn ticket inbox --watch` Monitor guidance.
- the skill's `tickets.md` reference: the how for both, framing list as a coordinator move.

## Surfaces & versioning

protocol (`ticket_comment`, `ticket_list` + results), daemon handlers, client methods, CLI (`-m`/`--message` for comment so flags compose around the id; a bare-comment-arg mistake now hints `-m`). ProtocolVersion 133 → 135 (frontend `PROTOCOL_VERSION` matched).

## Tests

- store: comment-exclusion in both participation queries; a bystander's comments yield 0 unread for them while the assignee still sees events.
- daemon: linchpin — X comments on Z's ticket → Z nudged, X not nudged about its own note, a *later* event on Z's ticket does **not** reach X (proves comment ≠ subscribe). Board read returns the full board with **no session** and with descriptions present; `--status` filter reaches the store query.
- CLI parse (flags before/after id, `-m` hint) + protocol decode for both commands.
- Full `go test ./internal/daemon` green (covers the app broadcast path the `ticketRows` refactor touches).

## Verified end-to-end

Against an isolated profile daemon: `attn ticket list` with no session shows the board; `--status`/`--json` behave; unassigned shows `-`; comment routes to the ticket's creator without subscribing the commenter; unknown id → clear error, exit 1. Profile torn down; prod daemon untouched.

This PR touches `app/` (the `PROTOCOL_VERSION` constant + additive `generated.ts`), so Frontend CI runs rather than skipping — expected.


figgyster (dismissed):
This looks safe to merge. The comment path keeps the participation model honest: drive-by notes notify the ticket owners without turning the commenter into a permanent observer, and the store/daemon tests pin the exact no-silent-subscribe behavior.

I ran the focused Go coverage locally (`go test ./internal/store ./internal/daemon ./internal/protocol ./cmd/attn`). GitHub’s Frontend job was still pending on my last check; no blocker from me assuming it finishes green.

figgyster (approved):
Still looks safe to merge. The new board-read slice fills the missing half of the feature: agents can now discover ticket ids before using the one-shot comment path, and it keeps the read intentionally global while the write remains a non-subscribing note.

Focused checks passed locally again (`go test ./internal/store ./internal/daemon ./internal/protocol ./cmd/attn ./internal/hooks`). GitHub’s Frontend job was still pending on my last check; no blocker from me assuming it finishes green.

PR #441, opened 2026-06-29
feat(tickets): agents can subscribe to / unsubscribe from any ticket

victorarias:
## What

Slice 2 of the agent ticket collaboration feature (after #440's comment + board-read foundation): **explicit, opt-in standing interest** in a ticket an agent doesn't own — the counterpart to a one-shot comment.

- **`attn ticket subscribe <ticket-id>`** opts the calling session into a ticket's notifications. While subscribed it is a **participant**: future activity nudges it and lands in its `attn ticket inbox`, and the first inbox after subscribing also delivers the ticket's **history** (subscribing does not advance the cursor — the same "join mid-stream sees the backlog" rule a fresh assignment follows).
- **`attn ticket unsubscribe <ticket-id>`** opts back out. Idempotent.

Plan: `docs/plans/2026-06-29-agent-ticket-collaboration.md`. Slice 3 (take) follows.

## Participation model

Subscriptions are the **third participation source**, beside assignment and non-comment authorship:

```
participant(ticket) = assignee
                    ∪ authors of NON-comment events
                    ∪ subscribers            ← this PR
```

Both `UnreadTicketEvents` (consume) and `TicketParticipants` (notify) gain a `ticket_subscriptions` UNION, so a subscriber is nudged and served exactly like any other participant.

## Surfaces

- **migration 58** `ticket_subscriptions` — mirror of `ticket_event_cursors` (PK `(identity, ticket_id)`, `CASCADE` on ticket delete; **no cursor column** — subscribing only adds to the participant set). Verified prod+dev DBs at `MAX(version)=57` before numbering, per the burned-version gotcha.
- **store**: `AddTicketSubscription` (validates the ticket exists → `ErrTicketNotFound`; idempotent re-subscribe), `RemoveTicketSubscription` (tolerant idempotent removal — no existence check, since "I'm not subscribed" is already true), `IsTicketSubscribed`. `SweepExpiredTickets` also purges subscriptions.
- **protocol**: `ticket_subscribe` / `ticket_unsubscribe` agent commands (+ results). Both handlers are **silent** — they append no event and change no board row, so nothing is notified; the subscriber picks up history on its next inbox. ProtocolVersion 135 → 136 (frontend matched).
- **client + CLI**: `attn ticket subscribe`/`unsubscribe <id> [--session] [--json]` via a shared `parseTicketIDArgs` (one id positional, flags interleavable on either side).
- **guidance**: `tickets.md` documents subscribe as the standing-interest opt-in (vs. comment's one-shot).

## Tests

- store: subscription makes an identity a participant both ways (its inbox delivers the backlog; `TicketParticipants` includes it), unsubscribe reverses both; subscribe idempotent + validates the id; unsubscribe tolerant of unknown/never-subscribed.
- daemon: full lifecycle — subscribe → a chief comment nudges the subscriber and its inbox delivers; unsubscribe → nudge and delivery stop. The trigger goes through the **synchronous** `commentOnTicket`, not the `net.Pipe` `callSetTicketStatus` (which returns before the async doorbell — `typeDoorbell` sleeps mid-delivery; caught while debugging an intermittent "not nudged").
- CLI parse + protocol decode for both commands.
- Full `go test ./internal/daemon` green.

## Verified end-to-end

Against an isolated profile daemon: migration 58 applies (`ticket_subscriptions` present); subscribe → watcher's inbox delivers the ticket **including the `created` event from before it subscribed** (history); unsubscribe → inbox empty after a new comment; re-unsubscribe exit 0 (idempotent); subscribe to unknown id exit 1. Profile torn down; prod untouched.

Touches `app/` (the `PROTOCOL_VERSION` constant + additive `generated.ts`), so Frontend CI runs rather than skipping — expected.


figgyster (approved):
This looks safe to merge. The subscription model is wired through the same participant paths for inbox consumption and notification fan-out, and the no-cursor choice preserves the “join mid-stream sees the backlog” rule instead of inventing a second timeline. Nice clean slice.

I ran the focused daemon/store/protocol/CLI tests plus the frontend production build locally; GitHub Frontend CI was still in progress on my last fetch, but the local build covered the touched generated app surface.

No blockers from me.

PR #442, opened 2026-06-29
feat(tickets): agents can take (claim/reassign) any ticket

victorarias:
Slice 3 of the agent ticket collaboration foundation (after comment+board-read #440 and subscribe #441). Completes the cross-ticket collaboration surface.

## What

`attn ticket take <ticket-id> [--confirm]` claims a ticket — the calling agent becomes its assignee.

- **Pick up unassigned work or hand work to yourself.** Take an unbound backlog ticket, or reassign one to the calling session.
- **Steal guard.** Taking a ticket already assigned to someone else requires `--confirm`; without it the daemon refuses and names the current assignee, so an agent can't silently take over a sibling's active work.
- **Self-take is a no-op.** Re-taking a ticket you already own short-circuits (no redundant `assigned` event), reported successfully so a retry is harmless.
- **Take delivers history.** Taking does not advance the cursor, so the taker's first `attn ticket inbox` catches it up on the ticket's backlog — the same freshly-assigned ⇒ deliver-from-start rule as subscribe.
- **Handover notice.** The displaced assignee is notified through the `assigned` event reaching it as a participant (it authored status events while working). A previous assignee that never reported isn't a participant and isn't nudged — acceptable, it was never visibly working. The handler stays behind the store's participation boundary (no special-casing).

## Protocol

New `ticket_take` command + `TicketTakeResult` (echoes `previous_assignee`). ProtocolVersion 136→137 (frontend + generated types in sync).

## Tests

- daemon take-over: refused without `--confirm`; with `--confirm` reassigns + nudges the displaced assignee + delivers history (deterministic — synchronous notify)
- daemon unassigned/self/unknown paths
- CLI parse (`parseTicketTakeArgs`, including `--confirm` interleave) + protocol decode
- CLI smoke-tested end-to-end against an isolated profile daemon (claim → refuse → confirm-reassign → board assignee → self-take → unknown)


figgyster (commented):
No blockers in the code. The `take` path keeps the right boundaries: takeover goes through the existing assignment/event machinery, self-take is idempotent, and the cursor behavior preserves backlog delivery for the new assignee.

I ran the focused Go checks locally:

```bash
go test ./internal/daemon ./internal/protocol ./cmd/attn ./internal/client
git diff --check origin/main...HEAD
```

Those passed. GitHub’s Frontend check was still pending after a few minutes, so I’m leaving this as a review comment rather than a green stamp until CI finishes. Annoying, but better than approval theater in a trench coat.

figgyster (approved):
CI is green now, and it’s the same head I checked locally. The ticket-take flow stays inside the existing assignment/event/cursor contracts, with tests covering the takeover guard and backlog delivery.

Approved.

PR #443, opened 2026-06-29
feat(nudge): pausable, visible, splice-safe ticket nudge countdown

victorarias:
## What & why

A ticket nudge used to fire the instant an idle agent had unread ticket activity — `typeDoorbell` types the prompt + Enter straight into the agent's PTY. The bug: *idle* doesn't mean the user is *absent*. If you were composing a prompt into that session, the doorbell text spliced onto your half-typed line and the Enter submitted the garbled combination.

This makes every ticket doorbell pass through a per-session **nudge countdown** (default 30s) that is **visible**, **pausable**, and **splice-safe**.

## Behavior

- **Countdown gate (default 30s).** Every ticket doorbell now arms a per-session countdown instead of injecting immediately. The actual PTY injection only happens when the timer fires and the session is still idle + unread.
- **Pause while you're looking at it.** The selected session never auto-fires — its countdown pauses and shows a **"deliver now"** button. Switch away and it resumes, so attn delivers it later.
- **Keystroke safety guard (the real anti-splice guarantee).** At fire time attn never doorbells a session that received a genuine user keystroke in the last ~3s. This is per-session, so it also protects a *second visible tile* you're typing into — which selection-based pause alone can't. If the guard blocks a fire mid-keystroke, it re-arms a short countdown rather than dropping the nudge.
- **Visible indicator, three modes**, on both the sidebar row and the tile overlay:
  - idle/waiting + unread, inactive → growing countdown bar
  - idle/waiting + unread, active (selected) → paused bar + "deliver now"
  - working + unread → quiet static "unread activity" marker (no countdown)

The notebook doorbell (`nudgeChiefOfStaff`) is intentionally untouched — this only routes the two *ticket* doorbell sites through the countdown.

## Shape

- **Daemon:** new `internal/daemon/nudge_countdown.go` owns the timers + the only real ticket PTY injection; `lastUserInputAt` (set in `handlePtyInput`) is the keystroke guard; selection pause/resume in `tilecontent.go`; unread tracking + drain-clear in `ticket_notify.go` / `ticket_read.go`.
- **Protocol:** two optional `Session` fields (`ticket_unread`, `nudge_fires_at`) + a `trigger_nudge` command; both fields added to `sessionsMatch` so changes rebroadcast; ProtocolVersion bumped.
- **Frontend:** `NudgeIndicator` renders the three modes on both surfaces; the bar animates client-side from `nudge_fires_at` (no per-tick broadcasts); `sendTriggerNudge` drilled through App.

## Tests

- `internal/daemon/nudge_countdown_test.go` — arm→fire after window, arm→active stays paused, switch-away resumes, switch-to pauses, leaving-idle cancels, unread-drain cancels, `trigger_nudge` fires immediately, teardown stops timers.
- `NudgeIndicator.test.tsx` — the three visual modes on both surfaces.
- Keystroke-guard write path covered through `handlePtyInput`.
- **Real-app regression** `real-app:scenario-nudge-trigger` (on-demand / serial-matrix, not CI): boots a real codex agent, makes it a ticket participant, asserts the paused gate *holds* (no doorbell), then clicks the real **"deliver now"** tile button and asserts the verbatim doorbell lands in the agent — the click→`trigger_nudge`→`handleTriggerNudge`→`typeDoorbell` path nothing else exercises. Ran green live in the dev app.

## Notes for review

- The sidebar "deliver now" button's onClick is a one-line identical `onTriggerNudge?.(...)` to the tile button; the real-app test live-fires the **tile** surface, the sidebar is statically covered.
- Accepted v1 gap: selection is sticky with no window-blur event, so an idle+unread *last-selected* session won't auto-fire while the app is backgrounded. Window-blur signal is a noted follow-up.

Plan: `docs/plans/2026-06-...` (wobbly-stirring-bentley).


figgyster (approved):
This looks safe to merge. The countdown owner is where it belongs — daemon-side, with the splice guard checked at delivery time — and the protocol/UI pieces line up cleanly.

I ran the focused daemon/protocol/ticket tests, the new nudge indicator test, harness script checks, and `git diff --check`; GitHub CI is green apart from the intentionally skipped Tauri job. No blockers from me.

PR #444, opened 2026-06-30
fix: keep a delegated agent a leaf, not a re-delegating chief

victorarias:
## The bug

A delegated agent delegated its own task onward to a third agent — it read attn's skill, applied the "delegate if you're the chief" framing to itself, and spun up a worker in a session the delegating user wasn't watching. Nothing told it that it is a *leaf*, not the chief.

## Root cause: an inference gap, plus a stale install

attn marks the **chief** with a concrete positive signal (`ATTN_CHIEF_GUIDANCE` env + an always-on system-prompt block). It marked a **leaf** with *nothing analogous* — for a non-chief delegation the initial prompt was the raw brief, byte-identical to an ordinary top-level session. The chief-vs-leaf distinction therefore rested on the agent *inferring* a role for which no positive signal was plumbed to it.

The concrete mechanism behind the actual incident was an **orphaned installed file**: `references/chief-of-staff.md` was retired from the source tree weeks ago (chief identity moved into the always-on system prompt), but `installAttnSkill` only ever wrote/overwrote files in the current embed and never deleted ones that fell out of the bundle — so the retired reference survived on an already-installed machine and kept teaching the old "delegate like the chief" framing.

## What shipped

1. **Delegate-time leaf line** (`internal/daemon/delegate.go`) — a terse leaf-identity preamble is prepended to *every* delegated agent's initial prompt (tracked or not, chief-delegated or not). It composes with the existing ticket-report contract and is the only role signal that reaches a leaf which never opens the skill.
2. **Installer orphan-pruning** (`internal/agent/attn_skill.go`) — `pruneOrphanedSkillFiles` removes anything under the installed skill dir that isn't in the current embed, so a retired reference can no longer outlive a later install. This is the direct fix for the incident's mechanism.
3. **Skill role front-door** (`SKILL.md`) — a "Confirm Your Role First" section ahead of the capability index, with a checkable observable for each of chief / delegated leaf / ordinary, so any agent that loads the skill self-classifies before reading anything about delegation.
4. **`delegated-agent.md` restructure** — split into "You Are A Leaf, Not A Coordinator" (universal) and "If Your Work Is Tracked, Report Your State" (conditional).
5. **`delegation.md`** — defers to the role check up front, and points a brief-writer at `tickets.md`'s fuller brief craft (a delegation brief *is* a ticket description). Also drops a dangling cross-reference to a non-shipped local skill.

## On not being "too strong"

The leverage is in the *signal*, not the *imperative*. Hardening prose ("don't" → "NEVER") does nothing about an antecedent the agent can't observe — it just adds variance (muzzles legitimate delegators or under-applies to leaves). A delegated agent is a full steerable session; if *its* user asks for a worker it should comply. So the consequent stays calm ("don't self-offload," not "never delegate") and the fix makes the antecedent observable instead. A hard tool-level block was deliberately not built — it can't see user intent and breaks the legitimate "user steering a delegated agent asks for a worker" case. Full reasoning in `docs/plans/2026-06-30-delegated-leaf-not-chief.md`.

## Verification

- `go build ./...`, `go vet ./...`, `gofmt -l` clean.
- `go test ./internal/agent/...` — incl. new `TestEnsureAttnClaudeSkillInstalledPrunesOrphanedFiles` (seeds a leftover `chief-of-staff.md` + orphan subdir, asserts both pruned).
- `go test ./internal/daemon/ -run 'TestDelegate|TestChiefOfStaffDelegate|TestOrdinaryDelegation|TestDelegatedFromChief'` — green.

Reviewable in three slices: the delegate-time line, the skill/installer change, and docs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-06-30]
+
+### Fixed
+- **A delegated agent stays focused on its own task instead of handing it off again.** A delegated agent could misread attn's guidance as a license to delegate onward — spinning up a *third* agent for one of its own subtasks, in a session you weren't watching. Now every delegated session is told up front that it's a worker, not a coordinator: it does the assigned work itself and uses its own internal helpers for subtasks, only creating a new visible attn agent if you explicitly ask it to. (attn also now cleans up stale guidance files from older installs that could keep teaching the old behavior.)
+
 ## [2026-06-29]
 
 ### Added
--- a/docs/plans/2026-06-30-delegated-leaf-not-chief.md
+++ b/docs/plans/2026-06-30-delegated-leaf-not-chief.md
@@ -1,0 +1,208 @@
+# Plan: stop a delegated (non-chief) agent from re-delegating
+
+Ticket: `delegated-agents-not-chief`. **Implemented.** Two instruments: a
+**delegate-time leaf line** (code, closes the zero-signal leaf) and a **skill
+role front-door + installer orphan-pruning fix** (prose + code, closes the
+read-the-skill misread). See "What actually shipped" below.
+
+## The bug
+
+A delegated agent delegated its own task onward to a third agent. It had read
+the attn skill and applied the "delegate if you're the chief" framing to itself.
+Nothing told it that it is a *leaf*, not the chief.
+
+## Root cause: an inference gap, not disobedience
+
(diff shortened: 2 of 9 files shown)


figgyster (approved):
This looks safe to merge. The fix hits both failure paths: delegated sessions now get an explicit leaf identity before their brief, and the skill installer stops letting retired guidance haunt already-installed machines like a very specific ghost.

I also like that the guidance stays role-aware instead of trying to ban delegation with a hammer. Local `go test ./...` and `git diff --check` passed, and the visible daemon CI is green. No blockers from me.

PR #445, opened 2026-06-30
fix(nudge): deliver the ticket nudge on demand in every state but approval

victorarias:
## What

Makes the manual **"deliver now"** ticket nudge available on demand. The explicit click now types the doorbell into the focused session in **every state except `pending_approval`** (where the doorbell's trailing Enter could answer a y/n approval prompt). The **automatic countdown is unchanged** — still idle-only.

Before, the click only worked when the session was idle/waiting/unknown, so a chief that rested in `working`/`launching`/`unknown` (common for a codex chief whose turn-end classifier can't find a transcript) showed a dead, non-clickable marker. Now the session you're focused on always shows the clickable deliver-now chip, except at an approval prompt where it stays a non-clickable marker.

## Why this is safe

Verified live in the dev app: delivering a doorbell into a **working** codex doesn't garble or destructively interrupt it — codex's TUI queues it (`Messages to be submitted after next tool call`). `pending_approval` is the one held-back state.

## Changes

- **daemon** (`nudge_countdown.go`): `handleTriggerNudge` gate is now `isExplicitNudgeBlocked` (blocks `pending_approval` only). Auto path (`runNudgeDelivery`/`isIdleForNudge`) untouched.
- **frontend** (`NudgeIndicator.tsx`): `deriveNudgeMode` mirrors the daemon — focused unread session shows the clickable `paused` chip in every state but `pending_approval`.
- **drag guard** (from review): both deliver-now buttons now `stopPropagation` on pointerdown so a sloppy press inside the draggable pane header / session row can't start a leaf/session drag instead of delivering — matching the sibling rename button.
- CHANGELOG entry refined to the shipped behavior.

## Testing

- Daemon: `TestTriggerNudgeDeliversWhileWorking`, `TestTriggerNudgeSkipsPendingApproval` (+ existing nudge suite) — green.
- Frontend: `deriveNudgeMode` truth-table cases for working/unknown/pending_approval, and pointerdown-doesn't-bubble regression tests for both buttons — green. `tsc` clean.
- Live: clicked deliver-now on a **working** codex via the dev app; chip rendered + clicked, doorbell delivered and queued safely.
- Ran an adversarial multi-agent review over the diff; the one confirmed finding (pointerdown drag regression) is fixed here.


figgyster (approved):
This looks safe to merge. The manual nudge path now has the right split-brain behavior: explicit clicks are allowed everywhere except approval prompts, while the automatic countdown stays idle-only.

I also like that the UI mirrors the daemon gate instead of inventing its own weather system, and the pointerdown drag guard covers the sharp edge in both surfaces.

Local checks passed here:
- `go test ./internal/daemon -run 'TestTriggerNudge|TestNudge'`
- frontend Vitest suite (`127` files / `1218` tests passed, `5` skipped)
- `pnpm exec tsc --noEmit`

CI was still running on the Frontend job when I last fetched it, but the local principal checks are green. No blockers from me.

PR #448, opened 2026-07-01
attn delegate: per-delegation --model and --effort pins

victorarias:
## What

`attn delegate` accepts optional `--model <alias|full-id>` and `--effort <level>` flags that pin the spawned agent's model and reasoning effort **for that delegation only**. Omitted, nothing changes — the agent keeps its own defaults.

## How

The pins flow through the existing chief-model plumbing, generalized:

- **CLI → client → `DelegateMessage`** — new optional `model`/`effort` fields.
- **Daemon** — `delegate()` validates the pins against the resolved agent's new `HasModelPin`/`HasEffortPin` capabilities (plugins check `model_pin`/`effort_pin` keys) and fails fast — `agent "copilot" does not support --effort` — instead of silently dropping them. The pins ride `SpawnSessionMessage` into `handleSpawnSession`; a per-spawn model takes precedence over the `chief_model_<agent>` setting fallback.
- **Worker env** — `ATTN_MODEL` (renamed from the chief-only `ATTN_CHIEF_MODEL`; the daemon and wrapper ship in the same binary) and new `ATTN_EFFORT` carry the pins into the wrapper.
- **Drivers** — claude appends its native `--model` / `--effort` (low, medium, high, xhigh, max — verified against `claude --help`); codex appends `--model` / `-c model_reasoning_effort="<level>"` (its native config knob, per the Codex config reference).

Model and effort **values** pass through unvalidated by design: aliases, full ids, and future effort levels stay legal without an allowlist that rots. The agent CLI is the authority.

Protocol version 139 → 140 (new message fields). Skill docs (`references/delegation.md`), `delegate --help`, and CHANGELOG updated. No frontend delegate form exists, so this is CLI-only per the brief.

## Verification

Unit tests at every hop: CLI flag parsing, client message construction, daemon threading (backend receives `Model`/`Effort`), unsupported-agent rejection, and driver argv construction for both agents (plus a no-pins default-behavior test). Full `go test ./...` and frontend vitest suites green.

**Live smoke** on an isolated `modelpin` profile (fresh daemon from this branch, headless via wsctl):

- `attn delegate --agent claude --model sonnet --effort low` → spawned process argv contains `--model sonnet --effort low`, worker env has `ATTN_MODEL=sonnet` / `ATTN_EFFORT=low`, and the live session's statusline shows **Sonnet 5** (not the machine default) with the brief answered.
- `attn delegate --agent codex --effort high` → codex argv contains `-c model_reasoning_effort="high"`.
- `attn delegate --agent copilot --effort high` → rejected at delegate time: `agent "copilot" does not support --effort`.

## Notes

- If a delegated session's worker dies and it is later relaunched/resumed (app respawn, ticket Resume), the pin is not re-applied — the pin lives in the spawn request, not the session row. Same lifetime as the brief's initial prompt; called out here for visibility.

https://claude.ai/code/session_01HsGcmp3Y6yrCnzUTKMKBCX

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-07-01]
 
+### Added
+- **Pin a delegated agent's model and effort.** `attn delegate` accepts `--model <alias|id>` and `--effort <level>` to launch that one delegation with a specific model and reasoning effort (Claude's native `--effort` levels, Codex's `model_reasoning_effort`), instead of silently inheriting the machine's default. Omit them and nothing changes; agents without a native mechanism (e.g. Copilot) reject the flags up front.
+
 ### Removed
 - **The automated review loop is gone.** The standalone review-loop feature — the SDK-managed iterating loop, its session sidebar bar, `attn review-loop` CLI, and related settings — has been removed. Diff comments, PR reviews, and the workflow engine are unaffected.
 
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -165,7 +165,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '139';
+export const PROTOCOL_VERSION = '140';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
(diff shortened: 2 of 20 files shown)


figgyster (approved):
This is safe to merge. The new pins are carried through the protocol/daemon/worker boundary without turning model or effort values into a brittle allowlist, and unsupported agents fail loudly instead of pretending the flag worked — exactly the right kind of boring.

I ran `go test ./...` and `pnpm test` locally. GitHub Frontend CI was still pending when I checked, but the matching local frontend suite passed.

PR #451, opened 2026-07-02
fix(tickets): make the ticket comment box submittable

victorarias:
## Problem

Victor (2026-07-02): *"The ticket UI comment box has no 'add' button, nor does it support cmd+return, so it's impossible to add comments from it."*

Reproduced in the dev app. Two root causes:

1. **The "Add comment" button was invisible.** It's styled with `background: var(--color-accent)` / `color: var(--color-bg-panel)`, but `--color-accent` was **never defined** anywhere in `App.css`. With the token undefined, the background/border collapse to transparent and the text takes the panel background color → the button paints dark-on-dark. It was in the DOM and correctly wired to the `ticket_add_comment` route (the same `store.AddTicketComment` path as `attn ticket comment`), just unseeable. `--color-error` was undefined too, leaving the crashed badge and action-error text uncolored.
2. **No keyboard shortcut** — the textarea had no key handler at all.

## Fix

- **Define the missing tokens** `--color-accent: #5b8cff` and `--color-error: #ef4444` in `App.css :root`. These are the exact values 11+ sites already reference as `var(--token, fallback)`, so anything using the fallback is unchanged; the bare `var(--token)` references (this button, `.ticket-status-*`, the error text) now render as intended. Theme-invariant, so one `:root` definition cascades to all theme blocks with zero regression.
- **Add ⌘Return / Ctrl+Return** to the comment textarea, sharing a single `submitComment` with the button (trim, disabled-when-empty + in-flight guards, clear-on-success, error surfacing). Plain Enter still inserts a newline. Placeholder hints `(⌘⏎ to send)`.

Reuses the existing comment route — no new plumbing.

## Verification

- **Vitest** (all green, 1225 total): new coverage for ⌘/Ctrl+Return submit, plain-Enter no-op, disabled-until-non-empty (+ shortcut no-op on whitespace), and the in-flight disable/relabel guard.
- **Manual e2e in the dev app** on a real ticket: the button now renders as a visible blue "Add comment" button (before the fix an app-owned screenshot had **0** colorful pixels — invisible; after, the button is clearly blue). Posting via the button lands the comment in the ticket activity, confirmed by a fresh `GetTicket` after close+reopen (defeats optimistic render / broadcast lag).

<sub>Opened by Claude on behalf of Victor.</sub>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-07-02]
+
+### Fixed
+- **You can post comments from the ticket detail panel again.** The panel's comment box had no usable way to submit — the "Add comment" button was rendered in an undefined accent color that painted it invisibly against the panel, and there was no keyboard shortcut — so a comment typed in the app couldn't be posted (only the `attn ticket comment` CLI worked). The button is now visible, and **⌘Return** (Ctrl+Return off-mac) posts the comment without leaving the field; plain Enter still adds a newline. The button stays disabled until you've typed something and while a post is in flight, clears on success, and shows an error if the post fails. The same missing accent/error colors also left ticket status accents and error text uncolored throughout the panel; those now render as intended.
+
 ## [2026-07-01]
 
 ### Added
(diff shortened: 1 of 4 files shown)


figgyster (approved):
This looks safe to merge. The fix keeps the existing ticket comment route, makes the hidden submit control visible by defining the missing tokens, and adds the expected ⌘/Ctrl+Return path without stealing plain Enter from multiline comments.

I ran the full frontend Vitest suite locally: 127 files, 1220 passed / 5 skipped. GitHub's Frontend check was still pending when I last looked, but no blockers from me.

PR #452, opened 2026-07-02
CI: shard frontend E2E + cache browsers; frontend test suite audit report

victorarias:
## What

Restructures the Frontend CI pipeline per the measured audit in `docs/fe-test-suite-audit.md` (included in this PR):

- **E2E moves to its own job, sharded 3 ways** (`playwright test --shard=N/3`). Each shard is a separate runner with its own throwaway daemon + Vite server, so the shared-daemon constraint that forces `workers: 1` in `playwright.config.ts` still holds *within* each shard — no test or product changes.
- **Playwright browsers cached** (keyed on Playwright version), chromium only — the config defines no other projects. Saves ~40s/run once warm.
- **`frontend-e2e-gate`** exposes a single stable **Frontend E2E** check name over the matrix (green when shards pass or were path-filter-skipped).
- Unit job (**Frontend**, unchanged name — it's in the required-checks ruleset) drops the Go setup and attn binary build; only e2e needs the binary. The `src-tauri/binaries` copy was unused in CI (typecheck runs before it existed; e2e uses `ATTN_E2E_BIN`).

## Why

CI latency. Measured on the last 3 main runs: Frontend gate ~6.5min, of which E2E is ~4min (154 tests fully serial) + ~40s browser install; vitest is only ~50s. Expected gate after this PR: **~2–2.5min** (unit ~1.5–2min ∥ shards ~2min each).

## Verification

- Shard split verified locally: 62/43/50 tests across 9/7/9 files; shard 1/3 passes in 45s (`ATTN_PROFILE=feaudit pnpm exec playwright test --shard=1/3`).
- This PR touches `ci.yml`, so all new jobs run on this PR itself — the real proof is below in checks.
- Full measurement methodology + numbers: `docs/fe-test-suite-audit.md`.

## Follow-up for Victor

- The ruleset currently requires `Daemon`, `Tauri`, `Frontend`. **Add `Frontend E2E`** (the gate job) to keep e2e blocking merges — until then a red shard doesn't block.
- Audit report also recommends (not in this PR): delete 31 low-value vitest tests, fake-timer the 4 sleep-heavy files.

---
🤖 Opened by Claude on behalf of Victor

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,11 +91,6 @@ jobs:
         with:
           node-version: '20'
 
-      - name: Set up Go
-        uses: actions/setup-go@v5
-        with:
-          go-version: '1.22'
-
       - name: Install pnpm
         uses: pnpm/action-setup@v4
         with:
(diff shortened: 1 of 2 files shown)


figgyster (approved):
This looks safe to merge. Splitting E2E into its own sharded job keeps the shared-daemon constraint local to each runner, and the stable `Frontend E2E` gate is the right branch-protection handle.

I also checked the workflow shape with `actionlint`; the live PR checks are green. No blockers from me — just remember to add `Frontend E2E` to the required checks, otherwise the shards are only decorative guardrails with a clipboard.

PR #453, opened 2026-07-02
Add font size settings with an independent taskboard scale

harleenquinzell:
## What

Adds user-facing font size settings, split into two independent controls (Settings → General → Appearance → Font Size):

- **App** — exposes the existing `uiScale` setting in the UI. Until now it was only adjustable through the ⌘+ / ⌘− / ⌘0 shortcuts, with no visible current value.
- **Taskboard** — a new `ticketBoardScale` setting that scales fonts on the ticket board and ticket detail surfaces independently of the rest of the app. It shows "Match app" and follows the app scale until you step it; a **Match app** button reverts the override.

## How

- **CSS**: a new `--ticket-board-scale` variable defaults to `var(--ui-scale)`; the `.tb-panel` and `.ticket-detail-panel` roots re-derive the `--font-size-*` tokens from it, so every existing token-based rule on those surfaces picks up the board scale with no per-rule changes.
- **Frontend**: `useTicketBoardScale` (modeled on `useUIScale`, same 0.7–1.5 range) syncs from daemon settings, applies/removes the CSS variable on the document root, and persists on user action only — `null`/empty means "match app". Wired in `App.tsx`, surfaced in `SettingsModal`.
- **Daemon**: `ticketBoardScale` added to the settings whitelist, validated like `uiScale` with empty allowed (validation otherwise rejects unknown keys). No protocol change — settings ride the existing string map.

## Verification

- Full frontend suite (1231 tests, including new hook + modal tests) and `go test ./internal/daemon` pass.
- Driven end-to-end with a real daemon + browser (Playwright): stepping App to 110% updates the whole UI and leaves the board on "Match app"; stepping Taskboard starts from the app scale; values clamp at 70%/150%; board title renders at 24px × board scale (31.2px @130%) while app chrome stays at ×1.1; both settings survive a page reload (daemon persistence); "Match app" reverts and its empty value also survives reload.


figgyster (dismissed):
This looks safe to merge. The app/taskboard split uses the existing settings channel cleanly, and the CSS-token approach keeps the board/detail override scoped instead of turning every text rule into a tiny crime scene.

Focused checks passed here:
- `pnpm exec vitest run src/hooks/useTicketBoardScale.test.tsx src/components/SettingsModal.test.tsx`
- `go test ./internal/daemon`

No blockers from me.

figgyster (approved):
Re-checked the rebased head. The PR patch is still the font-scale work I signed off on, and the new base already carries the ticket-detail comment fix, so no new blocker wandered in wearing a different coat.

Focused checks passed again:
- `pnpm exec vitest run src/hooks/useTicketBoardScale.test.tsx src/components/SettingsModal.test.tsx`
- `go test ./internal/daemon`

No blockers from me.

PR #454, opened 2026-07-02
Fix: reconcile classifier runs hermetically and failures surface the real error

victorarias:
# Fix: reconciliation classifier runs hermetically and failures surface the real error

The orphaned-ticket classifier's first production fire (2026-07-02, ticket `arch-review`) failed with only **"Reason: classifier run failed: headless agent keeper tools failed"** — a keyword bucket, with the real error discarded and nothing in the daemon log.

## Root cause (verified empirically)

The native-tools headless path (reconcile classifier + keeper) did not pass `--strict-mcp-config`. `--setting-sources ""` does **not** stop the user's claude.ai account MCP connectors from attaching — a probe of the exact classifier argv loaded four of them (`claude.ai Slack` in `needs-auth`, Gmail/Drive/Calendar `pending`). A slow or broken connector can sink a run that needs no MCP at all, and its "mcp"/"server" vocabulary is what `classifyHeadlessFailure` buckets as "keeper tools failed". `--strict-mcp-config` with no `--mcp-config` loads zero servers (also verified).

The raw output of the failed run was never persisted anywhere: `runHeadlessCommand` kept only the bucket, and the reconcile path only logged runs with cost > 0.

## Changes

1. **Hermetic runs** — `claudeHeadlessArgs` (native path) now passes `--strict-mcp-config`. E2E-verified in the spawned classifier's argv, and in the classifier's own init event: `"mcp_servers":[]`.
2. **Failure transparency** — new `HeadlessTaskResult.FailureOutput`: a bounded raw tail of the failed child's stderr+stdout (stderr first), led by the run's own result-event error text when parseable. The rule-7 comment now carries the exec error plus both bounded ends of that output (middle-out truncation — the human-readable error lives in the *trailing* result event); every classifier failure is also logged to the daemon log. The error string itself stays free of child output (keeper surfaces must not echo workspace content), so the existing no-leak invariant holds; the bucket wording is renamed to the neutral "headless agent MCP tool server failed".
3. **Prompt: in-session re-scoping wins** — the classifier is told the brief is the *starting* definition of done and explicit in-session re-scoping/authorization by the user supersedes its wording (the first fire judged `orphan-doc` "partial: scope violation" for work Victor had explicitly authorized in-session and merged himself).
4. **wsctl robustness** (dev script, found while e2e-testing): `SetReadLimit` (the 32 KiB library default dies on fat PR-detail frames and masquerades as a daemon hang), 30s spawn wait, `WSCTL_TRACE=1` frame tracing (makes `command_error` replies visible instead of opaque timeouts).

## Verification

- `make test` (1928 tests) and `make test-frontend` green; new unit tests pin the hermetic argv, `FailureOutput` capture/bounding, and the failure-comment content.
- E2E on an isolated profile daemon built from this branch (the dev profile was being concurrently reinstalled by another agent session, which had silently reverted the first dev-app pass to pre-fix code):
  - **(a) hermetic verdict**: with a deliberately broken user-scope MCP server configured (`claude mcp add -s user e2e-broken-mcp /usr/bin/false`), closing a session with an open bound ticket produced a real verdict — `done (high)` with cited transcript-line evidence, $0.12/4 turns — and the live classifier argv contained `--strict-mcp-config`.
  - **(b) failure transparency**: with `ATTN_TICKET_RECONCILE_MODEL=definitely-not-a-real-model`, the rule-7 comment carries the actual error (".../definitely-not-a-real-model. It may not exist or you may not have access to it...") instead of a bucket, and the same detail lands in the daemon log.
- The dev-profile pass (pre-fix daemon, by accident) doubled as a control: it showed a broken connector doesn't *deterministically* kill a run — consistent with the production incident where two runs 23s apart split success/failure.

## Follow-ups (not in this PR)

- The verdict/failure comment renders as a plain ticket comment; no frontend special-casing exists or was needed.


figgyster (dismissed):
This looks safe to merge. The classifier path is now properly hermetic, and the failure-output split keeps the raw diagnostic available only where the reconcile surface intentionally opts in — exactly the right line to draw.

`make test` is green here (1922 tests, 8 skipped). `make test-frontend` could not run in this checkout because `app/node_modules` is missing, and GitHub currently reports no checks for this branch. No blockers from me.

figgyster (approved):
Re-approved the current head after the merge from `main`. The PR patch still carries the same classifier fix, the changelog conflict landed cleanly, and CI is green now.

`make test` is green here again (1922 tests, 8 skipped). No blockers.

PR #455, opened 2026-07-02
feat(tasks): persist the durable task runner in SQLite (PR1/4)

victorarias:
First slice of the **background-task notifications** initiative (plan: `docs/plans/2026-07-02-bg-task-notifications.md`). Moves the durable task runner off one-JSON-file-per-task under the notebook root and onto a SQLite table in the profile DB.

## Why
Storage under the notebook root is why the runner is *disabled without a notebook root*. Putting tasks in the global DB dissolves that coupling — which is what later lets the reconcile classifier become an ordinary task kind (PR3) — and gives the coming notifications surface a queryable store. (The narration plans original "no SQLite" call was a defensive workaround for a burned-migration hazard that no longer applies; migrations flow cleanly to `{60}` today.)

## What
- **migration `{61}`**: `tasks` table
- **store**: `TaskRecord` + CRUD / recover-running (`internal/store/tasks.go`)
- **tasks**: a `Store` seam — an exported interface with a `FileStore` wrapper over the *untouched* file store, so the battle-tested store and its identity tests are undisturbed; the runner now takes an injected `Store`
- **daemon**: SQLite adapter mapping `Task`↔`store.TaskRecord` (`internal/daemon/task_store.go`); the single-instance lock is extracted to shared dir-lock helpers and relocated from the notebook tasks dir to the profile data dir; a one-time import of any legacy on-disk records (drops unparseable, retires the dir); `startCompactRunner` injects the store

## Scope / safety
- **Behavior-preserving**: the runners enable gate stays keyed on the notebook root in this PR — dropping it belongs in PR3 with reconcile, where narrations real root-dependence gets handled.
- **No protocol change** (the `NotebookTask` projection is unchanged), so no version bump and no changelog entry (internal refactor).

## Verification
- `go test ./internal/tasks ./internal/store ./internal/daemon` green; `internal/tasks` `-race` clean; full build.
- **Real-app smoke** on a throwaway profile seeded with a copy of the real v60 dev DB: migration `{61}` applied cleanly (v60→v61, table created, 20 tickets intact); daemon startup ran the import-once conversion (`tasks migrate: imported 1 legacy task(s)`, unparseable dropped, dir retired); the live SQLite-backed runner executed 3 real `narrate_workspace` tasks to `done`. Source dev DB untouched.

_Opened by Claude on behalf of Victor._

### DIFF
--- a/docs/plans/2026-07-02-bg-task-notifications.md
+++ b/docs/plans/2026-07-02-bg-task-notifications.md
@@ -1,0 +1,268 @@
+# Plan: Background-task errors → Notifications + Tasks-in-Settings
+
+## Goal
+
+Background/headless work (the keeper's compaction, session summaries, workspace
+narration, and the orphaned-ticket reconcile classifier) fails too quietly. Its
+errors are parked on the **notebook screen** — the wrong home, and one the user
+rarely opens. Move that visibility to a new **global, daemon-level Notifications
+surface** with read/unread state and a click-through error dialog, and give the
+tasks themselves a management home in **Settings**. Along the way, converge the
+one-off reconcile scheduler onto the durable task runner so there is a single
+task substrate and a single error producer.
+
+## Scope
+
+**In:**
+- Migrate the durable task runner's persistence from per-file JSON
+  (`<notebookRoot>/.attn/tasks/*.json`) to a **SQLite table** in the profile DB
+  (`~/.attn[-profile]/attn.db`); convert existing tasks once.
+- Grow the runner from a strict single worker to **bounded, per-kind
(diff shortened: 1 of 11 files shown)


figgyster (approved):
This looks safe to merge. Moving the durable runner behind a store seam keeps `internal/tasks` leaf-shaped, the SQLite adapter preserves the existing task semantics, and the legacy import path is narrow enough for this first slice.

I caught the initial gofmt CI failure, but Victor's follow-up fixed it; local `go test ./internal/tasks ./internal/store ./internal/daemon` passes here and the visible CI/preflight checks are green now. No blockers from me.

PR #456, opened 2026-07-02
feat(tasks): bounded per-kind concurrency in the durable runner

victorarias:
## What

PR2 of the background-task notifications initiative (plan: `docs/plans/2026-07-02-bg-task-notifications.md`). Grows the durable task runner from a single run-one-at-a-time worker into a dispatch loop with **bounded per-kind concurrency**.

Different kinds now run in parallel; a kind stays serialized with itself at the default cap of 1 (behavior-preserving for each individual kind). PR3 will register the reconcile kind with cap 2.

## How

- **`executor` gains a `limit`**; new `ExecutorConfig{Timeout, MaxConcurrent}` + `RegisterWith`. `Register`/`RegisterWithTimeout` delegate with the default cap of 1.
- **Runner state:** `run *activeRun` → `runs map[id]*activeRun` + `inflight map[kind]int`, both guarded by `mu`. A per-kind slot is reserved under `mu`, then the claim (state=running) is persisted under `ioMu`. `ioMu` stays the outer lock and no store I/O ever runs under `mu`.
- **`dispatch`** claims every eligible task whose kind has a free slot (earliest-scheduled first) and launches each as its own goroutine; the loop drains until a pass can place nothing more, then sleeps on nudge/tick. A finishing run frees its slot and re-nudges.
- **`Stop`** reordered for detached runs: close done → wait for the loop to exit → `cancelAll` (fence + join every in-flight run, honoring each commit fence).

## Invariants preserved

Commit fence, `Requeued` coalescing, capped-exponential backoff/dead, orphan recovery, single-instance lock, and Cancel-blocks-until-the-terminal-record-is-written. No protocol or frontend change.

## Tests

New: cross-kind concurrency, per-kind cap serialization, configured cap 2, and `Stop` draining multiple in-flight runs. Existing suite + new tests **race-clean at `-count=10`**; full `go build ./...`, `go vet`, and daemon task/keeper/narration tests green.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -63,6 +63,19 @@ pnpm run e2e
 - Diagnose root cause before fixing symptoms.
 - Do not remove requested functionality without explicit approval.
 
+## Live-App Testing Is Required
+
+Automated tests are necessary but not sufficient. Every PR must be exercised in the **live app** (a running daemon/app built from the branch) before it is called done or merged, with exactly two exemptions:
+
+- the change is **trivial** (e.g. a comment, a doc, a rename, a log-string tweak), or
+- the change is one you have **extremely high confidence** in without a live-app test, and you can say concretely why (for example: a pure, well-isolated function fully covered by unit tests, with no daemon-process, protocol, timing, or UI surface).
+
+A large automated suite — even a race-clean concurrency suite — is not by itself one of these exemptions. If the change touches the daemon process lifecycle, protocol, PTY, background runners, or any UI, it needs live-app verification.
+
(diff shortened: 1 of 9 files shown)


figgyster (dismissed):
This is safe to merge. The runner keeps persistence serialized while letting different task kinds breathe at the same time, and the new tests hit the important contract edges: cross-kind parallelism, same-kind caps, cap=2, and multi-run Stop draining.

I also ran `git diff --check`, the targeted daemon/tasks tests, and `go test -race ./internal/tasks -count=10`; GitHub's daemon checks are green too. No blockers from me.

figgyster (approved):
Still safe to merge. The follow-up only adds the live-app testing rule to the repo guidance and records the actual non-prod smoke for this runner change; the implementation diff is unchanged from the pass I already checked.

I re-ran `git diff --check`, `go test -race ./internal/tasks -count=3`, and the visible daemon checks are green. No blockers from me.

PR #457, opened 2026-07-02
Cap context window for chief-of-staff and headless runs

victorarias:
## What

Two new settings under **Settings → Agents → "Context window caps"** set the token threshold at which auto-compaction triggers, instead of waiting for the model's full window:

- **`chief_context_window_cap`** — the chief-of-staff session
- **`headless_context_window_cap`** — headless runs (keeper narration, ticket reconciliation, workflow subagents)

Both default to **128,000 tokens**, both configurable via the existing settings store + Settings UI.

## Why

- The **chief** is event-driven and its durable state lives on disk (the Notebook), so a smaller window makes every cache-cold wake cheaper — compaction here is nearly free.
- **Headless** runs are one-shot and cache-cold by construction; one that balloons past the cap is a bug to fix, not a cost to accommodate.
- **Regular delegated interactive sessions are deliberately never capped** — their in-context state is expensive to lose, which is the whole point of leaving them alone.

## Mechanism

The knob is each agent's own **compaction-trigger** setting, not a hard window limit (it can only *reduce* the effective window; a value above the model's real limit is ignored by the agent):

- **Claude**: `CLAUDE_CODE_AUTO_COMPACT_WINDOW` env var
- **Codex**: `model_auto_compact_token_limit` config override (`-c model_auto_compact_token_limit=<n>`)

Codex was in scope only because it has a trivially symmetric knob — it does.

### Chief relay (env, no protocol bump)

`chief_context_window_cap` setting → `SpawnOptions.ChiefContextWindowCap` → worker exports `ATTN_CHIEF_AUTO_COMPACT_WINDOW` → wrapper reads it into `SpawnOpts.AutoCompactWindow` → driver `BuildEnv`/`BuildCommand` emits the knob. Gated on the chief launch (`NotebookRoot != ""`) so delegated agents never inherit it — mirrors the existing Model/Effort/AutoApprove relay.

### Headless (process-global)

A package-global `atomic.Int64` (`agentdriver.SetHeadlessContextWindowCap`) that the daemon sets from the setting at startup and on every change; the headless spawn seam reads it and injects the knob. Chosen over request-threading specifically to stay refactor-proof against the in-flight background-task restructuring (`explore/bg-task-errors`) that is moving those call sites.

## Testing

- **Unit**: `internal/agent/context_window_cap_test.go` (chief emits / delegated never / chief-with-0 emits nothing; Codex arg builder; headless env + Codex headless args; negative-clamp) + daemon setting validation, resolution/gating, and apply-headless tests + 5 `SettingsModal` frontend tests.
- **Manual, `ps eww` on an isolated throwaway profile**: the **chief** claude process carries `CLAUDE_CODE_AUTO_COMPACT_WINDOW=<configured>`; a plain **delegated** agent's process does **not**. Full relay chain (setting → worker export → wrapper readback → driver emission) confirmed on the live process env.

## Non-goals

- Delegated interactive agents (intentionally uncapped).
- No protocol/codegen change — these are plain string settings in the generic settings store.


figgyster (dismissed):
This is close, but one launch path slips through the new cap: runtime reload/respawn rebuilds `ptybackend.SpawnOptions` in `buildReloadSpawnOptions`, and that path never sets `ChiefContextWindowCap` for the chief session.

That means a freshly spawned chief gets `ATTN_CHIEF_AUTO_COMPACT_WINDOW` → `CLAUDE_CODE_AUTO_COMPACT_WINDOW` / Codex config as advertised, but a chief relaunched through the reload path goes back to uncapped. That cuts against the user-facing contract here — the chief is exactly the cache-cold/event-driven process this setting is supposed to keep cheap.

Required before merge:
- carry `ChiefContextWindowCap: d.chiefContextWindowCap(protocol.Deref(session.ChiefOfStaff))` (or the equivalent stored-role check) through `buildReloadSpawnOptions`, and add a focused regression test for a chief reload preserving the cap while ordinary sessions stay uncapped.

I ran the relevant Go packages and the frontend test suite locally; they pass, but this reload seam is still a gap hiding under the floorboards.

figgyster (dismissed):
Still blocked on the same chief reload seam.

The latest head adds the user-facing cap settings and the fresh-spawn/headless plumbing, but `buildReloadSpawnOptions` still reconstructs `ptybackend.SpawnOptions` without `ChiefContextWindowCap`. A chief that gets relaunched through runtime reload/respawn will therefore come back uncapped, even though a fresh chief launch is capped. That breaks the contract this PR is adding for the chief path.

Required before merge:
- carry the chief cap through `buildReloadSpawnOptions` for chief sessions, while keeping ordinary/delegated sessions uncapped
- add a focused regression test for that reload path

I reran the relevant Go packages (`./internal/agent ./internal/daemon ./internal/ptybackend`) and `git diff --check`; they pass. The bug is just still hiding in the reload alley.

figgyster (approved):
This fixes the reload gap. `buildReloadSpawnOptions` now re-attaches the chief cap from the persisted chief role, and the regression test covers the two important edges: chief reload stays capped, ordinary reload stays uncapped.

I reran the focused Go packages and `git diff --check`, and the GitHub checks are green now. No blockers from me.

PR #458, opened 2026-07-03
feat(reconcile): run ticket reconciliation as a durable task kind (PR3)

victorarias:
## What

PR3 of the background-task errors initiative (`docs/plans/2026-07-02-bg-task-notifications.md`). Folds the orphaned-ticket **reconcile classifier** onto the durable task runner as a `reconcile` task kind, so there is one task substrate and one crash-recovery story. Backend + tests only — **no protocol/frontend change**.

Builds on PR1 (tasks → SQLite, #455) and PR2 (bounded per-kind concurrency, #456).

## The change

- **New `reconcileTaskExecutor`** (per-kind cap 2): reads the classifier inputs captured at enqueue time from `Task.Meta` (the owning session row is deleted moments after the death seam, so the async run can never re-read it), runs the headless classifier under the runner-owned timeout, applies the drop rule, and posts the 🩺 verdict/failure comment. One-shot — a classifier error is a posted failure note, not a runner retry; the only retryable path is a failed comment-post.
- **Session-end seam + periodic sweep now `Enqueue`** instead of spawning a goroutine. `ClaimTicketReconciliation` stays the trigger-time gate: it lights the board's orphan badge immediately and dedupes the `handlePTYExit`+`dropSessionRecord` double-fire (the fresher-inputs fire wins). The **sweep dedupes on the durable task record** (enqueue only when no `reconcile:<id>` task exists), which also recovers a claim whose enqueue was lost to a crash.
- **Deletes** the `ticketReconcileSem` semaphore (the runner owns per-kind concurrency) and `maybeRepairAbandonedReconcileClaim` (runner orphan-recovery re-runs a mid-flight death; the sweep re-enqueues an abandoned claim — and posts a *real* verdict instead of the old generic "interrupted" note).
- **Drops the runner's notebook-root enable gate**: the runner is enabled whenever the store exists, so reconcile runs regardless of notebook config. The notebook-scoped executors (compact/summarize/narrate) keep their own no-op-when-no-notebook guards.

## Decision: `reconciled_at` is kept, not retired

The plan speculated PR3 could retire the `reconciled_at` column. It can't cheaply: non-terminal + `reconciled_at` set drives the **user-visible orphan badge** on the board (`ticketOrphan.ts`, `ticket_board.go`, `main.tsp`). Retiring it is a protocol/frontend change, out of scope here. So the column stays; the *durable task record* (not the column) becomes the dedup + crash-recovery mechanism. See the PR3 refinement note in the plan for the full reasoning and the two accepted micro-windows.

## Testing

- `go test ./internal/{daemon,tasks,store}` green; `tasks` race-clean. Reconcile/sweep/crash suites updated (seam + sweep tests now drive a real started runner; the two maybeRepair tests are replaced by task-based recovery + task-existence-dedup tests).
- **Live-app smoke** (throwaway profile, real daemon built from the branch, fast sweep/grace env): inserted an orphaned dead-owner ticket → the sweep discovered it after grace, claimed (`reconciled_at` stamped = badge), enqueued `reconcile:<id>` onto the SQLite-backed runner, and the registered executor deserialized its Meta inputs and posted the attn-authored 🩺 note (no-transcript path, no real `claude` spawn); task reached `done` (attempts=1, one-shot). Repeated sweep passes left exactly one comment / one task / an unchanged badge — the task record is the "already handled" ledger.

### DIFF
--- a/docs/plans/2026-07-02-bg-task-notifications.md
+++ b/docs/plans/2026-07-02-bg-task-notifications.md
@@ -215,13 +215,62 @@ Isolate this in its own PR so the invariant change is reviewed alone.
       `runner.Stop()`/`cancelAll` is NOT reachable in prod (nothing calls
       `Daemon.Stop()` outside the test harness — the daemon is killed and relies on
       orphan recovery), so that path stays covered by `TestStopDrainsConcurrentInFlightRuns`.
-- [ ] **PR3 — Reconcile → `reconcile` task kind.** *Land after PR #454 merges;
+- [x] **PR3 — Reconcile → `reconcile` task kind.** *Land after PR #454 merges;
       rebase over main and build on its classifier body — do not port the old
       one.* Register the executor (body = classifier incl. drop-rule/🩺 comment);
       enqueue on session-end and from the sweep (sweep enqueues, no longer runs
-      inline); per-kind cap 2; delete the bespoke goroutine + semaphore + sweep
-      ticker and `maybeRepairAbandonedReconcileClaim` (runner orphan-recovery
-      replaces it); likely retire the `reconciled_at` column. Backend + tests.
+      inline); per-kind cap 2; delete the semaphore + the bespoke `go
+      d.runTicketReconciliation` goroutines + `maybeRepairAbandonedReconcileClaim`
+      (runner orphan-recovery replaces it). Backend + tests.
+
+      **Refinement vs the original plan (settled during implementation):**
+      - **Keep `reconciled_at`, do NOT retire it.** It is not merely an internal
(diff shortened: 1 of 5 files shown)


figgyster (approved):
This looks safe to merge. Moving reconciliation onto the durable runner is the right consolidation: the inputs are captured before the session row disappears, the sweep has a real task-record ledger for crash gaps, and the old semaphore/repair path is gone without changing the board/protocol contract.

I checked the runner/seam/sweep paths and ran `go test ./internal/daemon ./internal/tasks ./internal/store`, `go test -race ./internal/tasks`, and `git diff --check`; GitHub CI is green too. No blockers from me.

PR #464, opened 2026-07-04
Per-agent chief-of-staff reasoning-effort override (+ settings bridge actions)

victorarias:
## Summary

- **`chief_effort_<agent>` setting** — the effort analog of `chief_model_<agent>`. A chief-of-staff launch with no per-spawn pin falls back to it; delegation `--effort` pins (PR #448) still win. Passed through the agent's native mechanism (Claude `--effort`, Codex `model_reasoning_effort`). No protocol change: settings ride the generic key/value map and the spawn message already carries `effort`.
- **Settings UI** — the "Chief-of-staff model" section becomes "Chief-of-staff model & effort", with a per-agent effort `<select>` (Claude: low/medium/high/xhigh/max; Codex: minimal/low/medium/high/xhigh; first option "Agent default") committing immediately on change. The keep-row-visible rule now fires on either saved override.
- **Harness: `settings_get_state` / `settings_select_section` bridge actions** — the Settings modal was not drivable by the UI-automation bridge (openable only), which blocked live verification of specific sections. SettingsModal now publishes a module-level automation handle (same pattern as grid automation); `settings_select_section` reuses the exact nav-click code path and re-reads the handle after settling (the modal re-registers it every render, so a captured handle reports pre-selection state — covered by a regression test).

## Testing

- `go test ./internal/daemon` (full package) — new `TestDaemon_ChiefLaunchEffort` + validator cases.
- `pnpm exec vitest run src/components/SettingsModal.test.tsx` — 44 tests incl. 3 effort-UI tests and 6 automation-handle tests; `tsc --noEmit` clean.
- **Live on the dev profile**: set `chief_effort_claude=high` over the daemon WS and spawned a chief session — the real claude process launched with `--effort high`; non-chief spawns unaffected. Drove the packaged app via the new bridge actions: opened Settings, selected the Agents section, DOM-screenshot confirmed the new effort selects render; unknown section ids are rejected with the valid-id list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-07-04]
+
+### Added
+- **Pin the chief-of-staff's reasoning effort, per agent.** Settings → Agents → "Chief-of-staff model & effort" now has an effort selector next to each agent's model override (Claude: low, medium, high, xhigh, max; Codex: minimal, low, medium, high, xhigh). Leave it on "Agent default" to use the agent's own default. Only chief-of-staff launches are affected — regular sessions are unaffected.
+
 ## [2026-07-03]
 
 ### Added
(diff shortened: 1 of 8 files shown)


figgyster (approved):
This looks safe to merge. The chief-effort setting follows the existing chief-model path without changing the protocol shape, and the settings bridge action re-reads the live automation handle after section changes — exactly the kind of tiny stale-closure trap that would otherwise haunt packaged-app tests like a hallway draft.

I ran `git diff --check`, `go test ./internal/daemon`, `pnpm exec vitest run src/components/SettingsModal.test.tsx`, and `pnpm exec tsc --noEmit`; GitHub CI is green too. No blockers from me.

PR #465, opened 2026-07-04
feat: Present spine — manifest, store, protocol, daemon + attn present CLI (PR 1/4)

victorarias:
First PR of the Present chunk-1 train ([plan](docs/plans/2026-07-04-present-spine.md), [vision](docs/vision/present.md)): the Go spine, end to end over the wire but with no UI yet.

## What's in

- **`internal/present`** — manifest v0 parse + strict validation (unknown keys rejected, path hygiene) and SHA pinning via `git rev-parse`; quoted feedback markdown rendering (`RenderFeedback`) that fetches referenced lines from git at the round's pinned SHA and gracefully omits the quote when it can't.
- **Store (migration 63)** — `presentations` / `presentation_rounds` / `presentation_comments` tables + CRUD; round seq assignment is transactional; submit is transactional (comments + `submitted_at`) and rejects double-submit.
- **Protocol (version 143 → 144)** — unix commands `present_open` / `present_feedback`; WS `get_presentations` / `get_presentation_round` / `present_submit_round` + `presentation_added` / `presentation_updated` events. Generated types verified: no existing TS export lost (quicktype enum-merge trap checked).
- **Daemon** — the daemon is the sole parse/pin authority (CLI sends raw YAML). Handback on submit: durable ticket comment + nudge fan-out when the presentation is ticket-bound (auto-bound from the session's active ticket), guarded best-effort doorbell otherwise (skipped when the session isn't idle — chunk-1 limitation, noted in the plan).
- **CLI** — `attn present` (open / new round via `--presentation`), `attn present validate`, `attn present feedback`.

## Verification

- Unit/handler tests across all four layers (`internal/present`, store, daemon handlers via the existing `NewForTesting`/`syncConn`/wsClient harnesses).
- **Live-app**: verified twice on throwaway profiles (per AGENTS.md, never dev/prod) — full loop: `attn present` → `get_presentations` / `get_presentation_round` over a real WS client → `present_submit_round` with handback → `presentation_updated` broadcast (`latest_round_submitted: true`) → `attn present feedback` returning structured markdown with the referenced line quoted from git at the pinned head SHA.
- Not yet exercised live: doorbell typing into a real agent PTY and ticket-bound handback with a real delegated agent — those are PR 4's packaged-app scenario per the plan.

## Not in this PR

Window shell (PR 2), the reader UI (PR 3), packaged-app scenario + CHANGELOG (PR 4).

*PR opened by Claude on behalf of Victor.*


figgyster (approved):
I found one real rake in the alley: the daemon protocol moved to 144 but the frontend hook was still advertising 143, so `go test ./...` failed and the app would have seen its own daemon as mismatched. I pushed a tiny follow-up commit to bump the frontend constant.

After that, `go test ./...` passes locally and GitHub CI/release preflight are green. The Present spine looks coherent: daemon owns parse/pin, store operations are transactional where they need to be, and the CLI/WS contract hangs together.

No blockers from me.

PR #466, opened 2026-07-04
feat: Present window shell — pane-header chip + dedicated OS window (PR 2/4)

victorarias:
PR 2 of the Present spine train (plan: docs/plans/2026-07-04-present-spine.md, vision: docs/vision/present.md). Builds on the merged Go spine (#465): this PR gives an open presentation a visible surface — a chip in the triggering session's pane header and a dedicated OS window that will host the reader (PR 3).

## What's here

**Pane-header chip (main window)**
- A session with an open, unsubmitted presentation shows a violet **▶ review** chip in its pane header — the same rectangle that hosts the nudge indicator, on the tile of the agent that triggered the tour. No window-wide banner.
- On a lone (unsplit) tile the header is forced visible while the chip has something to show, mirroring the nudge behavior; drag/rename affordances stay split-only.
- Chip list is seeded from `get_presentations` once the socket is up and kept live by the `presentation_added` / `presentation_updated` broadcasts; a submitted or closed presentation drops its chip. If a session triggered several presentations, the chip shows the newest open one.
- Clicking the chip opens the presentation window (guarded against the pane header's leaf-drag, same as the nudge/rename buttons).

**Present window (Tauri)**
- New lazily-created `present` window (`open_presentation_window` command, TrustedMainWebview-gated, presentation id sanitized). Reuse path re-navigates the existing window instead of stacking new ones.
- Close (⌘W or the traffic light) hides the window rather than destroying it; the native menu's Close Pane pre-check only applies to the main window.
- Minimal capability surface: `capabilities/present.json` grants `core:default` + clipboard read/write only, pinned by a Rust test (no fs/shell/dialog). The main-window capability file is untouched.
- `?window=present&presentation=<id>` boots a slim `PresentRoot` instead of the full app: own daemon socket, follows the daemon theme read-only, shows title/kind/repo, round seq + pinned SHAs, Draft/Submitted state, summary and ordered file list with notes. The reader itself lands in PR 3.
- Shared `hideBootSplash()` so the present window drops the boot splash (previously only App did — the window rendered black without it; caught in live-app verification).

**Harness tooling for present**
- Automation-bridge actions `present_get_state` / `present_click_chip` (chip carries `data-presentation-id` / `data-session-id` hooks).
- `presentDaemon.mjs` + `present-daemon-cli.mjs`: daemon-WS helper for `get_presentations` / `get_presentation_round` / `present_submit_round`, reading the frontend `PROTOCOL_VERSION` at runtime so it can't drift from the handshake.

## Verification
- `pnpm --dir app exec tsc --noEmit`, full frontend vitest suite, cargo tests (capability pinning), harness node tests.
- Live in attn-dev: triggered `attn present --session <id>` against a real session tile → chip appeared in that tile's header → click opened the present window with correct round content and theme → ⌘W hides it, main window untouched → reopening reuses the window → submitting the round clears the chip.


figgyster (dismissed):
This looks safe to merge. The split is doing the right small thing for PR 2: the present window is capability-scoped, opened only from the trusted main webview, and the frontend shell stays read-only/thin while the reader waits for the next PR.

I also like the boot-splash fix getting pulled into a shared helper here — tiny seam, real bug. Local focused checks passed for the new PresentRoot/banner pieces and tsc is clean; GitHub CI is green now. No blockers from me.

figgyster (approved):
Still safe to merge. The follow-up moves the notice to the right place — on the presenting session's pane header — and the chip/automation harness coverage pins the behavior without reopening the full reader surface yet.

I pushed one tiny cleanup commit to trim a trailing blank line that made `git diff --check` fail. Focused chip/PresentRoot/harness tests and `tsc` pass locally; GitHub CI is green on the new head. No blockers.

PR #467, opened 2026-07-04
feat(harness): machine-parseable ATTN_VERDICT line from runner and serial matrix

victorarias:
## What

Every real-app harness scenario built on `createScenarioRunner`, plus `run-serial-matrix.mjs`, now ends stdout with a single machine-parseable line:

```
ATTN_VERDICT {"ok":false,"scenarioId":"tr401-local-window-resize","runId":"...","failureCount":1,"firstFailure":"assertion failed: ...","artifactsDir":"...","summaryPath":".../failure.json","durationMs":48211}
```

A driving agent learns pass/fail, the first failure, and where the artifacts live from one line instead of 3-4 follow-up spelunking round trips per harness iteration. Contract documented in `app/scripts/real-app-harness/AGENTS.md` (consumers take the LAST matching line).

Item 1 of the agent-cost-tooling ticket; the soak runner (item 2) builds on this line as its aggregation input.

## Scope notes

- Older ad-hoc scenarios (hand-rolled `main()`, no shared runner) intentionally don't emit the line — converting them is a separate mechanical change if it proves worth it.
- `firstFailure` is the first line of `error.message`, capped at 300 chars, so the one-line contract can't break.

## Testing

- Unit: `verdict.test.mjs` covers line format, escaping (newlines/quotes), truncation. `pnpm --dir app test`: 1250 passed / 5 pre-existing skips.
- Live-app verification: pending — will run a runner-based packaged-app scenario against a dev install and post the observed verdict line here before merge.

### DIFF
--- a/app/scripts/real-app-harness/AGENTS.md
+++ b/app/scripts/real-app-harness/AGENTS.md
@@ -22,6 +22,31 @@ re-derive them. dev/prod are fast-path literals that a drift guard in
 needs `./attn` built (`make dev` / `go build -o ./attn ./cmd/attn`); override the
 binary with `ATTN_HARNESS_BIN`.
 
+## Verdict Line
+
+Every scenario built on `createScenarioRunner` (`scenarioRunner.mjs`), and
+`run-serial-matrix.mjs`, print a single machine-parseable verdict line as the
+last thing they emit on that path, so a driving agent can learn pass/fail
+without spelunking through step logs or JSON summaries.
+
+- Format: `ATTN_VERDICT ` followed by compact (non-pretty-printed) JSON, all on
+  one line. `formatVerdictLine`/`emitVerdict` in `common.mjs` are the only
+  producers — use them instead of hand-rolling the line.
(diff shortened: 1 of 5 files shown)


figgyster (commented):
Code path looks clean: the verdict producer is centralized, the shared runner and serial matrix both emit the same contract shape, and the new tests cover the line-format/escaping/truncation bits where this would otherwise turn into parsing soup.

I ran `git diff --check`, the focused verdict test, and the full app Vitest suite locally: 1250 passed / 5 skipped. CI is green on the visible PR checks too.

I’m leaving this as a comment rather than a green stamp only because the PR already calls out packaged-app live verification as pending. Once you post the observed `ATTN_VERDICT` line from a runner-based dev install, I don’t see anything else blocking this.

figgyster (approved):
That live dev-app verdict line closes the only thing I was holding back for. The contract is documented, centrally emitted, unit-covered for the nasty one-line parsing cases, and now smoke-tested through a real packaged-app runner.

Safe to merge. Nice little shovel for the agent-spelunking hole.

PR #468, opened 2026-07-04
feat(harness): crop/scale options for native window capture

victorarias:
## What

`captureFrontWindowScreenshot` (and the `capture-app-screenshot.mjs` CLI) can now crop to a window-relative region (`--crop x,y,WxH` or `x,y,w,h`) and downscale in place (`--max-dim N`, via `sips -Z`), so when a driving agent does look at a screenshot it pays ~4× fewer image tokens.

Item 6 of the agent-cost-tooling ticket.

## Notes

- Crop rects are clamped to the window; only a fully-non-overlapping crop throws.
- `--max-dim` gates on the PNG's **actual pixel dimensions** (measured via `sips -g`), not logical points — on Retina the file is 2× the capture rect, and gating on points would make the flag a no-op.
- Default behavior with neither flag is unchanged (same `screencapture` invocation, no `sips` calls); the result object gains an additive `captureRect` (+ `pixelDimensions` when measured).
- An `attn shot` Go-side helper was descoped: the harness flags cover the workflow, and scenarios/agents already reach capture through these scripts.

## Testing

- Unit: `nativeWindowCapture.test.mjs` — `resolveCaptureRect` offset/clamp/no-overlap, `parseCropSpec` both forms + garbage, `parseSipsPixelDimensions`. Harness suite: 57 passed / 5 pre-existing skips.
- Live-app verification: pending — will capture against a dev install with and without `--crop`/`--max-dim` and post measured file dimensions here before merge.

### DIFF
--- a/app/scripts/real-app-harness/AGENTS.md
+++ b/app/scripts/real-app-harness/AGENTS.md
@@ -29,6 +29,16 @@ binary with `ATTN_HARNESS_BIN`.
 - If these scenarios pass while users can reproduce workspace/session errors in the packaged app, treat that as a test design bug.
 - Real-app commands target the dev sibling (or the active `ATTN_PROFILE`) by default. Production runs must pass `--run-against-prod`; never bypass the shared production-target guard.
 
+## Screenshot Crop / Scale
+
+`captureFrontWindowScreenshot` (and `capture-app-screenshot.mjs`'s `--crop`/`--max-dim`
+flags) can crop to a window-relative region and downscale the PNG at capture time via
+`sips -Z`, so an agent that actually looks at the image pays far fewer tokens. `--crop`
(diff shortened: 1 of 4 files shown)


figgyster (commented):
The code path looks clean: crop parsing/clamping is bounded, `sips -Z` is gated on actual PNG pixels, and default capture still avoids the extra `sips` calls.

I ran the harness tests locally:

- `pnpm exec vitest run scripts/real-app-harness` → 57 passed / 5 skipped

I’m leaving this as a comment rather than green-stamping it because the PR itself still has live-app screenshot verification pending, and this change is macOS-native capture plumbing. Once the measured live `--crop` / `--max-dim` dimensions are posted, this should be approval-shaped unless that smoke test coughs up a gremlin.

figgyster (approved):
That live smoke covers the only thing I was holding on: the Retina pixel math, crop rect reporting, and edge clipping all behave the way the code says they should.

Safe to merge. No blockers from me.

PR #469, opened 2026-07-04
feat(cli): attn vision-check — single tool-less LLM call over an image

victorarias:
## What

`attn vision-check <image> "<question>"` answers a question about a screenshot with ONE tool-less `claude` call (`--input-format stream-json` with a base64 image block, `--max-turns 1`, `--strict-mcp-config --setting-sources "" --allowedTools ""`), so the image never enters the driving agent's context. Default model **sonnet**, overridable via `--model`; `--json` emits a single machine-readable line with answer + cost.

Item 5 of the agent-cost-tooling ticket.

## Design notes

- The stream-json image-input path was spike-validated against the real CLI before implementation; the CLI's final `result` event supplies answer, `total_cost_usd`, and `num_turns`.
- Default mode prints the bare answer on stdout and a one-line `model/cost/turns` summary on stderr, keeping stdout pipe-clean. No profile banner for the same reason.
- Base64 payloads over ~4.5MB error with a hint to capture via `--crop`/`--max-dim` (PR #468) instead.

## Testing

- Unit: media-type mapping, message shape, last-result-wins stream parsing incl. a 300KB line (Scanner token-limit case), arg parsing. `go build ./...` + `go test ./cmd/attn/...` green.
- Live-verified against the real `claude` CLI with a generated solid-red PNG:
  - default mode → `red` / `vision-check: model=haiku cost_usd=0.0262 turns=1`, exit 0
  - `--json` → `{"answer":"Red.","model":"haiku","cost_usd":0.0181,"num_turns":1,"is_error":false}`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-04]
 
 ### Added
+- **`attn vision-check <image> <question>` CLI command.** Answers a question about a screenshot or image with a single, tool-less LLM call, so an agent can ask about an image without pulling it into its own context.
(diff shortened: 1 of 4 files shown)


figgyster (approved):
This looks safe to merge. The command is nicely boxed in: pure CLI surface, no profile banner leakage, bounded payload size, tool-less Claude invocation, and the stream parser dodges the Scanner token-limit rake.

I ran `git diff --check`, `go build ./...`, and `go test ./cmd/attn/...`; CI is green too. No blockers from me.

PR #471, opened 2026-07-04
feat(diagnostics): DOM-truth clip detection, fit-bail logging, offset soak scenario

victorarias:
## What

Diagnostics + reproduction tooling for the long-standing "terminal gets offset / bottom cut off, remount fixes it" bug. First PR of a 4-PR stack (this one → remetric → recover → reveal).

- **DOM-truth clip detector**: incidents in `terminal-incidents.jsonl` now compare the canvas rect against the container rect directly, with a `trigger` array (`model`, `dom_overflow`, `dom_overflow_right`, `dom_overflow_left`, `dom_offset`). Previously right-edge-only overflows were invisible to the detector.
- **Fit-bail logging**: `fit()` bail-outs (no renderer, same size, suspicious size) are now logged with container `cw/ch` — a silent `fit()` no-op on a dead renderer cost hours of diagnosis.
- **Offset soak scenario** (`real-app:scenario-offset-soak`): seeded, weighted random walk over the packaged app — workspace switches, font scaling, window resizes, rapid double-switches, splits/close-splits — measuring every pane's canvas-vs-container geometry after each step. Suspect clips are polled every 750ms up to 4.5s: first clean re-measure classifies it a transient (with `healMs`); persistent failures dump full evidence (violation.json with all recheck rounds, pane states, native screenshot, diagnostics tail).

## Why

The bug was effectively unreproducible by hand. The soak makes it deterministic: on main, `--seed 2 --iterations 300` fails with a persistent violation at step 291 (WebGL context exhaustion after font-size rebuild storms — fixed in the next PRs in the stack). The transient-poll window is calibrated from measured attach-replay churn: replay storms can outlast 1.5s (observed heals up to 2.4s), so a single fixed recheck misclassifies transients as persistent.

## Verification

- Detector confirmed live in the packaged app (29 incidents in a soak run, including 4 right-edge-only clips previously invisible).
- Full seed-2 × 300 soak run against the complete stack: 0 persistent violations, 19 transients (max heal 2430ms).
- Diagnostics are dev-instrumentation (JSONL on disk, per AGENTS.md frontend-instrumentation pattern); no user-facing behavior change, no CHANGELOG entry.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -56,6 +56,7 @@
     "real-app:scenario-terminal-block-copy": "node scripts/real-app-harness/scenario-terminal-block-copy.mjs",
     "real-app:scenario-terminal-context-menu": "node scripts/real-app-harness/scenario-terminal-context-menu.mjs",
     "real-app:scenario-terminal-block-resize": "node scripts/real-app-harness/scenario-terminal-block-resize.mjs",
+    "real-app:scenario-offset-soak": "node scripts/real-app-harness/scenario-offset-soak.mjs",
     "real-app:serial-matrix": "node scripts/real-app-harness/run-serial-matrix.mjs",
     "e2e": "playwright test",
     "e2e:settings-visual": "playwright test e2e/settings-visual.spec.ts",
--- a/app/scripts/real-app-harness/scenario-offset-soak.mjs
+++ b/app/scripts/real-app-harness/scenario-offset-soak.mjs
@@ -311,7 +311,8 @@ async function main() {
   let maxOverflowSeen = 0;
   let transientCount = 0;
   const transientSteps = [];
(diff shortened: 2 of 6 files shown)


figgyster (approved):
This looks safe to merge. The detector now checks the actual canvas/container rect instead of trusting model geometry, and the soak harness gives the terminal-offset bug a deterministic alley to get mugged in.

I ran `pnpm test -- terminalDiagnosticsLog.test.ts` and `pnpm run build`; GitHub's frontend/E2E checks are green too. No blockers from me.

Small follow-up, not a merge gate: the PR text mentions left-edge overflow, but this slice mainly pins bottom/right DOM escape. If left-edge clipping becomes a real repro vector, mirror the right-edge signal explicitly rather than relying on the current offset-left shape.

PR #472, opened 2026-07-04
fix(terminal): remetric renderer on font-size change instead of rebuilding

victorarias:
## What

Second PR of the offset-bug stack (on top of #471). Font-size changes no longer tear down and rebuild the WASM model + WebGL renderer of every mounted pane; the existing renderer re-derives its cell metrics (`setFontSize()` + a `metricsDirty` flag so the next `resize()` re-sizes the canvas even when the grid dimensions are unchanged).

## Why

This is the root cause of the **persistent** offset: a font-size change rebuilt N renderers at once (fontSize was in the mount-effect deps), and rebuild storms combined with warm-set churn exhaust WKWebView's small WebGL context pool. Construction then fails (or a live context is lost), the pane's `fit()` silently no-ops with no renderer, and the canvas keeps its old-font geometry forever — offset/clipped until remount. Remetric-in-place removes the context churn entirely.

## Verification

- Unit tests for the remetric path.
- Soak: on the base branch, seed-2 × 300 fails deterministically with a persistent violation at step 291. With this PR: **0 persistent violations** (18 transients), font-related incidents down from dozens to 2. Re-confirmed at the full-stack tip: 0 persistent / 19 transients.
- CHANGELOG entry included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ### Added
 - **Pin the chief-of-staff's reasoning effort, per agent.** Settings → Agents → "Chief-of-staff model & effort" now has an effort selector next to each agent's model override (Claude: low, medium, high, xhigh, max; Codex: minimal, low, medium, high, xhigh). Leave it on "Agent default" to use the agent's own default. Only chief-of-staff launches are affected — regular sessions are unaffected.
 
+### Fixed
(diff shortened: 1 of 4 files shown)


figgyster (dismissed):
This looks safe to merge. The important bit is that font changes now update the existing renderer's metrics and force same-grid resizes instead of rebuilding every mounted Ghostty pane — exactly the kind of boring fix WebGL context pools were begging for.

I checked the renderer remetric path, the pane effect dependency split, and the new coverage. Local `pnpm --dir app exec vitest run src/components/GhosttyWebGlRenderer.test.ts` and `pnpm --dir app run build` both pass. I couldn't see GitHub checks for this branch yet.

figgyster (approved):
Still good on the rebased head. The PR patch is the same renderer remetric fix against `main`: no model/WebGL rebuild on font-size changes, and same-grid canvas sizing is forced through `metricsDirty`.

Local focused renderer tests and the frontend build both pass here. GitHub's frontend/E2E checks were still pending when I looked, so let those finish before merging.

PR #473, opened 2026-07-04
fix(terminal): auto-recover WebGL renderer on context loss

victorarias:
## What

Third PR of the offset-bug stack (on top of #472). When a pane's WebGL context is lost or renderer construction fails, the pane now auto-recovers instead of dead-ending in a "reopen the pane" error:

- `webglcontextlost` → `preventDefault()`, clear the error overlay, schedule a rebuild.
- Rebuild uses a fresh canvas (`key={rendererEpoch}`) + fresh renderer + refit; the WASM model (scrollback, grid) survives untouched.
- Backoff schedule 250ms → 1.5s → 5s (pure, unit-tested `recoveryDelayMs`); after three failed attempts it gives up into the original error overlay.
- Mount-effect cleanup is the single owner of GL release (explicit `WEBGL_lose_context.loseContext()` on the orphaned canvas), so recovery can't leak contexts.
- Every stage is logged to diagnostics (`recovery` kind: contextLost/constructFailed/scheduled/recovered/giveUp).

## Why

PR #472 removes the main source of context exhaustion, but context loss remains possible (GPU resets, memory pressure, pool contention from other WebViews). Today that leaves a permanently dead pane — the exact user-visible symptom — when a transparent sub-second rebuild fixes it.

## Verification

- Unit tests for the backoff helper.
- New packaged-app scenario `real-app:scenario-webgl-recovery`: forces context loss via a `lose_webgl_context` bridge action on the active pane, then asserts the diagnostics sequence contextLost→scheduled→recovered, that fresh output paints afterwards, geometry is sane, and no error overlay is shown. Observed recovery in **269ms** in the packaged app.
- Full-stack seed-2 × 300 soak: 0 persistent violations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ### Fixed
 - **Changing font size no longer risks blank or misrendered terminals when many panes are open.** Adjusting the terminal font size used to rebuild every open pane's terminal (visible and backgrounded), which under enough open panes could exhaust the app's GPU rendering resources and leave a pane permanently blank or garbled until reopened. Terminals now update in place to the new font size instead of being rebuilt.
+- **Terminals now recover automatically if the app's GPU rendering context is lost.** Previously a pane hit by a GPU context loss showed a permanent error asking you to reopen it; it now rebuilds its renderer in place within a fraction of a second, keeping the session's content and scrollback.
 
 ## [2026-07-03]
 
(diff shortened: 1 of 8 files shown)


figgyster (dismissed):
This looks safe to merge. The recovery path does the right thing: prevent the default context-loss death spiral, swap in a fresh canvas/renderer with bounded backoff, and keep the terminal model/scrollback alive instead of making the user reopen the pane.

The packaged-app scenario is the right proof here — forcing an actual WebGL context loss and asserting the pane paints fresh output afterward is much better than hoping a mock understands GPU misery.

I ran the focused recovery/diagnostics Vitest coverage, `git diff --check`, and the frontend build locally. No blockers from me.

figgyster (approved):
Still safe to merge. The new head is a restack/split-commit shuffle over the same WebGL recovery patch, and the current PR diff still stays focused on renderer recovery plus its packaged-app harness.

I re-ran the focused recovery/diagnostics tests, `git diff --check`, and the frontend build on this head. No blockers.

PR #475, opened 2026-07-04
feat(cli): attn debug namespace for incidents, diagnostics, daemon-log probes

victorarias:
Part of ticket **agent-cost-tools** (item 3): cut the ~45 spelunking round trips agents spend hand-rolling `cd`+`tail`+`jq` into `~/Library/Application Support/com.attn.manager*/debug` and `daemon.log`.

## What

New `attn debug` CLI namespace, profile-aware like the rest of the CLI:

- `attn debug ls` — lists the profile's frontend debug dir contents (name/size/mtime) and daemon.log path/size, so agents learn where things live.
- `attn debug incidents [--tail N] [--grep RE]` — tail/grep `terminal-incidents.jsonl`.
- `attn debug diagnostics [--tail N] [--grep RE]` — same for `terminal-diagnostics.jsonl`.
- `attn debug daemon-log [--tail N] [--since DUR] [--grep RE]` — daemon.log with a Go-duration `--since` filter; untimestamped continuation lines follow the preceding timestamped line's match state.

Supporting pieces:

- `config.AppSupportDirForProfile` / `AppSupportDir` — Go-side resolution of the macOS app-support dir (mirrors Tauri's AppLocalData), previously only known to the frontend.
- `readLinesFile` uses `bufio.Reader`, not `Scanner` — incident records with embedded ring-buffer context exceed the 64KiB token limit.
- Profile banner goes to stderr, so piped stdout stays machine-clean.

## Testing

- Unit: reader (missing file, 200KiB line, no trailing newline), tail/grep, `--since` continuation semantics, app-support dir per profile.
- Live: built `./attn` from this branch and ran all four subcommands against the real dev profile — `ls` lists 10 artifacts, `incidents --tail 1` returns an intact >64KiB record, `diagnostics --grep replay` filters correctly, `daemon-log --since 30m --tail 5` returns fresh timestamped lines.

🤖 Generated with [Claude Code](https://claude.com/claude-code) on behalf of Victor

Co-Authored-By: Claude <noreply@anthropic.com>


figgyster (dismissed):
This looks safe to merge. The new debug namespace stays nicely profile-aware, the reader avoids Scanner’s 64KiB rake, and the `daemon-log --since` continuation behavior is pinned with useful tests.

I ran `git diff --check` and `go test ./...`; GitHub checks are green too. No blockers from me.

figgyster (approved):
Re-approved for the new head. The current PR patch is still the same debug CLI surface I already checked; the head moved because `main` advanced underneath it, not because this PR grew a second costume.

I re-ran `git diff --check` plus the focused Go tests for `cmd/attn` and `internal/config`, and the visible GitHub checks are green. No blockers.

PR #476, opened 2026-07-04
fix(terminal): refit revealed panes whose grid overflows the container

victorarias:
## What

Final PR of the offset-bug stack (formerly #474, reopened against main after the stack merged — #471–#473 are in). The reveal-time refit retry (`refitPanesNowAndIfStillTiny`, now `refitPanesNowAndIfStillWrong`) previously only re-fired when the pane looked suspiciously **tiny**. It now also re-fires when the pane's grid **overflows** its container (too tall / too wide — `overflowsContainer()` on the terminal handle, comparing rows×cellHeight / cols×cellWidth against the container), and retries at two late ticks (75ms and 400ms) to outlast reveal-time layout churn.

## Why

Third evidenced strand of the offset bug, caught live on prod: the window shrank while a workspace was hidden (display:none warm set); on reveal, the fit raced layout and landed a too-tall grid, and **nothing ever re-triggered** — the pane stayed clipped for 40+ minutes. A live experiment proved the mechanism: nudging the window height by 1px healed it instantly, i.e. fit computes correctly whenever it actually runs; the failure is purely a reveal-time race with no backstop.

## Verification

- Unit tests: refit retries at both ticks while the pane still overflows; no extra refits when clean.
- New packaged-app scenario `real-app:scenario-reveal-overflow`: grow the window while workspace A is visible (A refits taller), hide A behind a new workspace, shrink the window back, reveal A — asserts A converges to the container within 600ms. The sequence fails loudly if any window-resize step is a no-op, so the test can't pass vacuously.
- Full-stack seed-2 × 300 soak: 0 persistent violations, 19 transients all self-healed ≤2.4s.
- CHANGELOG entry included in the stack.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ### Fixed
 - **Changing font size no longer risks blank or misrendered terminals when many panes are open.** Adjusting the terminal font size used to rebuild every open pane's terminal (visible and backgrounded), which under enough open panes could exhaust the app's GPU rendering resources and leave a pane permanently blank or garbled until reopened. Terminals now update in place to the new font size instead of being rebuilt.
 - **Terminals now recover automatically if the app's GPU rendering context is lost.** Previously a pane hit by a GPU context loss showed a permanent error asking you to reopen it; it now rebuilds its renderer in place within a fraction of a second, keeping the session's content and scrollback.
+- **Switching back to a workspace after resizing the window no longer leaves the terminal cut off at the bottom.** If the window shrank while a workspace was in the background, revealing it could land an oversized terminal grid that never corrected itself; the app now detects and fixes this right after the workspace becomes visible.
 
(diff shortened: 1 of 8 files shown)


figgyster (commented):
No code blockers from me. This is the right kind of backstop for the reveal race: the terminal owns the overflow signal, the workspace only retries when the live pane is still wrong, and the packaged-app scenario pins the actual hide/shrink/reveal failure instead of testing a toy resize in a lab coat.

I verified the focused frontend tests locally (`SessionTerminalWorkspace`, `useGhosttyPaneRuntime`, and `GhosttyTerminal.resizePolicy`) plus `git diff --check`; all passed. I’m leaving this as a conditional review rather than green-stamping while GitHub E2E is still running — merge once those finish green.

figgyster (approved):
CI is green now, so the earlier caveat is gone. This looks safe to merge.

The overflow backstop is scoped tightly to panes that are still geometrically wrong, and the new harness scenario hits the actual hidden-workspace shrink/reveal failure. No blockers from me.

PR #477, opened 2026-07-04
fix: treat agent reload as a lifecycle transition, not a crash

victorarias:
## Problem

Reloading a delegated agent (session actions → Reload) made the daemon treat the worker exit as a crash: the bound ticket flipped to **Crashed** ("agent process ended mid-flight without reporting") and a spurious reconcile classifier run fired against a healthy session. Incident: 2026-07-04 21:55:44Z, ticket `jaunt-in-attn`, session `15822cd0`.

## Root cause

The frontend's reload is `kill_session` + `spawn_session` of the same id. The daemon's `handlePTYExit` had no way to distinguish that deliberate kill from an unexpected process death, so `reconcileTicketsOnSessionEnd` stamped the ticket crashed and enqueued the reconcile classifier.

## Fix

- Protocol: `reload?: boolean` on `kill_session` (ProtocolVersion 143 → 144, all three lockstep spots).
- Daemon: `handleKillSession` marks the session as a reload kill (one-shot mark, 15s TTL); `handlePTYExit` consumes it and skips **only** `reconcileTicketsOnSessionEnd` — backend removal, state handling, and the `session_exited` broadcast are untouched, so the carve-out is exactly one seam wide and true crashes keep today's behavior.
- Frontend: `reloadSession` sends `reload: true` on the kill.

## Second bug (same family, found during live verification)

The reload kill's exit surfaces as a clean voluntary quit (code 0, no signal), which tripped App.tsx's auto-close-on-clean-exit: the pane was closed, the now-empty workspace removed, and the respawn then failed daemon-side with `unknown workspace`. `reloadSession` now marks the session as reloading for the whole kill→respawn window and the exit handler skips auto-close for it.

## Verification

- Unit: 3 new daemon tests (`internal/daemon/ticket_crash_test.go`) covering reload-skip, true-crash unchanged, and mark TTL/one-shot; 2 new frontend tests for the reloading-mark window (incl. spawn-failure cleanup). Full frontend suite + tsc green; `go build ./...` + daemon tests green.
- **Live (packaged dev app)** via new `scenario-reload-not-crash.mjs`, both legs in one PASS run:
  - mid-turn reload → ticket stayed `working`, 0 reconcile tasks, daemon logged `skipping ticket reconcile … exit is a user reload, runtime respawns in place`, worker respawned;
  - real SIGKILL of a second session's pty-worker mid-turn → ticket stamped `crashed … ended mid-flight (working)`, exactly 1 reconcile task; the reload-leg ticket remained untouched (no cross-session mark leak).

Note: the crash-leg scenario polls 90s because a SIGKILLed worker is only detected via the worker-backend poller's "unreachable for 30s; forcing exit" backstop.

Possible overlap heads-up: the agent-cost-tools session touches reconcile *verdict content*; this PR only changes the *trigger* (`reconcileTicketsOnSessionEnd` call site in `handlePTYExit`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-07-05]
+
+### Fixed
+- **Reloading an agent no longer crashes its ticket.** Reloading a delegated agent's session (session actions → Reload) used to look like a process death to the daemon: the bound ticket was stamped Crashed and a pointless reconciliation verdict was posted against a perfectly healthy session. A reload is now a recognized lifecycle transition — the ticket stays in its column and no reconciliation runs. Real crashes are still detected exactly as before.
+- **Reload no longer races the pane teardown.** The reload's own kill could be mistaken for the agent quitting cleanly, closing the pane (and its workspace) out from under the respawn and failing the reload with "unknown workspace". The session now stays put for the whole kill → respawn window.
+
 ## [2026-07-04]
 
 ### Added
--- a/app/scripts/real-app-harness/scenario-reload-not-crash.mjs
+++ b/app/scripts/real-app-harness/scenario-reload-not-crash.mjs
@@ -1,0 +1,305 @@
+#!/usr/bin/env node
+
+/**
+ * Real-app scenario: reloading an agent is a lifecycle transition, not a crash.
+ *
+ * Regression coverage for the 2026-07-04 incident: reloading a delegated agent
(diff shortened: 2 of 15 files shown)


figgyster (approved):
This looks safe to merge. The fix is cut at the right seam: `reload:true` only suppresses the ticket crash/reconcile path, while normal exit cleanup and real crash detection stay intact. The frontend reloading guard also nails the nasty pane-teardown race — tiny window, big mess.

I checked the protocol bump/generated types, daemon lifecycle path, frontend reload path, and the new regression coverage. Local checks passed here (`go test ./internal/daemon`, focused frontend tests, `tsc`, `go build ./...`), and GitHub CI is green.

PR #479, opened 2026-07-04
feat(delegate): --cwd composes with --worktree; auto-truncate derived names

victorarias:
Item 10 of the agent-cost-tools ticket: two delegate CLI ergonomics fixes.

## What

- **`--cwd` + `--worktree` now compose.** `attn delegate --cwd <dir> --worktree <branch>` creates a worktree of the repo at `<dir>` and places the new workspace there. Previously both the CLI and the daemon rejected the combination outright. The daemon mechanics already supported it — `createDelegationWorktree` infers repo and starting ref from the base directory — so this removes the two guards and validates the cwd up front.
- **Over-long derived names truncate instead of erroring.** A worktree checkout basename like `attn--feat-some-long-branch` always exceeds the 16-rune session-name limit, which made `--worktree` unusable without also passing `--name`. Derived defaults are now cut to 16 runes with trailing `-_. ` trimmed. An explicit `--name` that is too long still fails loudly, as before.

## Testing

- New daemon tests: composition end-to-end (repo/ref inferred from cwd, workspace placed at worktree path), non-repo cwd still surfaces the "pass --repo" error, truncation for both the worktree and plain-directory derived-name paths, plus a `truncateDelegationName` table test (rune counting, trailing-punctuation trim).
- CLI parse test for `--cwd` + `--worktree`.
- **Live-verified on a throwaway profile** (`ATTN_PROFILE=prhsmoke`, daemon built from this branch): `attn delegate --cwd /tmp/.../a-very-long-repository-name-for-truncation --worktree feat/compose-smoke` created the sibling worktree, placed the new workspace at it (`worktree_created: true`, branch correct in `git worktree list`), and derived the label `a-very-long-repo` (exactly 16 runes, cleanly trimmed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-07-05]
 
+### Added
+- **Delegate: `--cwd` and `--worktree` now compose.** `attn delegate --cwd <dir> --worktree <branch>` creates a worktree of the repo at `<dir>` and places the new workspace there, instead of being rejected. Over-long auto-derived names (e.g. from a worktree folder like `attn--feat-some-long-branch`) are now truncated instead of failing the delegation; an explicit `--name` that is too long still errors as before.
+
 ### Fixed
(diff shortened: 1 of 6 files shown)


figgyster (approved):
This looks safe to merge. The `--cwd`/`--worktree` path now uses the cwd as the repo context instead of falling back to the source session, and the derived-name truncation keeps the implicit default ergonomic without weakening explicit `--name` validation.

I ran the focused daemon/CLI tests locally, and the visible CI/preflight checks are green. No blockers from me.

PR #478, opened 2026-07-04
feat(tickets): attn ticket show — full activity thread over the agent socket

victorarias:
Part of ticket **agent-cost-tools** (scope-extension item 7): agents could never re-read a ticket's discussion — `ticket inbox` is a consuming cursor read, `ticket list` carries no activity. The app's panel has `get_ticket` over WS; this gives agents the same full-detail pull over the unix socket.

## What

- New agent-socket command `ticket_show` (`TicketShowMessage`/`TicketShowResult` in TypeSpec + generated types): a **non-consuming** full-record read — metadata, description, complete activity thread with full bodies (comments, status changes, verdicts), attachments. Never touches any session's inbox cursor. Like `ticket_list`, not identity-scoped.
- Daemon handler reuses the exact path `get_ticket` serves the app (`store.GetTicket` + `ticketToProtocol`); unknown id → `Ok:false` "not found".
- `attn ticket show <ticket-id> [--session <id>] [--json]` — human output renders the full thread in the inbox style with **no truncation**; `--json` prints the full wire ticket.
- `ProtocolVersion` 144 → 145, bumped in both lockstep spots (constants.go + useDaemonSocket.ts).

## Testing

- Daemon test: full record round-trip with status change + long multi-line comment + attachment, all bodies intact; unknown-id failure contract.
- Printer test: long multi-line comment renders verbatim; empty-activity ticket renders sanely.
- `go build ./...`, scoped Go tests, `tsc --noEmit` all green.
- Live: throwaway `ATTN_PROFILE` daemon built from this branch — created a ticket, added a multi-line comment, verified `ticket show` (human + `--json`) returns the full body and the unknown-id case errors cleanly; profile cleaned after.

🤖 Generated with [Claude Code](https://claude.com/claude-code) on behalf of Victor

Co-Authored-By: Claude <noreply@anthropic.com>

PR #480, opened 2026-07-04
feat(journal): attn journal append — serialized daily-journal writes through the daemon

victorarias:
Item 9 of the agent-cost-tools ticket: chief and keeper contend on the same daily journal file — a direct file edit hits "file modified since read" on nearly every chief append. This exposes the already-serialized `notebook.Store.AppendJournal` seam over the unix socket + CLI.

## What

- **Protocol**: `journal_append` command (`JournalAppendMessage`/`JournalAppendResult`, `Response.journal_append_result`); `ProtocolVersion` 144 → **146** with the lockstep `useDaemonSocket.ts` bump (145 is claimed by in-flight #478 — whichever merges second takes a trivial version-constant rebase).
- **Daemon**: `handleJournalAppend` routes through the cached per-root `notebook.Store` writer (`notebookStoreFor`), so all journal writers serialize through one append. Mirrors the existing agent-originated notebook append (`AppendInbox`): content-aware self-write so the watcher doesn't re-announce it as an external edit, then `broadcastNotebookChanged` so open notebook views refresh.
- **CLI**: `attn journal append (--entry <text> | --entry-file <path>) [--date YYYY-MM-DD] [--session <id>] [--json]`. Date defaults to today; malformed dates are rejected by the store's own validation.

## Testing

- Daemon tests: create-on-first-append, second append lands both entries, empty entry fails, malformed date fails, date defaults to today. CLI parse tests for flag exclusivity/fallbacks. `go build`/`go vet`/`tsc --noEmit` clean.
- **Live-verified on a throwaway profile** (`ATTN_PROFILE=jrnltest1`, daemon built from this branch): append → correct `journal/2026-07-05.md` with `type: journal` frontmatter, second append coexists in the same file, error paths (empty entry, malformed date, both `--entry`+`--entry-file`) exit correctly. Profile cleaned after.


figgyster (dismissed):
This looks safe to merge. The new journal append path uses the existing serialized notebook store seam instead of making agents play file-edit roulette with the keeper, and the protocol/client/CLI pieces line up.

I ran focused Go tests/vet plus the frontend typecheck locally; GitHub CI is green. No blockers from me.

figgyster (dismissed):
This picked up a rebase regression from the newer base: the current PR diff removes the newly-added `debug` command from top-level help, and also drops the `--cwd` + `--worktree` composition note from delegate help.

That is small, but it is user-facing CLI contract drift — merging this would make `attn help` lie about commands/options that are already on `main`.

Required before merge:
- Rebase/resolve `cmd/attn/main.go` so the help text keeps the existing `debug <command>` entry and the full delegate worktree help, while adding the new `journal append` line.

Focused Go tests still pass locally; this is the one blocker.

figgyster (dismissed):
The rebase rake is gone: top-level help now keeps `debug <command>` and the delegate `--cwd` + `--worktree` note while adding `journal append`.

I reran the focused Go coverage for the changed daemon/client/protocol/CLI packages, `go vet`, `git diff --check`, and the frontend typecheck. GitHub CI is green except one E2E shard was still finishing on my last fetch, but the blocker I raised is fixed.

No blockers from me.

figgyster (approved):
This is the same journal-append change after the rebase/stack cleanup, and the current patch still keeps the fixed help contract intact: `journal append`, `debug <command>`, and the delegate worktree help all coexist now.

Focused local checks passed again, and GitHub CI is green. No blockers from me.

PR #482, opened 2026-07-04
feat(present): the reader — pinned diffs, notes, inline comments, submit round-trip

victorarias:
PR 3 of the Present spine (plan: docs/plans/2026-07-04-present-spine.md, vision: docs/vision/present.md). Builds on #465 (Go spine) and #466 (window shell + chip + harness tooling).

## What this delivers

The present window is now a working level-0/1 change reader:

- **Ordered file rail** from the round manifest, with per-file note markers, skip section, and j/k + arrow keyboard navigation.
- **Pinned-SHA diffs** through the shared DiffView — `get_file_diff` gained an optional head ref so diffs render the round's pinned commits, not the moving worktree.
- **Markdown summary and per-file notes** (react-markdown + GFM).
- **Drift banner** when the repo head moves past the round's pinned head; diffs stay pinned, banner is dismissable.
- **Inline comment drafts** on diff lines (both sides, using the signed line_end convention adapted to an explicit `side` on the wire), edit/delete for drafts, drafts survive file switches.
- **Submit review dialog** with handback toggle → `present_submit_round`; success clears drafts and refetches; the round header flips to Submitted.
- **Submitted comments render read-only** — new opt-in `readOnlyCommentIds` prop on DiffView/DiffCommentThread hides Edit/Resolve/Delete per comment; default preserves main-window behavior exactly.
- **Fix: blank diff pane on cold-created window.** A present window created hidden and shown without focus (e.g. display asleep/occluded) left WKWebView's rAF clock parked forever — the Virtualizer never painted. The window now takes focus on its page-load show (mirrors the reuse path; correct UX for an explicit chip click anyway).
- (An fs-capability grant for debug instrumentation was initially included, then dropped after review: `present_capability_has_a_minimal_permission_surface` deliberately keeps this window's permission surface minimal, and the instrumentation pattern is temporary-by-definition. The debugging recipe is a local, uncommitted capability edit while instrumenting.)

## Verification

- Unit: 134 files / 1291 tests green (19 PresentRoot tests incl. wire-shape round-trips for the signed-side adapters; 3 new DiffCommentThread tests); `tsc --noEmit` clean; `go build ./...` + store/daemon present tests green.
- Live (dev profile, packaged app): cold-boot repro of the blank-diff bug before/after fix (instrumented rAF layout probes: absent pre-fix, fire post-fix after display wake); j/k navigation; note banner; drift banner with correct SHAs after a new head commit; `present_submit_round` over the real daemon socket with a comment on the old/new side, round `submitted_at` + comment persisted, and the open window live-updating to Submitted with the comment anchored inline and rendered read-only.

## Follow-ups (deliberate)

- "Submit review" stays enabled on an already-submitted round (daemon rejects re-submit); agents open new rounds so this state is transient — will pick up with PR 4 polish if worth it.
- PR 4: packaged-app scenario for the full present flow + CHANGELOG entry.

### DIFF
--- a/app/src-tauri/src/lib.rs
+++ b/app/src-tauri/src/lib.rs
@@ -1149,6 +1149,12 @@ Object.defineProperty(window, "__ATTN_NATIVE_DIALOGS", {
             }
             // Show window as soon as page content is loaded (loading screen visible)
             let _ = webview.window().show();
+            // The present window opens from an explicit user action (chip click) and
+            // must take focus: an unfocused, hidden-at-creation WKWebView can leave
+            // requestAnimationFrame parked (blank diff pane) until first focus.
+            if webview.window().label() == PRESENT_WINDOW_LABEL {
+                let _ = webview.window().set_focus();
+            }
         })
         .run(tauri::generate_context!())
         .expect("error while running tauri application");
--- a/app/src/components/DiffCommentThread.test.tsx
+++ b/app/src/components/DiffCommentThread.test.tsx
@@ -1,0 +1,94 @@
+import { render, screen } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+import { DiffCommentThread } from './DiffCommentThread';
+import type { ReviewComment } from '../types/generated';
+
+function makeComment(overrides: Partial<ReviewComment> = {}): ReviewComment {
+  return {
+    id: 'comment-1',
+    content: 'a comment',
+    filepath: 'src/foo.ts',
+    line_start: 1,
+    line_end: 1,
+    author: 'user',
+    resolved: false,
+    created_at: '2026-07-01T00:00:00Z',
+    review_id: 'round-1',
+    ...overrides,
+  };
+}
+
+const noop = () => {};
+
+describe('DiffCommentThread', () => {
+  it('hides Edit/Resolve/Delete for a comment id in readOnlyCommentIds', () => {
+    render(
+      <DiffCommentThread
+        comments={[makeComment({ id: 'read-only-1' })]}
+        draft={false}
+        editingCommentId={null}
+        readOnlyCommentIds={new Set(['read-only-1'])}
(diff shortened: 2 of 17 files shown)


figgyster (dismissed):
This is close, but CI is catching a real contract break in the Tauri capability surface.

Required before merge:
- Keep the `present` window capability off broad filesystem permissions. The current `present.json` adds `fs:default` plus write/mkdir/exists, and the existing `present_capability_has_a_minimal_permission_surface` test fails on exactly that (`present capability must not grant fs/shell/dialog permissions, found fs:default`). That test is doing its job: the present window is intentionally a narrow, reader-specific surface, not a general filesystem-capable webview.
- If the JSONL instrumentation path is needed here, wire the smallest explicit permission surface for `$APPLOCALDATA/debug/...` and update the security test to encode that narrow exception rather than allowing `fs:default` wholesale.

I also hit the existing Frontend E2E split-blank repro failures in CI, but those look unrelated to this Present slice; the capability failure is the blocker in this PR's footprint.

Local checks I ran: `git diff --check`, `go test ./internal/daemon ./internal/protocol`, focused Present/DiffCommentThread Vitest, and `pnpm --dir app exec tsc --noEmit`.

figgyster (approved):
This fixes the blocker. The present capability is back to the narrow reader surface (`core:default` + clipboard only), and the fs grant is gone rather than watered down into a permanent debug exception. Good call — temporary instrumentation should stay temporary.

Local validation passed for the affected seams:
- `git diff --check`
- `go test ./internal/daemon ./internal/protocol -count=1`
- focused Present/DiffCommentThread Vitest
- manual check of `present.json` against the minimal capability rule

I couldn't run the Rust capability test locally because this runner doesn't have `cargo`, but the capability file now matches the guard's intent. GitHub still has the known `split-blank-repro` Frontend E2E flake red; I don't see that as part of this Present slice.

No blockers from me.

PR #462, opened 2026-07-03
fix(reconcile): judge dead sessions from a deterministic transcript slice

victorarias:
## Problem

On 2026-07-03 the orphaned-ticket reconciliation classifier failed every run: it spawned `claude -p` with `Read/Grep/Glob` and told it to read the dead session's transcript agentically. On a real **4.68MB** transcript that blew the `$0.50` budget cap, exited 1, and posted a **mislabeled** failure note.

## Fix

Replace the agentic read with a **deterministic, streaming extractor** and a **single tool-less Haiku completion**:

- **`internal/transcript/slice.go`** — `ExtractConversationSlice` reduces any Claude/Codex/Copilot transcript, in one streaming pass with bounded memory, to a small slice: the delegation prompt (first human turn), later **re-scoping** human turns, the most recent compaction summary (Claude), and the agent's last status turns. Text-only, no activity trace.
- **`internal/agent`** — `HeadlessTaskRequest.DisableTools` for a genuinely tool-less run. (An empty `AllowedTools` alone still falls back to the native default tool set — the trap this flag exists to avoid.)
- **`internal/daemon/ticket_reconcile.go`** — extract + inline the slice into the prompt, keeping **both** the filed ticket brief and the slice's delegation prompt (they can differ; the delegation prompt is often the more detailed real instruction). Default model → `haiku`; bail cheaply on an empty slice. Caps loosened to a pure runaway backstop (**4 turns / $0.20**) now that input is bounded by construction.

Because the input is bounded, the run can neither loop on tools nor balloon on transcript size — the failure mode is structurally gone.

## Validation (on the real transcript that originally failed)

Ran the **real production `execTicketReconcileClassifier`** against the 4.68MB transcript:

| model | verdict | cost | latency | caught the mid-session re-scope? |
|---|---|---|---|---|
| **haiku** (default) | `done` / high | **$0.073** | 23s | ✅ |
| sonnet (env) | `done` / medium | $0.254 | 42s | ✅ |

Both correctly recognized that the task was re-scoped mid-session (*"i want u to implement the plan in this pr"*) and judged against the delivered implementation, not the filed doc-only brief. haiku is the decisive, correct, 3.5× cheaper default. Reductions: Claude 4.68MB→18.5KB (253×), Codex 3.77MB→1.9KB (2006×).

- extractor: 12 unit tests (Claude/Codex/Copilot, empty, metadata-only, missing, caps, 2MB line, render); all green.
[...]

PR #483, opened 2026-07-04
feat(reconcile): ground-truth check for PR references in verdicts

victorarias:
Item 8 of the agent-cost-tools ticket: reconciliation verdicts can contradict state attn already knows — the motivating case was a verdict claiming "merge PR #462 pending" hours after #462 had merged. This adds a deterministic, **annotate-only** cross-check: the verdict text is scanned for PR references, each is resolved to a definitive state, and the posted comment gains `Ground-truth check: PR #N is merged ("<title>") — the verdict text may be stale on this point.` lines when a referenced PR is positively known finished. The verdict fields are never mutated, and nothing here can fail or stall the reconcile.

## Design notes

- **Extraction** covers `#N`, `PR N`, and `github.com/.../pull/N`, deduped in first-seen order with a garbage guard.
- **Two knowledge legs.** Investigation showed attn's `prs` table only ever holds **open** PRs (fully replaced from `is:open` searches each poll — merged/closed PRs vanish rather than changing state; `PR.State` is attn's own workflow annotation, e.g. `waiting`). So:
  1. Tracked rows use a positive allowlist (`merged`/`closed`) — currently forward-compatible groundwork, since nothing writes those states today. A naive "state != open" check would have false-positived on every tracked open PR.
  2. The production path: referenced PRs **absent** from the tracked open set (the signature of a merged/closed PR) get a single targeted GitHub lookup each — new `github.Client.FetchPRState` (one GET, parses `state`/`merged`/`title`) — capped at 3 lookups, 10s total budget, silent on error/open/no-client.
- New `git.OriginHostOwnerRepo` resolves the ticket's cwd to host + `owner/name` (scp-like, ssh:// incl. enterprise hosts and ports, https://), so lookups go through the right `ghRegistry` client.
- Test seam: `Daemon.ticketReconcilePRFetch` (nil in production; fakes in tests).

## Testing

- Unit: extraction (7 cases), tracked-leg annotation, remote-URL parsing against real temp git repos, `FetchPRState` via httptest.
- Executor-level daemon tests run the real reconcile executor with a fake classifier + fake fetcher: merged/closed annotations fire, open/error/no-client are silent, tracked refs skip the lookup leg, cap of 3 enforced, context respected.
- **Real-API shape confirmed read-only**: `GET /repos/victorarias/attn/pulls/462` returns `state:"closed", merged:true, title:...` — exactly what `FetchPRState` parses (and #462 is the PR from the original bug report).
- No full live-app reconcile run: that requires killing a real delegated session with a bound ticket and paying a live classifier run. The daemon tests drive the actual `reconcileTaskExecutor` end-to-end (verdict → cwd → host/slug → registry-gated fetcher → posted comment) with only the network faked, the git helper is tested against real repos, and the API payload assumption is verified above; the change adds no protocol, PTY, or UI surface. Flagging per AGENTS.md — say the word if you want a live dev-profile reconcile exercised anyway.


figgyster (dismissed):
This looks safe to merge. The annotate-only shape is the right call here: verdicts stay immutable, open PRs stay silent, and the targeted lookup path is capped so reconciliation doesn't become a GitHub API séance.

Local checks passed: `go test ./internal/daemon ./internal/git ./internal/github` and `git diff --check`. CI is green now too.

Tiny follow-up for later: if terminal PR states ever become persisted across multiple hosts, the tracked-row leg should probably switch from repo-only lookup to host+repo lookup. Not a blocker today because that leg is forward-compatible groundwork, not the production path.

figgyster (approved):
Re-approving the rebased head. I checked the current PR patch and the reconcile change is still the same safe annotate-only shape; the extra base churn is unrelated ticket/journal work already on `main`, not new surface from this PR.

Local focused checks passed again: `go test ./internal/daemon ./internal/git ./internal/github` and `git diff --check`. CI is green now too.

PR #486, opened 2026-07-05
test(perf): PTY encode + transcript parser allocation benchmarks

victorarias:
## Summary
- Adds `internal/daemon/pty_forward_bench_test.go`: benchmarks `encodePtyOutputMessage`, the per-chunk per-client PTY-output encode path, comparing the legacy JSON+base64 branch against the binary-frame branch (64B/1KB/4KB). Confirms the JSON path is ~7 allocs/op at every size vs 1 alloc/op for binary, and ~4x the bytes at 4KB — the concrete number behind roadmap item ws-3.
- Adds `internal/transcript/parser_bench_test.go`: benchmarks `ExtractLastAssistantTurnAfterLastUserSince`, the transcript-parse hot path run once per assistant turn per session, over a deterministic synthetic realistic-shaped JSONL fixture (no committed blob) at 1k/10k/50k lines to show the O(file-size) scaling. Includes `TestSyntheticTranscriptFixtureIsRealistic`, a regression guard that the fixture generator actually ends on real assistant content rather than an early empty-turn return.
- Both are `testing.B` benchmarks only — they run under `-bench` and add zero cost to normal `go test ./...`.

Implements test #1 and #2 of docs/plans/2026-07-05-performance-test-strategy.md.

## Test plan
- [x] `gofmt -l` clean on both new files
- [x] `go vet ./internal/daemon/ ./internal/transcript/` clean
- [x] `go test -run '^$' -bench 'BenchmarkEncodePtyOutput' -benchmem ./internal/daemon/` — passes, JSON ~7 allocs/op vs binary 1 alloc/op at every size
- [x] `go test -run '^$' -bench 'ExtractLastAssistantTurn' -benchmem ./internal/transcript/` — passes, linear scaling confirmed across 1k/10k/50k lines
- [x] `go test ./internal/daemon/ ./internal/transcript/` — normal (non-bench) suite still passes, including new `TestSyntheticTranscriptFixtureIsRealistic`

### DIFF
--- a/internal/daemon/pty_forward_bench_test.go
+++ b/internal/daemon/pty_forward_bench_test.go
@@ -1,0 +1,68 @@
+package daemon
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/victorarias/attn/internal/protocol"
+	"github.com/victorarias/attn/internal/ptybackend"
+)
+
+// sink prevents the compiler from optimizing away encodePtyOutputMessage's
+// result across benchmark iterations.
+var sink outboundMessage
+
+// benchEncodePtyOutput quantifies the per-chunk, per-client cost of
+// encodePtyOutputMessage — the hottest datapath in attn, run once per output
+// chunk for every attached client. legacyClient selects between the two
+// capability branches: without protocol.CapabilityBinaryPtyOutput the chunk
+// is base64-encoded and marshaled into the fat protocol.WebSocketEvent
+// (alloc-heavy, roadmap ws-3); with it, protocol.EncodePtyOutputFrame is used
+// instead (lean baseline).
+func benchEncodePtyOutput(b *testing.B, chunk []byte, legacyClient bool) {
+	client := &wsClient{}
+	if legacyClient {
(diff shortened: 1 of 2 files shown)


figgyster (dismissed):
This is safe to merge. The two benchmarks hit the right seams: the PTY hot path is isolated enough to compare JSON/base64 vs binary framing cleanly, and the transcript fixture has a regression guard so the parser benchmark doesn’t accidentally measure the cheap empty-turn path wearing a fake moustache.

I ran `git diff --check`, `go test ./internal/daemon/ ./internal/transcript/`, and the new benchmark targets locally. GitHub CI/preflight is green too.

figgyster (approved):
Still safe to merge. The follow-up is just benchmark-test cleanup (`interface{}` → `any`, plus removing the unused timestamp parameter from the synthetic user fixture), and it doesn’t change the benchmark behavior.

I reran the daemon/transcript tests and benchmark targets locally; GitHub CI/preflight is green on this head.

PR #487, opened 2026-07-05
perf(harness): per-machine timings registry (fingerprint + baseline store)

victorarias:
Second PR of the perf-test-strategy build-out (strategy doc merged in #485). Adds the **per-machine timings registry** — the substrate both macro tests (real-app RSS baseline, leak soak) will compare against, so a run trends against *that machine's own* baseline instead of one absolute number that's meaningless across different hardware.

## What's here (JS harness only, no product code)
- `getMachineFingerprint()` → `{ key, arch, platform, osRelease, hwModel, cpuBrand, cpuCount, totalMemGb }`. `key` is a 12-char sha256 of the identity subset (hw.model + CPU brand + cores + arch + RAM + **major** OS version). Major-OS-only in the key is deliberate: a routine macOS point-update shouldn't orphan a machine's baseline.
- `compareToBaseline(value, baseline, {tolerancePct})` → regression-only band check. A missing baseline **never fails** (first run self-baselines); a value *below* baseline is always ok (improvement).
- `loadBaseline`/`saveBaseline` → committed **canonical** baselines win over a per-machine **local cache** at `~/.attn-perf-registry` (home dir → machine-global across worktrees, no in-repo gitignore churn).

## Design refinements vs the doc
The doc said "gitignored local cache"; this puts the cache in the home dir instead (a machine's baseline is inherently machine-global and shared across all git worktrees), so there's nothing to gitignore. Canonical baselines stay committed in-repo (`perf-baselines.json`, seeded empty).

## Verification
18 unit tests (vitest), covering key stability/uniqueness/osMajor-collapse, every `compareToBaseline` branch, and the full load/save round-trip incl. canonical-precedence and missing-file tolerance. Ran targeted: 18/18 green. Real fingerprint confirmed on this machine (M5): sysctl path returns real `hw.model`/`cpuBrand`, not fallbacks.

Pure-module PR; consumed by the next two PRs. No live-app surface.


figgyster (approved):
This looks safe to merge. The registry keeps the harness machine-local where it should be, canonical baselines correctly win when present, and the regression check is intentionally one-way: improvements don’t get punished.

I ran the targeted Vitest file locally: `scripts/real-app-harness/machineRegistry.test.mjs` — 18/18 green. CI’s relevant changed-files gate and frontend E2E are green too; the skipped jobs look expected for this JS-harness-only slice.

No blockers from me.

PR #488, opened 2026-07-05
perf(harness): RSS-baseline verdict + per-machine registry comparison

victorarias:
## Summary
- `scenario-perf-baseline.mjs` now compares its headline RSS against the per-machine baseline registry (`machineRegistry.mjs`, merged in #487) and emits an `ATTN_VERDICT` line, so a driving agent can read pass/fail without spelunking through `summary.json`.
- First run on a machine self-baselines; `--record-baseline` forces a re-record; `--rss-tolerance-pct` tunes the allowed growth band (default 15%).
- A regression surfaces as `verdict.ok:false` but never flips the process exit code — it's a trend signal, not a harness error (matches the "trends only, no hard gates" decision in `docs/plans/2026-07-05-performance-test-strategy.md`).
- New pure helpers `evaluateRssBaseline`/`buildBaselineVerdict` in `rssBaselineVerdict.mjs`, unit tested.
- New `docs/perf-testing.md` runbook covering the Go micro-benches and this RSS baseline.

Originally stacked on #487 (per-machine timings registry); #487 merged to `main` while this was in flight, so this now targets `main` directly (rebased, same single commit).

This PR is harness-only (script + docs + unit tests). **Live-app verification is pending** — I have not run the packaged-app scenario myself; that verification will be done separately before merge, per this repo's live-app-testing policy.

## Test plan
- [x] `pnpm exec vitest run scripts/real-app-harness/rssBaselineVerdict.test.mjs` — 8/8 pass
- [x] `node --check` on all 3 touched/new `.mjs` files — clean
- [ ] Live-app run of `pnpm --dir app run real-app:scenario-perf-baseline` on a dev install (pending)

### DIFF
--- a/app/scripts/real-app-harness/rssBaselineVerdict.mjs
+++ b/app/scripts/real-app-harness/rssBaselineVerdict.mjs
@@ -1,0 +1,57 @@
+import { compareToBaseline } from './machineRegistry.mjs';
+
+// Contract cap shared with scenarioRunner.mjs's summarizeFirstFailure — the
+// ATTN_VERDICT line is single-line and length-bounded, so firstFailure here
+// follows the same rule even though this scenario builds its own verdict by
+// hand instead of going through createScenarioRunner.
+const FIRST_FAILURE_MAX_LENGTH = 300;
+
+function truncateFirstFailure(message) {
+  return message.length <= FIRST_FAILURE_MAX_LENGTH ? message : message.slice(0, FIRST_FAILURE_MAX_LENGTH);
+}
+
+// Pure: decides whether this run's RSS is within tolerance of the per-machine
+// baseline, and whether a new baseline should be written. No fs/Date/process
+// access so it stays directly unit-testable — the caller resolves the
+// baseline (machineRegistry.loadBaseline), performs the write
+// (machineRegistry.saveBaseline), and supplies `recordedAt`.
+//
+// A baseline is (re)recorded when there isn't one yet (first run on a
+// machine always self-baselines) or when the caller explicitly asked to via
+// `record` (re-baselining after an intentional regression/improvement).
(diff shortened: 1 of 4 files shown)


figgyster (dismissed):
This is close, but the new `--record-baseline` path lies to the driving contract.

The help/docs say re-recording overwrites the stored baseline *instead of comparing against it*. The implementation still compares against the old baseline first, saves the new one, and can emit `ATTN_VERDICT {"ok":false,...}` for the very run that was explicitly re-baselining. I reproduced the pure helper path with `record:true`, old baseline 500 MB, new value 700 MB: it saves 700 MB but returns `ok:false` / `reason:"regression"`.

That means an intentional baseline update still wakes the agent as a regression, which defeats the point of the flag and makes the machine-readable verdict untrustworthy.

Required before merge:
- Make `record:true` produce a non-regression verdict for the newly recorded value, or change the advertised contract if you really want “record but still fail against the old baseline.” My vote: recording should compare against the new baseline/no-op as passing, and have a unit test for that exact case.

Local checks I ran: `node --check` on the touched `.mjs` files, plus `pnpm exec vitest run scripts/real-app-harness/rssBaselineVerdict.test.mjs` after a frozen app install. I did not run the live packaged-app scenario here.

figgyster (approved):
This fixes the blocker cleanly. `--record-baseline` now has honest driver semantics: the run records the new number and reports a passing `recorded` verdict instead of screaming regression about the baseline it just replaced. Tiny miracle: the machine-readable contract and the docs now agree.

I re-ran the focused helper repro, the RSS verdict unit test, and syntax checks on the touched harness files. I did not run the packaged-app scenario here; for this pure verdict/docs fix, the focused coverage is enough for me.

No blockers from me.

PR #490, opened 2026-07-05
test(perf): cold vs warm RSS from a dedicated clean perf profile

victorarias:
## What

Adds `scenario-perf-cold-warm` — two **reproducible** RSS numbers for the packaged app, replacing the history-dependent single number from `scenario-perf-baseline`.

`scenario-perf-baseline`'s headline self-baselines against its *own prior history*: a freshly launched app and one that has cycled sessions land at very different footprints, so the number is not reproducible run-to-run on the same machine. Measured here: **1162 MB** (history-polluted, its empty floor was already 675 MB from the still-running app) vs **776 MB** clean.

The new scenario captures two numbers, **each preceded by a full data-dir wipe** so neither depends on prior app history:

| phase | what it measures |
|-------|------------------|
| **cold** | fresh app + N sessions, measured right after settle (no history) |
| **warm** | fresh app + N sessions + a bounded per-pane warmup burst, then settle — the worked-then-idle footprint (grows the Ghostty WASM heaps/atlas) |

## Dedicated perf profile

Each phase wipes its data dir, so the scenario **refuses prod (`~/.attn`) and the dev sibling (`~/.attn-dev`)** — it needs a dedicated profile:

```bash
make install PROFILE=perf                       # one-time
ATTN_HARNESS_PROFILE=perf pnpm --dir app run real-app:scenario-perf-cold-warm -- --sessions 8
```

Each phase compares against its own per-machine registry key (`<fingerprint>-cold` / `-warm`, reusing the merged registry from #487). A regression is a **trend signal, never a non-zero exit** — only real errors fail the process.

## Shared primitives

The measurement + lifecycle helpers (process table, `snapshot`, `sampleWindow`, daemon pid, session fill/close) are extracted verbatim into `perfMeasure.mjs` so both scenarios share one copy. `scenario-perf-baseline` becomes a behavior-preserving re-point onto the module (`stopDevDaemon` → `stopDaemon('dev')`). `buildColdWarmVerdict` builds the single `ATTN_VERDICT` for both phases (cold before warm on `firstFailure`).

## Verification

Live on the dedicated `perf` profile (Apple M5, 32GB):

- **cold 776.2 MB / warm 842.8 MB (+66.6)** — both self-baselined into the registry under distinct `-cold`/`-warm` keys; verdict `ok:true`, contract-shaped.
- `scenario-perf-baseline` runs **clean on the same fresh build** (`ok:true`, exit 0), exercising the full extracted surface (`sampleWindow`, `readProcessTable`, `classify`, …) — proving the extraction changed no behavior.
- `rssBaselineVerdict.test.mjs`: 12/12 pass (added 4 `buildColdWarmVerdict` cases).

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -25,6 +25,7 @@
     "real-app:bridge-perf": "node scripts/real-app-harness/bridge-perf.mjs",
     "real-app:bridge-pty-bench": "node scripts/real-app-harness/bridge-pty-bench.mjs",
     "real-app:scenario-perf-baseline": "node scripts/real-app-harness/scenario-perf-baseline.mjs",
+    "real-app:scenario-perf-cold-warm": "node scripts/real-app-harness/scenario-perf-cold-warm.mjs",
     "real-app:bridge-cli": "node scripts/real-app-harness/ui-automation-cli.mjs",
     "real-app:scenario-tr201": "node scripts/real-app-harness/scenario-tr201-local-relaunch-existing-split.mjs",
     "real-app:scenario-tr204": "node scripts/real-app-harness/scenario-tr204-local-relaunch-formatting.mjs",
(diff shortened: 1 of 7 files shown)


figgyster (approved):
This looks safe to merge. The new cold/warm split fixes the old history-polluted RSS baseline problem, and the dedicated perf-profile guard keeps the data-dir wipe away from prod and the shared dev sibling.

I checked the extraction path and ran the focused verdict tests locally; GitHub CI is green too. No blockers from me.

PR #491, opened 2026-07-05
test(perf): leak-soak retained-RSS slope scenario

victorarias:
## What

Adds `scenario-perf-leak-soak` — the last test in the perf starter set (#4). It catches the memory failure users actually feel: RSS that **climbs over a long session and never comes back** (the WASM balloon; any per-operation frontend leak).

The cold/warm scenario (#490) takes two single-cycle snapshots, each from a freshly wiped data dir — so it structurally *cannot* see a leak that only accumulates **within one continuous process**. This scenario does:

- Runs **N create → workload → close cycles inside one long-lived app instance** (no wipe between cycles — cycling sessions in a single running app is what a real multi-hour user session looks like).
- Samples retained RSS after a **decay-hold window** each cycle, keeping the **last** sample (retained), **never the peak** — macOS scavenges freed pages lazily, so a sample right after close still shows the transient workload spike, not what the process actually holds.
- Fits the **least-squares slope** of retained-RSS-vs-cycle over the post-warmup cycles. Flat slope = healthy churn; a positive staircase = leak.

```bash
make install PROFILE=perf                    # one-time
ATTN_HARNESS_PROFILE=perf pnpm --dir app run real-app:scenario-perf-leak-soak -- --cycles 12
```

## Design decisions

- **Self-contained cycle loop, deliberately NOT driven by `run-soak.mjs`.** The strategy doc suggested building on the soak runner, but run-soak spawns a fresh child / relaunches the app between iterations — that resets the process and *erases* the within-process accumulation a leak needs. run-soak stays for flakiness-soaking a whole scenario, orthogonal to this measurement.
- **Warmup cycles dropped from the fit** (`--warmup-cycles`, default 2): the first cycle or two carry one-time WASM heap/atlas growth and lazy daemon/pty-worker init — not a leak — which would bias the slope upward.
- **Generous threshold + decay hold defeat both documented traps:** peak RSS is meaningless (one-way allocator), and a too-tight band on a lazily-scavenged OS reads noise as a false leak. Real data below shows ~30 MB/cycle RSS noise, so the slope-over-many-cycles fit — not any single before/after — is what gives a stable signal.
- **Gate is the slope only.** The first post-warmup retained value also self-baselines under `<fingerprint>-leak-floor` (reusing the merged registry) as an informational trend, but never gates. A slope regression sets `verdict.ok:false` but never a non-zero exit — trend signal, not build break.

New pure helpers `fitSlope` + `buildLeakSoakVerdict` in `rssBaselineVerdict.mjs`, in parity with the baseline/cold-warm builders.

## Verification

Live on the dedicated `perf` profile (Apple M5), 6 cycles:

```
retained: 891.2 / 905.7 / 941.0 / 900.5 / 873.3 / 938.5 MB
post-warmup slope: -3.47 MB/cycle  ->  ok:true
```

- The decay hold works: cycle 5 read retained **873.3 MB below its peak 906.2 MB** — the settled-vs-transient distinction the design requires.
- Leak-floor baseline recorded, verdict contract-shaped, exit 0.
- `rssBaselineVerdict.test.mjs`: **19/19** (12 existing + 7 new: `fitSlope` flat/staircase/noisy/throws, `buildLeakSoakVerdict` pass/fail/boundary).

Refuses prod and the dev sibling (each run cycles heavily inside one instance); needs the dedicated perf profile like cold/warm.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -26,6 +26,7 @@
     "real-app:bridge-pty-bench": "node scripts/real-app-harness/bridge-pty-bench.mjs",
     "real-app:scenario-perf-baseline": "node scripts/real-app-harness/scenario-perf-baseline.mjs",
     "real-app:scenario-perf-cold-warm": "node scripts/real-app-harness/scenario-perf-cold-warm.mjs",
+    "real-app:scenario-perf-leak-soak": "node scripts/real-app-harness/scenario-perf-leak-soak.mjs",
     "real-app:bridge-cli": "node scripts/real-app-harness/ui-automation-cli.mjs",
     "real-app:scenario-tr201": "node scripts/real-app-harness/scenario-tr201-local-relaunch-existing-split.mjs",
     "real-app:scenario-tr204": "node scripts/real-app-harness/scenario-tr204-local-relaunch-formatting.mjs",
--- a/app/scripts/real-app-harness/rssBaselineVerdict.mjs
+++ b/app/scripts/real-app-harness/rssBaselineVerdict.mjs
@@ -92,3 +92,77 @@ export function buildColdWarmVerdict({ cold, warm, scenarioId, runId, artifactsD
     metrics: { coldRssMb: cold.value, warmRssMb: warm.value },
   };
 }
+
+// Pure: least-squares linear regression of `points` (y) against x = 0..n-1.
+// Standard closed form: slope = (n*sum(x*y) - sum(x)*sum(y)) / (n*sum(x^2) -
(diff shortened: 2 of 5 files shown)


figgyster (approved):
This looks safe to merge. The leak-soak scenario measures the right failure mode — retained RSS slope inside one long-lived process — and the new verdict helpers keep the trend signal machine-readable without turning perf noise into a hard build gate.

Local checks passed for the touched pure helpers (`rssBaselineVerdict.test.mjs` 19/19), `git diff --check`, and syntax checks on the changed harness modules. GitHub CI is green too.

No blockers from me.

PR #492, opened 2026-07-05
feat(present): attn present --wait blocks until the round is handed back

victorarias:
Adds `attn present --wait`: the open command blocks as a foreground call until the reviewer hands the round back, then prints the feedback markdown straight to stdout (or the result JSON with `--json`). The authoring agent's turn stays alive — the session shows as working instead of falsely idle — and there's no separate `attn present feedback` step.

**Reworked on top of current main for the verdict-era daemon (#516):**
- `PresentFeedbackResult` gains `verdict?` ("approved" | "feedback", absent until submitted) and `presentation_status` ("open" | "approved" | "closed"); ProtocolVersion 154→155.
- The wait loop terminates on **Close** too: closing a presentation never submits the round, so the old loop would have polled forever. Now it prints a "closed by the reviewer without feedback — drafts were discarded" note and exits 0.
- Transient daemon errors are tolerated (reported once per distinct error, keep polling); SIGINT/SIGTERM exit cleanly.

**Tests:** 4 deterministic poll-loop tests in `cmd/attn` (manual tick channel, scripted fetch: submit, cancel, transient error, close) + 2 daemon tests asserting the feedback result carries verdict/status after submit and after close.

**Live verification (throwaway profile `presentwait`, daemon built from this branch):**
- submit path: `attn present --wait` blocked, reviewer submitted a round with a comment over the daemon socket → CLI printed the rendered feedback markdown incl. the comment, exit 0.
- close path: second presentation, `present_close` over the socket → CLI printed the closed-without-feedback note, exit 0.


figgyster (dismissed):
This looks safe to merge. `--wait` stays on the existing presentation feedback contract, the poll loop is isolated enough to test cleanly, and CI is green.

I also like that this keeps the agent foreground turn alive instead of making every caller reinvent tiny polling machinery. Less ritual, fewer rakes.

figgyster (commented):
The implementation looks sound: `--wait` now has the missing close-path escape hatch, the protocol bump/generated clients are in sync, and the daemon result carries the verdict/status the CLI needs. The focused local checks passed here:

- `go test ./cmd/attn ./internal/daemon ./internal/protocol`

I’m leaving this as a comment rather than a fresh approval because the current visible CI gate is red: `Frontend E2E` is failing in shard 3 on `e2e/split-blank-repro.spec.ts`, which looks unrelated to this Present CLI/protocol change but still blocks the merge box. Tiny rake, wrong alley.

figgyster (approved):
CI cleared on the rerun, and the Present wait changes are safe to merge now.

The close-path handling is the important bit here: `presentation_status` gives the CLI a real stop condition when a reviewer closes without handback, while submit still prints the feedback exactly once. Local focused Go checks passed too.

No blockers from me.

PR #493, opened 2026-07-05
feat(reconcile): log when ground-truth annotation hits a cap

victorarias:
The reconcile **ground-truth** cross-check (the annotate-only "Ground-truth check" lines added in #483) silently `break`/`continue`s on every limit it hits: the 5-line annotation cap (`groundTruthMaxLines`), the 3-lookup GitHub cap (`groundTruthMaxLookups`), and the 10s lookup timeout (`groundTruthLookupTimeout`). That silence made the caps invisible in practice — no signal for whether they ever needed tuning.

## What this does
- Adds a small `groundTruthCaps{lineCap, lookupCap, timeout}` threaded out of `reconcileGroundTruthLines` and `groundTruthUntrackedLines`.
- Splits the previously-combined break condition so each cap is attributed to the right flag (same stop points and priority, no behavior change).
- Emits **one** `d.logf` breadcrumb from `(*Daemon).reconcileGroundTruth` when any cap fires, tagged with repo slug, ref count, and line count.

The **annotation itself is unchanged** — same lines appended, same silence-on-failure. This only adds a debug log line so the caps can be tweaked later.

## Testing
- `go build ./...` clean; `gofmt -l` clean.
- `go test ./internal/daemon -run GroundTruth -count=1` — all pass, including 3 new assertions that `lineCap` / `lookupCap` / `timeout` are set when their limit is reached.
- No live-app run: internal-only debug logging on the reconcile path, pure control-flow attribution + one `d.logf`, fully covered by unit tests. Internal instrumentation, so no CHANGELOG entry and no protocol change.

### DIFF
--- a/internal/daemon/ticket_reconcile_groundtruth.go
+++ b/internal/daemon/ticket_reconcile_groundtruth.go
@@ -57,6 +57,17 @@ type prStateFetcher func(repo string, number int) (state string, merged bool, ti
 // rather than a real PR reference.
 const groundTruthMaxPRNumber = 100000
 
+// groundTruthCaps records which best-effort limits the ground-truth
+// cross-check hit while building annotation lines. It drives a single debug
+// log line in reconcileGroundTruth; it never changes what gets annotated.
+type groundTruthCaps struct {
+	lineCap   bool // groundTruthMaxLines reached
+	lookupCap bool // groundTruthMaxLookups reached
+	timeout   bool // lookup budget (ctx) expired
+}
+
+func (c groundTruthCaps) any() bool { return c.lineCap || c.lookupCap || c.timeout }
+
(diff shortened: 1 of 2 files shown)


figgyster (approved):
This looks safe to merge. The cap attribution stays on the annotate-only path, preserves the existing silence-on-failure behavior, and now leaves a useful daemon breadcrumb when the guardrails actually bite.

I verified `go build ./...`, `go test ./internal/daemon -run GroundTruth -count=1`, the broader ground-truth focused daemon tests, and `git diff --check`. GitHub CI is green for the daemon/release jobs I could see; the Bench A/B job was still running, but this logging-only path doesn’t need to wait on benchmark theater in a raincoat.

PR #494, opened 2026-07-05
chore(harness): dedup helpers across the agent-cost-tooling stack

victorarias:
Post-merge cleanup surfaced by a `/simplify` review of the whole agent-cost-tooling stack. **Pure dedup — no behavior change:** same test coverage, same emitted `ATTN_VERDICT` JSON, same perf-scenario log strings and saved baseline files.

## Changes
- **daemon test** — drop `runGitDaemonTest`, a byte-identical duplicate of the same-package `runGitDaemon` (`worktree_naming_test.go`); reuse the existing helper and drop the now-unused `os/exec` import.
- **perf harness** — single-source `FIRST_FAILURE_MAX_LENGTH` in `common.mjs` instead of two copies previously kept in sync by comment only.
- **perf harness** — extract one private `buildVerdictEnvelope()` so `buildBaselineVerdict` / `buildColdWarmVerdict` / `buildLeakSoakVerdict` stop repeating the shared 6-key envelope (names/signatures unchanged).
- **perf harness** — extract `recordOrCompareBaseline()` and collapse the four copy-pasted "record-or-compare-and-log" blocks across the three perf scenarios into one call each.

## Not done (reviewed, deliberately skipped)
- Consolidating the two remote-URL parsers in `internal/git/resolve.go` — would change the pre-existing `ResolveRepoDir` parsing path; the split is sound as-is.
- Fanning out the ≤3 GitHub lookups in the reconcile ground-truth check — over-engineering for a capped, infrequent path.
- Pointing the untouched `bridge-perf.mjs` / `bridge-pty-bench.mjs` at the new `perfMeasure.mjs` — correct in spirit but edits files outside this stack's scope.

## Testing
- `go build ./...` + `go test ./internal/daemon -run GroundTruth` — green.
- `vitest run` over the harness suite — 37/37 on the two touched test files; 132 pass / 0 fail across the full harness suite.
- `node --check` clean on all 7 touched `.mjs` files.
- The record-branch value substitution in `recordOrCompareBaseline` was verified: `evaluation.comparison.value` is provably equal to each site's former local (it's the same `totalRssMb` fed into `evaluateRssBaseline`), so the log output is byte-identical.
- No live-app run: the perf **scenario** files have no unit coverage, so their correctness here is by-inspection behavior-preservation (harness scripts, not shipped product); the verdict/registry logic that *is* unit-tested stays green.

### DIFF
--- a/app/scripts/real-app-harness/common.mjs
+++ b/app/scripts/real-app-harness/common.mjs
@@ -95,6 +95,12 @@ Options:
 // naive line-oriented parsing never splits it.
 export const ATTN_VERDICT_PREFIX = 'ATTN_VERDICT ';
 
+// Contract: verdict.firstFailure is capped at this length (see ATTN_VERDICT_PREFIX
+// above) so it can never contain a newline or otherwise break the one-line
+// contract. Shared by every producer of a firstFailure field (scenarioRunner.mjs,
+// rssBaselineVerdict.mjs).
+export const FIRST_FAILURE_MAX_LENGTH = 300;
+
 export function formatVerdictLine(verdict) {
   return `${ATTN_VERDICT_PREFIX}${JSON.stringify(verdict)}`;
 }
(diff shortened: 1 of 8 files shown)


figgyster (approved):
This is safe to merge. The helper extraction preserves the verdict envelope and the perf baseline log shapes, and the daemon test cleanup just reuses the same-package git helper instead of carrying a duplicate.

I verified the touched harness modules with `node --check`, the focused harness tests (37 pass), `go test ./internal/daemon -run GroundTruth`, `go build ./...`, and `git diff --check`. CI was green except Bench A/B was still running when I checked, so keep an eye on that one before landing.

PR #495, opened 2026-07-05
fix(present): close window on submit + hover "+" follows pointer

victorarias:
Two fixes to the Present review window's reader UX, both proven with tests I ran myself.

## 1. The presentation window now closes when you submit

`PresentRoot.handleSubmit` calls `getCurrentWindow().hide()` after submitting a review round, but `capabilities/present.json` was missing `core:window:allow-hide`, so the IPC call rejected and the window stayed open. Adds `allow-hide` (+ `allow-is-visible`).

**This shipped broken twice** because no test drove the real present window — the unit test mocked `getCurrentWindow()` so `.hide()` always "succeeded". So this PR adds a real-app scenario (`scenario-present-submit-closes-window.mjs`) that drives the actual second Tauri window end to end: open a presentation on a shell session → click the pane-header chip → click Submit → confirm → **assert the window reports `visible` → `hidden`**. That last assertion is the regression guard; it only passes if `allow-hide` is present.

The scenario routes through a second automation bridge (`usePresentAutomationBridge`) that answers only `present_window_*` actions, so the main-window and present-window bridges never race on the broadcast request.

## 2. The gutter hover "+" follows the pointer after you add a comment

`DiffView` passed no `selectedLines` to `@pierre/diffs`' `MultiFileDiff`, so the library ran **uncontrolled**: the first gutter-"+" click committed an internal `selectedRange`, and `InteractionManager.placeUtilityFromSelection` then re-anchored the hover "+" to that stale range on every pointer move instead of following the mouse. Fix: pass a controlled `selectedLines` reflecting our own popup/draft state, cleared to `null` when neither is open — matching Pierre's own `DiffsHubViewer` pattern.

Adds an e2e repro (`diff-view-comment-interactions.spec.ts`): the hover test pins the "+" to a newly hovered line after a comment was added (was failing `4 != 5`, now green).

## Note on the scroll-jump bug (deferred)

The companion "clicking to add a comment scrolls the diff to the top" bug does **not** reproduce at the DiffView level — the three scroll-preservation tests in the same spec pass through all three triggers (gutter "+", line-click popup, background content refresh). Its root cause lives one level up (PresentRoot/DiffDetailPanel or a file-switch trigger), so it's a follow-up. The passing tests are left in as documentation of that negative result.

## Testing
- **Real-app (ran locally against `attn-dev.app`):** `real-app:scenario-present-submit-closes-window` → green; window went `visible` → `hidden` after submit.
- **e2e:** `diff-view-comment-interactions.spec.ts` (4 passed, hover green), `diff-view.spec.ts` (23 passed — drag-select-range and all comment paths intact).
- **Unit:** frontend suite 1380 passed; `tsc --noEmit` clean.


figgyster (approved):
This looks safe to merge. The present-window permission fix is backed by the right kind of regression test now — real packaged-window behavior, not a mock politely lying to us — and the controlled `selectedLines` change matches the `@pierre/diffs` state model instead of fighting it.

I ran the relevant frontend suite locally (`pnpm test` and the new DiffView Playwright spec), and GitHub CI is green. No blockers from me.

PR #496, opened 2026-07-05
fix(tickets): daemon-owned ticket_resume fixes racy Resume rollback

victorarias:
## Problem

Pressing **Resume** on a ticket whose bound session was closed/unregistered failed with `Session spawn arguments were not prepared.` and rolled back.

**Root cause:** racy *frontend* resume orchestration. `createWorkspaceSession` seeded a local zustand placeholder, then made two WS round-trips before `takeSessionSpawnArgs`. `syncFromDaemonSessions` pruned the unprotected placeholder in that window — near-deterministic while the chief churns `session_state_changed` broadcasts.

## Fix

Move the resume composite **daemon-side** as one request/result command, mirroring `delegate()`'s `register → pane → spawn → rollback`. The frontend now just sends `ticket_resume` and focuses the result; the fragile `ticketResume` orchestration + `resumePicker` plumbing are deleted.

- **protocol:** `TicketResumeMessage` / `TicketResumeResultMessage`; `ProtocolVersion` → 149 (constants + `useDaemonSocket` lockstep)
- **daemon:** `internal/daemon/ticket_resume.go` composite with rollback; a `ResumeAvailable` downgrade in `handleSpawnSession` so a stale/unwritten resume id falls back to the picker instead of exiting non-zero
- **frontend:** `sendTicketResume` request/result; `handleResumeTicket` focuses the returned session. `onResume(ticketId)` contract unchanged.

## Verification

- **Automated:** daemon unit tests (respawn closed session, already-running focus, picker fallback when transcript gone, validation, pane rollback on spawn failure, fresh-id mint, reply envelope); `useDaemonSocket` request/result tests; full daemon + frontend suites green; `go build`/`vet` + `tsc` clean.
- **Live dev-app (non-exemptible daemon-lifecycle surface):** new `real-app:scenario-ticket-resume` drives the **real** packaged app — bootstrap chief → delegate codex worker (bound ticket) → **close the worker session** (store row gone) → confirm the panel still offers Resume → click the real Resume button → daemon respawns the session under the ticket's id, ticket stays bound. **Passed** against `attn-dev.app`.
- Rebased onto latest `main` (resolved a conflict in `App.tsx` against the concurrently-landed ticket-pane-overlay PR #497); re-ran `go build`/`vet`, `tsc --noEmit`, `go test ./internal/daemon ./internal/protocol ./internal/store`, and the full frontend vitest suite (143 files / 1398 tests) green post-merge.

Ready to merge.


figgyster (commented):
This is the right fix shape: moving Resume into a daemon-owned composite removes the frontend placeholder race instead of papering over the zustand prune window. The rollback coverage is solid, and the request/result frontend path matches the rest of the protocol surface.

I don’t see a code blocker. I’m leaving this as a comment rather than an approval because the PR body says “Do not merge.” Local validation here: `git diff --check`, targeted daemon resume tests, `go test ./internal/daemon ./internal/protocol ./internal/store`, `pnpm test -- useDaemonSocket.test.tsx`, and `pnpm exec tsc --noEmit`. Visible GitHub checks are green now too, aside from the intentionally skipped Tauri job.

figgyster (commented):
Still clean after the merge-from-main refresh. The current PR patch is the same ticket-resume fix shape, with the App conflict resolved toward the daemon-owned `ticket_resume` path and the ticket overlay now calling that path directly.

No blockers from me. I’m keeping this as a comment, not an approval, because the PR still says “Do not merge.” Local re-validation passed: `git diff --check`, targeted ticket-resume daemon tests, `go test ./internal/daemon ./internal/protocol ./internal/store`, `pnpm test -- useDaemonSocket.test.tsx`, and `pnpm exec tsc --noEmit`. At last fetch, GitHub still had Frontend E2E (3/3) and Bench A/B running.

figgyster (approved):
Now that the PR is marked ready, this is safe to merge.

The daemon-owned `ticket_resume` path is the right seam for this: one request/result command, rollback stays daemon-side, and the frontend no longer has to juggle a placeholder session while broadcasts are flying around like drunk pigeons. The merge-from-main conflict resolved cleanly toward that model, and CI is green.

No blockers from me.

PR #497, opened 2026-07-05
feat(tickets): ticket chip + in-pane overlay on the agent pane header

victorarias:
## What

Surfaces a session's **bound ticket as an ambient chip** on its agent-pane header — title, status tint, and an unread dot. Clicking the chip expands the existing `TicketDetailPanel` as an **in-pane overlay** (over the terminal, pane-scoped, non-modal) to edit status / add a comment / edit the description / resume.

This makes the "reverse channel" — steering an agent by editing its ticket — **ambient**, instead of the current `Cmd+K → board → find the card` hop. It is the gating experiment for `ticket-rich-content`.

## Scope

**Frontend only.** No protocol / daemon / `ProtocolVersion` changes. Reuses `TicketDetailPanel` as-is and the existing ticket websocket actions (`sendTicketChangeStatus` / `AddComment` / `EditDescription`). 562 insertions.

- `PaneTicketChip` (+ css/test) — the header chip, mirrors `HeaderNudgeIndicator` (pointerdown/click stop-propagation so it never triggers the leaf drag handle).
- `boundTicketForSession` (`utils/tickets.ts`) — single home for the `assignee === sessionId` binding rule; the automation bridge now reuses it too.
- `SessionTerminalWorkspace` — overlay state (`ticketOverlayPaneId`), `useEscapeStack` wiring, ambient-header logic, and the overlay render.
- `App.tsx` — `ticket` on each workspace-session entry + a memoized `ticketActions` prop group.

## Focus ownership (critical pattern #6)

Every close path — **Escape** (via `useEscapeStack`), **chip re-click**, **panel ✕**, and **Resume** — funnels through `closeTicketOverlay() → focusActivePane()`, so focus returns to the active pane's terminal (main or utility split), never blindly to main.

## Coordination with `ticket-resume-fix`

The sibling branch owns the resume rewrite. This PR treats `onResume: (ticketId) => void` as the contract and calls it after closing the overlay. **Assumption:** the merged `handleResumeTicket` keeps that signature. Expect a trivial rebase on `App.tsx`'s `ticketActions` memo.

## Verification

**Automated (green):** full frontend suite `141 files / 1362 tests`, `tsc --noEmit` clean, `pnpm run build` clean. New specs: `PaneTicketChip.test.tsx` (render/unread/aria/no-bubble) and `SessionTerminalWorkspace.ticketOverlay.test.tsx` (chip renders for bound session only, overlay opens + fetches once, status routes through `onChangeStatus`, and Escape / chip-reclick / ✕ all close **and restore focus** via the mocked terminal handle).

**Live app (packaged, dev profile):** drove the running app end-to-end — created a session, bound a ticket, and verified via the automation bridge's per-pane `inputFocused`:
- ✅ chip renders as an ambient header for a bound-ticket session
- ✅ opening the overlay takes input focus **off** the PTY
- ✅ the panel actually mounts over the terminal
- ✅ **Escape** closes it and restores focus to the terminal
- ✅ **chip re-click** closes it and restores focus

**Not verified live:** the utility-**split** focus-restore leg. Driving a split in the harness hits a **pre-existing** race in `createSplitSession` (App.tsx: `createSession` → `takeSessionSpawnArgs` returns null before the store ingests the session → "spawn arguments were not prepared" rollback, leaving orphan shell sessions) — unrelated to this PR (which doesn't touch `createSplitSession`) and not reproducible without stealing OS focus. That leg is the same **unchanged** `focusActivePane()` reading `activePaneId` (existing production focus authority) and is covered by the unit specs above.


figgyster (approved):
This looks safe to merge. The ticket chip keeps the reverse-channel work pane-local, reuses the existing `TicketDetailPanel`/ticket actions instead of inventing a new path, and the focus-return rule is covered at the seam that matters.

I also like that the utility-split leg is called out honestly: the unchanged `focusActivePane()` authority is the right contract here, and the pre-existing split creation race doesn’t make this PR unsafe.

No blockers from me.

PR #498, opened 2026-07-05
refactor(present): remove standalone diff-review panels and comment popups

victorarias:
## Summary
- Removes the old main-window diff-review UI (`DiffDetailPanel`, `ChangesPanel`, the comment popup) and its backend support (`internal/daemon/ws_review.go`, `internal/store/review.go`, branch-diff review routing/protocol) now that guided review lives fully in the Present window.
- Bumps `ProtocolVersion` 148 -> 149 for the protocol surface removal.

## Verification
- `go build`/`go vet`/`go test` green; `make generate-types` idempotent; `tsc` clean.
- 1318 frontend unit tests green; 28 DiffView e2e specs green, including all popup-removal behaviors.
- Live-app verified on a fresh dev build: the real Present diff surface renders correctly (screenshot-confirmed) — the removal did not break it.
- `ProtocolVersion` confirmed in three-spot lockstep: `constants.go`=149, `useDaemonSocket.ts`=149, guard test passes.


figgyster (dismissed):
This is close, but I found two real blockers in the current head.

Required before merge:
- `app/scripts/real-app-harness/scenarioCatalog.mjs` still includes the `diff-review` scenario, but this PR deletes `app/scripts/real-app-harness/scenario-diff-review.mjs` and removes the `real-app:scenario-diff-review` package script. The catalog-driven `real-app:serial-matrix` path will still try to run that missing script and fail. Remove or replace that catalog entry.
- `PresentRoot` can still let an old pinned diff response overwrite the refreshed round UI when the same path remains selected. The fetch captures `requestedRoundId`, but the apply guard only checks `selectedPathRef.current === requestedPath`, so a stale response from the previous round can call `setDiffState(entry)` after `presentation_updated` reloads the round. The guard needs to include the current round/base/head identity too, not just the path.

Validation I ran on the latest head: `go test ./...`, `pnpm exec tsc --noEmit`, targeted `PresentRoot`/review-hash Vitest, `git diff --check`, and a catalog/script consistency probe. The tests are green; the catalog probe is the little rake in the alley.

figgyster (approved):
The blockers are fixed. The stale catalog entry is gone, and the Present diff guard now keys the async apply path to the active round instead of just the selected file path. The focused regression test is the right kind of trap for that bug — small, mean, and useful.

No blockers from me. Local Go/TS/Present tests pass; GitHub checks are green except the non-blocking bench job was still running at last fetch.

PR #499, opened 2026-07-07
Install attn skill for Copilot sessions too

harleenquinzell:
The attn skill (delegation, tickets, workspace context, notebook, workflow, and chief-of-staff guidance) was only ever installed to ~/.claude/skills/attn and ~/.agents/skills/attn. Copilot-driven sessions never received it, so they had no way to learn what "chief of staff" or attn's other vocabulary meant.

Add ensureAttnCopilotSkillInstalled(), writing to
~/.copilot/skills/attn (the directory the Copilot CLI already reads skills from), and wire it into settingsWithAgentAvailability's per-agent switch alongside Claude and Codex. Extend the install tests to cover Copilot: identical tree assertion, orphan pruning, and inclusion in the cross-agent identical-install check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-07-07]
+
+### Fixed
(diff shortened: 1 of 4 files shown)


figgyster (approved):
This is safe to merge. Copilot now gets the same bundled attn skill tree as Claude and Codex, and the tests cover the right failure modes: install location, stale-file pruning, and cross-agent tree equality.

I ran `go test ./internal/agent ./internal/daemon` and `git diff --check`; visible CI is green for the daemon/release-preflight jobs, with Bench A/B still running when I checked. No blockers from me.

PR #500, opened 2026-07-07
Fix terminal text selection getting stuck mid-drag

harleenquinzell:
startSelectionDrag() tracked an in-progress selection via document-level mousemove/mouseup listeners, with a fallback that force-finalizes on the next mousemove if event.buttons no longer reports the button held (to recover from a mouseup that happened outside the app). That buttons bitmask can go stale in WebKit (the engine Tauri uses on macOS) following a release outside the webview, or a real mouseup can be swallowed by a descendant calling stopPropagation() before it bubbles to document. Either way selectingRef stayed stuck true and the selection kept extending on the next pointer movement, even with no button held.

Add window 'blur' and 'pointercancel' listeners as a second, independent safety net, alongside a new cancelSelectionDrag() that finalizes (copies any threshold-crossing selection, or clears an unfinished click-only one) without needing a real MouseEvent to interpret a click/link/block hit against. This mirrors the existing pattern in
SessionTerminalWorkspace/leafDrag.ts, which already uses the same blur/pointercancel combination to guarantee pane-drag cleanup.


figgyster (approved):
This looks safe to merge. The extra blur/pointercancel cleanup is exactly the right kind of second net for WebKit’s “button state is fine, don’t look behind the curtain” nonsense, and it preserves the existing click/link/block handling by only using the synthetic cancel path when there’s no real release event to interpret.

Local `pnpm run build` and `pnpm test` passed here, and the visible GitHub frontend/E2E checks are green. I couldn’t do a macOS live-app drag smoke from this Linux runner, but the change is narrow and matches the existing pane-drag cleanup pattern.

PR #501, opened 2026-07-07
fix(chief): stop comparing agent direction changes against the original brief

victorarias:
## Why

The chief-of-staff ticket-comeback guidance told the chief to surface what the agent reported "against what the brief asked." This caused it to treat user-driven scope changes, plan revisions, and PR closures as deviations — comparing them against the original delegation brief and editorializing about whether the agent went rogue. The user drives delegated agents interactively, so direction changes should be reported neutrally.

## Summary

- **`internal/hooks/hooks.go:140`** — Replaced "against what the brief asked" with guidance that directional changes (scope revisions, plan pivots, PR closures, marking failed) are status updates reported neutrally, with the default assumption the user drove the change
- **`docs/plans/2026-06-28-delegated-ticket-awareness.md`** — Same revision for consistency with the plan doc that sourced this text

The rest of the paragraph (technical claims as agent assertions, prose review exception, small-and-reversible autonomy) is untouched.

## Test plan

- [x] `go build ./internal/hooks/` compiles clean
- [ ] Verify chief-of-staff no longer flags user-driven direction changes as brief deviations


figgyster (approved):
This is the right fix. The chief guidance now treats scope pivots/PR closures/failed marks as neutral ticket status instead of putting the agent on trial against the original brief, and the plan doc matches the shipped hook text.

I checked the tiny diff and ran `go build ./internal/hooks/` plus `git diff --check`; GitHub CI is green. No blockers from me.

PR #502, opened 2026-07-07
feat(present): multiple simultaneous comment drafts + pin diff scroll until user input

victorarias:
## Summary
- Multi-draft commenting in the present reader's `DiffView` (GitHub-style): per-anchor draft map keyed `${side}:${start}`, the hover "+" keeps following the pointer while boxes are open (controlled selection now always null — anchors are communicated via `lineAnnotations`/`renderAnnotation` instead), threads keyed by anchor key so the library's index-keyed annotation slots can't leak one draft's text into another, Escape closes the most-recently-opened draft first, and save/cancel is isolated per anchor.
- Cold-load scroll fix: `@pierre/diffs`' `Virtualizer` scroll-anchoring (`scrollFix`) captures its anchor against unmeasured row geometry during the throttled first render of a fresh Present webview and autonomously scrolls the pane to the bottom (instrumented live: `scrollTo 8792 → clamp 8265 → 5565`). The first click then remaps rows under a stale `scrollTop`, producing a visible jump and anchoring the comment on the wrong (nonexistent) line. `DiffView` now pins the scroller to top until real user input (wheel/touch/pointer/key) and then disarms permanently.

## Test plan
- [x] 6 new multi-draft e2e tests + 2 scroll-pin tests
- [x] The three previously-red scroll-jump repro tests now pass exactly
- [x] Full suite green: `tsc`, 1337 vitest tests, 164 e2e tests
- [x] Live-app verified on the `dev` profile with native CGEvents input — 4/4 cold-relaunch runs open at top and place the draft on the clicked line
- [x] Victor's manual pass
- Note: the e2e `scrollDown` helper now uses a real mouse wheel because programmatic scrolls are (correctly) indistinguishable from the library's autonomous scrolls and get pinned.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-07-07]
 
+### Added
+- **Comment boxes in Present are now GitHub-style.** Multiple comments can now be open at once: the hover "+" button keeps working while a draft is open, each comment saves or cancels independently, and Escape closes the most recently opened box first.
+
 ### Fixed
+- **The Present window now opens scrolled to the top.** A freshly opened diff no longer arrives scrolled away from the beginning, and the first click won't jump and misplace a comment — the diff stays anchored until you scroll it.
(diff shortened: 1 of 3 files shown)


figgyster (approved):
This looks safe to merge. The multi-draft state is properly anchored per line, and the scroll pin is scoped narrowly: it blocks the library's cold-load scroll weirdness until real user input, then gets out of the way.

I ran the targeted DiffView e2e file and `tsc --noEmit`; GitHub CI is green too. No blockers from me.

PR #503, opened 2026-07-07
feat(harness): InputDriver --window-title targeting + scroll command

victorarias:
## Summary
- Add a global `--window-title <substring>` option to the real-app harness InputDriver, backed by a unified `resolveWindow(bundleId:titleSubstring:)` (PID filter -> case-insensitive kCGWindowName substring match -> layer-0-first/largest-area selection) that replaces the duplicated mainWindowBounds/mainWindowID logic. Honored by `windowid`, `click`, `right_click`, `window_park`, and `scroll`; no-match errors list the titles found.
- Motivation: macOS Accessibility never enumerates secondary Tauri windows (e.g. the "attn — present" window), so CGWindowList + title filtering is the only reliable way to target them.
- Add a new `scroll` command: warps the cursor to `--relative-x/--relative-y`, posts pixel-unit scroll-wheel events split across `--steps` (~16ms apart, remainder rounded into the last step); positive `--delta-y` scrolls up per CGEvent convention.
- `macosDriver.mjs` gets a pure `withWindowTitleArgs` helper (unit-tested), trailing-opts plumbing on existing methods (back-compat preserved), and a `scrollWindow` method.

This is a harness-internal, scripts-only change — no CHANGELOG entry needed.

## Test plan
- [x] `swiftc` compiles clean
- [x] `pnpm test` — 136 passed
- [x] Live-verified against `attn-dev`: `windowid` correctly distinguishes the main window from the Present window; `scroll` moves the Present diff in both directions

### DIFF
--- a/app/scripts/real-app-harness/InputDriver.swift
+++ b/app/scripts/real-app-harness/InputDriver.swift
@@ -37,6 +37,10 @@ struct Options {
     var modifiers = [String]()
     var menuPath = [String]()
     var visiblePx: Int?
+    var windowTitle: String?
+    var deltaX: Double = 0
+    var deltaY: Double?
+    var steps: Int = 1
 }
 
 func parseOptions() throws -> Options {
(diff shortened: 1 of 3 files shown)


figgyster (approved):
This looks safe to merge. The CGWindowList/title-targeting path is the right fix for secondary Tauri windows, and the JS wrapper keeps the existing harness calls compatible while adding the new scroll surface cleanly.

I ran `pnpm test` locally: 142 files / 1336 tests passed. I couldn’t compile `InputDriver.swift` in this Linux runner, so I’m leaning on your macOS `swiftc`/live-harness verification for that part.

PR #504, opened 2026-07-07
docs(present): reader UX gap map (jaunt parity)

victorarias:
Maps the jaunt-parity gaps for the Present reader UX: must-have tour scroll/progress/keyboard/rail, critical annotations/grouping/diagrams, and deliberate simplifications we're accepting for now. Sequences the remaining work into 8 small, independently shippable slices.

Also updates `docs/vision/present.md` status checkboxes: Spine and Window shell are done, Change reader is in progress. Slice 1 of the plan is already in flight on a separate branch.

Docs-only change; no code touched.

### DIFF
--- a/docs/plans/2026-07-07-present-ux-gap.md
+++ b/docs/plans/2026-07-07-present-ux-gap.md
@@ -1,0 +1,154 @@
+# Plan: Present — reader UX gap map (jaunt parity)
+
+## Why / Alignment
+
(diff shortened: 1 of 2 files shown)


figgyster (approved):
This is safe to merge. The gap map is explicit about what gets jaunt parity, what stays deliberately simpler, and the slices are small enough to review without turning Present into a haunted migration fog.

Docs-only change; `git diff --check` is clean and the visible docs-filter CI passed. No blockers from me.

PR #505, opened 2026-07-07
docs(present): capture jaunt authoring-skill port in the UX gap plan

victorarias:
Adds an "Authoring guidance (the other half)" section to docs/plans/2026-07-07-present-ux-gap.md capturing the port of jaunt's authoring skill (voice rules, reading-order doctrine, annotation hygiene, thread: form, mermaid-by-default, validate discipline) plus attn-specific deltas (manifest frame instead of PR refs, doorbell/CLI instead of jaunt's server sentinels, future `attn present validate`, daemon-embed distribution). Inserted as slice 7 in the implementation list, bumping Submit and Polish to 8-9.

### DIFF
--- a/docs/plans/2026-07-07-present-ux-gap.md
+++ b/docs/plans/2026-07-07-present-ux-gap.md
@@ -79,6 +79,64 @@ Victor's priorities, verbatim buckets:
 - Collapse-on-review persistence across reload.
 - Deleted-line context in whole-file views (jaunt `lib/hunkOverlay.ts`).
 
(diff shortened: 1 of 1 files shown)


figgyster (approved):
This is safe to merge. The plan now captures the missing authoring half of Present clearly: voice, reading order, annotation/thread discipline, validation, and the attn-specific delivery loop are separated from jaunt instead of copy-pasted blindly.

No blockers. CI's doc-shaped path passed the relevant gate; the skipped heavier jobs make sense for a plan-only change.

PR #506, opened 2026-07-07
feat(present): continuous scroll-tour reader (slice 1 of the Present UX arc)

victorarias:
## Summary

- Present window now renders every manifest file as one continuous virtualized scroll (`@pierre/diffs` `CodeView`) in the presenter's reading order, replacing the per-file `DiffView` pane. Rail click / j/k scroll the tour; scrolling updates the rail highlight. Per-file presenter notes render as callouts above their file; round summary is pinned above the scroller; there's an end-of-tour footer.
- Inline commenting ported to per-anchor drafts keyed `${path}:${side}:${start}` — multiple open comment boxes across different files, Escape closes most-recent-first.
- Diffs for all files are fetched eagerly with bounded concurrency (4) and stale-round guards; the PR #498 round-guard regression test is preserved against the new fetch-all effect.
- Notable discovery encoded in code comments: `CodeView`'s controlled items reconcile by `version` — without bumping it, new annotations silently never render.
- New Playwright component harness + spec (PresentTour) covering rendering order, the scroll pin, rail-driven scroll, cross-file multi-draft, and gutter interactions.
- Known parity drop (deliberate, slice 1): comments anchored to non-visible lines no longer get `DiffView`'s "not visible" banner. Follow-ups tracked in `docs/plans/2026-07-07-present-ux-gap.md` (slices 2+).
- The committed real-app harness driver binary (`.build/attn-real-input-driver`) is rebuilt in this PR: PR #503 changed `InputDriver.swift` without rebuilding the tracked binary, leaving a stale binary whose fingerprint matched source. This rides along as the fix.

## Test plan

- [x] Unit tests: 1341 passing
- [x] New Playwright component spec (PresentTour): 10/10 passing
- [x] Live-verified in `attn-dev` (non-prod profile)
- [x] Reviewed line-by-line and live-verified by Claude on behalf of Victor

Claude-Session: https://claude.ai/code/session_01MvgEqMMAkmqJQezpgLNjzQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-07-07]
 
+### Changed
+- **Present is now a continuous tour, not a file-by-file viewer.** The Present window renders every file in the round as one scrollable sequence, in the presenter's reading order — scroll (or click a file in the rail) to move through the whole change set instead of clicking each file to swap the view. Per-file presenter notes appear as callouts above their file, the round summary stays pinned at the top, and an end-of-tour footer confirms how many files you've reviewed. Inline commenting works exactly as before, including multiple open comment boxes across different files at once.
+
 ### Added
 - **Comment boxes in Present are now GitHub-style.** Multiple comments can now be open at once: the hover "+" button keeps working while a draft is open, each comment saves or cancels independently, and Escape closes the most recently opened box first.
 
--- a/app/e2e/present-tour.spec.ts
+++ b/app/e2e/present-tour.spec.ts
@@ -1,0 +1,329 @@
+import { test, expect, type Page, type Locator } from '@playwright/test';
+
+/**
+ * PresentTour (the multi-file `@pierre/diffs` CodeView tour reader) component
+ * harness coverage. Mirrors e2e/diff-view-comment-interactions.spec.ts's
+ * conventions (goto, wait for `__HARNESS__.ready`, wait for real rendered
+ * content before interacting — Shiki highlighting is async, and CodeView
+ * mounts each file's `<diffs-container>` lazily as it scrolls into the
+ * virtualizer's window).
+ *
+ * The harness (test-harness/harnesses/PresentTourHarness.tsx) seeds a fixed
+ * 3-file manifest (src/alpha.ts, src/beta.ts, src/gamma.ts), each with several
+ * separate hunks so the tour has real scroll range even with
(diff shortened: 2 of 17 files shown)


figgyster (dismissed):
I’m going to block this one. The direction is good — continuous Present tour is the right shape — but the async edge cases have teeth.

Required before merge:

- The round-staleness guard does not hold on the real WebSocket path. `PresentRoot` now issues same-path diff requests across refreshed rounds with pinned `baseRef/headRef`, but `sendGetFileDiff` still correlates pending requests only by `path` (`app/src/hooks/useDaemonSocket.ts:4518`) and resolves results only by `data.path` (`:2507`). If round A has `src/foo.ts` in flight, round B starts and requests the same path, then A’s late `file_diff_result` can resolve B’s promise. The component-level `activeRoundKeyRef` check then sees “round B” and accepts stale diff content. The new round-guard test mocks `sendGetFileDiff` with independent promises, so it proves the React guard under a cleaner contract than the real wire has. Fix needs request correlation that includes a request id or at least the exact `{path, baseRef, headRef}` identity.

- Rail/j-k scroll requests made while the tour is still loading get dropped. `PresentRoot` renders the rail immediately and calls `requestScroll()` on click/key, but `PresentTour`’s scroll effect only depends on `[scrollToPath, scrollNonce]` and does `handleRef.current?.scrollTo(...)` while `CodeView` is not mounted during `!allSettled`. Once diffs finish loading, nothing retries the latest request. That breaks the advertised rail navigation in the normal async-loading path, not just a lab-grown edge case.

Local validation: `vitest` for `PresentRoot` passed (21 tests). The targeted Playwright component run was not clean here: `e2e/present-tour.spec.ts -g "Escape closes"` consistently timed out waiting for the second file’s gutter utility button. CI is green, so that may be environment sensitivity, but the failing local principal spec is another sign this slice needs one more pass rather than a rubber stamp.

No taste objections. Just two contract holes hiding in the loading/race alley, wearing tiny scrollbars.

figgyster (approved):
This fixes the blockers I raised. The diff responses are now request-correlated instead of path-correlated, the protocol version bump keeps old daemon/new frontend from quietly speaking past each other, and the pending rail/j-k scroll request is replayed once the tour actually mounts. That’s the right contract boundary.

Local focused checks passed for the wire-level stale-round case (`PresentRoot` Vitest + daemon request-id echo test), and GitHub’s frontend/daemon/E2E checks are green. My local Playwright runner still times out on the Escape multi-draft spec, but CI covers that path cleanly, so I’m treating that as runner weather rather than a merge blocker.

No blockers from me.

PR #507, opened 2026-07-07
Open Claude Code's OSC 8 hyperlink labels with Cmd+click in the terminal

victorarias:
Implements Option B from [docs/plans/2026-07-07-terminal-clickable-links.md](https://github.com/victorarias/attn/blob/main/docs/plans/2026-07-07-terminal-clickable-links.md): real OSC 8 hyperlink support end-to-end, so label-style links like Claude Code's "Learn more" (where the URL never appears on screen) open on Cmd+click.

## The three root-cause links, each fixed

1. **Nondeterministic emission** — attn PTYs only carried `TERM_PROGRAM=ghostty` when the daemon happened to be launched from a ghostty shell. Now pinned deterministically in `buildSpawnEnv` (and a leaked `TERM_PROGRAM_VERSION` is scrubbed), so Claude Code always emits OSC 8 hyperlinks. Honest: the VT core *is* ghostty.
2. **No URI accessor in the wasm API** — the vendored ghostty VT core parses and stores OSC 8, but the wasm surface only exported a boolean hyperlink flag. The compat patch now adds `ghostty_render_state_get_hyperlink_uri` / `ghostty_terminal_get_scrollback_hyperlink_uri`, mirroring the grapheme accessors; wasm rebuilt reproducibly via `build-ghostty-vt-wasm.sh` (SHA recorded in the vendor README).
3. **Null stub in ghostty-web's JS API** — rather than a 1.2MB pnpm patch on the minified dist, a small typed shim (`ghosttyHyperlinks.ts`) reads the new exports through the pinned package's runtime-stable internals, unit-tested against the real vendored wasm.

Frontend wiring checks OSC 8 first in the hover/click pipeline (labels can contain spaces, so text-derived detection can't find them), underlines the label extent by matching URI, keeps the existing gesture (Cmd/Ctrl+click opens, plain click selects), and routes `file://` targets through `openPath`.

## Testing

- Unit: `hyperlinkRangeAt` scan (4 cases) + shim against the real wasm (5 cases); full frontend suite 1350 passed; `go test ./internal/pty` green.
- Browser e2e: hidden-URI label — hover cursor states, plain click opens nothing, Cmd+click opens the hidden URI (18/18 in `terminal-interactions.spec.ts`).
- **Live packaged app** (new harness scenario `terminal-osc8-link`, run green against a dev install built from this branch): real PTY echoes `TERM_PROGRAM=ghostty`, native plain click produces no navigation, native Cmd+click (new `--modifiers` support in the HID driver) lands a GET on a scenario-owned local HTTP probe.


figgyster (approved):
This is the right fix: the terminal now treats OSC 8 as terminal state instead of trying to reverse-engineer a hidden URI from visible text. Good. Claude-style “Learn more” links were never going to be solved by making the URL regex squint harder.

I checked the PTY env pinning, the vendored wasm/export shim, the active-vs-scrollback row mapping, `file://` routing, and the real-app harness coverage. Local focused checks passed (`go test ./internal/pty`, terminal hyperlink Vitest coverage, TypeScript), and GitHub CI is green.

No blockers from me.

PR #509, opened 2026-07-08
feat(present): rail anatomy — per-file ± stats, comment chips, pinned Summary row

victorarias:
Extends the Present reader's left file rail, jaunt-style: per-file +N/−M line stats sourced daemon-side from `git diff --numstat` over the round's pinned base..head SHAs (exactly consistent with what FileDiff renders); a comment-count chip counting submitted comments + local drafts per file; a pinned Summary row that scrolls the tour back to the summary card and highlights when the tour is scrolled above the first file.

Protocol: PresentFile gains optional `additions`/`deletions`; ProtocolVersion bumped 150 → 151 (all three lockstep spots: constants.go, generated files, useDaemonSocket.ts PROTOCOL_VERSION).

Accepted limitation: binary files and renames are left stats-less (numstat `-` counts and `=>` rename paths are skipped, documented in code).

Testing: Go daemon tests (fixture repo incl. binary-file case + pure numstat parser test incl. both rename forms), jsdom PresentRoot tests (stats render/omit, chip counts, Summary click → scroll-request contract), full frontend suite green, tsc clean. Live-verified in attn-dev: stats match git numstat exactly on a real round, Summary row highlights at scrollTop 0 and yields to file rows on scroll, slice-2 J/K/R progress behavior intact.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ### Added
 - **Present now tracks your review progress and gives you a fuller keyboard model.** Mark a file reviewed with `R`, or just keep moving with `J`/`K` — leaving a file with `J` marks it reviewed automatically, jaunt-style. Progress shows in the file rail (a header count and thin progress bar, plus a checkmark and dimmed styling on reviewed rows) and in a new bottom drive bar with its own progress meter, keyboard hints, and the "Submit review" button (now bound to `S`). Each file's diff header also gets its own Reviewed toggle. The submit dialog lists any files you haven't walked yet as an advisory note — it never blocks submission. Marks are saved locally per round, so they survive a window reload but start fresh on a new round.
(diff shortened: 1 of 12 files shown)


figgyster (dismissed):
This is close, but the Summary-row navigation is wired to the CodeView scroller while the summary card itself is rendered outside that scroller.

Required before merge:
- Put the summary region inside the same scroll container that drives `onActivePathChange`, or otherwise make the Summary row’s click/highlight logic observe the actual container that contains both summary and files.

Right now `scrollTop === 0` means “the file scroller is at its first file,” not “we are above the first file in the summary region,” and `scrollTo({ top: 0 })` on that container can’t scroll to the summary card because the card is a sibling above `CodeView`. The tests mock `PresentTour`, so they don’t catch the geometry mismatch. Tiny UI rake, but it’s in the headline feature.

Validated: `go test ./internal/daemon ./internal/protocol`, `pnpm test` in `app`, and `git diff --check` passed locally. CI was still in progress when I checked.

figgyster (approved):
Thanks for the context — that changes the judgment. I was reading the Summary row as “scroll to a summary stop inside the document”; this slice intentionally uses the pinned-summary fallback, where Summary means “reset the file scroller to stop 0.” Given the `PresentTour` header, the CodeView item constraints, and the live attn-dev verification, that’s coherent.

No blockers from me. The protocol bump, daemon-side numstat wiring, rail stats/chips, and tests all line up, and CI is green now.

PR #511, opened 2026-07-08
feat(present): manifest author annotations — schema, anchor resolution, protocol carriage

victorarias:
## Summary

Slice 5A of `docs/plans/2026-07-07-present-ux-gap.md` (Present reader UX). This is the first of two annotation PRs — it lands the backend/protocol plumbing for author annotations; rendering them in the reader is slice 5B, a follow-up PR.

- Manifest file entries gain an `annotations` field: each entry pins a `note` (single comment) or `thread` (ordered list of comments) to exactly one of a substring `anchor` (minimum 3 characters), a single `line`, or a `start`/`end` range. The manifest decoder validates this strictly (exactly one anchor kind, no overlaps in author input).
- New `internal/present/resolve.go` — `ResolveAnnotations` resolves each annotation against the pinned head SHA via `git show`. A zero-match anchor is an error and the annotation is dropped; a multi-match anchor is a warning that pins to the first match and names the other matching lines; an out-of-bounds line or range is an error; an unreadable file collapses to a single file-level issue rather than one per annotation.
- Daemon: `present open` resolves annotations after pinning the round — any error-severity issue rejects the open, listing `path[index]: message` for each problem. Warnings are returned in the open result and printed to stderr by the CLI. `get_presentation_round` re-resolves annotations best-effort and carries them (`line_start`/`line_end`/`comments`) per file in the round payload; if resolution fails, the round still loads with annotations simply omitted.
- `attn present validate` resolves anchors locally when the manifest has annotations, exiting 1 on errors or unresolvable references.
- Protocol bump 152 → 153: new `PresentAnnotation` model, `PresentFile.annotations`, and `PresentOpenResult.warnings`.

## Verification

- `go test ./internal/present/... ./internal/daemon/... ./internal/protocol/...` — green
- `pnpm test` (vitest) — 1363 passed
- `tsc` — clean
- Live-verified on `attn-dev`: `validate` warning/error paths, a broken anchor rejected by the daemon on open, an ambiguous anchor producing a warning on both stderr and JSON, a round payload carrying all three anchor forms (substring, line, range) with correct resolved lines and thread ordering, and the reader loading annotated rounds with no regression.


figgyster (dismissed):
This is close, but there’s one contract hole big enough to drive the reader UI through.

`annotations` are documented as non-overlapping author input, but the implementation never enforces that after resolution. `validate()` only checks each annotation independently, and `ResolveAnnotations` appends every resolved `[LineStart, LineEnd]` without comparing it to earlier annotations for that file. So a manifest with `line: 2` and `start: 1 / end: 3` is accepted and carried over the protocol even though the new schema contract says those ranges cannot overlap.

I proved it with a temporary focused test in `internal/present`: the overlapping manifest parsed cleanly when it should have failed. The existing focused suite still passes:

- `go test ./internal/present/... ./internal/daemon/... ./internal/protocol/...` ✅

Required before merge: reject overlapping resolved annotation ranges at validate/open time, and add coverage for line/range and anchor-derived overlaps. Once that invariant is enforced, this slice looks good.

figgyster (approved):
The overlap hole is closed now. Static line/range overlaps fail at manifest validation, anchor-derived overlaps fail after pinned-head resolution, and the later overlapping annotation gets dropped instead of leaking into the protocol. That matches the contract I was blocking on.

I reran the focused backend/protocol suite:

- `go test ./internal/present/... ./internal/daemon/... ./internal/protocol/...` ✅
- focused repro for the old `line: 2` + `start: 1/end: 3` overlap ✅

No blockers from me. Some GitHub E2E/bench checks were still running at last fetch, but the changed fix is in the Present Go path and the relevant local checks are clean.

PR #512, opened 2026-07-08
feat(present): render author annotations in the reader — threads, reply, N/P

victorarias:
This is slice 5B of the Present reader UX plan (docs/plans/2026-07-07-present-ux-gap.md) — the reader side of author annotations, on top of the backend that landed in #511.

- PresentRoot synthesizes read-only ReviewComments from `round.manifest.files[].annotations` (author renders as "Claude"), prepended so author notes sit above reviewer replies at shared anchors; counts merge into the per-file chips.
- DiffCommentThread gains an optional caption and a Reply button on read-only threads; Reply opens a normal draft at the rendered line.
- Annotations that fall outside the visible diff re-anchor to the nearest visible additions-side line with a muted "refers to line N, outside the visible diff" caption (non-annotation comments keep the existing drop behavior).
- N/P hop between annotations in document order with wrap; DriveBar shows the hint only when the round has annotations.
- Tests: PresentTour.test.tsx (new, mocks only CodeView so grouping/anchor logic stays real) + 4 new PresentRoot tests.

Verified live on attn-dev: all annotation forms render, Reply opens drafts, outside-diff re-anchoring, N/P with wrap both directions.

Also fixes a flaky new test found during local verification: the N/P first-hop test waited on a prop being "defined" (always true) rather than on the parent's annotation-anchors state actually landing from the mock's async effect — retried the first keydown inside the waitFor poll instead.

Claude-Session: https://claude.ai/code/session_01MvgEqMMAkmqJQezpgLNjzQ


figgyster (approved):
This is safe to merge. The annotation path is wired through the existing comment/thread machinery instead of inventing a parallel reader lane, and the outside-diff re-anchor + N/P behavior has targeted coverage at the right seam.

I ran the focused PresentRoot/PresentTour tests and the frontend build locally; GitHub’s frontend/E2E checks are green too. No blockers from me.

PR #513, opened 2026-07-08
feat(present): mermaid diagrams via shared markdown renderer

victorarias:
## Summary

Slice 6 of `docs/plans/2026-07-07-present-ux-gap.md`: fenced ```mermaid``` code blocks now render as diagrams in Present round summaries, file notes, and annotation/comment bodies, via a new shared Markdown component (react-markdown + mermaid, lazy-loaded so the mermaid chunk only loads when a diagram actually appears in the content). Broken diagrams fall back to showing the error plus the original source. Rendering is theme-aware (light/dark).

`TicketDetailPanel` and `WorkspaceContextNavigator` are migrated onto the same shared Markdown component, so ticket descriptions and comments now also render as markdown.

Folded in an architectural fix needed to support notes as diagram-capable content: file notes moved out of `CodeView`'s `renderHeaderMetadata` into a synthetic first-line "note" annotation. `CodeView` computes item heights analytically against a single global `diffHeaderHeight` constant, so variable-height header content (like a rendered diagram) broke its layout math; annotation slots, by contrast, are DOM-measured by its `ResizeManager`. Notes with no visible diff line to anchor to (e.g. errored diffs) still fall back to header rendering. Note annotations are excluded from N/P hops and rail comment counts. The summary card is capped at 40vh with its own scroll so a diagram-heavy summary can't squeeze the tour out of the viewport.

Diagram async settle bumps `CodeView` item versions via a rAF-coalesced tick (the callback is threaded through context so renderer identity stays stable and doesn't cause a remount loop).

## Test plan

- [x] Live-verified on attn-dev: summary flowchart renders, broken-diagram fallback shows error + source
- [x] Live-verified: note-as-annotation rendering a sequenceDiagram
- [x] Live-verified: annotation stateDiagram, N/P hops skip note annotations
- [x] Live-verified: j/k navigation and reviewed marking still work with notes-as-annotations
- [ ] CI (frontend/unit/e2e) — pending


figgyster (approved):
This looks safe to merge. The shared Markdown renderer is a good consolidation point, and moving file notes into measured annotation slots is the right fix for CodeView’s height math instead of trying to make variable header content behave like a fixed constant with a fake mustache.

I ran the frontend build and the Vitest suite locally: `pnpm run build` and `pnpm test` passed. GitHub E2E was still running when I checked, so let that finish before merging.

PR #515, opened 2026-07-08
fix(tickets): intentional session close never crash-stamps the bound ticket

victorarias:
## Problem
When Victor intentionally closes a delegated session (pane close, delegate teardown, workspace close) whose last runtime state read mid-flight (working / pending_approval / launching), the ticket reconcile seam stamped the bound ticket **Crashed** — even in the common case where the agent had finished and left the ticket In Review. `hasForcedStopMark` already distinguished a close from a spontaneous death but was only used to word the reconcile comment, not to gate the crash stamp — and it is in-memory with a 30s TTL, so any late-running seam (startup reap after a daemon restart) lost the signal entirely.

## Fix
- `terminateSession` now also records the close **durably** on the session row (`sessions.closed_intentionally_at`, migration 64, idempotent columnExists guard) before the kill. The mark travels with exactly the state the restart reap reads back and is garbage-collected when the row drops.
- The session-end seam gates the Crashed stamp on `sessionCloseWasIntentional` (in-memory forced-stop mark OR durable row mark). An intentional close leaves the ticket where the agent last reported it; reconciliation still runs and posts its verdict with the existing clean-close framing.
- Startup recovery clears the mark on sessions it adopts as live (daemon died between mark and kill, worker survived), so a later genuine crash is still stamped.
- Spontaneous mid-flight deaths are unchanged: still Crashed.

Design note: the durable signal lives on the sessions row rather than a separate table or the ticket because (a) both seam call sites run with the row still present, (b) the restart reap reads exactly this row, and (c) the periodic sweep never crash-stamps, so nothing needs the signal after the row is gone.

## Tests
- Store: mark persists across a store reopen (daemon restart), clears, unknown session reads false.
- Daemon: intentional close at `working` via the real `unregisterSession` path → not crashed, ticket stays In Review, verdict comment posts with clean-close framing; restart-reap on a second daemon sharing only the store (fresh in-memory map) → not crashed; cleared mark re-arms crash stamping. Both behavioral tests verified to FAIL on pre-fix code.
- Live-verified on a throwaway profile with the branch daemon: closing a working claude session left the ticket un-crashed with a clean-close verdict; `kill -9` of the agent mid-run still stamped Crashed.


figgyster (dismissed):
This looks safe to merge. The durable close mark sits on the exact session row the restart reap already reads, and the tests cover the important traps: normal intentional unregister, restart with only persisted state, and clearing the mark so real later crashes still stamp `Crashed`.

I ran the focused daemon/store suites locally. No GitHub checks were reported for this branch, so no CI tea leaves to read here.

figgyster (approved):
Re-approved the current head after the merge-from-main refresh. The actual PR patch is still the same close-vs-crash fix, with the changelog rebased cleanly onto the newer Present entries.

Local daemon/store tests pass, and the visible required-looking checks are green. Bench A/B was still pending when I checked, but nothing in this follow-up changes the reviewed ticket/session path.

PR #516, opened 2026-07-08
feat(present): submit verdicts — approve, feedback, close

victorarias:
## Summary
- Slice 8 of the Present UX plan: the Present submit dialog gains three verdict actions — Approve, Submit feedback, and Close.
- Store migration 64 adds verdict tracking on `presentation_rounds` / `presentations`; protocol bumped to version 154 with the corresponding schema and daemon handler changes; feedback renderer and frontend wiring updated end-to-end.
- Follow-up fix in the second commit: repairs a stale/broken UI-automation-bridge selector for the submit dialog discovered during live verification, and adds per-action targeting so the harness can drive each verdict independently.

## Test plan
- [x] `go build ./...`, `go vet ./...`
- [x] `go test ./internal/store ./internal/daemon ./internal/present ./internal/protocol`
- [x] `pnpm test`, `pnpm exec tsc --noEmit`
- [x] Live-app verification against a packaged `attn-dev` build: all three verdict paths (feedback / approve / close) exercised end-to-end via the real-app harness automation bridge, confirmed via direct SQLite inspection of `presentation_rounds.verdict` and `presentations.status`.


figgyster (dismissed):
This looks safe to merge. The verdict state now has the important lifecycle edge covered: a fresh round reopens previously approved/closed presentations, so the reviewer chip does not disappear into the alley with a fake moustache.

Backend, protocol, generated types, frontend wiring, and the automation bridge all line up from what I can see. CI is green, and I re-ran the focused store/daemon/present/protocol checks plus the frontend test suite and typecheck locally.

Non-blocking follow-up: the new submit/close websocket results still resolve through static pending-action keys even though the result payloads carry `round_id` / `presentation_id`. I would tighten that to correlate by identity before this surface gets more concurrent/retry-heavy, but it does not need to hold this PR.

figgyster (approved):
Re-approving the refreshed head. The only substantive change since my last pass is the migration-base shuffle: the verdict migration moved behind the new sessions migration (`65` now), and the guarded migration path still holds together.

Focused local checks passed again (`git diff --check` plus the store/protocol/daemon/present Go tests), and the visible CI jobs are green apart from the Bench A/B job still running. No blockers from me.

PR #518, opened 2026-07-08
refactor(app): route dock-tile markdown through the shared Markdown renderer

victorarias:
Slice-9 follow-up from docs/plans/2026-07-07-present-ux-gap.md: WorkspaceDockTile maintained its own direct ReactMarkdown+remark-gfm stack. It now renders through the shared `app/src/components/Markdown` component (the one the Present reader and ticket surfaces use), so notebook/dock markdown tiles gain mermaid diagram rendering and we stop maintaining a parallel markdown setup.

- The tile's heading-slug and link-resolution overrides pass through unchanged via the `components` prop; it defines no `code` component, so the shared renderer's mermaid fence handling applies.
- The shared renderer adds one wrapper div: the only structure-sensitive CSS rule (`> :first-child` top-margin reset) now targets the new `.workspace-dock-tile-markdown` wrapper class; all other markdown styles are descendant selectors and are unaffected.
- New test renders a mermaid fence through the dock tile using the same mermaid mock pattern as Markdown.test.tsx; existing 21 dock-tile tests unchanged and green.

Verification: tsc clean; full frontend suite 1403 passed / 5 skipped. Live-app check of a rendered markdown tile on a dev build to follow before merge (will confirm in a comment).


figgyster (commented):
The refactor looks mechanically sound: the dock tile keeps its link/heading overrides while letting the shared Markdown renderer own Mermaid fences, and the focused dock-tile test plus `tsc --noEmit` pass locally.

I’m leaving this as a comment rather than green-stamping it because this is a UI rendering path and the repo rules call for live-app verification before merge. The PR already says that smoke is still to follow; once you’ve confirmed the rendered tile in the dev app and CI finishes, I don’t see a code blocker here.

figgyster (approved):
This is safe to merge now. The shared renderer path keeps the dock tile’s local overrides intact, the focused test/TS check passed on this head, CI is green, and Victor’s dev-app smoke confirms the actual tile renders Mermaid without layout weirdness.

No blockers from me.

PR #519, opened 2026-07-08
fix(present): keyboard reaches summary, default to summary on load, fix N-hop under-scroll

victorarias:
Slice-9 reader polish from docs/plans/2026-07-07-present-ux-gap.md. Three fixes:

- **K reaches the summary stop.** `moveSelection` clamped at file index 0, so the keyboard could never reach the pinned "00 · summary" row. K from the first file now drives the exact same `requestScroll(null)` path as the rail's summary-row click (when the round has a summary); K on the summary stays a no-op; without a summary the old clamp holds. J from the summary lands on the first file, as before.
- **Summary is the initial stop.** On round load, `activePath` defaulted to the first file — mismatching the actual scroll position (top = summary). It now defaults to the summary stop when the round has one.
- **First-N-press under-scroll.** The annotation-hop effect retried locating the anchor for at most 10 animation frames, but a smooth cross-file scroll routinely outlasts that, so the first hop into a far, unmounted file gave up before the anchor mounted. The retry now runs on a ~1500ms wall-clock budget.

Tests: new cases for the default stop (with/without summary) and K-to-summary (incl. no-op edges); six pre-existing tests that baked in the old first-file default updated to the new intended behavior. The retry-budget change is not unit-tested — PresentTour's CodeView is fully mocked in jsdom and no existing harness drives the rAF loop; covered by live-app verification instead (will confirm in a comment before merge).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-08]
 
 ### Fixed
+- **Present reader polish.** The keyboard (`K`) now reaches the summary stop from the first file, the summary is the initial stop when a round opens, and jumping to an annotation in a far file no longer under-scrolls on the first press.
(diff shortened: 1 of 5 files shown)


figgyster (commented):
Code path looks clean: the summary stop is now modeled consistently as `activePath === null`, `j/k` handles the summary edge without disturbing the no-summary case, and the annotation hop retry switches from frame-count roulette to a bounded wall-clock budget.

I ran the focused PresentRoot suite plus the frontend build locally, and GitHub's frontend/E2E checks are green. I’m leaving this as a comment rather than an approval only because the PR explicitly calls out live-app verification still pending for the N-hop under-scroll fix, and this repo treats that as merge evidence for UI timing work. Once that live check is confirmed, no code blockers from me.

figgyster (approved):
This is safe to merge now. The follow-up fixes the footer to use the same progress model as the rail — tour + other count, skipped files stay out — and the live-app verification closes the one timing gap I was waiting on.

No blockers from me.

PR #520, opened 2026-07-08
fix(chief): make ticket waiting runtime-aware

victorarias:
## Intent / Why

Chief-of-staff guidance told every runtime to start `attn ticket inbox --watch` after delegating. That is the supported path for Claude, whose Monitor observes watcher output. For Codex, the watcher consumed ticket events without waking a new turn, clearing the unread state before attn could deliver its existing ticket nudge.

This change aligns the injected guidance with the notification capability attn already uses: self-monitoring runtimes watch; nudge-driven runtimes wait for attn and consume their inbox when nudged.

## Summary

- make `ChiefGuidance` select its delegated-ticket waiting clause from `HasSelfMonitor` while keeping the rest of the chief prompt shared
- preserve Claude's existing `ticket inbox --watch` Monitor instruction and give Codex explicit nudge-to-inbox guidance
- keep notebook guidance and the bundled delegation reference consistent with the runtime split
- strengthen the real-app chief benchmark so Claude must arm a watcher, while Codex must remain watcher-free, receive the fixed doorbell, and react

## Test plan

- [x] `go test ./internal/hooks ./internal/agent ./internal/daemon ./internal/ticketnotify ./cmd/attn`
- [x] `node --check app/scripts/real-app-harness/scenario-chief-ticket-watch.mjs`
- [x] `git diff --check`
- [x] Build, sign, and install the isolated app with `make dev`
- [x] Packaged Codex chief: delegated with no watcher, received the fixed ticket doorbell, and began handling `ticket inbox`
- [x] Packaged Claude chief: launch process carried the preserved Monitor guidance

The Claude instruction-following benchmark did not reach attn delegation in its latest run because Opus chose a native teammate; the harness stopped at its existing `did-not-delegate` boundary. The watcher implementation itself is unchanged, and its CLI/daemon regression coverage remains green.


figgyster (approved):
This looks safe to merge. The important bit is the seam: chief ticket waiting now follows the runtime capability instead of pretending Codex can self-monitor like Claude, and the tests pin both launch-time guidance and live notebook guidance.

I ran the focused Go packages, `node --check` on the real-app scenario, and `git diff --check`; all passed. Bench A/B was still pending at my last fetch, so let that finish its lap before merging — tiny stopwatch goblin and all.

No blockers from me.

PR #522, opened 2026-07-08
feat(tickets): move any ticket by id with attn ticket status --ticket <id>

victorarias:
## Summary
- Adds an optional `ticket_id` to the `set_ticket_status` protocol message so a ticket can be moved by id, not just via the session's currently-bound ticket. Protocol version bumped 154 -> 155.
- `attn ticket status <state> --ticket <id>` skips session-bound resolution in the daemon and moves the referenced ticket permissively. This is a deliberate product decision: no ownership/authorship gating is added on the by-id path (any session can move any ticket it names explicitly). The acting session remains the recorded activity author, so self-reports, attn reconciliation stamps, and manual moves-by-id stay distinguishable in ticket activity history.
- CLI flag parsing, daemon handler branch, docs (delegated-agent reference), and CHANGELOG entry included.

## Test plan
- [x] `go test ./internal/daemon -run TestSetTicketStatus -v` (new by-id daemon tests pass: bound-ticket move, unbound-session move-by-id, unknown-ticket error)
- [x] `go test ./cmd/attn -run TestParseTicketStatusArgs -v` (flag parsing incl. `--ticket` before/after positional)
- [x] `gofmt -l .` clean, `go build ./...` clean
- [x] Live-verified: CLI `attn ticket status --ticket <id>` against a throwaway daemon profile, plus packaged dev-app UI check that the move reflects correctly


figgyster (dismissed):
This looks safe to merge. The by-id path stays explicit (`--ticket`), keeps attribution on the acting session, and the old bound-ticket self-report path is still covered.

I ran the focused daemon/CLI tests, `gofmt`, `go build ./...`, and `git diff --check`; all passed. GitHub still had E2E/bench jobs pending when I checked, but no blockers from me.

figgyster (approved):
Still safe after the rebase. The PR patch is the same ticket-status change, now correctly riding protocol version 156 after `main` claimed 155.

I reran the focused daemon/CLI tests, `gofmt`, `go build ./...`, and `git diff --check`; all passed. CI was still coming up on the new head when I checked, but no blockers from me.

PR #523, opened 2026-07-08
fix(tickets): keep chief awareness across session changes

victorarias:
## Intent / Why

Chief-of-staff ticket awareness was attached to the session that created a delegation. Transferring the chief role therefore left the new chief without the ticket participation or cursor needed to see later reports, while the retired session could remain a notification target. The chief role is durable, so its ticket ownership and unread position need to be durable too.

## Summary

- add durable `ticket_role_owners` records for chief-delegated tickets while retaining concrete session IDs as event authors for audit
- read a chief session through both its ordinary session identity and the durable chief role identity, merging overlapping events once while advancing both relevant cursors
- resolve role notifications to the current chief session and clear stale countdown/backstop state when the role transfers
- keep assignee and explicit-subscriber behavior session-scoped and unchanged
- preserve runtime-specific delivery: self-monitoring agents may drain `ticket inbox --watch`, while other runtimes receive attn nudges and use one-shot inbox reads

## Migration

Migration 66 creates the role-ownership table and backfills active tickets created in the pre-fix born-assigned delegation shape. It does not create, copy, or advance a role cursor, so unread events remain unread across upgrade and handoff. Ordinary tickets assigned after creation are not reclassified.

## Test plan

- [x] Chief A consumes an update, transfers the role to B, and B receives the next unread update plus a nudge
- [x] B receives no skipped or duplicate events, and A receives no further chief-role nudge
- [x] overlapping explicit subscription and chief-role ownership still deliver once
- [x] migration backfills an active legacy delegation without seeding its cursor and leaves ordinary assignment alone
- [x] Claude/Codex runtime-specific waiting guidance remains covered
- [x] `go test -race ./internal/store ./internal/ticketnotify ./internal/daemon -run "Test(Migration66|ChiefTicketContinuity|ChiefRoleAndExplicit|TicketInboxConsumes|TicketSubscribeLifecycle|TicketBackstopSuppressedAfterWatchDrains|NotebookGuideUses)" -count=1`
- [x] `go vet ./internal/store ./internal/ticketnotify ./internal/daemon`
- [x] `go test -p 1 ./... -count=1`
- [x] signed `attn-dev.app` install and packaged-app ticket lifecycle scenario

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-08]
 
 ### Fixed
+- **Your chief keeps every delegated ticket when you switch chief sessions.** Unread ticket activity now belongs to the durable chief-of-staff role instead of the session that happened to create the ticket. The new chief picks up at the same unread position and receives future nudges, while the retired chief stops receiving role notifications. Existing active delegations are carried forward automatically.
 - **Present reader polish.** The keyboard (`K`) now reaches the summary stop from the first file, the summary is the initial stop when a round opens, and jumping to an annotation in a far file no longer under-scrolls on the first press.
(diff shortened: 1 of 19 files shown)


figgyster (approved):
This looks safe to merge. The chief ticket identity is now durable without muddying the audit trail: events still come from the concrete session, while the role cursor and delivery target survive a handoff.

I checked the role-owner migration/backfill shape, the merged personal+chief inbox consumption, cursor monotonicity, and the runtime-specific nudge/watch path. `go test ./...` passes locally; GitHub still had a few daemon/bench jobs pending when I last checked, so let those finish their lap around the block.

No blockers from me.

PR #524, opened 2026-07-08
fix(tickets): revive crashed ticket to working when its session reloads

victorarias:
## Summary
- A session stamped Crashed can be brought back to life (dead-pane reload, ticket Resume, CLI re-register, or recovery adopting a still-live worker), but nothing moved its bound ticket out of the terminal Crashed column — Victor had to flip it by hand.
- The same authority that stamps Crashed now un-stamps it: every session-revival seam calls `reviveCrashedTicketsForSession`, which moves the ticket crashed -> working (attn-authored, like the crash stamp) and puts it back on the crash/reconcile seam's radar, so a later genuine crash re-stamps it.
- Archived crashed tickets stay archived (the user dismissed them), and non-crashed columns are never touched.

## Test plan
- [x] `go test ./internal/store ./internal/daemon` (green)
- [x] Live-tested on a throwaway `ATTN_PROFILE` (non-prod)
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-07-09]
+
+### Fixed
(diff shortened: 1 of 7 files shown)


figgyster (approved):
This looks safe to merge. The fix puts the ticket lifecycle back under the same daemon authority that crash-stamps it, and the guardrails are right: archived tickets stay buried, non-crashed columns are untouched, and revived tickets get back onto the crash/reconcile radar for a later real failure.

I ran `go test ./internal/store ./internal/daemon` locally; the relevant daemon/release CI was green when I checked, with only the benchmark job still running.

PR #525, opened 2026-07-09
feat(present): reclaim vertical space — compact topbar + summary fold

victorarias:
## Why

Victor's review-window feedback: there's no room to read the code. The stacked title/changes/round/drift header burned ~20% of the height, and the summary card (often with a mermaid diagram) is pinned outside the CodeView scroller so it could never be scrolled away — together ~60% of the window was chrome.

## What

- **One slim topbar** replaces the header stack: title · kind · repo path, with round seq, pinned shas, draft/submitted status on the right. Repo-drift is now a dismissible amber pill in the topbar (full tooltip preserved) instead of a full-width banner.
- **Summary folds away off the Summary stop.** The card is a flex sibling of CodeView (the library owns its scroll container and positions items at analytic offsets, so the card can't live inside the scroller). Instead its visibility now derives from tour position: expanded on the pinned Summary stop (`activePath === null` — initial load, rail Summary click, K from the first file, scroll-to-top), folded the moment you move into a file. Animated max-height/opacity collapse, kept mounted so it can animate, `prefers-reduced-motion` honored, `aria-hidden` while folded.
- **End-of-tour progress strip removed** — it duplicated the rail and pinned more chrome at the bottom.
- Root padding and DriveBar spacing tightened.

Net effect: once you're reading files, everything except the slim topbar and the drive bar is code.

## Verification

- `pnpm test` PresentRoot+PresentTour: 66/66 (new tests: summaryVisible wiring from PresentRoot, fold class/aria states in PresentTour, drift-pill behavior)
- `tsc` clean
- Live packaged app: `real-app:scenario-present-flow` green end-to-end against a dev build of this branch (session → chip → present window → comment → submit → feedback CLI)
- Manual fold captures (stop-0 vs after J) to follow in a PR comment — the target machine's display is asleep; a watcher grabs them when it wakes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-07-09]
 
+### Changed
+- **The Present review window reclaims vertical space for the diff.** The title, kind, repo path, round info, drift notice, and comment count now share one compact line at the top instead of a tall stacked header. The summary card folds away once you scroll into the tour and reappears when you jump back to the Summary stop, and the redundant "End of tour" strip — which just repeated the drive bar's reviewed count — is gone.
+
 ### Fixed
 - **Reloading a crashed session moves its ticket back to Working automatically.** When a delegated session died mid-run its ticket was stamped Crashed — but bringing the session back to life (reloading the dead pane, resuming the ticket, or the daemon re-adopting a still-live worker after a restart) left the ticket sitting in the Crashed column until you moved it by hand. Reviving the session now flips the ticket back to Working on its own, and crash detection is re-armed so a later genuine crash is still stamped. Intentional closes are still never treated as crashes.
 
--- a/app/e2e/present-tour.spec.ts
+++ b/app/e2e/present-tour.spec.ts
@@ -88,12 +88,26 @@ async function findContainerByTitle(page: Page, path: string): Promise<Locator |
   return null;
 }
 
-/** Opens a draft on the given line via the gutter "+" (hover then click), scoped to one file's container. */
+/**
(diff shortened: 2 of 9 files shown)


figgyster (dismissed):
This is close, but the summary-fold trigger is wired to a state that the CodeView scroll handler immediately resets at the top of the first file.

Required before merge:
- Keep the summary collapsed once the user intentionally enters a file, including the first file at CodeView scrollTop 0. Right now `requestScroll(firstFile)` sets `activePath`, but the next `handleScroll(0)` reports `null`, so the summary can pop back open instead of freeing the reading space this PR is meant to reclaim.

I verified the targeted PresentRoot/PresentTour tests and `tsc --noEmit` locally; they pass, but they don't cover this first-file/top-of-scroller interaction. Tiny rake, very much in the path of the feature.

figgyster (dismissed):
The original fold bug looks addressed now — Summary is explicit-only, and the targeted PresentRoot/PresentTour tests plus `tsc --noEmit` pass locally.

Still red, though: the required Frontend E2E check is failing on current head. The failure is in `e2e/present-tour.spec.ts`: one test still waits for `present-tour-footer`, but this PR intentionally removed that footer, and another PresentTour gutter/draft test times out in the same shard.

Required before merge:
- Update/fix the PresentTour E2E expectations for the removed footer / new layout, and get the Frontend E2E shard green.

This is no longer the old blocker; it’s the test suite waving a different little red flag.

figgyster (approved):
This is safe to merge now. The Summary fold fix holds, the stale footer E2E expectation is gone, and the gutter retry is scoped to the flaky hover/click path instead of watering down product behavior.

Checked locally: PresentRoot/PresentTour unit tests, the full `e2e/present-tour.spec.ts`, `tsc --noEmit`, and `git diff --check`. GitHub’s Frontend and Frontend E2E checks are green too.

No blockers from me.

PR #527, opened 2026-07-10
feat(present): collapsible summary card + fix dead passive-scroll fold

victorarias:
## Why

Victor: the summary/diagram card still eats 30–40% of the present window's vertical space and scrolling doesn't fold it — it needs to be collapsible, and to auto-collapse (animated) when you start scrolling the files.

Investigating live showed the #525 fold-on-scroll was **never working in the packaged app**: the takeover/cold-window-pin listener effect has `[]` deps, but on first render diffs are still loading over the WS, CodeView isn't mounted, and `containerRef.current` is null — the effect bails and never re-runs. Passive scroll never arms `userTookOverRef`, so `handleScroll` returns early forever: no fold, no rail tracking, and the cold-window scroll pin was dead too. Unit tests and the e2e harness feed pre-loaded diffs (`allSettled` true on first render), which is why CI stayed green.

## What

- **Fix:** the listener effect now depends on `allSettled`, attaching the moment CodeView mounts (and re-attaching if diffs reload).
- **Disclosure card:** the summary keeps a slim always-visible "Summary" strip with a chevron; the body folds. Click toggles it manually.
- **Dead zone removed:** wheel-down with the cursor over the card collapses it once the card's own inner scroll is exhausted (tall summaries stay readable).
- Navigation semantics unchanged: arriving at any file stop collapses; rail-Summary / K-to-top expands; manual toggle wins until the next navigation.

## Verification

- Unit: 75/75 (PresentTour + PresentRoot), including a regression test for the loading→settled listener attach — **flip-checked**: it fails with the old `[]` deps.
- E2E: 14/14 `present-tour.spec.ts`, including a new browser-level fold spec that uses the harness's `deferred=1` mode to reproduce the live loading transition, real `mouse.wheel`, and the toggle click.
- Live packaged app (attn-dev rebuilt from this branch, synthetic CGEvent wheel): wheel over the code folds the card and moves the rail highlight; K back to top re-expands; wheel over the card collapses it. Screenshots in the session log.
- `tsc --noEmit` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ### Added
 - **Durable ticket handovers keep plans alive after the producing agent stops.** `attn ticket handover` copies one or more Markdown files into a visible `tickets/<ticket-id>/` Notebook directory, can change ticket state in the same operation, records a retry-safe handover receipt, and returns the canonical paths. Ticket reads now derive their current artifact list directly from that directory, so ordinary edits, renames, and deletions are reflected without reconciliation. The ticket panel can hand over files, open them in the Notebook editor, copy their paths, rename them, and delete them. Chiefs and delegated agents receive role-specific guidance for continuing work from the same canonical plan.
 
+### Fixed
+- **The Present window's summary card now actually collapses when you scroll the code.** Wheel-scrolling the diff — including wheel gestures over the card itself — folds the summary to a slim strip, which you can also collapse or expand by hand at any time; previously, scrolling never folded it.
+
 ---
 
 ## [2026-07-09]
--- a/app/e2e/present-tour.spec.ts
+++ b/app/e2e/present-tour.spec.ts
@@ -382,6 +382,37 @@ test.describe('PresentTour draft and comment gutter interactions', () => {
   });
 });
 
+test.describe('PresentTour summary fold', () => {
+  // Browser-level regression for the listener-never-attached bug: the harness's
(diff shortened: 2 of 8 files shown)


figgyster (dismissed):
This is close, but I’m blocking on a regression in the PresentTour browser spec.

Required before merge:
- Fix the new layout/summary-collapse interaction so `PresentTour multiple simultaneous drafts across files › opening a draft on one file and one on another keeps both open independently` passes again.

Evidence from the PR-scoped checkout:
- `pnpm exec tsc --noEmit` ✅
- `pnpm exec vitest run src/components/PresentTour/PresentTour.test.tsx src/components/PresentRoot/PresentRoot.test.tsx` ✅ 75/75
- `pnpm exec playwright test e2e/present-tour.spec.ts` ❌ 13/14, timing out at `e2e/present-tour.spec.ts:295` while trying to click the first form’s `.save-btn`
- The same focused spec passes on `origin/main` in the same runner, so this isn’t just local Playwright weather wearing a false moustache.

CI is green, which is interesting, but the local A/B makes this look like a real interaction with the PR’s layout/virtualized draft behavior. I’d fix or harden that path before shipping the fold.

figgyster (approved):
The blocker is fixed. Pre-collapsing the summary in the draft/gutter specs is the right boundary: those tests aren’t trying to prove the fold animation, and the fold still has dedicated coverage.

I reran the relevant local checks, including `present-tour.spec.ts --repeat-each 3` (42/42), and the previously failing multi-draft path held up. CI E2E was still running at my last fetch, but this is safe to merge from my side once GitHub finishes green.

PR #528, opened 2026-07-11
fix(workspaces): remove orphan agent panes

victorarias:
## Intent / Why

Closing `json-inc-fable` terminated its agent but left an agent pane referencing the deleted session. Moving that stale pane created a sessionless workspace that repeatedly appeared and disappeared in the sidebar as session and workspace state reconciled.

This change makes session removal and pane removal one coherent lifecycle transition, and repairs orphan panes left by older runs.

## Summary

- persist the pane-free workspace layout before terminating or unregistering its session, so later events cannot replay the closed pane
- reject cross-workspace moves of ready agent panes whose backing session no longer exists
- remove historical orphan agent panes during startup reconciliation while preserving legitimate pending spawns and docked tiles
- add regression coverage for close ordering, stale-pane moves, and startup repair
- document the lifecycle invariant and user-visible fix

## Test plan

- [x] `pnpm --dir app test` — 1,426 passed, 5 skipped
- [x] `pnpm --dir app run real-app:scenario-workspace-close-one-session-keeps-selection` against the packaged `attn-dev` app
- [x] `git diff --check`

## Out of scope

- Production data cleanup before this build is installed; startup reconciliation performs that repair when the updated daemon starts.
- Broader workspace/sidebar redesign.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-07-11]
+
+### Fixed
+- **Closed sessions no longer leave workspace tiles that reappear in the sidebar.** Closing now removes the pane before announcing the session's departure, stale agent panes cannot be moved into another workspace, and startup repairs any orphan panes left by older runs.
+
 ## [2026-07-10]
 
 ### Added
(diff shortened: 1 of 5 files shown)


figgyster (approved):
This looks safe to merge. The lifecycle ordering is the important bit here: the pane-free layout is now persisted before session teardown can broadcast anything, and the startup reconciliation covers the old stale state without eating pending spawns.

I ran `go test ./internal/daemon` and `git diff --check` locally; the visible daemon/Linux checks are green. Bench A/B was still pending when I signed off, but that shouldn’t block this fix unless it catches something weird in the alley.

PR #529, opened 2026-07-11
fix(present): keep Codex waiting for review feedback

victorarias:
## Intent / Why

Codex currently ends its turn after opening a Present review and depends on a later nudge to retrieve feedback. That nudge can be suppressed while the user is watching the session, leaving the review loop stalled even though the agent is otherwise available.

This makes the supported Codex workflow synchronous: the Present command stays in the foreground until the reviewer submits or closes the round, then returns that outcome directly to the same turn.

## Summary

- update the embedded Present guidance so Codex always runs each round with `attn present --wait` and continues from the returned result
- reshape the packaged-app Present scenario to keep one CLI child alive from open through submission and assert that the same process receives the feedback
- explicitly target the harness profile socket so an attn-managed shell cannot leak its production socket into the isolated dev run
- document the agreed scope: synchronous Codex handback changes, with nudge delivery left unchanged

## Test plan

- [x] `go test ./cmd/attn/... -run Present -count=1`
- [x] `go test ./internal/daemon/... -run Present -count=1`
- [x] JavaScript syntax checks for the modified real-app harness files
- [x] Build, sign, install, and launch the isolated `attn-dev` app
- [x] Exercise two real Present rounds with foreground `--wait`; both returned their handback synchronously and round two was approved
- [ ] Re-run `real-app:scenario-present-flow` once native Present-window enumeration is working locally. The waiting CLI opened correctly and the daemon logged the Present window loading its diff, but native window enumeration timed out before the scenario could submit.

## Out of scope

Nudge scheduling and delivery semantics are unchanged. Non-Codex callers may continue using the explicit feedback command when they intentionally manage handback themselves.

### DIFF
--- a/app/scripts/real-app-harness/scenario-present-flow.mjs
+++ b/app/scripts/real-app-harness/scenario-present-flow.mjs
@@ -3,11 +3,11 @@
 /**
  * Real-app scenario: the full "Present" loop in the packaged app.
  *
- * An agent (via the attn CLI, `attn present`) opens a presentation from a
+ * An agent (via the attn CLI, `attn present --wait`) opens a presentation from a
  * manifest -> a notice chip appears in the main window -> clicking the chip
  * opens the second Tauri window (titled "attn — present") -> a reviewer
- * submits a round over the real daemon socket -> the authoring agent reads
- * the feedback back (`attn present feedback --json`).
+ * submits a round over the real daemon socket -> the same blocking CLI call
+ * returns the feedback to the authoring agent.
  *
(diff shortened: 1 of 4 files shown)


figgyster (approved):
This looks safe to merge. The harness now proves the important contract: the same `attn present --wait` process opens the round, stays alive through review, and receives the submitted feedback instead of relying on a later nudge.

I also like the explicit harness socket targeting — boring in exactly the right way, which is how socket isolation should dress.

Checked locally: targeted Present Go tests, JS syntax for the modified harness files, and `git diff --check`. GitHub CI is green.

PR #535, opened 2026-07-12
fix(notebook): native undo/redo in the editor + dead ⌘⌥N shortcuts

victorarias:
## Summary

First PR of the notebook editor polish plan (`docs/plans/2026-07-11-notebook-editor-polish.md`, committed in this branch's first commit).

- Removes the native Edit > Undo menu item (Redo was already removed for ⇧⌘Z/`toggleZoom`) so ⌘Z/⇧⌘Z reach the WebView; CodeMirror's history handles undo/redo in the notebook editor; ⇧⌘Z still resolves to `terminal.toggleZoom` outside editable targets. Note: React-controlled inputs (e.g. the finder input) never had working native undo — controlled value updates break WebKit's undo stack — so removing the menu item is not a regression there.
- Fixes dead ⌘⌥N/⇧⌘⌥N notebook shortcuts in the packaged app: with ⌥ held, WebKit reports the dead-key character as `e.key`, so key-only matching never fired; added `code: 'KeyN'` fallbacks (same pattern as the Digit1..9 workspace shortcuts).
- New packaged-app scenario `real-app:scenario-notebook-editor-undo` (native typing + disk-based undo/redo assertions + no-zoom assertion). Harness robustness fixes: presence probes only treat the bridge's "selector not found" error as absence; probe selectors scoped to `.terminal-wrapper.active` (warm-set workspaces keep stale finders mounted); Esc posted as keycode 53 (InputDriver's `--key` map only covers printable keys — `pressKey('Escape')` never worked).

## Test plan

- [x] `pnpm run test` — 1431 tests green
- [x] `pnpm run build` — tsc + vite build clean
- [x] `cargo build` (src-tauri) — clean
- [x] Verified live in attn-dev: manual ⌘Z/⇧⌘Z in the editor, ⇧⌘Z zoom outside editors, plus both packaged scenarios (`notebook-editor-undo`, `notebook-tile-finder`) green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-07-12]
+
+### Fixed
+- **Undo and redo now work in the Notebook editor.** ⌘Z and ⇧⌘Z were swallowed by the native Edit menu in the packaged app and never reached the editor; they now undo/redo your edits when the editor has focus. ⇧⌘Z still zooms the active pane when you're not in a text field.
+- **⌘⌥N and ⇧⌘⌥N open the Notebook again.** With Option held, macOS reports a dead-key character instead of the letter, so the notebook shortcuts never matched in the installed app and the keystroke fell through into the terminal. They now match on the physical key.
+
 ## [2026-07-11]
 
 ### Fixed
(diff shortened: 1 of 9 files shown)


figgyster (approved):
Looks good — this hits the packaged-app shortcut trap at the native menu boundary and adds the right kind of real-app coverage for it.

I especially like that the new undo scenario proves the behavior through native input plus disk persistence rather than DOM-only assertions; that is exactly the class of regression browser e2e would miss. The `code: 'KeyN'` fallback also matches the existing localized-key pattern for workspace digits and should make the Option-modified notebook shortcuts robust against WebKit's dead-key `e.key` values.

CI was still pending when I reviewed, but I don't see a blocker in the diff.

PR #541, opened 2026-07-12
feat(diagnostics): self-healing watchdog repairs persistently clipped terminal panes

victorarias:
## Summary

Layer 2 on top of #537 (just merged). #537 fixed the root cause found in prod telemetry (stale replay geometry suppressing the corrective fit). This PR adds a structural guarantee on top: the existing 1.5s DOM-truth bottom-clip sweep in `terminalDiagnosticsLog.ts` now actively **repairs** a persistently clipped pane instead of only detecting it.

- After 3 consecutive clipping sweeps (~4.5s — beyond the longest observed legitimate transient), the watchdog invokes the pane's registered repair handler (a refit).
- Backoff: 3s, then 10s between repair attempts, max 3 attempts, then a one-time `bottom_clip_repair_gave_up` incident.
- `bottom_clip_resolved` telemetry now carries `repairAttempts`.
- Repairs only run for visible, active panes.

Net effect: any future stale-geometry path (not just the one fixed in #537) converges within seconds instead of sticking until remount.

## Verification

Done by the lead prior to handoff:
- Full frontend unit suite green (146 files / 1443 tests) + `tsc --noEmit` clean, re-verified after rebase (42/42 targeted tests).
- Live packaged-app verification on a throwaway `offsetfix` profile built from this branch: `real-app:scenario-reveal-overflow` passed, and `real-app:scenario-offset-soak --seed 3 --iterations 120` passed (0 persistent violations, 3 transients). The profile's incident log showed 4 transient `bottom_clip` onsets and ZERO `bottom_clip_repair` events — the watchdog correctly stays quiet on clips that heal within its 3-sweep threshold; no repair loops or false positives.

Re-verified locally before opening this PR:
- `pnpm vitest run src/utils/terminalDiagnosticsLog.test.ts src/components/GhosttyTerminal.resizePolicy.test.ts` — 42/42 passed.
- `pnpm exec tsc --noEmit` — clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ### Fixed
 - **Terminal panes no longer get stuck at a stale size.** Some panes could stay rendered at a stale, historical size after a session reattach — bottom or right edges painted outside the pane — until the pane was reopened. They now recover on their own.
+- Terminals that still end up painted outside their pane now self-heal within seconds — the app detects the mismatch and refits the terminal automatically, and records what it repaired.
 - **Undo and redo now work in the Notebook editor.** ⌘Z and ⇧⌘Z were swallowed by the native Edit menu in the packaged app and never reached the editor; they now undo/redo your edits when the editor has focus. ⇧⌘Z still zooms the active pane when you're not in a text field.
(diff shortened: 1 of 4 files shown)


figgyster (approved):
This is a clean structural backstop for the stale-geometry fix: the repair state is episode-scoped, respects active/visible panes, backs off and terminates cleanly, and leaves the original diagnostic edge signals intact. Focused diagnostics/resize tests pass (42/42) and TypeScript is clean locally. CI’s Frontend and all three sharded Frontend E2E jobs are green; the aggregate E2E check is still pending.

PR #542, opened 2026-07-12
feat(notebook): render whole-line images inline via daemon-resolved sources

victorarias:
PR6 of the notebook-editor-polish plan (docs/plans/2026-07-11-notebook-editor-polish.md).

Renders whole-line Markdown images (`alt`) inline in the live markdown editor, with image bytes resolved through a new daemon `fs_read_asset` command rather than direct filesystem access from the frontend.

- daemon: `fs_read_asset` command to read image bytes for a note-relative or absolute path
- editor: image widget renders resolved images in place of the raw markdown line
- editor: click handling reads the image widget's click position from the view

Bumps `ProtocolVersion` to 159 (new daemon command).

Verified locally: `go build ./...`, `go vet ./...`, `gofmt` clean, full `go test ./...`, `tsc --noEmit`, and `pnpm exec vitest run` (149 files / 1467 tests) all green after rebase onto main. This PR was live-verified in a dev install during development; the rebase onto main applied cleanly with no conflicts, so that verification stands per repo policy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 - **Richer live-preview rendering in the Notebook editor.** Code fences now get language syntax highlighting, and blockquotes and horizontal rules render styled instead of as raw markers.
 - **Markdown tables in the Notebook editor** now render as real tables, with raw source revealed for editing when the cursor is inside (click a row to edit it).
 - **⌘B/⌘I/⌘E toggle bold, italic, and inline code in the Notebook editor.** Wraps the selection or the word under the cursor; press again to remove.
+- **Images now render inline in the Notebook editor.** A `![alt](path)` on its own line shows the picture itself, with a "not found" placeholder if the file can't be located; click the image to see (and edit) its raw markdown.
 
 ### Fixed
 - **Terminal panes no longer get stuck at a stale size.** Some panes could stay rendered at a stale, historical size after a session reattach — bottom or right edges painted outside the pane — until the pane was reopened. They now recover on their own.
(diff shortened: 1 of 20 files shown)


figgyster (dismissed):
The advertised 10 MiB asset cap is not safe for the current WebSocket transport. `fs_read_asset` base64-encodes the entire file into one JSON response, so a 10 MiB asset becomes roughly 13.3 MiB plus JSON overhead. Ordinary clients use a 1 MiB **read** limit, but neither `sendToClient` nor the write pump impose a corresponding outbound cap; more importantly, this permits a notebook image to queue a very large frame and causes avoidable memory/connection pressure.

Please cap the *encoded response* to a transport-appropriate size (or add a bounded/chunked asset transport), and add a boundary test that proves the maximum accepted asset still fits the resulting WebSocket message.

figgyster (dismissed):
The current 4 KiB envelope allowance still does not guarantee the stated 8 MiB wire cap for all accepted assets. `maxAssetBytes` consumes the entire payload budget, while `assetEnvelopeSlack` is only validated with `assets/max.png`; a near-platform-limit valid path plus the fixed JSON fields can exceed that slack. “Paths are far shorter in practice” cannot serve as the transport bound.

Please reserve a documented worst-case envelope (including the maximum accepted path), or marshal/check the complete response before sending and reject it when it exceeds the cap. Add the corresponding long-path boundary test.

figgyster (dismissed):
The long-path suspect has an alibi now. `assetMessageFits` measures the actual request ID, path, MIME type, and base64 length before encoding, so the complete outbound message is bounded rather than merely estimated. The focused long-path test covers the regression, and the current-head CI is green.

Approved. Case closed.

figgyster (dismissed):
The rebase preserves the reviewed image/asset patch; the only material adjustment is the protocol regeneration against current main, with the version moving cleanly to 161. The exact outbound-message bound and long-path regression coverage are unchanged.

Still safe to merge. The rain moved the commit hashes, not the evidence.

figgyster (approved):
The final rebase preserves the reviewed image/asset patch; the only material adjustment is protocol regeneration against current main, with the version moving cleanly to 164. The exact outbound-message bound and long-path regression coverage are unchanged, and the current daemon checks are green.

Still safe to merge. The rain moved the commit hashes, not the evidence.

PR #543, opened 2026-07-12
fix(terminal): worker-owned OSC 10/11/12 color-query answering

victorarias:
## Problem

Two responders raced to answer terminal OSC 10/11/12 color queries: a worker-side startup-window fallback (could re-send/double-fire) and the frontend (unbounded latency via the ws round-trip, plus a dedup bug that answered only one query per burst chunk). Late or duplicate replies land as stray PTY input. User-visible fallout in shell panes:

- `glow file.md` printed garbage like `^[]11;rgb:1e1e/1e1e/1e1e^[\` before rendering (its 11,10,11,11 query burst answered late/incompletely)
- `gh pr create`'s template-chooser survey crashed with `could not prompt: unexpected escape sequence from terminal: ['\x1b' ']']`

## Fix

The PTY worker's read loop is now the single, always-on responder for OSC 10/11/12 — every query occurrence answered immediately, in query order (positional-pairing clients depend on ask order). The theme flows down from the app:

- new `set_terminal_theme` ws command (ProtocolVersion 160); daemon stores the theme globally, fans it out to live sessions, and seeds future spawns
- new `set_theme` worker RPC (no RPC version bump, MethodSnapshot precedent; old workers rejected non-fatally) + `--theme-*` spawn flags so a shell's first startup queries get correct colors
- the app re-pushes the theme on every reconnect (the daemon holds it in memory only, so a daemon restart would otherwise revert colors to defaults)
- the frontend answers no terminal queries anymore: OSC color responses from the local terminal model are stripped exactly like CPR/DA1; the old fallback/dedup machinery in `internal/pty` is deleted

AGENTS.md pattern #7 updated to the new ownership split; CHANGELOG entry included.

## Testing

- `internal/pty`: socketpair read-loop tests — exactly-once answering incl. queries split across read chunks, per-chunk multi-query counts in query order, SetTheme/spawn-seeded themes, OSC SET (no `?`) never answered
- `internal/ptyworker` / `internal/ptybackend`: RPC dispatch observed via a real PTY answering with the new color; real-worker-subprocess SetTheme round-trip; spawn argv `--theme-*` flag round-trip
- `internal/daemon`: handler stores + fans out + blanks invalid colors; real `handleSpawnSession` carries the stored theme
- frontend: queue-until-handshake flush and reconnect re-push regression tests; full suite green; e2e 188 passed
- live on a dev install: glow burst answered 4/4 in query order instantly; `#ff00ff` pushed over ws reflected in the next OSC11 reply; real `glow` rendered clean; workers answered exactly-once with the app quit; after killing the daemon the app re-pushed a light theme to the respawned daemon with no user action

Note: sessions running from before an upgrade keep their spawn-time colors (old workers reject the new RPC by design, logged and ignored) — self-resolves on session restart.


figgyster (dismissed):
One blocker from the terminal beat: `writeOSCColorResponses` writes directly to `s.ptmx` without `writeMu`. `Session.input` uses that mutex, but these replies can now be emitted at any time while frontend input is being written. Concurrent writes to the PTY master can interleave bytes, turning either a user keystroke/paste or an OSC response into malformed terminal input—the very class of stray escape-sequence failure this PR is fixing. Please serialize the full OSC reply burst through the same write path/mutex (and add a concurrent-input regression if practical).

figgyster (approved):
The suspect cleaned up nicely. `118f326a` serializes the complete OSC reply burst through `writeMu`, extends the same protection to CPR/DA1, and adds a behavioral regression proving replies wait for an in-flight writer. The original blocker is resolved.

All current GitHub checks pass. My local sandbox could not rerun Go tests because its Go 1.24 image is older than the repository's required Go 1.25.3, but the daemon CI passed on this head.

PR #544, opened 2026-07-13
fix(tickets): submit Codex nudges atomically

victorarias:
## Intent / Why

Ticket nudges could appear in Codex's composer without being submitted, leaving the agent idle and the inbox unread. The doorbell still needs to remain one approval-fenced PTY write so a nudge cannot race with a permission prompt.

## Summary

- frame the bounded doorbell prompt as an explicit bracketed-paste event, with Enter outside the paste terminator but inside the same atomic PTY write
- update daemon coverage to assert the exact framed input while preserving the pending-approval state fence
- strengthen the packaged-app scenario so text injection alone cannot pass: it now requires a real Codex turn, inbox consumption, overlapping-event deduplication, and busy-session queue behavior
- document the user-visible fix and the transport boundary

## Test plan

- [x] `go test ./internal/daemon -count=1`
- [x] focused nudge tests under `go test -race`
- [x] `pnpm --dir app exec vitest run scripts/real-app-harness` (142 tests)
- [x] `go vet ./internal/daemon ./internal/ticketnotify ./internal/ptybackend`
- [x] signed `attn-dev.app` live scenario with real Codex v0.144.3: idle nudge submits, overlapping activity produces one doorbell, busy nudge queues and drains the inbox, no composer text remains stranded

`go test ./...` passed every package except a load-sensitive timeout in `TestCodexResumeMappingEndToEnd`; that test passed immediately in isolation, and the full `internal/daemon` package passed independently.

## Out of scope

- nudge eligibility, countdown timing, selection behavior, and recent-user-input guards are unchanged
- no plugin or agent-driver contract changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-07-13]
+
+### Fixed
+- **Ticket nudges now submit in Codex instead of stopping in the composer.** The
+  bounded inbox prompt is explicitly framed before Enter, so the nudge starts a
+  turn while retaining the approval-prompt safety fence.
+
 ## [2026-07-12]
 
 ### Added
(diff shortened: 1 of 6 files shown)


figgyster (approved):
Case closed. The bracketed-paste terminator gives Codex the missing semantic boundary while keeping prompt + Enter inside the existing approval-fenced PTY write. The strengthened packaged-app scenario now proves the behavior that matters: a real turn starts, the inbox drains, overlapping activity deduplicates, and a busy turn queues the nudge instead of leaving text in the composer.\n\nI also checked the shared Present/Notebook callers; the transport change preserves their bounded doorbell contract.  is clean, Release Preflight and Frontend E2E are green; Daemon and Bench A/B were still running at my last look. This Linux review runner lacks Go and pnpm, so I couldn't independently rerun those suites here; the PR records focused daemon/race/harness checks plus a signed attn-dev live run with Codex 0.144.3.

PR #545, opened 2026-07-13
fix(terminal): keep canvas anchored in pane

victorarias:
## Intent / Why

Production diagnostics captured terminal panes whose model geometry fit the container while the rendered canvas was displaced downward inside the editable host. That left the bottom rows offscreen, and the existing watchdog repeatedly called `fit()` even though it correctly bailed with `sameSize`.

This change keeps the rendered terminal surface structurally contained and narrows automatic repair to geometry that a refit can actually change.

## Summary

- anchor the direct terminal canvas to the pane origin so browser-managed editable content cannot move rendered rows offscreen
- add an interaction regression that inserts editable-flow content before the canvas and verifies it remains contained
- remove DOM-rectangle polling and DOM-offset incident fields from terminal diagnostics
- retain delayed repair for actionable model overflow in both rows and columns, preserving existing incident event names

## Test plan

- [x] `pnpm --dir app test` — 148 files, 1,463 tests
- [x] terminal interaction Playwright spec — 19 tests
- [x] packaged dev app offset soak — 300 steps, no persistent violation
- [x] packaged dev app post-cleanup smoke — 60 steps, no persistent violation; one resize transient self-healed in 750 ms

## Out of scope

- Production was not rebuilt or restarted. All live verification used the isolated dev profile.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 - **Ticket nudges now submit in Codex instead of stopping in the composer.** The
   bounded inbox prompt is explicitly framed before Enter, so the nudge starts a
   turn while retaining the approval-prompt safety fence.
+- **Terminal content stays inside its pane even when the editable input host gains browser-managed content.** The rendered canvas is now anchored to the pane origin, preventing composition or editing layout from pushing its bottom rows offscreen.
 
 ## [2026-07-12]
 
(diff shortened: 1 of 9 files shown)


figgyster (approved):
Case closed. The canvas is now structurally anchored to the terminal host, while the watchdog keeps its useful row/column geometry repair path and drops the DOM-offset work that could never be fixed by refitting. The editable-flow regression covers the original displacement mechanism.\n\nLocally verified: all 1,463 frontend tests passed, the frontend production build passed, and git diff --check was clean. The Playwright browser was unavailable on this runner, but the PR records the 19-test interaction spec and packaged dev-app offset soak; CI frontend/E2E jobs were still running at my final fetch.

PR #546, opened 2026-07-13
docs(plans): close ticket attach ergonomics plan

victorarias:
## Summary

- mark the ticket attach ergonomics plan as not pursuing
- record that bidirectional by-ticket attaching already shipped
- document why delegate-time attaching is not worth the lifecycle coupling
- update the chief-awareness umbrella status

## Verification

- git diff --check
- documentation-only change; no runtime tests needed

### DIFF
--- a/docs/plans/2026-07-02-chief-awareness-wave.md
+++ b/docs/plans/2026-07-02-chief-awareness-wave.md
@@ -21,7 +21,7 @@ exactly one status (`crashed`); doorbells stay content-free.
 
 ## The plans
 
-> Legend: ⬜ not started · 🟡 in progress · ✅ merged.
+> Legend: ⬜ not started · 🟡 in progress · ✅ merged · ⛔ not pursuing.
 
 | plan | scope (one line) | dependencies / seams | wave | status |
 |---|---|---|---|---|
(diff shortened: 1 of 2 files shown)


figgyster (dismissed):
The plan closure is the right call, but the umbrella still sends future executors after the closed case.

`docs/plans/2026-07-02-chief-awareness-wave.md` still lists `ticket-attach-ergonomics` as Wave 2 work (lines 65–76), says brief-craft unblocks/de-conflicts it (lines 50–53), and assigns it future ownership in the seam registry (lines 107–109 and 114). That contradicts the new ⛔ row and leaves actionable instructions to implement the declined `delegate --attach` lifecycle.

Please remove or rewrite those remaining Wave 2/dependency/seam entries so the umbrella consistently treats this plan as historical/not pursuing. The alley needs one set of directions; right now half the signs point at a building we demolished.

Verification: `git diff --check` passes. I also checked the live `ticket attach --ticket` CLI/protocol/guidance surfaces; the outcome's by-id claim matches the shipped implementation.

figgyster (approved):
The follow-up removes the stale Wave 1/Wave 2 dependency text and future seam ownership, leaving only the historical ⛔ row. The umbrella and closed plan now tell the same story.

Verified on `5e898612`: `git diff --check` passes, the remaining umbrella references are intentional, and CI is green. The old directions have been taken down. Case closed.

PR #547, opened 2026-07-13
Centralize daemon session state transitions

victorarias:
## Summary

- centralize every persisted daemon session-state transition behind a typed applyState door
- preserve classifier timestamp ordering, plugin run/sequence CAS, doorbell fencing, lifecycle ordering, and source-specific side effects
- add pre-refactor characterization coverage plus an AST guard that prevents future production bypasses
- rewrite the session-state plan and strengthen the repository architecture guidance

## Verification

- go build ./...
- targeted store and daemon state-door suites
- full daemon, store, classifier, and transcript suites
- repeated characterization, state-door, and doorbell concurrency tests
- scoped go test -race suite
- make test
- make dev
- packaged real-app workspace shell lifecycle scenario

## Live-test note

The packaged Codex nudge scenario reached the real working transition on two runs, but the spawned Codex CLI then emitted task_complete with last_agent_message null and no normal stop hook. The scenario therefore could not reach its nudge assertions and is documented as an explicit live-verification gap, not passing evidence.

No user-visible behavior change is intended, so this PR has no changelog entry.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -162,7 +162,11 @@ Canonical example: `sendPRAction` in `app/src/hooks/useDaemonSocket.ts`.
 
 ### 3. Classifier Timestamp Protection
 
-Classifier work is asynchronous and can be slow. Capture the timestamp before classification starts and update state via `UpdateStateWithTimestamp()` so stale classifier results cannot overwrite fresher runtime state.
+Every persisted daemon session-state transition must go through `applyState` in
+`internal/daemon/session_state.go`; direct store state writes bypass required
+effects and are structurally forbidden. Classifier work is asynchronous and can
+be slow, so capture its observation timestamp before classification starts and
+pass it through `classifierObservation` so the door rejects stale results.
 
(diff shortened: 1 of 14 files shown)


figgyster (approved):
The state door walked into the precinct with every old side effect accounted for. Clean work.

I checked the cause-specific commit rules and effect profiles against the previous paths: classifier timestamp rejection, plugin run/sequence CAS, doorbell fencing, long-run tracking, `LastSeen` touching, startup silence, process-exit ordering, and session/workspace broadcasts all remain source-appropriate. The characterization suite and AST guard give this refactor a sturdy lock on the door it just installed.

`git diff --check` is clean. I also attempted the focused daemon/store suite in the review sandbox, but its Go 1.24 image cannot build this repository's Go 1.25.3 module; visible CI is the authoritative test evidence here, with the daemon job still running at my last fetch. The documented packaged-app Codex nudge gap is explicit and does not undermine the preserved concurrency contract covered by the focused tests.

PR #548, opened 2026-07-13
fix(present): cache per-file diff items so typing and J/K stay responsive

victorarias:
## Problem

Two user-visible bugs in the Present review window, worst on large changesets:

1. **Typing latency** — 1s+ per character when writing a comment.
2. **Broken keybindings** — J/K navigation behaved erratically (landed on the wrong file, state lagged far behind keypresses).

## Root cause (one bug)

`PresentTour`'s `items` useMemo had `drafts` (updated per keystroke) and `reviewedPaths` (updated per `j`) in its deps, and every recompute re-ran `parseDiffFromFile` for **every** file and bumped **every** item's `version`. `@pierre/diffs` CodeView treats a version bump as "re-layout this file", so each keystroke re-parsed and re-laid-out the entire changeset. The resulting layout/scroll storms also corrupted `activePath`, which is why J/K navigated relative to the wrong file.

## Fix

- **Draft content out of React state.** The comment form owns its live value locally (a ref map is remount insurance only), so keystrokes never touch the items memo.
- **Per-file caching in the items memo.** Parse results are cached on content identity; item objects are cached on a per-file signature (comments, draft anchors, reviewed membership, file-scoped editing, mermaid layout tick). Unchanged files return the *same object* with the *same version* — zero work for CodeView. Only genuinely changed files get a version bump. Caches are pruned to current paths and cleared when files aren't settled.

## Testing

- 5 new regression tests (with a `parseDiffFromFile` spy): typing doesn't rebuild items or re-parse; reviewed toggle / opening a draft / editing a comment bump only the affected file; draft content survives item remount. Full PresentTour suite 81/81 green; `tsc --noEmit` clean.
- **Live before/after** on a throwaway profile with a synthetic ~62k-line changeset (30 modules × 2500 lines, 28 changed files):
  - *Before:* 6×j left the header on file 3 with a blank pane, draining to the correct file only after ~15s; 43 keystrokes typed in ~1s showed an **empty** textarea, full text appearing ~7s later.
  - *After:* header/state correct instantly on every keypress; the complete typed text visible immediately.

Initial-load parse of a very large diff still takes a few seconds on first open — that's inherent to parsing the whole changeset up front and untouched here (possible future work: worker-side/virtualized parsing).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-13]
 
 ### Fixed
+- **The Present review window no longer lags on large changesets.** Typing a comment re-parsed and re-rendered every file's diff on each keystroke, making characters take a second or more to appear and J/K navigation land on the wrong file. Each file now only re-renders when its own content or comments change, so typing echoes instantly and navigation tracks keypresses correctly.
 - **Ticket nudges now submit in Codex instead of stopping in the composer.** The
   bounded inbox prompt is explicitly framed before Enter, so the nudge starts a
   turn while retaining the approval-prompt safety fence.
--- a/app/src/components/PresentTour/PresentTour.test.tsx
+++ b/app/src/components/PresentTour/PresentTour.test.tsx
@@ -27,6 +27,18 @@ vi.mock('mermaid', () => ({
   },
 }));
 
+// Wraps the real `parseDiffFromFile` in a spy so the per-file item cache's
+// "never re-parses an unchanged file" claim is observable, without faking the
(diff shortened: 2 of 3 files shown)


figgyster (approved):
The rain came down in memoized buckets, but the case stayed clean. The per-file cache matches `CodeView`'s version contract, and keeping draft text local removes the hot keystroke path without losing remount recovery.

Verified:
- `git diff --check` passes.
- Frontend sandbox run passes, including all 27 `PresentTour` tests and the new cache/draft regressions.
- GitHub CI is green: Frontend plus all E2E shards.

One unrelated limitation: the generic repository test helper selected a Go 1.24 sandbox while this repository requires Go 1.25.3. That is test infrastructure, not evidence against this PR.

No blocking findings. Case closed.

PR #549, opened 2026-07-13
perf(present): open the tour instantly, stream file diffs in progressively

victorarias:
## Problem

On large changesets the Present window showed "Loading tour…" for several seconds before rendering anything. Two causes, both frontend:

1. **All-settled barrier** — the tour refused to build its CodeView items until *every* file's diff fetch had settled.
2. **Parse-all pass** — once settled, one synchronous memo pass ran `parseDiffFromFile` over every file, stalling the main thread.

The daemon side is fine (per-file fetches already run concurrently); the delay was purely render gating + a monolithic parse.

## Fix

- **CodeView mounts as soon as the manifest has files.** Each not-yet-ready file renders as a zero-hunk placeholder card (same item id and type, so the later swap is an in-place `version` bump — never a teardown/remount).
- **Frame-budgeted admission**: settled files are parsed ~8 ms per animation frame (always ≥1 per frame) and swapped in progressively, so the first files are readable immediately while the rest stream in.
- Error cards render immediately without waiting for admission. Pending cards get a subtle opacity pulse (reduced-motion respected). Pending files are excluded from header notes so CodeView's analytic header-height math stays correct.
- Effects that previously gated on "all diffs settled" (scroll pin, j/k scrollTo, annotation hop, card-class sync) now gate on the tour being mounted.
- Round reloads reuse the content-keyed parse cache — identical re-settled content is never re-parsed.

## Testing

- 6 new regression tests (placeholder-before-settle, single-file in-place swap with identity assertions on untouched placeholders, exactly-one-parse-per-file across a 12-file batch, placeholder identity stability under unrelated prop changes, immediate error cards, round-reload parse-cache reuse). PresentTour suite 33/33.
- Full frontend suite: 1477/1477 (148 files); `tsc --noEmit` clean.
- **Live-app verification** on a throwaway profile (`presprog`) against a 62k-line synthetic fixture (28-file diff, 2.5k-line modules): window content fully rendered at t0 (previously several seconds of blank "Loading tour…"), 14 rapid `j` presses from cold open all processed correctly, comment-form typing echoed a full sentence instantly. Profile cleaned up after.

## Future work (out of scope)

- @pierre/diffs ships a worker pool for off-main-thread Shiki highlighting, but nothing mounts its `WorkerPoolContextProvider` — a possible next lever if highlighting cost ever shows up.

https://claude.ai/code/session_01LrwfEhDc8tfoGykC6961gW

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-13]
 
 ### Fixed
+- **The Present review window opens instantly on large changesets.** The tour used to show "Loading tour…" until every file's diff had been fetched and parsed, which took several seconds on big diffs. It now opens immediately — each file appears as a lightweight card that fills in with its diff as it loads, so you can start reading the first file right away.
 - **The Present review window no longer lags on large changesets.** Typing a comment re-parsed and re-rendered every file's diff on each keystroke, making characters take a second or more to appear and J/K navigation land on the wrong file. Each file now only re-renders when its own content or comments change, so typing echoes instantly and navigation tracks keypresses correctly.
 - **Ticket nudges now submit in Codex instead of stopping in the composer.** The
   bounded inbox prompt is explicitly framed before Enter, so the nudge starts a
--- a/app/e2e/present-tour.spec.ts
+++ b/app/e2e/present-tour.spec.ts
@@ -436,13 +436,19 @@ test.describe('PresentTour summary fold', () => {
 
 test.describe('PresentTour deferred-load scroll replay', () => {
   test('a scroll request issued while still loading is replayed once files settle', async ({ page }) => {
-    // `&deferred=1` starts every file `{loading: true}` (no `<diffs-container>`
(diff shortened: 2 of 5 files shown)


figgyster (approved):
The old all-settled barrier finally gave up the address. This keeps the placeholder-to-real-item transition stable, budgets parsing across frames, preserves pending/error behavior, and covers reload/cache semantics without disturbing unrelated item identities.

I traced the round-fetch lifecycle and the annotation/navigation handoff, and the focused PresentTour suite passes (33/33). The PR also records packaged-app verification against a large synthetic round. No blocking findings from me.

CI was still finishing when I reviewed.

figgyster (approved):
The new commit closes the loose end from progressive mounting: a rail/j-k request made against zero-hunk placeholders is replayed once every real item has been admitted, while error cards still count as settled and do not hold navigation hostage. The updated deferred-load E2E now exercises the actual placeholder state instead of the retired full-screen loading state.

I reviewed the incremental `58b39def..35db7297` diff, reran the focused PresentTour suite (33/33), and confirmed the current-head CI is green, including all three E2E shards and the aggregate E2E job. No blocking findings. The suspect cleaned up nicely.

PR #550, opened 2026-07-14
feat(tiles): cmd+click .md opens a session-bound markdown tile

victorarias:
## Summary

PR 1 of the markdown reader + annotator plan (`docs/plans/2026-07-14-md-reader-annotator.md`, included in this branch).

- Cmd+click on a `.md` path in a session terminal now opens the file in an in-app markdown tile bound to that pane's session, instead of just OS-opening it.
- Multiple markdown tiles can coexist; tile id is `sha256` of the cleaned absolute path. Reopening an already-open file reuses its existing tile.
- Daemon exposes `open_markdown` over the websocket using the standard request/result async pattern (`request_id` / `open_markdown_result`), `ProtocolVersion` bumped 160 → 161.
- Falls back to the OS-open behavior if the daemon call fails.
- Adds a packaged-app harness scenario (`real-app:scenario-terminal-md-link`) plus a harness fix that dismisses the first-launch What's New modal so it doesn't block other scenarios.

## Verification

- `go build ./...`, `go vet ./...`, `go test ./...` all pass (note: `internal/daemon` has a pre-existing unrelated data race in `TestGitStatusScheduler` — not run with `-race` here, per repo convention).
- Frontend: `pnpm run build` (tsc + vite) and `pnpm test` (148 files / 1484 tests) pass.
- Live-app verification: `real-app:scenario-terminal-md-link` was run against a packaged build on a throwaway profile and passed — plain click is a no-op, cmd+click docks a session-bound tile (asserted `tile_session_id` in the daemon layout), opening a second file docks a second tile, and re-clicking the first file reuses its existing tile.

## Test plan

- [x] `go test ./...`
- [x] `pnpm run build` / `pnpm test`
- [x] Packaged-app scenario `real-app:scenario-terminal-md-link` on a throwaway profile


figgyster (dismissed):
One blocker remains on the current head.

- **[P1] Route `open_markdown` through the pane's remote session.** `CmdOpenMarkdown` is declared `ScopeSession` here, but `remoteCommandSessionID` in `internal/daemon/websocket.go` has no `CmdOpenMarkdown` case. For a terminal pane owned by a remote endpoint, routing therefore gets an empty session id; `OpenMarkdownMessage` also has no path-based resolver fallback, so the hub handles the command locally. The local store cannot find the remote session, and the frontend fallback then tries to OS-open the remote absolute path on the hub machine. Please return `protocol.Deref(typed.SessionID)` for this command in `remoteCommandSessionID` and add it to `TestRemoteCommandSessionID`, so Cmd+click preserves the advertised session-bound behavior for remote panes.

The follow-up formatting commit fixed the earlier red Daemon gate, and the current CI suite is green. This routing branch is not covered by those checks yet.

figgyster (approved):
Fix verified. `open_markdown` now routes through the pane's `session_id` before hub forwarding, and `TestRemoteCommandSessionID` covers both the remote-session and missing-session cases. The incremental diff is clean and the current-head CI matrix is green. The old blocker has gone straight; approving this head.

PR #551, opened 2026-07-14
feat(md-reader): reader core — typography, source anchoring, highlighting, frontmatter card

victorarias:
This PR turns the markdown tile into a proper reading surface (PR2 of docs/plans/2026-07-14-md-reader-annotator.md, stacked on feat/md-tile-cmdclick / PR #550). It adds reader typography adapted from plannotator to attn's theme variables, syntax-highlighted code blocks with a hover copy button, GitHub-style heading slugs with dedup and clickable anchors resolving within the tile, link sanitization (javascript:/data: URLs render as plain text), and a YAML frontmatter metadata card. Underneath, a new rehype anchoring pass stamps data-block-id/data-source-line/data-source-line-end onto every top-level block and list item — the source-line groundwork later annotation PRs attach to. Frontend-only; the shared Markdown component is unchanged for other consumers.

## Test plan
- [x] Full frontend suite green (1526 tests)
- [x] tsc clean
- [x] Go untouched (frontend-only change)
- [x] CHANGELOG entry included
- [x] Live-app verification on a throwaway profile: reader tile renders typography, shiki highlighting, copy button chrome, frontmatter card, sanitized links (confirmed via screenshots)

https://claude.ai/code/session_01TkqDvJnLVPJCU5udzSdcdP

PR #552, opened 2026-07-15
feat(reader): markdown reader polish — callouts, prose transforms, task lists, images, raw HTML

victorarias:
## Summary

Builds on the reader core (#551) with the polish layer for markdown rendering:

- GitHub-style alert callouts (`[!NOTE]`, `[!TIP]`, `[!WARNING]`, `[!IMPORTANT]`, `[!CAUTION]`) with icons and per-kind palettes
- Prose transforms: smart punctuation (curly quotes, en/em dashes) and emoji shortcodes, with a guarantee that `--flags` and other double-dash tokens inside code spans/blocks are never touched
- Task-list checkbox styling and GitHub-flavored table styling
- Inline local images rendered via the Tauri asset protocol, with a portal-based lightbox that's integrated into the app's Escape-key stack
- `rehype-raw` + `rehype-sanitize` with a hardened schema so embedded raw HTML in markdown renders safely (`<details>`, etc.) while scripts and other unsafe elements are stripped
- A content-hash re-render gate so live file reloads don't blow away transient DOM state like an open `<details>` disclosure

## Notable / packaged-app-relevant

- `app/src-tauri/tauri.conf.json` gains an `assetProtocol` scope and `app/src-tauri/Cargo.toml` gains the `protocol-asset` feature — needed to serve local images through the asset protocol. This is a packaged-app-relevant config change.
- Deliberate divergence from the plannotator reference implementation: remote `http(s)` images/media are stripped rather than rendered. This reader enforces a no-network policy — only local images are shown.

## Test plan

- [x] `pnpm run lint` — clean
- [x] `pnpm run typecheck` — clean
- [x] `pnpm test -- --run` — 1591 tests passing across 153 files
- [x] Live-verified on a throwaway packaged-app profile (`mdp9x4`), screenshot-confirmed:
  - Alert callouts render with correct icons/palettes for each kind
  - Smart punctuation applies (en-dashes) while `--flags` inside code are preserved verbatim
  - Emoji shortcodes render
  - Task-list checkboxes render correctly
  - Inline image renders via the asset protocol
  - Lightbox opens on click and closes on Escape (correctly integrated into the Escape-key stack)
  - `<details>` renders collapsed and survives live reload without resetting
  - `<script>` tags are stripped by the sanitize schema
  - Smart quotes render correctly inside table cells


figgyster (dismissed):
This looks safe to merge. The no-network media boundary is enforced at both the sanitizer and renderer seams, and the memoized reader body preserves DOM-owned state without letting the lightbox trigger a document reparse.

The full frontend suite (1,586 passed, 5 skipped), focused reader coverage, TypeScript check, production frontend build, and diff check are clean. The packaged-profile evidence in the PR covers the asset-protocol and Escape-stack behavior I cannot reproduce on this Linux review host. No blockers from me.

figgyster (dismissed):
One blocker surfaced in the anchoring foundation.

`app/src/components/MarkdownReader/anchoring/rebase.ts:186` accepts the highest-scoring candidate whenever its absolute score is at least `0.5`, even if the runner-up has the same score. For a whole-block quote, both stored context windows are empty; `similarity('', '')` is `1`, so two identical paragraphs at equal line distance both score about `0.98`. The stable sort then silently attaches the anchor to whichever duplicate appears first, despite the function's contract to orphan rather than paint the wrong text.

I reproduced this with a throwaway regression: anchor the second of two identical whole-paragraph blocks, move the duplicates to positions equally distant from the old source line, then rebase. The result selected the first duplicate instead of `{ state: 'orphan', reason: 'ambiguous' }`.

Please require a meaningful lead over the runner-up when there are multiple candidates (the absolute confidence floor can still be used in addition), and add the equal-score duplicate regression. The rest of the new anchoring suite is green: 65 focused tests, 1,651 full frontend tests with 5 skipped, and the production build all pass.

figgyster (dismissed):
The ambiguity fix closes the blocker cleanly. Multiple candidates now need both the confidence floor and a meaningful lead over the runner-up, so equal or near-equal duplicates orphan instead of inheriting identity from sort order. The new regression covers the equidistant whole-block case, and the decisive-proximity case preserves useful rebasing when the evidence is actually strong.

The focused anchoring suite (67 tests), production TypeScript/Vite build, and incremental diff check are clean. No blockers remain.

figgyster (dismissed):
One restack blocker surfaced. `app/package.json:102-103` still adds `rehype-raw` and `rehype-sanitize`, but `app/pnpm-lock.yaml` no longer contains either dependency. The previously approved head carried the corresponding 122-line lockfile update; the merge restack dropped it.

As a result, both the focused anchoring tests and the production frontend build fail during frozen installation with `ERR_PNPM_OUTDATED_LOCKFILE`, before any project code runs. The live Frontend and all Frontend E2E checks fail as well.

Please regenerate and commit `app/pnpm-lock.yaml` for the current `app/package.json`, then rerun the checks. The ambiguity fix and the rest of the previously approved reader patch survived the restack unchanged; the main-only local-image CSS and changelog/hub additions were preserved cleanly.

figgyster (approved):
The restack blocker is closed. This push restores exactly the 122 lockfile lines for `rehype-raw`, `rehype-sanitize`, and their transitive dependencies; no reader or anchoring code changed.

The frozen install, all 67 focused anchoring tests, the production TypeScript/Vite build, and `git diff --check` pass locally. Frontend, every Frontend E2E shard, and Tauri are green on the current head. No blockers remain.

PR #553, opened 2026-07-15
feat(reader): anchoring core — W3C-selector anchor/rebase engine + highlight paint layer

victorarias:
## Summary

Pure anchoring module for the markdown reader (`app/src/components/MarkdownReader/anchoring/`), stacked on #552. No annotation UI in this PR — that lands in PR 5.

- Anchors are `{blockId, startLine, endLine, exact, prefix, suffix, start, end, contentHash}` — a TextQuote + TextPosition selector pair, Hypothesis-style.
- Content-hash state machine drives resolution:
  - hash unchanged → zero-heuristic slice verification (no fuzzy matching needed)
  - hash changed → one fuzzy rebase pass (same-block exact match → document-wide search → whitespace-normalized match, scored by prefix/suffix overlap and line proximity), then the anchor re-baselines against the new content so fuzz never compounds across edits
  - no confident match → the anchor orphans; it is never wrong-painted
- Headless pipeline-parity block-text extraction: hast text extraction and the live DOM walker are pinned to produce identical text via a fixture test, so anchors created from either source resolve consistently.
- Swappable `HighlightPainter`: CSS Custom Highlight API is preferred (per-root registry names so multiple reader tiles don't clobber each other's highlights), with a mark-insertion fallback for environments without the API.
- Brutal fixture suite covering duplicated paragraphs, text rewrites, line shifts, smart-punctuation normalization, UTF-16 surrogate-pair offsets, and repeated rebase (5x) without compounding drift.

Also includes a **throwaway spike** (`anchoring/spike.ts`) that proves the engine live against `<!-- attn-anchor-spike: "text" -->` markers in a document, exposed via `window.__attnAnchorSpike` and a `markdown_get_anchor_spike_state` bridge action. This is deleted in PR 5 once the real annotation UI replaces it.

Frontend-only. No protocol bump. No changelog entry (internal/no user-visible behavior yet).

## Fix included

Found and fixed during review: `anchoring/rebase.ts` had a raw NUL byte embedded directly in a template-string dedupe key (`` `${blockId}\0${start}` ``) instead of the `\0` escape sequence. This made git treat the entire file as binary — no diffs, no blame, no line-based review possible. Replaced with the proper escape sequence; behavior is unchanged.

## Test plan

- [x] `pnpm test` — 1655 frontend tests pass, including 64 anchoring-specific tests (create/resolve/rebase, DOM range mapping, block extraction, painter, spike)
- [x] `pnpm run build` — tsc typecheck + vite build clean
- [x] Live-verified on a throwaway packaged-app profile via the spike bridge + screenshots:
  - markers paint on open under `custom-highlight` mode (comment tint + deletion strikethrough visible)
  - inserting lines above a marker re-baselines its anchor (blockId b2→b4, startLine 8→12) and it stays painted through a live reload
  - destroying the target text orphans that anchor (`text-not-found`) while an unrelated anchor stays painted


figgyster (dismissed):
One blocker: the ordinal `blockId` can cause a quote to be painted on the wrong passage after an insertion.

`rebaseAnchor` stops at the first tier with any candidate. If a new paragraph is inserted before the anchored paragraph, it can inherit the anchor's old block ID; when that new paragraph also contains `exact`, `sameBlockCandidates` returns it and the single-candidate path accepts it unconditionally. The document tier never gets to compare the original paragraph's matching prefix/suffix under its new ID.

I reproduced this with an anchor on `Beta before. The shared sentence sits here. Beta after.`, then inserted `Intruder before. The shared sentence sits here. Intruder after.` immediately above it. Resolution rebased to the intruder (`prefix === "Intruder before. "`) instead of the original Beta paragraph. The focused regression failed exactly there, while the existing 64 anchoring tests and TypeScript/diff checks passed.

Required before merge: do not treat an ordinal same-block hit as conclusive after the content hash changes. Compare it with document candidates (or require enough context confidence before short-circuiting), and keep a regression for this ID-reuse case. The engine's orphan-over-wrong-paint contract needs to hold here; highlights are unreliable witnesses otherwise.

figgyster (approved):
The ordinal-ID collision is fixed. Exact matches now compete in one document-wide pool, so the inherited old block ID no longer gets an identity shortcut; the regression pins the original `Beta before` passage over the inserted intruder.

Focused anchoring suite: 65/65. TypeScript and diff checks are clean. No blockers remain from me.

PR #554, opened 2026-07-15
feat(reader): annotation UI with daemon-persisted drafts

victorarias:
## Summary

Annotation UI for the markdown reader, built on the PR4 anchoring core, with drafts persisted through the daemon.

- Selection surfaces a floating toolbar (copy / redline / comment / quick-label / thumbs-up). Type-to-comment: pressing any printable character while text is selected opens the comment popover seeded with that character. Alt+1..0 apply quick labels. Code blocks support whole-block annotation. An annotation sidebar lists global comments; popovers support inline editing; Cmd+Enter submits.
- Drafts persist in the daemon: new store `internal/store/markdown_annotations.go` (migration 68), one row per absolute file path holding a JSON annotations blob, a generation counter, and a tombstone generation. New commands `markdown_annotations_get`/`_save`/`_clear`. `ProtocolVersion` bumped 161→162, updated in all three lockstep spots (`internal/protocol/constants.go`, generated files, `useDaemonSocket.ts` `PROTOCOL_VERSION`).
- Annotations rebase on reload via content-keyed anchors from the PR4 anchoring engine. The PR4 paint spike is fully deleted and replaced by this real UI.
- Packaged-app fix found during live verification: macOS WebKit doesn't move focus on mousedown for non-focusable content, so after drag-selecting in the reader the terminal's hidden input kept `document.activeElement` — the toolbar's editable-element guard then swallowed type-to-comment keystrokes (and they'd otherwise leak into the shell). Fix: the reader root is now focusable (`tabindex=-1`) and selection gestures claim focus explicitly, while plain clicks never steal focus, preserving terminal focus ownership. Regression test included.
- Harness addition: a real `drag` command in `InputDriver.swift` and `dragWindow()` on `MacOSDriver` (interpolated `leftMouseDragged` CGEvents) — WebKit requires genuine drag events to produce a text selection.

## Verification

- `cd app && pnpm exec tsc --noEmit && pnpm vitest run` — 1734 tests green
- `go build ./...` and `go test ./internal/store ./internal/daemon -run 'Markdown|Annotation'` — green
- Live packaged-app verification on a throwaway profile, 3 legs:
  1. Drag-select → type-to-comment → Cmd+Enter creates a painted comment annotation, asserted via the read-only `markdown_get_annotations_state` bridge action.
  2. App relaunch hydrates the draft from the daemon (generation intact).
  3. Editing the file (inserting a paragraph above the annotated block) rebases the anchor (blockId b1→b2, startLine 3→5); annotation stays painted and is re-persisted at generation 2.

This PR is stacked on `feat/md-reader-polish` (parent of `main` via #552). CI only triggers on PRs based on `main`, so zero checks here is expected, not a failure.


figgyster (changes requested):
I’m blocking on two draft-safety contracts. The ordinary annotation paths look good, but these failures can silently lose or cross-wire a user’s review notes.

Required before merge:

- Preserve annotations created while initial hydration is unavailable. In `useAnnotations.ts`, the UI remains usable after a failed `get`, but the successful retry assigns the daemon snapshot directly to `annotationsRef.current`. Any comment/redline created during the outage is erased at recovery. The existing failed-hydration test creates “made during outage” and then never asserts it survives or appears in the next save. Either disable mutation until hydration succeeds or merge/reconcile the local pending records after seeding the generation floor, with a regression test for the outage → local edit → successful hydrate sequence.
- Give annotation persistence an endpoint/workspace identity. The new get/save/clear messages carry only `path`; they are marked session-scoped, but neither `remoteCommandSessionID` nor `remoteCommandPath` recognizes them. On a hub, remote markdown-tile requests therefore fall through to the hub-local store. Besides putting ownership on the wrong daemon, two endpoints with the same absolute path share one draft/generation/tombstone row and can erase each other’s annotations. Route by the tile/workspace/session owner (and key drafts by endpoint + path), with a remote-routing/collision regression.

Frontend typecheck and the focused annotation/socket suites pass locally (176 focused assertions across the two runs). The visible Bench A/B check is green. I could not independently rerun the Go tests because Figgy’s isolated Go image is 1.24 while this branch requires 1.25.3; that is runner weather, not a PR failure.

figgyster (changes requested):
The parent-stack merge and ambiguous-anchor fix look clean, but neither draft-safety blocker changed on this head.

Still required before merge:

- Preserve annotations created while initial hydration is unavailable. The successful retry still replaces `annotationsRef.current` with the daemon snapshot, erasing comments/redlines created during the outage. Add the outage → local edit → successful hydrate regression and either lock mutation until hydration or reconcile the pending local records.
- Carry endpoint/workspace ownership through annotation get/save/clear. The messages still contain only `path`, and remote routing still has no annotation case, so remote tiles fall through to the hub-local store and identical absolute paths on different endpoints share generations/tombstones.

The incremental anchoring and annotation suites pass (49 tests), and `git diff --check` is clean. This is the same contract blocker moved to the current head, not a new complaint wearing a fresh commit SHA.

figgyster (approved):
Both original draft-safety blockers are fixed on this head.

Initial hydration now preserves records created during an outage and persists the merged union, while stale-generation re-syncs still replace from the authoritative store so tombstoned annotations stay buried. Annotation get/save/clear requests and results also carry workspace ownership through client correlation, hub routing, and endpoint response forwarding.

The focused annotation/socket suites pass (90 tests), TypeScript is clean, and `git diff --check` is clean. The local Go sandbox is still on Go 1.24 while this repository requires 1.25.3, so I could not independently rerun the focused Go packages here. No blockers from me.

figgyster (approved):
The restack is clean. The current `main...HEAD` patch has the same stable patch fingerprint as the previously approved version, the obsolete anchor-spike files and bridge command remain removed, and protocol v162 is aligned between daemon and app.

The focused annotation/socket suites still pass (90 tests), TypeScript is clean, and `git diff --check` is clean. Re-approved for the current commit identity; no new blockers found in the conflict resolution.

PR #555, opened 2026-07-15
feat(reader): send annotations to the bound session (protocol 163)

victorarias:
## Summary

Sends markdown annotation drafts to the tile's bound session:

- New protocol command `markdown_annotations_submit` + result event, ProtocolVersion bumped 162→163 (all three lockstep spots updated: `constants.go`, generated files via `make generate-types`, and `app/src/hooks/useDaemonSocket.ts` `PROTOCOL_VERSION`).
- Delivery reuses the daemon's typeDoorbell path (bracketed paste + `\r`); a pure-Go payload formatter orders annotations by anchor line with document-wide notes last; orphans are labeled "(~line N, moved)".
- If the target session is `pending_approval`, the submit is skipped with status `skipped_pending_approval` and drafts are retained (tombstone-clear only happens on successful delivery).
- Retarget: the tile session picker updates `tile_session_id` via `workspace_layout_update_tile` (reuses `rebindTileSession`).
- Frontend: ⌘Enter (`markdown.sendAnnotations`) is gated on tile focus-within; `flushPendingSave()` runs before submit; the popover's ⌘Enter only submits the comment, never sends.

## Verification

Live-verified on a throwaway profile (`mdsend1`, fresh packaged app build):

1. Annotation created via drag-select + type-to-comment; popover ⌘Enter did NOT deliver.
2. ⌘Enter with tile focus delivered the full payload verbatim into a `cat` pane via bracketed paste, and the sidebar emptied.
3. App relaunch hydrated 0 drafts (send tombstone held).
4. Submit targeting a live claude session parked at `pending_approval` returned `skipped_pending_approval` with drafts retained (driven over the daemon WS protocol).

Not live-driven: the session-picker dropdown UI itself (screen locked mid-run); covered by automated frontend tests.

Automated gates green locally: `go build`/`go vet`/`go test ./...` (Go build/vet clean; one pre-existing flaky test in `internal/ptybackend`, unrelated to this change, confirmed passing in isolation), `tsc --noEmit`, and 1754 vitest tests.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...`
- [x] `tsc --noEmit`
- [x] `pnpm test` (1754 tests)
- [x] Live-app verification on throwaway profile (see above)


figgyster (dismissed):
The local flow is carefully fenced, but the remote-workspace path falls through a trapdoor. This crosses the product-contract line: remote daemons are a first-class attn surface, yet Send is enabled there and cannot reach the selected session.

Required before merge:

- Make annotation draft ownership and submit delivery endpoint-aware. `markdown_annotations_get/save/clear` are currently path-only and execute on the local hub, while `markdown_annotations_submit` is not recognized by `remoteCommandSessionID`; the local handler then calls `d.store.Get(target)` and returns `session not found` for a remote session. Forwarding only Submit is not sufficient, because the draft it formats would still be stored on the hub. Carry enough workspace/endpoint context through the draft operations and submit so one authoritative daemon owns both the draft and target PTY, and add a hub→remote regression test.

The focused frontend suite is green (59 tests) and `tsc --noEmit` passes. I could not run the Go suite here because the review sandbox has Go 1.24 while this branch requires Go 1.25.3. The code path above is direct, though; the function said it was session-scoped, but the routing switch never learned the suspect's name.

figgyster (dismissed):
The routing fix closes the blocker I raised. Draft get/save/clear now follow the owning workspace to its endpoint daemon, submit follows the target session to that same daemon, and the hub relays every result event back. The focused routing/annotation tests pass, along with 118 frontend tests and `tsc --noEmit`.

One required proof remains before merge: the repository requires live-app verification for protocol/UI changes, and the live run currently recorded on the PR predates this remote-routing fix. Please exercise one remote-workspace loop on the current head: create/persist an annotation on a remote markdown tile, send it to the bound remote session, confirm the payload reaches that PTY, and confirm the draft clears. Once that witness checks in, I see no remaining code blocker. The old routing suspect has gone straight; now we just need the lineup photo.

figgyster (commented):
The live witness clears #555's own blocker: with #556 layered onto this unchanged head, remote draft save/get, submit delivery to the owning PTY, and remote draft clearing all passed end to end.

I am withholding approval only until #556 lands. That separate pre-existing `client_hello` fix is required for any hub-forwarded remote command to reach the endpoint at all, so today's head cannot reproduce the live loop by itself yet. No further change to #555 is requested; once #556 is merged, this case is ready for approval without another broad test-pageant. The annotation suspect checked out. Its getaway car is still in another PR.

figgyster (dismissed):
#556 has landed, satisfying the only condition left by my prior review. This head is unchanged; its remote annotation save/get, PTY delivery, and draft clearing were already exercised end to end with that hub handshake fix in the tested composition.

Approved. The getaway car has been impounded.

figgyster (commented):
The restack preserves the reviewed 33-file send-flow slice and correctly reconciles #554's workspace-qualified pending keys: submit now stores and resolves under the same `submit::<path>` key. Remote routing remains endpoint-owned, #556's hub handshake is present through `main`, `git diff --check` is clean, the focused daemon/hub tests pass, all 118 focused frontend tests pass, and `tsc --noEmit` is green.

I am withholding re-approval only because principal CI is red. Frontend E2E shard 3 failed two existing split-render tests after receiving lines 1–49 but missing line 0; that looks timing-sensitive and is outside this PR's semantic slice, but it still needs a green rerun before the badge goes back on. No code change to #555 is requested from the evidence currently on record. The restack checked out; one flaky witness is still changing its story.

figgyster (approved):
CI is green on the unchanged restacked head, including shard 3. The exact-head live loop also confirms save/get, delivery to the bound PTY, and tombstone clearing, so the earlier code validation still applies.

Approved. The flaky witness finally remembered where it was.

PR #556, opened 2026-07-15
fix(hub): send client_hello on remote endpoint connections

victorarias:
Since commit 2deea077, the daemon rejects every `workspace_sessions`-gated command from a connection that never sent `client_hello` (closing it with `StatusPolicyViolation`). The hub's persistent SSH connection to a remote endpoint never sent one, so ALL hub-forwarded commands to remote endpoints (`register_workspace`, `spawn_session`, workspace layout ops, forwarded client payloads such as markdown annotations) were rejected and dropped the connection — confirmed live against a real SSH endpoint whose daemon log showed:

```
rejecting websocket command register_workspace: client must send client_hello with "workspace_sessions" capability
```

The fix sends the hello (`client_kind` "hub", capability `workspace_sessions`, version `protocol-<ProtocolVersion>`) right after connecting, before the consume loop. Regression test: `TestSendClientHelloDeclaresWorkspaceSessions`.

Discovered while live-verifying PR #555 on a hub profile with a remote endpoint.


figgyster (dismissed):
The hello still is not guaranteed to be the first frame on this connection. `runEndpointLoop` publishes `runtime.conn` through `setConnection` before calling `sendClientHello`; meanwhile `ForwardEndpointCommand` can observe that connection and acquire `runtime.writeMu`, but the hello write does not participate in that lock. A concurrent forwarded command can therefore win the race, hit the daemon before the capability declaration, and trigger the same policy-violation close this PR is fixing.

This crosses the merge line because first-frame ordering is the protocol contract here. Please make connection publication and the hello atomic with respect to forwarded writes (or send the hello before publishing the connection, with proper failure cleanup), and cover that ordering through the manager/forwarding seam rather than only testing the helper payload in isolation.

figgyster (commented):
The ordering blocker is fixed on this head. Publishing the connection while `m.mu` is held and acquiring `writeMu` before releasing it prevents any forwarder from observing the connection until the hello owns the write lane; the new manager-level race test covers the right seam. Daemon CI and the relevant preflight checks are green.

I’m holding approval only on the repository’s required live verification for daemon/protocol/timing changes. I can’t run the macOS app or a real SSH endpoint from this Linux reviewer, and the PR currently documents live reproduction of the pre-fix failure rather than a successful post-fix smoke. Please verify on a non-production hub profile that a remote endpoint connects and accepts representative gated traffic (for example workspace registration/session spawn) without the policy-violation disconnect.

figgyster (approved):
That closes the case. The first-frame ordering is now enforced at the manager/forwarding seam, the regression test covers the race, CI is green, and the live SSH smoke exercised the actual gated workspace/session path end to end.

Safe to merge. The connection finally showed its papers before trying the door.

figgyster (approved):
The merge from `main` leaves the reviewed hub implementation and regression test byte-identical, and the current PR patch is unchanged apart from upstream changelog context. Daemon CI and preflight are green, so the exact-head live evidence still applies to the same code.

Re-approved. GitHub changed the coat; the suspect underneath is the same.

PR #557, opened 2026-07-15
Disable Copilot's mouse tracking to fix stuck selection/scroll

harleenquinzell:
Copilot's TUI independently enables SGR mouse tracking (DECSET 1000/1002/1003) in alt-screen mode. attn's "forward mouse to app" path (sendTrackedMouse in GhosttyTerminal.tsx) has no fail-safe for a mouse-release event that lands outside the terminal's own DOM node, unlike the hardened text-selection path (startSelectionDrag), which explicitly listens on document/window for mouseup/blur/pointercancel. When the release is dropped, Copilot's TUI keeps believing the button is held, producing a runaway selection, plus scroll ending up routed to the accessibility contentEditable host instead of the terminal output after a reload/reattach.

attn already owns text selection and scrolling itself, so Copilot's own mouse tracking is unnecessary. Launch it with --no-mouse to avoid the whole class of bug.


victorarias (approved):

figgyster (approved):
This is safe to merge. `--no-mouse` is Copilot's documented way to preserve terminal-native selection and scrollback, and applying it at command construction covers fresh, resumed, yolo, and prompted sessions without disturbing their existing arguments.

The `main` conflict is resolved with both changelog entries preserved, and the relevant CI/preflight checks are green. No blockers from me.

PR #559, opened 2026-07-15
feat(wsctl): profile-aware daemon targeting and app-visible sessions

victorarias:
## Summary

- `scripts/wsctl` previously hardcoded the dev daemon `ws://localhost:29849/ws` unless `ATTN_WS_URL` was set, which silently sent commands to the wrong daemon when working on a named `ATTN_PROFILE`.
- It now resolves the target daemon in this order: `ATTN_WS_URL` explicit override > `ATTN_PROFILE`-derived port (`config.WSPortForProfile`) > dev fallback. The resolved target is printed to stderr on every invocation. wsctl never targets prod implicitly — prod requires an explicit `ATTN_WS_URL`.
- `add-session` previously created a workspace layout pane client-side before calling `spawn_session`, so wsctl-spawned sessions render in the app instead of leaving a blank main area. Per review feedback, that guarantee now lives in the daemon: `spawn_session` ensures a session's layout pane on its success path (adopting a pre-created pane when the app or another flow already made one, otherwise creating one with default placement). A failed or timed-out spawn therefore has nothing to roll back — there's no window where a pane can outlive a spawn that never succeeded. This is a single mechanism shared by the app's own pre-create flow and bare spawns (wsctl, scripts), and it made the `--no-pane` flag meaningless, so it was removed.
- Unit tests cover the URL resolution logic (including a "never implicit prod" guard) and the pane-guarantee behavior, including a regression test asserting a failed bare spawn leaves no ghost pane.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./scripts/wsctl`
- [x] `go test ./scripts/wsctl/... ./internal/daemon/...`
- [x] Live-verified against a throwaway profile (`mdclick1`): the profile-derived banner `[wsctl → ws://localhost:25696/ws]` printed correctly, and `add-workspace` + `add-session --agent shell` rendered a live shell pane in `attn-mdclick1.app` with streaming output.

### DIFF
--- a/internal/daemon/workspace_session_protocol_test.go
+++ b/internal/daemon/workspace_session_protocol_test.go
@@ -77,6 +77,140 @@ func TestWorkspaceSessionProtocolLifecycleMatchesAppOrder(t *testing.T) {
 	}
 }
 
+// Bare spawn_session (no pre-created pane — wsctl, scripts) must still yield a
+// rendered session: the daemon ensures a layout pane on the spawn success path.
+func TestWorkspaceSessionProtocolBareSpawnEnsuresLayoutPane(t *testing.T) {
+	d := NewForTesting(filepath.Join(t.TempDir(), "test.sock"))
+	d.ptyBackend = &fakeSpawnBackend{}
+	client := newWorkspaceProtocolTestClient()
+	workspaceID := "workspace-bare-spawn"
+	sessionID := "session-bare-spawn"
+	cwd := t.TempDir()
+
+	d.handleRegisterWorkspace(client, &protocol.RegisterWorkspaceMessage{
+		Cmd:       protocol.CmdRegisterWorkspace,
+		ID:        workspaceID,
+		Title:     "Bare Spawn",
+		Directory: cwd,
+	})
+	d.handleSpawnSession(client, &protocol.SpawnSessionMessage{
+		Cmd:         protocol.CmdSpawnSession,
+		ID:          sessionID,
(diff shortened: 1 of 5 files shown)


figgyster (dismissed):
The profile targeting is a good correction, but the pane-first path leaves durable debris when spawning fails.

`add-session` persists a `spawning` workspace pane before `spawn_session`. If the spawn is rejected (bad cwd/agent) or times out, the function returns without closing that pane. The app's corresponding flow explicitly rolls the pane back in its catch path; `wsctl` should do the same, using the returned `pane_id`, so a failed command cannot leave a ghost pane that renders a session which never started. Please cover the failure/rollback path too—the new tests currently exercise URL selection only.

Required before merge:

- close the newly-created pane when `spawn_session` fails or times out (while leaving `--no-pane` unchanged)
- add a regression test for that failure cleanup

CI is green. My local focused test attempt could not run because the isolated runner has Go 1.24 while this repository requires Go 1.25.3.

figgyster (approved):
The ghost-pane path is fixed. Making the daemon ensure a session pane only after a successful spawn gives bare callers the same rendered-session guarantee without creating rollback debt, and the idempotent adoption path preserves the app's existing pane placement and title.

The new daemon tests cover successful bare spawn, failed spawn with no pane, and adoption without duplication. Current-head Daemon and release-preflight checks are green; the trend-only benchmark is still running. No blockers from me.

PR #560, opened 2026-07-15
fix(frontend): capture session spawn args before pane-add await

victorarias:
## Summary
- Fixes an intermittent "Session spawn arguments were not prepared." failure when opening a new session while other sessions are busy.
- A just-created session's workspace has no `spawning` pane yet, so the launching-session preservation guard doesn't cover it. A daemon sessions broadcast that lands during the `sendWorkspaceAddSessionPane` round-trip could prune the session before `takeSessionSpawnArgs` ran.
- Fix: take the spawn args synchronously right after `createSession`, before the pane-add await — mirroring the ordering `createSplitSession` already uses.

## Test plan
- [x] Added regression test in `app/src/store/sessions.test.ts` that reproduces the race at the store level (spawn args available immediately post-create, gone after a mid-flight prune).
- [x] `pnpm exec tsc --noEmit` clean.
- [x] `pnpm run test` — 165 files / 1756 tests pass.
- [x] Verified the fixed packaged-app session-create path end-to-end on a throwaway profile.


figgyster (approved):
The race was hiding in the await, exactly where these things like to smoke. Capturing the spawn arguments before the pane-add round trip closes the prune window and matches the established split-session ordering without weakening rollback.

The focused session-store suite passes locally (14/14), and the frontend/E2E checks are green. No blockers from me.

PR #561, opened 2026-07-15
feat(plugin): accept newer OpenCode releases

victorarias:
## Intent / Why

The OpenCode plugin currently rejects every release except 1.17.16 and 1.17.18, forcing an attn update even when a newer stable OpenCode release preserves the server contract. This change makes compatibility forward-looking while retaining the strict checks that protect native-session resume and per-run server identity.

## Summary

- Replace the exact release allowlist with strict stable-version parsing and a minimum supported version of 1.17.16.
- Allow existing native sessions to resume after a forward OpenCode upgrade while rejecting downgrades.
- Keep each launched server pinned to the exact version recorded for that run, so wrong-process or wrong-port connections still fail closed.
- Document the compatibility policy and its user-visible behavior.

## Test plan

- [x] `bun test` in `plugins/attn-opencode` — 40 tests, 93 assertions.
- [x] `go test ./internal/daemon -run TestPluginDriver -count=1`.
- [x] `make dev` — full isolated app build, signing, installation, and launch.
- [x] Installed this branch plugin into the dev profile; daemon reported the driver registered and healthy.
- [x] Launched a real OpenCode 1.17.18 TUI in `attn-dev`, received `OPENCODE_VERSION_LIVE_OK`, and observed the attn session settle to `idle`.

## Out of scope

Launch-instruction parity, prose stop classification, plugin supervision, and bundled distribution remain separate PRs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   approval prompt, nothing is sent and your annotations are kept so you can retry.
 
 ### Changed
+- **OpenCode upgrades no longer require an attn plugin update.** The plugin now
+  accepts stable OpenCode releases at or above 1.17.16 and can resume existing
+  sessions after a forward upgrade. Older, preview, and downgrade versions are
+  still rejected, while a newer incompatible server reports the concrete API
+  failure through plugin health.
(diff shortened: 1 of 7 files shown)


figgyster (approved):
This compatibility gate looks safe to open up. Stable releases are parsed strictly at the 1.17.16 floor, resume can only move forward, and the launched server still has to match the exact version pinned to that run. The function said it was compatible; this time it brought identification.

The focused plugin suite passes (40 tests, 93 assertions), and the upgrade, downgrade, and wrong-server cases cover the contract at the right seam. No blockers from me.

PR #562, opened 2026-07-15
fix(notebook): three editor interaction fixes

victorarias:
## Summary
- The "Send to chief" pill now hangs below the current selection instead of covering the line above it.
- Typing works immediately after resolving a save conflict banner, with no extra click needed to regain editor focus.
- A note's YAML frontmatter no longer shows stray markdown list bullet decorations while editing it.

## Test plan
- `go build ./...`
- `go test ./internal/daemon ./internal/protocol`
- `pnpm exec tsc --noEmit`
- `pnpm exec vitest run` (166 files / 1766 tests)
- Live-app verification pending in this PR flow (part of stacked notebook-editor-polish series; part of the larger epic already exercised live per prior PRs in the stack)

### DIFF
--- a/app/e2e/notebook-browser.spec.ts
+++ b/app/e2e/notebook-browser.spec.ts
@@ -385,4 +385,35 @@ test.describe('NotebookBrowser (fs surface)', () => {
     await expect(page.locator('.cm-md-code')).toHaveCount(0);
     await expect(content).toContainText('a horizontal rule');
   });
+
+  test('anchors the send-to-chief pill below the selection end, not over the line above it', async ({ page }) => {
+    await page.goto('/test-harness/?component=NotebookBrowser');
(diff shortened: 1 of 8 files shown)


figgyster (approved):
These three interaction fixes hang together cleanly: editor focus returns after conflict resolution, the selection action uses CodeMirror's live end coordinates, and YAML frontmatter no longer picks up Markdown decorations. The focused unit tests and typecheck pass, and the Frontend/E2E checks are green.

One non-blocking edge for a follow-up: `frontmatterEnd` scans only the first 4 KiB while `frontmatterCard` parses the whole document, so unusually long but valid frontmatter can still leak Markdown decorations while raw editing. Sharing one frontmatter boundary would keep the two extensions from drifting.

Safe to merge. The code knew where focus belonged; it just needed directions home.

PR #563, opened 2026-07-15
chief prompt: awareness stays on when the user is driving

victorarias:
## Summary
- The chief-of-staff prompt framed a "recommended next step" as an action the chief was staging for the user's approval, so telling it "I'm driving / don't intervene" read as a rejection and it stopped relaying agent status updates.
- Reframes the guidance so a recommended next step is explicitly advice for the user to act on (or route to a delegation), never a move the chief holds pending sign-off. Awareness/reporting continues regardless of who acts.
- Adds a CHANGELOG entry describing the user-visible fix.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/hooks/`
- [x] `gofmt -l internal/hooks/hooks.go` (clean)
- No behavior/code changes beyond a prompt string and changelog entry; not a daemon-process/protocol/PTY/UI surface change, so no live-app verification performed.


figgyster (approved):
This looks safe to merge. The recommendation is now explicitly advice for the user, while the chief's existing awareness-and-upkeep duty stays intact. That closes the reported failure mode without widening the chief's authority or touching runtime/protocol behavior.

No blockers from me. Non-blocking: `TestChiefGuidance` could pin the new "never a move you stage" clause so this exact regression cannot quietly wander back in later.

PR #564, opened 2026-07-15
feat(notebook): relative note links resolve, images fixed, heading jumps

victorarias:
## Summary
- Relative links between notes now resolve against the linking note's own folder instead of being treated as external URLs, and `#heading` links jump the editor to that heading in the same note.
- Centralizes href classification (note/fragment/external) and directory-relative path resolution (with root-clamped `..` handling) into `resolveNotebookLink`, shared by broken-link detection and NotebookSurface's link-following.
- Fixes a double-resolution bug in `imageWidget` where an image's raw `src` was pre-resolved against an implicit empty baseDir before the real baseDir resolution ran downstream, mis-clamping `..`-relative image paths.

Out of scope: cross-note anchor scrolling (mod-click a link to another note's heading) — `loadFile` is fire-and-forget with no return value and no non-invasive way to know when the target note finishes loading, so scrolling to a heading in a *different* note is left for a follow-up.

Part of the notebook editor polish stack; based on origin/main (post PR7 #562).

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/daemon ./internal/protocol`
- [x] `pnpm exec tsc --noEmit`
- [x] `pnpm exec vitest run` (167 files / 1777 tests)
- [ ] figgyster review
- [ ] CI green

### DIFF
--- a/app/e2e/broken-links.spec.ts
+++ b/app/e2e/broken-links.spec.ts
@@ -28,9 +28,11 @@ test.describe('LiveMarkdownEditor broken links', () => {
     expect(color).not.toBe(realColor);
 
     // The existence check ran for the in-notebook links but NEVER for the external one.
+    // notebookLinkPath now returns a normalized, no-leading-slash root-relative path
+    // (via resolveNotebookLink) rather than the href's raw leading-slash form.
     const checked = await page.evaluate(() => window.__HARNESS__.getCalls('existsFile').map((c) => c[0]));
-    expect(checked).toContain('/knowledge/areas/missing.md');
-    expect(checked).toContain('/knowledge/areas/real.md');
+    expect(checked).toContain('knowledge/areas/missing.md');
+    expect(checked).toContain('knowledge/areas/real.md');
     expect(checked.some((p: string) => p.includes('example.com'))).toBe(false);
 
     await page.screenshot({ path: 'test-results/broken-links.png' });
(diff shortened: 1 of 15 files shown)


figgyster (dismissed):
One blocker before merge: the new heading-jump E2E fails in the required CI path. After finder navigation opens `foo.md`, the test expects `scrollTop < 40` but gets `635` before the link is clicked, so it never establishes the precondition needed to prove the click moved the editor.

Please reset/wait for the scroller to reach a controlled starting position (or assert movement from an explicitly established position) and rerun the E2E job. Shard 3's split-pane failure is unrelated to this patch; shard 2 is the new test added here.

The resolver/unit seam looks sound: the focused 62 Vitest cases pass locally and `git diff --check` is clean. I couldn't perform the required macOS live-app check from this Linux review runner, so please include that evidence with the next push as well.

figgyster (dismissed):
The CI blocker is fixed: the test now establishes a controlled scroll position, the focused 62 Vitest cases pass, and all current E2E shards are green.

One merge blocker remains from the repository's verification contract: this changes interactive Notebook UI behavior, but the PR still records no macOS live-app check. Please exercise the relative-note navigation, same-note heading jump, and `..`-relative image path in a non-production profile and record the result. Once that evidence is here, I see no remaining code blocker.

figgyster (dismissed):
The packaged-app harness now clears the relative-note navigation and same-note heading-jump parts of the live-app gate, and the current Frontend/E2E matrix is green. The focused 62 Vitest cases pass as well.

One requested live-app check remains: the scenario never seeds or renders a `..`-relative image, and the recorded output contains no image-path verification. Please extend the committed harness (or record equivalent packaged-app evidence) with a note in a nested directory that renders an image through a `..` path. That is the only blocker I still see.

figgyster (approved):
That closes the remaining gap. The packaged-app run now exercises the nested-note `..` image path through the real asset-read IPC flow, and the full current CI matrix is green.

Safe to merge. The suspect finally brought a valid alibi.

PR #565, opened 2026-07-15
feat(plugin): inject attn guidance into OpenCode

victorarias:
## Intent / Why

OpenCode sessions launched through attn's plugin driver did not receive the workspace, workflow, ticket, or chief guidance that built-in agents rely on. That left OpenCode outside attn's coordination model and made chief promotion unsafe because the running conversation could not adopt its new role.

## Summary

- extend plugin API v3 with an attn-composed launch-instructions bundle, including pre-persistence workspace-context checkout and safe rollback
- keep each OpenCode run's guidance in a private file and inject it through a per-process system hook that rereads the file on every prompt
- support fail-safe OpenCode chief promotion and demotion by preflighting resume before replacing the live worker, while preserving the native conversation, model, effort, and yolo settings
- document the plugin contract and user-visible OpenCode guidance behavior

## Test plan

- [x] `go test ./...`
- [x] `bun test` in `plugins/attn-opencode` (45 tests, 108 assertions)
- [x] Build, sign, install, and launch the isolated `attn-dev.app`
- [x] Verify a fresh real OpenCode session receives its private workspace-context guidance
- [x] Update workspace context through the real OpenCode session and verify the resumed native conversation receives the current checkout
- [x] Promote and demote the same native OpenCode conversation, verifying chief guidance and restored workspace guidance after each live reload

## Out of scope

- refreshing guidance in the middle of an already-running model turn
- a reusable SDK helper for plugin-owned private instruction files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-15]
 
 ### Added
+- **OpenCode sessions now receive attn's launch guidance.** Ordinary sessions
+  get their workspace context plus workflow and ticket instructions; OpenCode
+  chiefs get Notebook guidance. Promotion, demotion, and resume refresh that
+  guidance without changing the native OpenCode conversation or modifying the
+  user's repository and global OpenCode configuration.
 - Markdown tiles can now be annotated: select text to comment, redline, or apply
   quick-labels (with keyboard shortcuts), add document-wide comments, and review
   everything in a sidebar. Annotations survive file edits by re-anchoring to the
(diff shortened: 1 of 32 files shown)


figgyster (dismissed):
The guidance transport itself is careful, but the chief-toggle transaction can still leave attn lying about the active role.

`handleSetChiefOfStaff` persists and broadcasts the new role, retargets ticket delivery, and returns a successful `chief_of_staff_result` before the asynchronous `reloadSessionAgent` has run `driver.resume`. If that preflight fails, `preparePluginReload` deliberately preserves the live OpenCode worker—but that worker keeps its old instruction file while the store/UI already says it is chief (or demoted). The new failure-path test calls `reloadSessionAgent` only after pre-seeding the role, so it does not cover this public request/result mismatch.

Required before merge: make plugin reload preflight part of the role-change transaction. Either preflight before committing/broadcasting the role, or roll the role and ticket-delivery target back and return an error when preflight cannot prepare the replacement. Please cover promotion/demotion through `handleSetChiefOfStaff` with a failing `driver.resume` and assert both the result event and persisted role remain honest.

The function said the promotion succeeded. The running conversation had an older badge in its pocket.

figgyster (approved):
The fix closes the blocker. Plugin role changes now preflight the desired guidance before persisting the role, and the public-handler coverage proves both promotion and demotion failures leave the stored role and live worker untouched.

Daemon and release-preflight checks are green. Bench A/B was still running at the last refresh, but it does not gate this lifecycle correction. No blockers from me.

The badge and the conversation have compared notes. Case closed.

PR #567, opened 2026-07-15
feat(plugin): classify OpenCode prose stops

victorarias:
## Intent / Why

OpenCode can finish a turn by asking for direction in ordinary assistant prose without invoking its native question tool. attn then needs semantic stop classification to distinguish those turns from completed answers while preserving OpenCode's authoritative question and permission signals.

## Summary

- Normalize the newest completed assistant prose from OpenCode while excluding synthetic, ignored, reasoning, tool, and file parts.
- Classify explicit idle turns in a temporary same-server OpenCode session with every discovered tool disabled, strict JSON verdicts, bounded execution, and mandatory cleanup.
- Reserve report sequences before slow classification so newer busy, question, permission, error, or close signals cancel and overtake stale work.
- Cache verdicts by native message ID and exact text hash in the migrated run registry, and report uncertainty as unknown.
- Advertise classifier support after a 5/5 real-model fixture evaluation and linked-session live-app verification.

## Test plan

- [x] `bun test` — 61 tests, 0 failures
- [x] `go test ./...`
- [x] Non-prod `attn-dev.app`: a prose-only question moved the linked session to `waiting_input`
- [x] Non-prod `attn-dev.app`: a completed follow-up moved the same linked session to `idle`
- [x] Verified the linked TUI remained selected and no temporary classifier session survived cleanup
- [x] Real OpenCode 1.17.18 evaluation: 0/2 false negatives and 0/3 false positives

## Out of scope

A generic daemon classifier and classifier cost/latency telemetry remain deferred unless real usage shows the plugin-owned boundary needs them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-15]
 
 ### Added
+- **OpenCode prose questions now surface as waiting for input.** Native question
+  and permission requests remain authoritative, while an explicit idle turn can
+  also distinguish ordinary assistant questions from completed answers without
+  changing the visible OpenCode conversation.
 - **OpenCode sessions now receive attn's launch guidance.** Ordinary sessions
   get their workspace context plus workflow and ticket instructions; OpenCode
   chiefs get Notebook guidance. Promotion, demotion, and resume refresh that
--- a/docs/plans/2026-07-15-opencode-prose-classifier.md
+++ b/docs/plans/2026-07-15-opencode-prose-classifier.md
@@ -1,0 +1,60 @@
+# OpenCode prose stop classification
+
+## Why
+
+OpenCode already reports native questions and permissions accurately, but an
+assistant can still ask for direction in ordinary prose and leave attn showing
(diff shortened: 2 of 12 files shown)


figgyster (dismissed):
One contract hole remains in the prose extractor. `lastAssistantTurn()` sorts all completed assistant messages, but only normalizes `candidates[0]`. OpenCode can produce a completed assistant message containing tool/reasoning parts and no prose; if that message is newer than a completed prose question, `normalizeAssistantTurn()` returns `undefined`, and the driver reports `idle`. That recreates the false-done state this feature is meant to prevent.

Required before merge: select the newest completed candidate that actually contains normalized prose (while preserving the malformed-metadata fail-closed behavior), and add a regression with an older completed text question followed by a newer completed tool-only assistant message.

The rest of the shape looks solid. Local sandbox verification: `bun test` passed 61/61, and `git diff --check` was clean.

figgyster (approved):
Fixed cleanly. The extractor now skips completed non-prose envelopes and the regression covers the exact older-question/newer-tool-only ordering that caused the false `idle` result.

No blockers from me. Local sandbox verification passed 62/62 tests, the full patch is whitespace-clean, and the visible checks are green.

PR #568, opened 2026-07-15
feat(plugins): supervise installed plugin processes

victorarias:
## Intent / Why

An installed plugin that exits or loses its daemon connection currently stays
offline until attn restarts. For OpenCode, that also stops state monitoring even
though the attn-owned TUI and native OpenCode session are still alive.

This change makes plugin availability a daemon-owned responsibility and lets a
restarted OpenCode plugin resume monitoring the exact run it previously owned.

## Summary

- Add a generic installed-plugin supervisor with a five-second connection
  grace, bounded 250 ms–30 s restart backoff, a 60-second stability reset, and
  generation guards that reject stale processes.
- Expose starting, connected, backoff, and stopped phases plus restart and exit
  diagnostics in Settings.
- Bump the plugin wire API to v4. `driver.register` now returns the registering
  plugin's store-authoritative active runs, and bundled manifests and the SDK
  carry the daemon-issued process generation.
- Reconcile OpenCode's private registry against those active runs after a plugin
  restart, remove orphaned records, and rebuild monitoring against the same
  authenticated server and native session without launching another TUI.

## Test plan

- [x] `go test ./...`
- [x] Race-enabled supervisor, generation, and real child-process tests
- [x] OpenCode plugin: 63 tests, 175 assertions
- [x] Plugin SDK: 9 tests, 16 assertions
- [x] Settings: 44 tests
- [x] Build, sign, and install the isolated app with `make dev`
- [x] In `attn-dev`, complete an OpenCode turn, terminate only the confirmed dev
      plugin process, observe a replacement process register healthy, and
      complete a second turn in the same visible TUI. The attn run ID and native
      OpenCode session ID remained unchanged, and ordered reporting advanced
      from sequence 8 to 13. The production plugin process was separately
      identified and left untouched.

## Out of scope

- Manual “restart now” controls
- Health-triggered restarts for a connected but unresponsive plugin

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-15]
 
 ### Added
+- **Installed plugins now recover automatically after crashes and lost connections.**
+  Attn restarts them with bounded backoff, shows their recovery state in
+  Settings, and lets OpenCode resume monitoring the same native sessions without
+  launching replacement TUIs.
 - **OpenCode prose questions now surface as waiting for input.** Native question
   and permission requests remain authoritative, while an explicit idle turn can
   also distinguish ordinary assistant questions from completed answers without
--- a/app/src/components/SettingsModal.css
+++ b/app/src/components/SettingsModal.css
@@ -326,7 +326,8 @@
 .settings-pill.warn,
 .endpoint-status-badge.status-connecting,
 .endpoint-status-badge.status-bootstrapping,
-.plugin-status-badge.starting {
+.plugin-status-badge.starting,
+.plugin-status-badge.backoff {
   color: #f59b4a;
   border-color: rgba(245, 155, 74, 0.42);
   background: rgba(245, 155, 74, 0.11);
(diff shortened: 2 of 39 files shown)


figgyster (dismissed):
The supervisor itself is careful, but the remove failure path leaves an installed plugin intentionally offline. `handleRemovePluginWS` moves the supervisor to `stopped` before `plugins.Remove`; if that filesystem operation fails, the handler returns the error without restoring desired state. The manifest is still installed, but its process will no longer be restarted until the daemon itself restarts. That cuts across this PR's availability contract and regresses the old remove-on-success ordering.

Required before merge: make failed removal restore/retain supervision (or otherwise commit the stopped state only after deletion is known to succeed), and cover the filesystem-failure path with a regression test.

The rest of the lifecycle shape looks solid: generation ownership, disconnect grace, stable-window backoff reset, and store-authoritative OpenCode recovery all line up. OpenCode (63), SDK (9), and Settings (44) focused tests passed locally; Daemon and the full visible CI suite are green.

figgyster (approved):
The failed-removal path is fixed cleanly: deletion now succeeds before supervision is stopped, and the regression test proves a permission failure leaves the installed process desired and running.

Daemon, Frontend, release preflight, and two E2E shards are green; the last E2E shard was still running at my final check, but this follow-up is confined to the daemon removal path and its test. No blockers from me.

PR #569, opened 2026-07-16
feat(plugins): bundle OpenCode as an opt-in plugin

victorarias:
## Intent / Why

OpenCode support currently requires an attn source checkout and a Bun-based local plugin install. This makes a first-party integration feel like a development workflow and prevents each profile from opting in cleanly. This change ships the adapter with the app while preserving the out-of-process driver boundary and keeping it inactive until the user chooses to install it.

## Summary

- add a merged bundled/user plugin catalog with durable per-profile install state, collision handling, active-run uninstall protection, and immediate start/stop lifecycle
- support validated standalone executable entrypoints alongside the existing Bun manifest shorthand
- add explicit Settings and CLI Install/Uninstall actions and expose availability, installation, runtime, and allowed-action state through protocol v166
- compile the OpenCode adapter and guidance plugin into app resources, include their inputs in source fingerprints, and sign the nested executable before the app
- run the packaged adapter without a Bun runtime while preserving source-form development installs

## Test plan

- [x] `go test ./...`
- [x] `pnpm --dir app test` (1,768 tests)
- [x] `bun test --cwd plugins/attn-opencode` (64 tests, 179 assertions)
- [x] regenerate TypeSpec outputs and verify shell syntax for packaging/fingerprint scripts
- [x] build and sign isolated `bundle-a` and `bundle-b` app profiles
- [x] verify a fresh profile lists OpenCode as Available/stopped and does not launch it
- [x] install from the signed app resource, complete a real OpenCode 1.17.18 session, reload and resume the same native session, and verify healthy runtime state
- [x] reject uninstall during an active delegated run, then close and uninstall while preserving plugin data
- [x] reinstall the app while opted in and verify the profile retains its installation choice; verify another profile remains untouched
- [x] rebuild from the final tree and repeat the packaged Available -> Installed/healthy -> Available lifecycle

## Out of scope

- remote or third-party plugin catalogs
- destructive removal of plugin-owned run data


figgyster (dismissed):
The bundled catalog and supervisor split is coherent, but one public path still bypasses the daemon-owned safety contract.

Required before merge:

- Route the legacy CLI `plugin install` / `plugin remove` mutations through the daemon action path, or enforce the same collision and active-run guards there. `runPluginInstall` still writes directly to the profile plugin directory (`cmd/attn/plugin.go:75`), so a same-name user plugin can be installed while the bundled entry is opted in; after restart the user copy shadows the bundled artifact while the allowlist still says the bundled copy is installed. `runPluginRemove` likewise deletes directly (`cmd/attn/plugin.go:135`), bypassing `ListAgentDriverRuns`, so it can remove a plugin that still owns a live delegated run. The Settings/new `uninstall` path is guarded, but the old commands leave a side door open in the same public CLI.

I refreshed onto `25fd30d`, including the clean-checkout resource-root fix. Focused Go packages, the Settings/socket tests, TypeScript checking, the 64-test OpenCode adapter suite, and `git diff --check` passed locally; CI was green apart from one E2E shard still running at the last fetch. The remaining issue is the mutation contract, not the packaging shape.

figgyster (approved):
The side door is closed. Legacy CLI install/remove now use the daemon’s guarded action path, with focused coverage for bundled-name collisions and active-run removal. Keeping bundled run data outside the user-plugin install root also makes the source-override transition coherent instead of leaving two owners in one directory.

Focused Go suites, the 64-test OpenCode adapter suite, and `git diff --check` pass locally. CI is fully green. No blockers remain from me.

PR #570, opened 2026-07-16
fix(app): capture black-screen diagnostics

victorarias:
## Intent / Why

Agent and workspace switches can occasionally leave the app window black and unresponsive. Existing terminal diagnostics captured a matching `blank_after_resize` incident for the long-lived chief session, but they cannot tell whether React, the app-shell DOM, a native browser WebView, or the terminal renderer failed.

This change preserves enough bounded, disk-backed evidence to identify the failing layer the next time the issue occurs. It deliberately does not add automatic recovery yet, because reloading could erase the state needed for diagnosis.

## Summary

- add a bounded `ui-diagnostics.jsonl` stream for frontend errors, event-loop stalls, visibility changes, and agent/workspace switch snapshots
- capture root/app geometry, active terminal wrappers, focus ownership, the element covering the window center, and native browser-host visibility or suspicious full-window geometry
- show a persistent error screen when React rendering fails instead of leaving an unexplained black window
- keep terminal contents and user text out of the new diagnostics

## Test plan

- [x] Run all frontend tests: 167 files, 1,759 tests
- [x] Run TypeScript compilation and the production frontend build
- [x] Build, sign, install, and launch the exact branch as the isolated `attn-dev` profile
- [x] Verify the packaged app writes boot and 0/250/1500 ms switch snapshots to `ui-diagnostics.jsonl`
- [x] Exercise multiple real workspace switches and inspect the captured app-shell geometry
- [ ] Grant Accessibility access to the real-app input driver and rerun the full workspace-switching scenario end to end

## Out of scope

Automatic reload or renderer recovery remains deferred until the next captured occurrence establishes which layer failed. The packaged switching scenario produced valid switch snapshots, then stopped when its native input driver lacked macOS Accessibility permission.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-07-16]
+
+### Added
+- **Black or stuck UI failures now leave useful evidence on disk.** Agent switches capture delayed app-shell health snapshots alongside the existing terminal renderer diagnostics, including frontend crashes, event-loop stalls, and native browser visibility. A React render failure now shows a persistent error screen instead of leaving an unexplained black window.
+
 ## [2026-07-15]
 
 ### Added
--- a/app/src/App.css
+++ b/app/src/App.css
@@ -190,6 +190,29 @@ html, body, #root {
   overflow: clip;
 }
 
+.app-fatal-error {
+  height: 100%;
+  width: 100%;
+  padding: 48px;
+  background: var(--color-bg-app, #17181c);
+  color: var(--color-text, #f3f3f4);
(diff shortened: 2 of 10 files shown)


figgyster (dismissed):
The diagnostics path is headed in the right direction, but the patch is not clean yet.

Required before merge:
- Remove the extra trailing blank lines in `app/src/components/AppErrorBoundary.tsx:28` and `app/src/utils/uiDiagnosticsLog.test.ts:70`. `git diff --check origin/main...HEAD` currently fails on both files.

I stopped the heavier local test loop at the static failure; the visible frontend checks are otherwise passing, with E2E shard 3 still running at the time of review.

figgyster (approved):
The cleanup landed: the two trailing blank lines are gone, `git diff --check origin/main...HEAD` is clean, and the full visible frontend/E2E matrix is green on this head.

The app-shell diagnostics and persistent React failure surface look safe to merge. Case closed.

PR #571, opened 2026-07-16
feat(tickets): allow arbitrary file attachments

victorarias:
## Intent

Agents need to hand over durable artifacts such as self-contained HTML prototypes directly through `attn ticket attach`, rather than creating a Markdown substitute and copying files outside the supported workflow.

## Summary

- Allow any regular, non-symlinked file through the CLI, daemon staging, and canonical ticket artifact listing.
- Remove Markdown-only file selection and rename constraints in the ticket detail panel, so attached files remain discoverable and manageable in-app.
- Keep opening safe: HTML is shown as Notebook source text rather than executed; known binary formats retain the existing non-executing preview placeholder.
- Update agent guidance, CLI help, protocol documentation, glossary, changelog, and focused regression coverage.

## Test plan

- [x] `go test ./...`
- [x] `pnpm --dir app test -- --run src/components/TicketDetailPanel.test.tsx`
- [x] `ATTN_HARNESS_PROFILE=htmlattach pnpm --dir app run real-app:scenario-ticket-lifecycle`
- [x] Built and ran the isolated `htmlattach` profile; attached a self-contained HTML fixture with the real CLI, verified `ticket show`, and byte-compared the canonical Notebook copy.
- [x] Repeated the real CLI flow with a Markdown fixture to confirm existing behavior.

## Why safe viewing matters

This change broadens transport and listing, not execution. The existing Notebook file-kind policy displays HTML as editable source and leaves known binary formats as a non-executing placeholder.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,8 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ### Added
 - **Tickets can now carry any regular file.** Attach HTML prototypes, images,
-  archives, and other artifacts alongside Markdown; text formats open as source
-  in the Notebook, while binary formats remain non-executing previews.
+  archives, and other artifacts alongside Markdown; recognized text formats open
+  as source in the Notebook, while opaque files remain non-executing previews.
 - **OpenCode is now an opt-in plugin included with the app.** Each profile can
   install it from Settings or the CLI without an attn checkout, Bun, or a daemon
   restart. Uninstalling preserves OpenCode run data and is blocked while the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-15]
 
 ### Added
+- **Tickets can now carry any regular file.** Attach HTML prototypes, images,
+  archives, and other artifacts alongside Markdown; text formats open as source
+  in the Notebook, while binary formats remain non-executing previews.
 - **OpenCode is now an opt-in plugin included with the app.** Each profile can
   install it from Settings or the CLI without an attn checkout, Bun, or a daemon
   restart. Uninstalling preserves OpenCode run data and is blocked while the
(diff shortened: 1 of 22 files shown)


figgyster (dismissed):
The attachment path is coherent across the CLI, daemon, ticket listing, and UI, but the new arbitrary-extension surface opens one data-loss trap.

`TicketDetailPanel` now accepts every file type, while `app/src/components/notebook/fileKind.ts` uses a binary-extension denylist and classifies every unknown extension as editable text. Common opaque formats outside that list—such as `.docx`, `.dmg`, and `.pkg`—therefore go through `fs_read`, where raw bytes are converted to a string, and a later editor save can overwrite the canonical artifact with transformed text. The same path reads the whole file before responding, so a large unknown binary can also put unbounded memory pressure on the daemon.

Required before merge:

- Make editability fail closed for unknown/opaque files (for example, a positive text allowlist or daemon-provided/content-sniffed kind), and bound the text-read path.
- Add a regression using a common binary extension outside the current denylist that proves opening it shows the non-editing placeholder and never calls `fs_read`.

The frontend suite is otherwise clean here: 168 test files and 1,767 tests passed. The local Go sandbox could not execute because its Go 1.24 image is older than this repo's Go 1.25.3 requirement; the live `Daemon` CI check is green.

figgyster (dismissed):
The original blocker is fixed on this head: opaque/unknown extensions now fail closed, the `.docx` regression never calls `fs_read`, and the focused frontend seam passes 66/66 tests.

The read-cap fix introduced a new contract regression, though. Putting the 2 MiB limit in shared `fsdoc.Store.Read` also applies it to `fs_read_asset`, whose endpoint intentionally supports image payloads up to its separately derived ~6 MiB raw limit. The principal `Daemon` CI job now fails `TestFsReadAssetMaxSizeFitsMessageCap` because the previously valid max-size asset is rejected before the asset-specific transport cap runs.

Required before merge: bound the text `fs_read` path without shrinking the existing asset-read contract—for example with an endpoint-specific limited read or a store API that accepts the caller's cap—and restore the daemon suite to green.

figgyster (approved):
The fix is clean now. Opaque attachments fail closed, text reads keep the 2 MiB cap, and image assets retain their separate transport-derived limit through `ReadWithLimit`.

Daemon CI and all frontend/E2E checks are green on this head. No blockers from me.

PR #572, opened 2026-07-17
fix(runtime): prioritize active attn binary on PATH

victorarias:
## Why

An attn-launched process could resolve bare `attn` to a stale login-shell installation. Managed PTYs had a local workaround, while terminal panes and headless agents did not share that guarantee.

## Summary

- Centralize active-wrapper resolution and PATH normalization in `internal/launchenv`.
- Apply the invariant to managed PTYs, ordinary shell panes, and all Claude/Codex headless launches.
- Keep shell panes free of managed-session identity while ensuring their bare `attn` resolves correctly.
- Update the canonical bundled attn skill to prefer bare `attn`, retaining `ATTN_WRAPPER_PATH` as recovery.

## Test plan

- [x] `go test ./internal/launchenv ./internal/pty ./internal/agent`
- [x] `go vet ./internal/launchenv ./internal/pty ./internal/agent`
- [x] Compiled adjacent PTY backend, daemon, workflow, and CLI packages.
- [x] Built and exercised the isolated `attn-dev.app`.
- [x] `pnpm --dir app run real-app:scenario-workspace-shell-lifecycle`


figgyster (dismissed):
There’s one runtime-contract blocker here. The active directory is moved to the front of `PATH` in `buildSpawnEnv`, but the process is then started through a login shell (`shell -l`, or `shell -l -c` for managed agents). Login startup files run after this environment is constructed and can prepend a stale user install directory again, recreating the exact failure this PR is meant to close—especially for ordinary shell panes.

The current tests stop at the pre-shell environment, and the cited `workspace-shell-lifecycle` scenario exercises pane creation/typing/teardown without invoking bare `attn`, so neither catches that reorder.

Required before merge: enforce the active-attn precedence after login-shell initialization, and add a spawn-level regression whose login startup file prepends a stale `attn`, then proves bare `attn` resolves to the active wrapper.

figgyster (dismissed):
The managed-agent path is fixed, but the ordinary shell-pane contract still has the same gap. After restoring `PATH`, this bootstrap `exec`s a fresh `shell -i`; that final interactive shell then runs `.zshrc`/`.bashrc`, which can prepend the stale installation again. The new fake shell’s `-i` branch checks `command -v` immediately, so it does not model the startup phase that can undo the restoration.

This also changes the pane from the previous single interactive login shell to a non-login interactive shell, so login identity/logout behavior and shell state established by login startup files are no longer equivalent.

Required before merge: preserve the interactive login-shell contract while reasserting the active path after all startup files have run, and make the spawn regression reorder `PATH` during the final interactive startup before checking bare `attn`.

figgyster (dismissed):
The zsh/bash/fish startup paths now address the original ordering bug, but this introduces a separate shell-selection regression. `GetUserLoginShell` accepts any configured user shell; when `prepareShellPaneLaunch` does not recognize it, `Spawn` silently skips that shell and launches a fallback zsh/bash instead. Before this change, the configured shell itself was started with `-l`.

That means users of tcsh, ksh, nushell, or another custom login shell now receive a different interpreter in terminal panes, with different syntax and startup state.

Required before merge: do not silently replace the configured login shell. Preserve it through an appropriate post-startup strategy, or keep its prior launch behavior when the PATH hook is unsupported, with regression coverage proving an unrecognized primary shell is not substituted.

figgyster (approved):
You took the long road through shell-startup country, but this closes the case cleanly. Known shells restore the active `attn` path after their real startup hooks, while unsupported configured shells keep their original login launch instead of being silently replaced. The focused zsh, fish, and custom-shell coverage pins the contracts that mattered.

CI is green. No blockers from me.

figgyster (approved):
You added the one witness this case was still missing: real Bash startup now proves that a user profile can prepend the stale install and the final interactive login shell still resolves bare `attn` to the active wrapper. Production code is unchanged, the incremental diff is clean, and CI is green.

Current head is safe to merge.

PR #575, opened 2026-07-18
docs(vision): pi as the daily-driver harness — the attn plugin suite

victorarias:
## Summary

Vision doc for building attn's own plugin suite for the pi coding agent: pi as the daily-driver harness carrying the capabilities that matter (autonomy with a safety envelope, subagents, skills, compaction, background eyes, multi-model), tightly integrated with attn as a native citizen.

Key decisions captured:
- **Plugin suite, not a fork** — one main attn package for pi composing focused extensions; attn wants signals, steering, and capabilities, never loop ownership (fork and RPC-first embedding considered and rejected; openclaw's vendoring history is the price list).
- **Two plugin systems meet in the middle** — an attn agent-driver plugin (per the existing plugin system, opencode precedent) launches pi; a pi-side attn package declares state and takes doorbell/ticket steering.
- **Declared state, no classifier fallback** — with a stop-enrichment service (done vs. waiting-on-reply) surviving as an attn API the plugin calls.
- Skills trial as-is then curate; safety envelope designed in its own doc as the third rock.

Grounded in pi's actual extension surface and openclaw's embedding history (research summarized into the doc's decisions, detail deferred to per-capability implementation docs).

## Testing

Docs-only change; no runtime surface.

### DIFF
--- a/docs/vision/pi-attn-plugins.md
+++ b/docs/vision/pi-attn-plugins.md
@@ -1,28 +1,28 @@
 # Vision: pi as the daily-driver harness — the attn plugin suite
 
-> Grounded in pi's actual extension surface ([earendil-works/pi](https://github.com/earendil-works/pi),
-> `packages/coding-agent`) and in openclaw's history of embedding pi
-> ([openclaw/openclaw](https://github.com/openclaw/openclaw)). Builds on attn's
-> plugin system ([docs/plans/2026-04-16-plugin-system.md](../plans/2026-04-16-plugin-system.md)),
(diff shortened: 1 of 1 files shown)


figgyster (approved):
The direction hangs together. It builds on attn's existing external-driver ownership boundary, and pi's package/extension surface supports the single-install, composable suite described here.

No blockers from me. The safety-envelope design is where the teeth will need to show—especially process-level containment for bash and writes beyond the worktree—but this vision correctly leaves that implementation contract for its own alignment doc instead of pretending the problem is already solved. The bridge is drawn where the river actually is.

PR #576, opened 2026-07-18
docs: plan editor-over-arbitrary-roots epic

victorarias:
Plan/tracking doc for making the notebook UI surface (becoming "Editor") display arbitrary folder roots, defaulting to the selected workspace directory. Lays out a six-PR sequence; the storage-side Notebook concept is unchanged.

### DIFF
--- a/docs/plans/2026-07-18-editor-arbitrary-roots.md
+++ b/docs/plans/2026-07-18-editor-arbitrary-roots.md
@@ -1,0 +1,62 @@
+# Plan: Editor Over Arbitrary Roots
+
+## Why
+
+The notebook UI proved itself, but pinning it to `notebook.root` walls the user
+off from the rest of their working tree: a repo that *contains* the notebook
+can't be browsed, so the tool blocks the user's own system. The UI surface
+becomes a folder editor over any root, defaulting to the selected workspace's
+directory. The storage concept (keeper, journal, knowledge base) stays bound to
+`notebook.root` and keeps the Notebook name.
+
+## Decisions
+
+- Naming: storage keeps **Notebook**; the UI surface's user-facing name becomes
+  **Editor**. Internal identifiers (`tile_kind: 'notebook'`, component/file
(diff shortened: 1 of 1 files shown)


figgyster (dismissed):
This is the right product split: Notebook keeps custody of the knowledge system, while Editor stops pretending the rest of the working tree does not exist. But the plan hands the filesystem API a master key without naming who is allowed to hold it.

Required before this plan becomes the contract:

- Add an authorization boundary for every non-default `root` operation, including `fs_index`, `fs_watch`, and `fs_unwatch`. Today `fs_*` dispatches for ordinary accepted WebSocket clients; `IsBrowserHost()` is the existing authenticated Tauri-client boundary. Once `root` can name an arbitrary absolute directory, origin/path validation alone does not prevent another accepted local web client from reading, overwriting, deleting, indexing, or watching user files outside the Notebook. Either require the authenticated app identity (or an equivalently explicit capability/grant) for arbitrary roots while preserving omitted-root compatibility, and record that decision here.
- Put negative protocol tests in the daemon slice proving an ordinary/untrusted WebSocket client cannot operate on an arbitrary root. Cover at least read/write plus watch/index authorization, so later PRs cannot quietly widen the door again.

That is a security-contract issue, not architecture taste. Once the plan assigns custody of the key, the six-PR shape looks sound. The lazy tree, bounded index, root-tagged events, and persisted-tile compatibility are the right bones.

figgyster (approved):
The master key has a named custodian now. The plan puts every explicit-root operation behind the authenticated app identity at one resolution chokepoint, preserves omitted-root Notebook compatibility, and makes each daemon slice prove its denial path. That closes my blocker.

This is a sharp next act after the Notebook polish stack closed: keep Notebook as the durable knowledge system, let its proven UI grow into a general Editor, and refuse to blur the trust boundary along the way. The lazy tree, bounded index, root-tagged events, and persisted-tile compatibility give the arc good bones.

No blockers. Case closed.

PR #577, opened 2026-07-18
feat(daemon): fs commands accept an explicit root

victorarias:
PR1 of the editor-over-arbitrary-roots epic (docs/plans/2026-07-18-editor-arbitrary-roots.md).

Every `fs_*` WS command (list/read/write/exists/rename/delete/read_asset) gains an optional absolute `root` field — omitted defaults to the notebook root, preserving full backward compatibility. `fs_changed` gains a matching `root` field. Root validation is shared with `notebook.root` (`normalizeExternalRoot`), with a per-root `fsdoc.Store` cache. Notebook coupling (self-write suppression + `notebook_changed`) is now root-conditional.

ProtocolVersion bumped 166 -> 167 (constants.go and useDaemonSocket.ts in lockstep).

No changelog entry — protocol-internal, no user-visible behavior change yet.

## Test plan
- [x] `go build ./...`
- [x] `gofmt -l .` clean
- [x] `go test ./internal/daemon -run 'TestFs|TestNotebook|TestWebsocket|TestSettings|TestTicketAttach'` — all pass
- [x] `tsc --noEmit`
- [x] `vitest run src/hooks/useDaemonSocket.test.tsx` — 58 passed

### DIFF
--- a/app/src/hooks/useDaemonSocket.test.tsx
+++ b/app/src/hooks/useDaemonSocket.test.tsx
@@ -2935,15 +2935,15 @@ describe('useDaemonSocket fs surface', () => {
     unmount();
   });
 
-  it('invokes onFsChanged with origin and paths', async () => {
+  it('invokes onFsChanged with origin, paths, and root', async () => {
     const onFsChanged = vi.fn();
     const { unmount } = renderFsHook({ onFsChanged });
     const ws = await waitForOpenSocket();
 
     act(() => {
-      ws.emit({ event: 'fs_changed', origin: 'ui', paths: ['notes/todo.txt'] });
+      ws.emit({ event: 'fs_changed', origin: 'ui', paths: ['notes/todo.txt'], root: '/Users/x/attn-notebook' });
     });
-    expect(onFsChanged).toHaveBeenCalledWith('ui', ['notes/todo.txt']);
+    expect(onFsChanged).toHaveBeenCalledWith('ui', ['notes/todo.txt'], '/Users/x/attn-notebook');
     unmount();
   });
 });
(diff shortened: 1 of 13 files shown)


figgyster (dismissed):
I like the direction here. Making the filesystem surface root-explicit is the right foundation for the editor arc, and the per-root store plus root-tagged events keep ownership legible. But the new boundary has one unlocked side door.

`internal/daemon/ws_settings.go:511-516` excludes the attn data directory with a lexical comparison only. A root symlink such as `/tmp/editor-root -> ~/.attn` passes `normalizeExternalRoot`, while `fsdoc.Store` deliberately resolves and permits a symlinked root. That makes a request such as `fs_delete` for `attn.db` operate inside the protected data directory through the alias. This crosses the data-loss line.

Required before merge:

- Resolve the root (or its deepest existing ancestor) and compare that canonical target against the canonical attn data directory before accepting it. Preserve legitimate symlinked roots that resolve elsewhere.
- Add a regression test with an external-looking symlink into `config.DataDir()` and prove the `fs_*` surface rejects it, especially the mutating paths.

The alley is otherwise clean, but this lock needs to work from both sides.

figgyster (dismissed):
The lock catches aliases now. Canonicalizing the deepest existing ancestor on both sides closes the symlink route into the profile, and the regression covers the dangerous trusted-client delete path, nested profile aliases, and the legitimate external-symlink case.

The authenticated-app gate from the preceding fix is a worthwhile second lock too: this filesystem surface is becoming the foundation for the arbitrary-root editor, so narrowing both *who* can name a root and *where* that root may resolve is exactly the boundary I wanted to see after the notebook polish arc closed.

No blockers. Case closed.

figgyster (approved):
Good follow-up. The symlink regression now builds its profile under a temporary `HOME` and fails closed if `config.DataDir()` escapes that sandbox, so the test can prove destructive-path safety without ever putting the real profile in the witness chair.

The approved implementation is unchanged, the incremental patch is clean, and this strengthens the evidence at exactly the dangerous seam. No blockers. Case closed again; GitHub made me say it twice.

PR #578, opened 2026-07-18
feat(daemon): rotating SQLite backups

victorarias:
## Summary
Prod `attn.db` was destroyed today with zero backups. The daemon now takes a `VACUUM INTO` snapshot at startup and every 6h into `<data-dir>/backups`, keeping the last 12, plus an unpruned pre-migration snapshot before any schema migration. Backup failures never block daemon startup.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./internal/daemon/...`
- [x] Diff reviewed line-by-line and approved by the tech lead
- [x] Local `go test` green (tests remain hermetic under `t.TempDir()` HOME, never touching real `~/.attn`)
- [ ] CI green
- [ ] figgyster review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-07-18]
+
+### Added
+- **Automatic rotating database backups.** The daemon now snapshots its SQLite
+  database every 6 hours (keeping the last 12) and takes an extra safety
+  snapshot right before any database migration, so session, ticket, and
+  workspace state can be recovered if the database is ever lost or corrupted.
+
 ## [2026-07-16]
 
 ### Added
--- a/internal/daemon/backup.go
+++ b/internal/daemon/backup.go
@@ -1,0 +1,58 @@
+package daemon
+
+import (
+	"path/filepath"
+	"time"
+)
+
+// backupDir returns the directory rotating SQLite backups are written to:
+// a "backups" subdirectory of the daemon's runtime data root. In production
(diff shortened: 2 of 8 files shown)


figgyster (dismissed):
I wanted this one bulletproof. After losing the production DB, rotating snapshots are exactly the right repair—but the degraded-start path turns the safety net into a slow shredder.

Required before merge:

- `internal/daemon/backup.go:52`: do not run backup/rotation when `New` has fallen back to `store.New()` after the durable DB failed to open. That fallback is an in-memory SQLite database, so the current loop writes empty/degraded snapshots into the real `backups/` directory every six hours and prunes the oldest genuine snapshots. After twelve ticks, the recovery copies can all be gone precisely because the source DB was corrupt or unavailable. Track whether the daemon store is durably backed (or otherwise make `BackupNow` refuse the fallback store), skip rotation while persistence is degraded, and add a regression test proving degraded startup leaves existing backups untouched.

The backup and migration tests are otherwise aimed at the right seams, and the visible Daemon CI is green. My focused local run could not start because the review sandbox has Go 1.24 while this module requires Go 1.25.3; that limitation does not affect the data-loss path above, which follows directly from the existing fallback wiring.

One real blocker. Keep the empty witness out of the evidence locker, and this becomes the recovery story the incident deserved.

figgyster (approved):
The degraded store has been shown the door before rotation starts. Marking durability on the store and refusing `BackupNow` before directory creation or pruning closes the exact failure chain: a corrupt durable DB can no longer spend the next three days replacing its own recovery copies with empty fallback snapshots.

The regression test sits on the right seam, the current Daemon and release-preflight checks are green, and the remaining Bench A/B job is trend-only. The safety net now protects itself. No blockers—case closed.

PR #579, opened 2026-07-18
feat(daemon): per-root fs watchers via fs_watch/fs_unwatch

victorarias:
PR2 of the editor-over-arbitrary-roots plan (docs/plans/2026-07-18-editor-arbitrary-roots.md):

- Refcounted per-root watcher registry (cap 16)
- `fs_watch`/`fs_unwatch` commands gated through `resolveFsRoot` so explicit roots require the trusted app client
- Notebook root dual-broadcasts `fs_changed` + `notebook_changed`
- Dot-prefixed atomic-write temp files so generic watchers never see swap files
- Watcher cleanup on client disconnect
- Protocol version 168

## Test plan
- [x] go build/vet/test green locally
- [x] pnpm test green locally
- [x] protocol version 168 verified in lockstep (constants.go + useDaemonSocket.ts)

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -170,7 +170,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '167';
+export const PROTOCOL_VERSION = '168';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 interface PRActionResult {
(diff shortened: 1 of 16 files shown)


figgyster (dismissed):
The watcher registry is the right second slice. The refcounting, generic cleaner, and self-write suppression fit together cleanly. The formatting push cleared CI's Daemon check, but one trust-boundary rake is still lying in the doorway.

Required before merge:

- `internal/daemon/fs_watch.go:88` sends arbitrary-root changes through `broadcastFsChanged`, and `internal/daemon/fs.go:104` broadcasts that event to every websocket client. The trusted-app gate prevents an ordinary client from *starting* an explicit-root watch, but once the app starts one, every accepted client can receive the absolute root and changed paths. Route non-notebook watch events only to clients subscribed to that root (or otherwise restrict them to the trusted app), and add a two-client regression proving a non-subscriber/untrusted client receives nothing. Notebook-root compatibility can remain global.

Fix that audience leak and this should be a short second look. The architecture is sound; the broadcast audience is the suspect. Case stays open.

figgyster (approved):
The audience list now has a bouncer. Generic-root events snapshot only that root's active subscribers, delivery happens outside the registry lock, and the notebook root keeps its legacy global fan-out. The new two-client regression catches the exact leak: a connected non-subscriber hears nothing.

Daemon and Frontend are green on this head; the remaining E2E shards were still running at last look, but this fix is covered at the daemon boundary where the contract lives. This is the right continuation of the Notebook polish arc: opening the editor to arbitrary roots without opening the user's filesystem gossip to every local client.

No blockers. Case closed.

PR #580, opened 2026-07-18
feat(daemon): fs_index bounded recursive file index

victorarias:
PR3 of the editor-over-arbitrary-roots plan (docs/plans/2026-07-18-editor-arbitrary-roots.md) — new `fs_index` command returning a bounded recursive index of a root for the ⌘P finder:

- server-side 25k entry cap with a truncated flag
- skips dot-dirs/dot-files/node_modules/symlinks
- rejects nonexistent or non-directory roots
- inherits the explicit-root auth gate via the `resolveFsRoot` chokepoint, with its own denial test
- protocol version 169
- `sendFsIndex` socket helper

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -170,7 +170,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '168';
+export const PROTOCOL_VERSION = '169';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 interface PRActionResult {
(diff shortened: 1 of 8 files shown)


figgyster (dismissed):
This is a clean third step in the arbitrary-root editor arc: the auth gate stays at `resolveFsRoot`, the request/result correlation is wired end to end, and the bounded finder contract is the right shape. The new walker had two loose threads, though, and one of them leads somewhere users can trip.

Required before merge:

- `internal/daemon/fs_index.go:99` excludes symlinks, but every other non-directory entry is appended. FIFOs, sockets, and device nodes therefore enter a result documented as “regular files only”; the existing `fs_read` path rejects those paths as non-regular, so the finder can advertise entries it cannot open. Check `DirEntry.Info().Mode().IsRegular()` (or an equivalent regular-file test) before appending, and pin it with a special-file regression test.
- The required Daemon check is red because `internal/daemon/fs_index_test.go` is not gofmt-clean (`.hidden-dir/inside.md` comment alignment). Tiny rake, real gate.

The split-blank E2E shard failure looks unrelated to this filesystem/protocol slice, so I’m treating that witness as flaky rather than inventing a third charge. Fix the two concrete items above and this should be a short walk back to green.

figgyster (approved):
The finder now keeps the same definition of “file” as the reader, and the FIFO regression puts a lock on the door instead of a polite note beside it. The formatting is clean, the Daemon job is green, and the terminal flake from the first run vanished on rerun like the unreliable witness it was.

This third arbitrary-root slice preserves the auth chokepoint from the earlier work while giving the coming Editor UI a bounded index it can trust. No blockers on the new head. Case closed.

PR #581, opened 2026-07-18
feat(ptyworker): self-reap orphaned workers after idle TTL

victorarias:
## Summary
- Per-session PTY workers previously ran forever once their daemon was gone for good (e.g. a torn-down test profile, or a daemon killed and never restarted), leaking `attn pty-worker` and agent processes.
- Workers now self-reap after 12 hours with no authenticated daemon connection **and** no PTY output, ending the child agent process and removing the registry/socket files.
- Daemon restarts/upgrades are unaffected — the daemon's reattach (seconds away in the normal path) cancels the countdown, and a busy agent producing output defers the deadline until it goes idle.
- TTL is overridable via `ATTN_WORKER_ORPHAN_TTL` (Go duration; `0` disables reaping).

## Test plan
- [x] Unit tests in `internal/ptyworker/runtime_test.go` covering: idle-unowned reap, cancellation by an authed connection (and re-arm after it drops), deferral by ongoing output activity, disable via zero TTL, and no-arm when the session already exited (exit-cleanup owns that path).
- [x] `go test ./internal/ptyworker/...`, `go vet ./internal/ptyworker/...`, `gofmt -l` all clean.
- [x] Live-app verification already performed on a throwaway `ATTN_PROFILE` (not dev/prod): confirmed orphan reap actually kills the leaked worker/agent process, confirmed a daemon restart during the TTL window cancels the countdown and the worker survives, and confirmed `attn profile clean` no longer leaves an orphaned worker running.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,16 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-07-18]
 
+### Fixed
+- **Orphaned session workers now clean themselves up.** A per-session PTY worker
+  whose daemon is gone for good (e.g. a torn-down test profile, or a daemon that
+  was killed and never restarted) previously kept running forever, accumulating
+  stray `attn pty-worker` and agent processes. Workers now stop themselves after
+  12 hours with no daemon attached and no terminal output, ending their agent
+  process and removing their registry/socket files. Daemon restarts and upgrades
+  are unaffected — a reattaching daemon cancels the countdown, and a busy agent
+  producing output defers it.
+
 ### Added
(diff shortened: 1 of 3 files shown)


figgyster (approved):
This gives abandoned workers a bounded afterlife without booking ordinary daemon restarts for murder. The authenticated-connection gate, output-driven deadline deferral, and separate exited-session cleanup are the right safety shape; the focused lifecycle coverage sits on the seams where this could otherwise kill useful work.

Daemon CI is green, and the throwaway-profile live check covers the part unit tests cannot: the worker actually ends the child process and removes its registry/socket evidence. No blockers. This is a clean continuation of the durability hardening from #578—less glamorous than a new surface, more important than it looks. Case closed.

PR #582, opened 2026-07-18
feat(daemon): backup restore CLI, pre-migration snapshot cap, last-backup surfacing

victorarias:
## Summary
Implements §1 follow-ups of `docs/plans/2026-07-18-db-loss-mitigation.md`:

- Pre-migration snapshots are now capped at 5 (pruned by trailing timestamp — being exempt from rotation no longer means immortal).
- `attn db restore [path|latest]` restores `attn.db` from a rotating or pre-migration snapshot. It refuses while a daemon holds the pid-file flock or when liveness is undeterminable, preserves the previous `attn.db` and its `-wal`/`-shm` sidecars as `attn.db.pre-restore-<ts>`, and copies (never moves) the snapshot into place.
- `db.last_backup_at` is a computed, read-only settings key broadcast after each successful backup.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)
- [x] `go test ./internal/store ./internal/daemon ./cmd/...`
- [x] Restore live-smoked end-to-end on a throwaway `ATTN_PROFILE`: normal restore, and refusal while a daemon is live
- [x] CHANGELOG.md updated

Known pre-existing flake unrelated to this change: `TestDaemon_PrunesSessionsWithoutLivePTYOnStart` can fail on machines with a live daemon (order-dependent).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   database every 6 hours (keeping the last 12) and takes an extra safety
   snapshot right before any database migration, so session, ticket, and
   workspace state can be recovered if the database is ever lost or corrupted.
+- **Restore the database from a backup.** `attn db restore [path|latest]`
+  restores `attn.db` from a rotating snapshot (the newest one by default),
+  refusing while the daemon is running and always preserving the previous
+  database file rather than deleting it. Settings now also surface the
+  timestamp of the most recent successful backup.
 
 ## [2026-07-16]
 
--- a/cmd/attn/db.go
+++ b/cmd/attn/db.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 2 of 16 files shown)


figgyster (dismissed):
There’s one recovery-path blocker here. The rest of the durability follow-up is a strong continuation of #578: bounded pre-migration snapshots, a deliberately boring operator CLI, and a read-only freshness signal are exactly the right shape after a database-loss incident. CI is green, and `git diff --check` is clean. My local Go run could not start because the review sandbox still carries Go 1.24 while this branch requires 1.25.3.

The restore itself can strand the profile without a usable `attn.db`. `restoreDatabase` only proves that an explicit source exists, then renames the current database and sidecars before it opens/copies the source. For example, `attn db restore <data-dir>/attn.db` renames the source out from under itself and then fails to open it; passing a directory also gets through `Stat`, creates/truncates the destination, and fails during `io.Copy`. Either command returns an error after the live database path has already disappeared or become empty. A second restore within the same second can also reuse `attn.db.pre-restore-<timestamp>` and `os.Rename` will replace the earlier preserved copy.

Please make the replacement fail-safe before mutating the live path: validate a regular source and reject `os.SameFile` with the destination, copy it to a uniquely named temp file beside `attn.db` first (including close/sync errors), choose a collision-proof preservation name, and roll back the old database/sidecars if the final swap cannot complete. Regression coverage for source==destination, a non-regular/unreadable source, and preservation-name collision would put teeth in the promise. Recovery code only gets one umbrella; right now it opens after the rain starts.

figgyster (dismissed):
The new head fixes the source-validation, staging, same-file, and main preservation-name cases from my first review. That’s the right direction, and the new tests cover those doors cleanly. One narrower atomicity gap remains before I can turn the light green.

`preserveExistingDB` renames `dbPath` at line 209, then returns immediately if either sidecar rename at lines 212–215 fails. At that point `restoreDatabase` removes the staged copy on return, but nothing moves `preservedAs` back to `dbPath`, so a WAL/SHM filesystem error still strands the profile without its live database. The “collision-proof” selection also checks only `preservedAs`; if `preservedAs+"-wal"` or `+"-shm"` already exists while the main target does not, `os.Rename` can silently replace that older preserved sidecar.

Please reserve a name that is free for the main file and both sidecar targets, and make the three-file preserve step rollback every earlier move if a later one fails (with rollback failures surfaced, not discarded). A deterministic test with an injected clock/rename operation—or a pre-created conflicting sidecar target—should prove both no clobber and restoration of `dbPath` after a sidecar failure. The suspect moved to a smaller room, but it still has the database keys.

Current-head `git diff --check` is clean, and Daemon CI is green. The blocker above is independent of the test result.

figgyster (dismissed):
The sidecar-target collision and rollback work from the previous review is now fixed cleanly: all three destination names are checked, forward moves are tracked, rollback runs in reverse order, and the injected clock/rename tests exercise the failure rather than hoping the filesystem feels dramatic. That part of the case is closed.

There is still one restore-safety blocker in the wider path. `isDaemonRunningAt` acquires the daemon’s `attn.pid` flock only as a point-in-time probe, explicitly unlocks it at line 391, and returns. `restoreDatabase` then stages the backup and renames the live DB/sidecars with no lock held. The daemon’s `acquirePIDLock` uses that same flock, so the app can restart a daemon immediately after the probe and have it open/use `attn.db` while restore is moving the files. Two concurrent `attn db restore` processes can likewise both pass the probe, choose the same preservation target, and race through the swap.

Please acquire-or-refuse the pid-file flock and hold its file descriptor for the entire restore critical section, including when the pid file did not previously exist; release it only after the final swap or rollback completes. Be careful not to unlink the lock path on release in a way that creates a second-inode flock race. A focused concurrency test should pause restore after lock acquisition, prove a daemon-style second flock cannot be acquired until restore exits, and prove a second restore is excluded too. Recovery code needs custody of the room, not a photograph showing it was empty once.

Current-head `git diff --check` is clean, and Daemon plus both release-preflight checks are green. Bench A/B was still running at my last fetch; this blocker is independent of its result.

figgyster (dismissed):
You closed the point-in-time probe gap, and the restore now holds its flock across staging, preservation, and swap. That is the right shape. But the lock-file lifecycle still leaves the same safety contract break through the daemon's other hand.

`acquireDaemonLock` locks the inode currently named `attn.pid`, while `Daemon.releasePIDLock` unlocks and closes that inode and then removes the pathname. Restore can acquire the old inode in the interval after the daemon unlocks but before `os.Remove`; the daemon then unlinks it, and a new daemon startup creates and locks a different `attn.pid` inode while restore continues under the orphaned lock. The comment here promises that callers always contend on the same inode, but that is not true while the daemon release path still deletes the file.

Please make the lock-file lifecycle stable on both sides—normally by leaving the pid file in place when the daemon releases it—and add a regression that exercises unlock → restore-acquire → daemon cleanup/startup ordering. The invariant has to be one persistent lock inode, or the flock is just a badge pinned to yesterday's coat.

CI is green, including Bench A/B, and `git diff --check` is clean. The focused local Go test could not start because the review sandbox has Go 1.24 while this branch requires 1.25.3.

figgyster (dismissed):
You fixed the local half of the inode-lifecycle race cleanly: daemon shutdown leaves `attn.pid` in place, local ensure no longer sweeps it away, and the new contention test proves a restore holder excludes the next daemon. That is exactly the durability shape this #578 follow-up needs. But attn has another hand on the same lock file, and it is still carrying a knife.

The remote-daemon lifecycle scripts still unlink `attn.pid`: stale cleanup in `internal/hub/bootstrap.go:569`, `stopRemoteDaemonScript` at line 654, and restart cleanup at line 672 all run `rm -f "$pid_path"`. If `attn db restore` is holding that remote profile's flock, any of those paths can orphan its inode and then start a daemon that creates and locks a new inode at the same pathname. That reopens the exact restore-vs-daemon database race this push closes locally.

Please make the stable-lock-file invariant global: remote lifecycle cleanup may remove the socket, but it must leave the PID path in place, and the generated-script tests should pin that behavior. One lock pathname has to mean one persistent inode everywhere a daemon can be started; otherwise recovery safety depends on which doorway the daemon used.

`git diff --check` is clean, and Daemon plus both Linux preflight checks are green. The focused local Go run is still unavailable because the review sandbox has Go 1.24 while this branch requires 1.25.3; Bench A/B was still running at the last fetch.

figgyster (approved):
You chased this recovery lock all the way through the building instead of fixing only the door I happened to point at. Local shutdown/ensure and remote stale/stop/restart now preserve one persistent `attn.pid` inode, and the generated-script tests pin the remote half without being fooled by ordinary PID reads. That closes the restore-vs-daemon race cleanly.

This is the right continuation of #578's durability arc: backups are only real when the recovery path survives the ugly timing, not just the sunny-day copy. `git diff --check`, Daemon, and both Linux preflights are green; Bench A/B is still running but remains trend-only. Safe to merge. Case closed.

PR #583, opened 2026-07-18
feat(app): notebook tile root plumbing — tileParams {root,path}, root-bound fs, per-root change signals

victorarias:
## Summary
PR4 of the editor-arbitrary-roots plan (`docs/plans/2026-07-18-editor-arbitrary-roots.md`).

Frontend-only, no protocol change:
- `tileParams` upgraded to `{root, path}` JSON with legacy bare-path back-compat
- `NotebookSurfaceDaemon` becomes a root-binding factory (notebook-storage commands deliberately stay notebook-rooted)
- Per-root `fs_changed` signals
- `fs_watch`/`fs_unwatch` lifecycle wired to root-bound tiles

## Test plan
- [x] `pnpm exec tsc --noEmit` — clean
- [x] `pnpm test -- --run` — 1806 tests passing
- [x] Live-smoked on a throwaway profile: created a file externally and confirmed it appeared in the open notebook's tree

### DIFF
--- a/app/src/App.fsChangeSignals.test.ts
+++ b/app/src/App.fsChangeSignals.test.ts
@@ -1,0 +1,64 @@
+import { describe, expect, it } from 'vitest';
+import { bumpFsChangeSignal, fsChangeSignalKey } from './App';
+
+// Pure-logic coverage for the per-root fs_changed routing that feeds every
+// notebook surface's changeSignal (see makeNotebookSurfaceDaemon in App.tsx).
+// App.tsx has no dedicated render-level test suite for this wiring (see
+// App.presentationNotices.test.ts for the same pattern applied to another
+// piece of App.tsx state), so this exercises the extracted key/reducer
+// functions directly rather than standing up a full App render + mocked
+// WebSocket.
+
+const NOTEBOOK_ROOT = '/Users/victor/attn-notebook';
+const ROOT_A = '/Users/victor/code/repo-a';
+const ROOT_B = '/Users/victor/code/repo-b';
+
+describe('fsChangeSignalKey', () => {
+  it('uses the event root verbatim when present', () => {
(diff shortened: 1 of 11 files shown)


figgyster (changes requested):
The Notebook spent the last polish arc learning to be a good room; this starts turning it into a whole building. The `{root, path}` compatibility shape and resolved-root signal routing are clean work. But one watcher still gets lost whenever the socket goes dark.

Required before merge:

- In `NotebookTile.tsx`, the `fs_watch` effect runs only when `root` or `effectiveNotebookRoot` changes. The daemon deliberately drops every explicit watcher ref when that WebSocket client disconnects, while the normal frontend reconnect leaves mounted tiles and these callback identities intact. After a reconnect, an arbitrary-root tile therefore stays open but receives no more external `fs_changed` events until it is remounted. Please thread a reconnect/handshake generation into this lifecycle, re-establish the balanced subscription after each fresh connection, and add a regression proving disconnect → reconnect sends a second `fs_watch` without leaking the old ref.

The focused 35 tests pass, `git diff --check` is clean, and CI is green. Those tests cover mount, unmount, late watch resolution, and root changes; they do not cover the connection lifecycle that destroys the server-side subscription. That stale editor crosses the product-correctness line, so this one stays red for now.

figgyster (approved):
The reconnect finally has papers. The new connection generation reaches the tile lifecycle, the old subscription is cleaned up, and the regression proves the exact `watch → reconnect → unwatch/watch` sequence that was missing. That closes my blocker without disturbing the root-normalization and legacy-params work already in place.

The focused component/socket suites pass (94 tests), typecheck is clean, and CI plus all frontend E2E shards are green. This arbitrary-root editor arc now has the connection footing the next switcher/finder slices will need. Approved. Case closed.

PR #584, opened 2026-07-18
feat(config): ATTN_DATA_DIR test scoping — tests can never touch the real data dir

victorarias:
## Summary
Implements §2 of docs/plans/2026-07-18-db-loss-mitigation.md (included in this PR).

- `ATTN_DATA_DIR` is now the highest-precedence data-dir source at the `attnDir()` chokepoint.
- Added a `testing.Testing()` presence-check backstop that panics when tests resolve the data dir unscoped — this is the failure class behind the 2026-07-18 incident where a daemon test overwrote the real `~/.attn` database.
- `TestMain` scoping added in `internal/daemon`, `internal/daemonctl`, `internal/client`, `internal/config`.
- `config.json` loading made lazy so the backstop doesn't fire at package init.
- The three PR#577 HOME-redirect tests migrated to `ATTN_DATA_DIR`.
- A subprocess regression test proves the panic actually fires when a test resolves the data dir unscoped.

Note: `DataDirForProfile`/`SocketPathForProfile`/`StatePath` deliberately bypass the chokepoint (documented in code and `docs/profiles.md`).

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)
- [x] `go test ./...` (all packages pass, including internal/daemon, internal/daemonctl, internal/client, internal/config)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -108,9 +108,17 @@ Tests must never resolve `config.DataDir()` (or anything derived from it —
 2026-07-18 daemon test did exactly that and destroyed the production
 database. Scoping goes through `ATTN_DATA_DIR`, never by redirecting `HOME`.
 Every package whose tests reach `config.DataDir()` sets a per-run temp dir as
-the default via `TestMain` (`os.Setenv("ATTN_DATA_DIR", <temp dir>)` before
-`m.Run()`); individual tests layer `t.Setenv("ATTN_DATA_DIR", t.TempDir())`
-on top for extra isolation. `config.DataDir()` panics under `go test` if
+the default via `TestMain`, calling `config.ScopeTestEnvironment(<temp dir>)`
+before `m.Run()` — not a raw `os.Setenv("ATTN_DATA_DIR", ...)`, because
+`DBPath`, `SocketPath`, `PluginDir`, and the config-file path each check
+their own env-var override (`ATTN_DB_PATH`, `ATTN_SOCKET_PATH`,
+`ATTN_CONFIG_PATH`, `ATTN_PLUGIN_DIR`) before ever reaching the
+`ATTN_DATA_DIR`-scoped chokepoint, so a stray inherited value from the
+calling shell (e.g. a leftover `ATTN_DB_PATH` pointed at the real
+`~/.attn/attn.db`) can bypass scoping entirely unless it's cleared.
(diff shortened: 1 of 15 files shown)


figgyster (dismissed):
This is the right response to the incident: one attn-owned scope knob plus a fail-loud backstop is much stronger than asking every future test to remember where the bodies are buried. But the backstop still has a side door, and on a data-loss patch that crosses the line.

Required before merge:

- `internal/config/config.go`: `requireExplicitDataDirUnderTest()` is only reached through `attnDir()`, while `DBPath()`, `SocketPath()`, and `PluginDir()` return their explicit env/config overrides first. The new package `TestMain`s set `ATTN_DATA_DIR`, but they do not sanitize inherited `ATTN_DB_PATH`, `ATTN_SOCKET_PATH`, `ATTN_CONFIG_PATH`, or `ATTN_PLUGIN_DIR`. So a developer running tests from a scoped shell can still make daemon tests open the real database (for example, inherited `ATTN_DB_PATH=~/.attn/attn.db`) even though `ATTN_DATA_DIR` points at a temp directory; an inherited config file can do the same. That defeats the advertised “tests can never touch the real data dir” safety contract in exactly the process that opens and mutates SQLite.

Please make test-mode resolution impossible to escape from the explicit `ATTN_DATA_DIR` scope—whether by giving that scope true precedence, rejecting out-of-scope path overrides under `go test`, or equivalently sanitizing the inherited override surface with a central guard—and add a subprocess regression that seeds hostile `ATTN_DB_PATH` / `ATTN_CONFIG_PATH` values and proves resolution either stays inside the temp scope or fails before opening them.

CI is green and `git diff --check` is clean. I could not rerun the focused packages locally because the isolated runner currently has Go 1.24.13 while this branch requires Go 1.25.3; that runner mismatch does not change the static escape path above.

figgyster (approved):
The backstop locks the side door now. `ScopeTestEnvironment` gives the temp data root one owner, strips the inherited DB/socket/config/plugin overrides before tests run, and the subprocess regression brings hostile paths to the scene on purpose. Daemon CI is green, and the plugin helper subprocesses keep their trusted parent-scoped wiring without reopening production state.

This pairs cleanly with #578: backups make the database loss recoverable; this makes a repeat from the test suite much harder to cause. No blockers on `fde49d82`. Case closed.

PR #585, opened 2026-07-18
feat(app): editor tile roots — workspace-dir default, root switcher, finder over fs_index

victorarias:
## Summary

PR5 of the editor-over-arbitrary-roots epic (`docs/plans/2026-07-18-editor-arbitrary-roots.md`; follows merged PR1 #577, PR2 #579, PR3 #580, PR4 #583).

- ⌘⌥N (`notebook.openTile`) in a workspace whose directory differs from the effective notebook root now docks the notebook tile root-bound to the workspace directory at dock time (`tile_params {"root":...}`) — no wrong-root flash.
- New protocol 170: `workspace_layout_dock_tile` gains optional `tile_params`; the dock handler prefers it when non-empty and preserves existing params on a paramless re-dock (move). This was needed because the daemon previously dropped the frontend's `tile_params` silently.
- Dock-tile header gains a root switcher (Notebook / Workspace / current-other-root / Browse…) for notebook tiles; switching rewrites the tile's params envelope (path deliberately dropped so the tile returns to the tree view).
- ⌘P finder and the tile file listing now go through `fs_index` scoped to the tile's root; fullscreen notebook browser keeps the rootless notebook listing.

## Testing

- Unit: `resolveEditorTileRoot`, `fsIndexToNotebookEntries`, 7 root-switcher component tests.
- Go: protocol parse round-trip + daemon dock-handler params tests.
- Local: `go build`/`vet`/`test` on protocol+daemon, full frontend vitest (172 files / 1821 tests), `tsc --noEmit`.
- Verified live in the packaged app on a throwaway profile: dock-time params persisted, finder listed the workspace root's files (including nested), file open persisted `{"root","path"}`, and a full app relaunch restored the root-bound tile with the file loaded.

## Note for reviewer

Changelog entry is deliberately deferred to PR6, which finishes the user-facing Editor rename and off-root gating.

### DIFF
--- a/app/src/App.fsIndexToNotebookEntries.test.ts
+++ b/app/src/App.fsIndexToNotebookEntries.test.ts
@@ -1,0 +1,53 @@
+import { describe, expect, it, vi } from 'vitest';
+import { fsIndexToNotebookEntries } from './App';
+import type { FsIndexResult } from './hooks/useDaemonSocket';
+
+// Pure-logic coverage for the fs_index -> NotebookEntry adapter that backs the
+// ⌘P finder (both notebook tiles and the fullscreen NotebookBrowser) since PR5
+// switched the finder off notebook_list onto the root-scoped fs_index command.
+// See makeNotebookSurfaceDaemon / notebookBrowserListFiles in App.tsx.
+
+describe('fsIndexToNotebookEntries', () => {
+  it('maps each fs_index path to a NotebookEntry with size 0', () => {
+    const result: FsIndexResult = {
+      root: '/Users/victor/code/attn',
+      files: ['README.md', 'docs/plan.md'],
+      truncated: false,
+    };
+
+    expect(fsIndexToNotebookEntries(result)).toEqual([
+      { path: 'README.md', size: 0 },
+      { path: 'docs/plan.md', size: 0 },
+    ]);
+  });
+
+  it('returns an empty list for an empty index', () => {
+    const result: FsIndexResult = { root: '/Users/victor/code/attn', files: [], truncated: false };
+    expect(fsIndexToNotebookEntries(result)).toEqual([]);
+  });
+
+  it('still returns entries for a truncated result (partial beats none) but warns', () => {
+    const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    const result: FsIndexResult = {
+      root: '/Users/victor/code/attn',
+      files: ['README.md'],
+      truncated: true,
+    };
+
+    expect(fsIndexToNotebookEntries(result)).toEqual([{ path: 'README.md', size: 0 }]);
+    expect(warnSpy).toHaveBeenCalledTimes(1);
+    expect(warnSpy.mock.calls[0][0]).toContain('[App]');
+
+    warnSpy.mockRestore();
+  });
+
+  it('does not warn when the result is not truncated', () => {
+    const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    const result: FsIndexResult = { root: '/Users/victor/code/attn', files: ['a.md'], truncated: false };
+
+    fsIndexToNotebookEntries(result);
+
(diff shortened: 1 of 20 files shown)


figgyster (changes requested):
You’re carrying the arbitrary-root stack from #577/#579/#580/#583 into the editor users will actually touch. The local path is tidy, the protocol-170 persistence fix is the right seam, and the focused frontend suite passed here (77 tests). One endpoint boundary is still standing in the doorway.

**Blocker — keep remote workspace paths out of the local filesystem client** (`app/src/App.tsx:1987`)

`handleOpenNotebookTile` finds the active workspace by bare `workspace.id` and serializes its `directory` as the tile root. The root switcher likewise receives `workspace.directory` for every workspace. But endpoint-aware workspace views can represent remote workspaces, while `sendFsIndex` and the other `sendFs*` calls carry no `endpoint_id`; they always execute against the connected local daemon. In a remote workspace, ⌘⌥N therefore persists a remote path as a local root. If that path is absent, the new default is broken; if the same absolute path exists locally, the editor can silently show or modify the wrong machine’s files. A duplicate workspace id across endpoints can also make the bare `.find()` select the wrong directory before the tile is even docked.

Please either gate the workspace-directory default and Workspace switcher option to local workspaces, or make the filesystem/root pipeline endpoint-aware end to end. Add coverage for a remote workspace (and, since endpoint-disambiguated ids are already a project invariant, duplicate ids across local/remote endpoints) so this path cannot cross machines by accident.

CI is green. I could not rerun the Go packages locally because the review sandbox has Go 1.24 while this checkout requires 1.25.3; the GitHub daemon job covers that side.

figgyster (changes requested):
The active-remote-session path is fixed on `5244e2e9`; both call sites now suppress that session’s remote directory, and the focused frontend set passed here (84 tests). The endpoint leaves one forwarding address behind, though.

**Blocker — a sessionless remote workspace still passes its remote directory to local fs** (`app/src/types/workspace.ts:512`)

`localWorkspaceDirectory` proves remoteness only by finding a currently-live session. attn deliberately retains pinned and tile-only workspaces after their last session leaves, and `listWorkspaces` builds the app snapshot by appending `hubManager.RemoteWorkspaces()` to the local workspace list independently of sessions (`internal/daemon/workspace.go:707-714`). Once a retained remote workspace is sessionless, this helper sees no matching session and returns its remote `directory`; both ⌘⌥N and the root switcher then hand that path back to the local-only `fs_*` pipeline, recreating the wrong-machine risk from the original review.

The duplicate-id explanation in the comment also does not match the current model: remote workspaces live in per-endpoint hub runtime maps and are appended to local workspaces, and the existing frontend regression at `app/src/utils/workspaceViewModels.test.ts:141-158` explicitly supports two endpoint-less workspace snapshots with the same id across distinct endpoints. The bare-id lookup can therefore still select the wrong snapshot while sessions exist.

Please derive locality and selection from the endpoint-aware workspace identity (or conservatively omit the Workspace root whenever locality cannot be proven), and cover a retained sessionless remote workspace plus the existing duplicate-id shape. The first fix got the common case straight; these are the two alleys where it still loses the machine.

CI for the new head is still running; the focused endpoint/workspace/root-switcher tests are green locally.

figgyster (approved):
The endpoint finally gave up the alias. `944ec985` closes both blockers: locality is now locked until a live session positively proves local ownership, a remote twin keeps the gate closed for duplicate raw workspace ids, and sessionless workspaces no longer get mistaken for local just because the last witness left.

That conservative boundary is the right shape for PR5. The acknowledged cost — sessionless local pinned/tile-only workspaces temporarily falling back to Notebook until `protocol.Workspace` carries endpoint identity — is a safe, visible limitation, and folding the endpoint-aware fix into PR6 keeps this slice honest without letting remote paths wander into local `fs_*` calls.

Focused validation passed here: 90 tests across the root gate, endpoint-disambiguated workspace models, dock-tile switcher, and finder adapter. Approved.

PR #586, opened 2026-07-18
test(daemon): harden pid-lock orphan regression; classify restore lock errors

victorarias:
## Summary
Two review follow-ups from #582 (backup/restore pid-lock work, merged as ba613ae6):

- `internal/daemon/pidlock_test.go`: resequence `TestDaemon_ReleasePIDLock_DoesNotOrphanAConcurrentHolder` so the restore stand-in opens its fd (no `O_CREATE`) *while the daemon still holds the lock*, before `releasePIDLock` runs. The prior sequencing opened that fd after release, so the test would have passed even under the old unlink-on-release bug — it never actually held the daemon's original inode across the release. Verified by deliberately reintroducing `os.Remove` into `releasePIDLock`: the resequenced test fails at the exclusion assertion as expected; the old sequencing did not catch it.
- `cmd/attn/db.go`: `acquireDaemonLock` now classifies flock errors — `EWOULDBLOCK` keeps the existing "daemon is running" message (a live daemon holds the lock), any other error becomes "cannot determine daemon state: %w" instead of misattributing it. Both paths still fail closed and refuse the restore.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/daemon/... -run TestDaemon_ReleasePIDLock -v` — both pass
- [x] `go test ./cmd/attn/... -run TestAcquireDaemonLock -v` — all three pass
- [x] `go test ./...` — full suite green
- [x] `gofmt -l .` clean
- No live-app verification: this is a test-sequencing fix plus error-classification logic fully covered by unit tests, with no daemon-process, protocol, PTY, or UI surface touched.

### DIFF
--- a/cmd/attn/db.go
+++ b/cmd/attn/db.go
@@ -388,7 +388,14 @@ func acquireDaemonLock(pidPath string) (release func(), err error) {
 	}
 	if flockErr := syscall.Flock(int(lockFile.Fd()), syscall.LOCK_EX|syscall.LOCK_NB); flockErr != nil {
 		lockFile.Close()
-		return nil, fmt.Errorf("the attn daemon is running; stop it first (quit the app, or `attn daemon stop`) before restoring the database")
+		if errors.Is(flockErr, syscall.EWOULDBLOCK) {
+			return nil, fmt.Errorf("the attn daemon is running; stop it first (quit the app, or `attn daemon stop`) before restoring the database")
+		}
+		// Any other flock failure means we cannot determine whether a
+		// daemon holds the lock. Fail closed rather than risk racing a
+		// live daemon: never proceed with the restore on an inconclusive
+		// lock result.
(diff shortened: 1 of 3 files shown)


figgyster (approved):
You came back to #582 and made the regression test prove the inode race instead of merely looking stern at it. Good catch: opening the restore stand-in before daemon release, without `O_CREATE`, is the detail that turns this from test theater into evidence.

The `flock` error split is right too: contention gets the useful operator message; an inconclusive lock failure stays fail-closed without falsely accusing a live daemon. The daemon CI job is green, and `git diff --check` is clean. My local sandbox could not rerun the Go suite because its image is still on Go 1.24 while this module requires 1.25.3; that is runner weather, not a defect in this patch.

Safe to merge. This is the kind of follow-up durability work I like: small, specific, and aimed directly at the assumption that failed last time. Case closed.

PR #587, opened 2026-07-18
test(daemon): harden pid-lock and restore-lock coverage (PR #582/#586 follow-up)

victorarias:
## Summary

Purely additive hardening on top of #586's flock-error classification and resequenced orphan-holder test — no reverts of #586's structure or wording, just the untested branches/assertions layered on:

1. **`flockFn` injectable seam** in `cmd/attn/db.go` over `syscall.Flock`, so a non-EWOULDBLOCK flock failure (e.g. `ENOLCK`) can be exercised deterministically. #586 added the EWOULDBLOCK/other-error classification but left the "other error" branch untested.
2. **`TestAcquireDaemonLock_NonContentionFlockErrorIsNotMisreportedAsRunning`** (new) — covers that previously-untested branch via `flockFn`.
3. **Negative assertion** in `TestAcquireDaemonLock_HeldByAnotherProcess` — EWOULDBLOCK contention must not also match the indeterminate-state message.
4. **`os.SameFile` inode assertion** in `TestDaemon_ReleasePIDLock_DoesNotOrphanAConcurrentHolder` — on top of #586's existing flock-success check, explicitly proves the pathname still resolves to the daemon's original inode after release.

## Test plan

- [x] `gofmt -l` clean
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -count=1 ./internal/daemon ./cmd/...` — all green

### DIFF
--- a/cmd/attn/db.go
+++ b/cmd/attn/db.go
@@ -381,12 +381,19 @@ func stageBackupCopy(src, dstPath string) (stagedPath string, err error) {
 // The lock file is never unlinked here — only unlocked and closed — so a
 // second acquireDaemonLock always contends on the same inode rather than
 // racing a delete-then-recreate against a concurrent acquirer.
+//
+// flockFn is indirected to syscall.Flock so tests can inject a non-EWOULDBLOCK
+// failure (e.g. standing in for ENOLCK) without needing real OS-level
+// conditions to trigger it — EWOULDBLOCK is the only signal that means a live
+// daemon actually holds the lock.
+var flockFn = syscall.Flock
+
 func acquireDaemonLock(pidPath string) (release func(), err error) {
 	lockFile, err := os.OpenFile(pidPath, os.O_RDWR|os.O_CREATE, 0644)
 	if err != nil {
 		return nil, fmt.Errorf("open pid file %s: %w", pidPath, err)
 	}
-	if flockErr := syscall.Flock(int(lockFile.Fd()), syscall.LOCK_EX|syscall.LOCK_NB); flockErr != nil {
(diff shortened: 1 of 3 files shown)


figgyster (dismissed):
You went back to a merged durability fix and made the regression test capable of convicting the old bug. That is unglamorous work, which is usually where the bodies are buried.

The corrected ordering now pins the restore stand-in to the daemon's original inode before release, and `os.SameFile` makes an unlink/recreate regression fail for the right reason. The `flock` error split is equally sound: real contention gets the operator action, indeterminate kernel failures stay fail-closed without inventing a live daemon.

Daemon CI, release preflight, Changes, Frontend E2E, and the trend-only benchmark are green. My isolated local runner could not execute because its Go 1.24 image is behind this repository's Go 1.25.3 requirement; I inspected the exact lock paths and tests instead. Safe to merge. Case closed.

figgyster (approved):
You restacked this the clean way: #586 now owns the behavior, and this PR owns the proof it was missing. That makes the durability trail easier to follow instead of keeping two versions of the same crime scene in one diff.

The remaining patch is narrow and useful. The injected `flock` seam exercises the fail-closed non-contention branch, the negative assertion keeps contention classification honest, and `os.SameFile` proves the release path preserves inode identity instead of merely observing a successful flock on whatever file the pathname happens to name.

Daemon CI, release preflight, Changes, and Frontend E2E are green; the benchmark is still running but remains trend-only. My local sandbox still cannot run this Go 1.25.3 repository with its Go 1.24 image. Safe to merge. Case closed—again, after the suspect changed jackets.

PR #588, opened 2026-07-18
feat(app): editor tile — off-root gating, Editor labels, changelog, workspace-root scenario

victorarias:
## Summary

PR6 of the editor-arbitrary-roots epic (`docs/plans/2026-07-18-editor-arbitrary-roots.md`). Frontend only.

- **Off-root affordance gating**: `NotebookSurface`'s `backlinksNotebook`/`sendToChief` props become optional. `NotebookTile` computes off-root the same way its `fs_watch` effect already does (`root !== effectiveNotebookRoot`) and omits both for off-root tiles, hiding the backlinks/outline rail and the "Send to chief" button off-root. Root-scoped `listFiles` (`fs_index`) is unaffected.
- **Editor labels on the tile surface only**: fallback tile title (`Notebook` → `Editor`), the `notebook.openTile` shortcut label, the App.tsx action-menu/palette entry, and the notebook-tile-finder scenario assertion. Fullscreen labels, storage/settings copy, Sidebar, and NotebookBrowser are unchanged.
- **CHANGELOG.md**: dated entry summarizing the epic's net user-visible outcome so far.
- **New packaged-app scenario**: `scenario-editor-workspace-root.mjs` (registered in `package.json` + `scenarioCatalog.mjs`), docking an off-root editor tile against a session whose cwd is a temp directory, asserting root-scoped `fs_read` works and the backlinks rail never renders, plus a positive-control tile pinned to the Notebook root where the rail does render.
- Plan-doc updates (PR1-5 marked done, PR7 fast-follow noted; unrelated db-loss-mitigation follow-ups marked done per coordinator note).

## Verification

- Live packaged-app verification on a throwaway profile: off-root rail withheld, Notebook-root positive control rendered the rail, root-scoped `fs_read` served real file content.
- `vitest`: 173 files, 1830 passed.
- `tsc`: clean.

This diff was already reviewed line-by-line and approved by the coordinating session prior to opening this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,13 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   database file rather than deleting it. Settings now also surface the
   timestamp of the most recent successful backup.
 
+### Changed
+- **The Notebook tile is now an Editor that can open any folder.** ⌘⌥N opens
+  it on the active workspace's directory, the tile header shows the open root
+  with a switcher (workspace / Notebook / browse), and ⌘P fuzzy-finds across
+  the whole open root. Notebook-only affordances — backlinks and send-to-chief
+  — still appear only when you're browsing your actual Notebook.
+
 ## [2026-07-16]
 
 ### Added
--- a/app/package.json
+++ b/app/package.json
@@ -53,6 +53,7 @@
     "real-app:scenario-notebook-tile-close": "node scripts/real-app-harness/scenario-notebook-tile-close.mjs",
     "real-app:scenario-notebook-editor-undo": "node scripts/real-app-harness/scenario-notebook-editor-undo.mjs",
     "real-app:scenario-notebook-link-nav": "node scripts/real-app-harness/scenario-notebook-link-nav.mjs",
+    "real-app:scenario-editor-workspace-root": "node scripts/real-app-harness/scenario-editor-workspace-root.mjs",
     "real-app:scenario-autoclose-on-exit": "node scripts/real-app-harness/scenario-autoclose-on-exit.mjs",
     "real-app:scenario-present-flow": "node scripts/real-app-harness/scenario-present-flow.mjs",
     "real-app:scenario-ticket-lifecycle": "node scripts/real-app-harness/scenario-ticket-delegation-lifecycle.mjs",
--- a/app/scripts/real-app-harness/scenario-editor-workspace-root.mjs
+++ b/app/scripts/real-app-harness/scenario-editor-workspace-root.mjs
@@ -1,0 +1,344 @@
+#!/usr/bin/env node
+
+// Editor tile over an arbitrary workspace root, in the packaged app
+// (editor-arbitrary-roots epic, PR6). Docks a fresh editor tile with the REAL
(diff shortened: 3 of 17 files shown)


figgyster (dismissed):
You’re closing the arbitrary-roots editor arc with the right authorization boundary: notebook-only powers stay home, while ordinary filesystem editing travels. That split is clean. But the root switcher still leaves one loaded gun on the desk.

Required before merge:

- `NotebookSurface` initializes a tile only when `active` changes (`app/src/components/NotebookSurface.tsx`, initialization effect around line 563, dependency list around line 654). Changing the header root only updates the existing tile’s params; the React key remains the tile id, so the surface stays mounted. `NotebookTile` immediately swaps to a new root-bound daemon (`app/src/components/notebook/NotebookTile.tsx`, lines 31–41), but the old `selectedPath`, `note`, and draft remain visible. Worse, `writeBuffer` adopts the new `writeFile` callback, so the next edit can write the old document’s buffer to the same relative path under the newly selected root. The new affordance gating updates correctly, but the document itself is still from the previous root. Please make a root change clear/reinitialize the surface (or remount it by root) and add a rerender/root-switch regression proving the old buffer cannot survive into the new root.

The focused frontend slice is otherwise clean: `git diff --check` passed, the three touched component suites passed (57 tests), and live CI is green. This one is a wrong-root write risk, so I’m keeping the door shut until the editor changes jurisdiction without carrying the old case file along.

figgyster (changes requested):
You fixed the wrong-root write exactly where the state boundary lives, and the remount tests prove React really tears the surface down. Credit where it’s due. But I sent you toward “remount it by root” too casually in my first review; I missed the dirty-buffer half of the editor contract, and the bill arrived immediately.

Required before merge:

- `key={root ?? ''}` unmounts `NotebookSurface` as soon as the root param changes (`app/src/components/notebook/NotebookTile.tsx`, around line 108). A dirty buffer is only flushed by `loadFile`/`requestClose`; the 700 ms autosave effect in `NotebookSurface.tsx` cancels its timer during unmount and has no unmount flush. Edit a note and choose another root before that debounce fires, and this fix drops the unsaved edit instead of writing it to the wrong root. Please make root switching an awaited transition: persist against the old root first, abort the switch on conflict/error, then update params/remount. Add a regression that edits, switches roots inside the debounce window, and proves the old-root write completes (and a conflict prevents the root change).

The follow-up itself is mechanically clean: `git diff --check` passed and the focused `NotebookTile`/`NotebookSurface` suites pass (17 tests). The new mount-count tests cover remount identity, but mocks cannot see the dirty document they are destroying. I called the original corruption risk correctly and the remount shape incompletely; this red stays until both sides of the file-ownership handoff are safe.

figgyster (approved):
You made the root switch earn its new jurisdiction this time: the outgoing surface saves through the same hash-CAS path used by navigation and close, and only then may the tile replace it. Conflicts and write errors keep the old root mounted with its banner visible. That closes both halves of the case—the wrong-root write and the disappearing draft—without adding a rival persistence path.

The focused surface/tile/switcher suites pass here (66 tests), the full GitHub matrix is green, and the existing packaged-app proof still covers the arbitrary-root boundary this PR is closing. Safe to merge. The editor-roots arc can leave the station without somebody's unsaved note under the wheels. Approved.

PR #589, opened 2026-07-18
feat(session): answer questions from prior sessions

victorarias:
## Intent

Let users ask a free-form, evidence-bounded question about a previous native Codex conversation, without creating an authorization ledger or workflow-specific policy model.

## Summary

- Adds attn session instructions <session-id> --question "..." with concise human and JSON output.
- Finds only the exact native session transcript, bounds and fingerprints it, and validates every quoted excerpt against the source. Assistant text may add context but cannot independently establish authorization.
- Runs Luna with no shell, unified execution, web, MCP, or workspace access; preserves the default CODEX_HOME for packaged-daemon Codex authentication.
- Includes malformed, missing, oversized, fabricated/ambiguous citation, assistant-only, model-failure, CLI, and transcript-discovery coverage.

## Test plan

- [x] go test ./...
- [x] git diff --check
- [x] Installed isolated non-production attn app smoke with a synthetic native transcript: returned Yes with validated excerpts, transcript path/fingerprint, and gpt-5.6-luna low effort.

## Out of scope

This reports what the recorded conversation supports. It does not establish external human identity, create an authorization ledger, or interpret PR/workflow-specific policy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ### Added
 - **Black or stuck UI failures now leave useful evidence on disk.** Agent switches capture delayed app-shell health snapshots alongside the existing terminal renderer diagnostics, including frontend crashes, event-loop stalls, and native browser visibility. A React render failure now shows a persistent error screen instead of leaving an unexplained black window.
+- **Ask a bounded question about another Codex session.** `attn session instructions <session-id> --question "..."` reads that session's conversation and returns a concise Luna answer with exact supporting excerpts, transcript identity, and a fingerprint. It fails closed when the transcript, model response, or evidence cannot be verified.
 
 ## [2026-07-15]
 
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -170,7 +170,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '170';
+export const PROTOCOL_VERSION = '171';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 interface PRActionResult {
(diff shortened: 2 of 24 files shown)


figgyster (changes requested):
You're turning old Codex conversations into a queryable evidence surface. That's a useful continuation of attn's multi-harness arc, and the exact-transcript/fingerprint/tool-free shape is the right bet. But the evidence boundary is the product here, and two holes sit directly on it.

Required before merge:

- `internal/sessioninstructions/service.go:290` waives user-authored evidence whenever the question merely contains `agent` or `assistant`. That means an assistant-only transcript saying “I was authorized to merge” can answer **“Was the agent authorized to merge?”** with `Yes.` and pass validation—the subject word disables the very guard this PR promises. Please distinguish explicit speech-reporting questions (for example, “Did the agent say…?”) from claims about user authorization, and add the adversarial regression.
- This adds `session_instructions` plus new result types to the shared TypeSpec protocol, but `internal/protocol/constants.go:13` and `app/src/hooks/useDaemonSocket.ts:173` remain at v170, and `app/src/types/generated.ts` was not regenerated. A newly rebuilt CLI can therefore meet a still-running v170 daemon that does not know the command, while the app sees no version mismatch to restart/refuse it. Bump the daemon/frontend protocol version together and regenerate every committed client artifact from the schema.

CI is green and `git diff --check` is clean. My local focused test run couldn't start because Figgy's sandbox image still has Go 1.24 while this repo now requires 1.25.3; that tooling fog doesn't change either code-path finding. Fix these two contracts and the rest of the shape looks sound. The court reporter can stay. It just can't accept the witness's own forged warrant.

figgyster (approved):
Both suspects went straight.

- The evidence guard now exempts only explicit questions about what the agent said, while “Was the agent authorized?” still requires user-authored evidence. The adversarial regression covers the exact hole.
- Protocol v171 is synchronized across the daemon, frontend socket, and regenerated TypeScript artifacts, so mixed old/new installs fail visibly instead of wandering into an unknown command.

This is a useful continuation of the OpenCode/pi multi-harness arc: old native Codex conversations become a bounded evidence surface without quietly turning transcript prose into an authorization ledger. The exact-transcript, fingerprinted, tool-free boundary is honest now.

All visible CI is green and the follow-up diff passes `git diff --check`. Approved. The rain can bother somebody else.

PR #590, opened 2026-07-18
feat(protocol): endpoint_id on Workspace — hub stamps remote workspaces, frontend gate reads the record (protocol 171)

victorarias:
## Summary

Final PR of the editor-arbitrary-roots epic (`docs/plans/2026-07-18-editor-arbitrary-roots.md`), the fast-follow committed in #585's review thread.

- Adds `endpoint_id?: string` to `protocol.Workspace` (protocol 170→171, bumped in both lockstep spots: `internal/protocol/constants.go` and `app/src/hooks/useDaemonSocket.ts`).
- Hub daemons stamp `EndpointID` on every workspace mirrored from a remote endpoint at ingest (`replaceRemoteWorkspaces`/`upsertRemoteWorkspace` in `internal/hub/manager.go`), **unconditionally overwriting** any remote-supplied value so a buggy or compromised remote can't spoof locality.
- `localWorkspaceDirectory` now reads the workspace record directly instead of cross-referencing sessions, restoring the workspace-directory ⌘⌥N default for sessionless local pinned workspaces.
- The workspace lookup in `handleOpenNotebookTile` is now endpoint-aware, to correctly disambiguate duplicate-id remote twins.
- `daemonSessionsRef` deleted (no longer needed once workspace lookups don't cross-reference sessions).

## Test plan

- `go build ./...`, `go vet ./...` — clean
- `go test ./internal/hub/... ./internal/daemon/... ./internal/protocol/...` — pass, including hub-stamping coverage (spoof-overwrite on both `replaceRemoteWorkspaces`/`upsertRemoteWorkspace`, and change-detection short-circuit preserved), and confirming the local (non-hub) daemon never stamps `EndpointID`.
- `pnpm test` — 173 files / 1843 tests pass, including frontend gate-semantics coverage for the new field.
- `pnpm exec tsc --noEmit` — clean.
- Live packaged-app verification on a throwaway profile: ⌘⌥N workspace-dir default → root-scoped editor tile + protocol-171 handshake (`scenario-editor-workspace-root`) — passed.
- **Not live-verified**: the hub-stamping half was not exercised against a real remote endpoint (no cheap rig available) — it is covered by unit tests only.

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -1973,12 +1973,6 @@ sendFetchPRDetails,
   // default actually needs.
   const daemonWorkspacesRef = useRef<DaemonWorkspace[]>([]);
 
-  // Same reasoning as daemonWorkspacesRef: lets handleOpenNotebookTile read the
-  // current session list (to gate a workspace's directory to local-only, see
-  // localWorkspaceDirectory) without rebinding the callback on every session
-  // change.
-  const daemonSessionsRef = useRef<DaemonSession[]>([]);
-
   // Dock a fresh Notebook tile into the active workspace. A unique tile id every
   // time: the daemon treats a duplicate id as a move, so a shared id would just
   // relocate the first tile instead of opening a second. Defaults the tile's root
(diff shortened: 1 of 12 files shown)


figgyster (dismissed):
You’re closing the seven-PR arbitrary-editor-roots arc by making locality a workspace property. That’s the right architectural move, and the hub-side overwrite of remote-supplied `endpoint_id` is exactly the sort of distrust this boundary deserves. But the final frontend lookup takes the local twin’s papers without first proving that twin is the one the user selected.

Required before merge:

- `handleOpenNotebookTile` still tracks the active workspace as a bare id in `activeWorkspaceIdRef`. With a remote workspace active and a local workspace sharing that id, `find((w) => w.id === workspaceId && !w.endpoint_id)` selects the local twin anyway, so its local directory is trusted and the bare-id dock command can mutate/open the wrong workspace. Carry the active endpoint identity through this lookup (or refuse a local root when the active record is remote), and add a regression that starts from an active remote twin plus a same-id local twin. The new unit test only searches a list for the local record, so it proves the filter can find local—not that it matches the active workspace.

The protocol/stamping half holds, and CI is green. This locality hole is the last suspect in the room; close it and the epic can leave with a clean coat.

figgyster (approved):
You took the smaller fix, and it’s the right one. `soleWorkspaceForId` refuses to invent endpoint identity from a bare id, so duplicate twins lose the workspace-directory default while the sessionless local-pinned case still gets it. The exact regression now passes locally — 33/33 focused workspace tests — and Daemon plus Frontend CI are green.

That closes the locality hole without turning this capstone into a selection-model rewrite. The arbitrary-editor-roots arc can finally put down the shovel. Approved.

PR #591, opened 2026-07-18
feat(tickets): establish canonical plan artifact lifecycle

victorarias:
## Intent / Why

Delegated plans can currently remain editable in both a repository worktree and the attn Notebook after attachment. That leaves follow-on agents without an obvious source of truth and allows the copies to drift silently.

This change makes plan authority contextual and explicit: repository conventions win when the affected component keeps plans in Git; otherwise the Notebook becomes canonical after promotion.

## Summary

- Add `attn ticket attach-plan` with monorepo-aware convention detection and explicit `repository` or `notebook` overrides.
- For repository-owned plans, require a committed, clean source and attach a Notebook reference containing the repository path, branch, and introducing commit.
- For Notebook-owned plans, verify the attached copy byte-for-byte before retiring an untracked staging file; never delete a tracked source implicitly.
- Preserve divergent legacy copies for explicit reconciliation while retiring byte-identical duplicates safely.
- Update delegation prompts and bundled agent guidance so follow-on work receives the canonical location and ticket completion follows strong terminal evidence.
- Make `profile-env` clear inherited socket and resource overrides before selecting a profile, which keeps managed sessions from silently talking to the wrong daemon.

## Out of scope

Generic ticket attachments remain Notebook snapshots. The authority lifecycle in this PR applies specifically to durable Markdown plans and design artifacts.

## Test plan

- [x] `go test ./cmd/attn -count=1`
- [x] `go test ./internal/agent ./internal/hooks -count=1`
- [x] `go test ./internal/daemon -run TestChiefOfStaffDelegateBindsTicketAndPrompt -count=1`
- [x] `go vet ./cmd/attn ./internal/agent ./internal/hooks ./internal/daemon`
- [x] Exercise repository-reference and Notebook-promotion outcomes against a non-production attn profile.
- [x] Build the CLI and verify `profile-env dev` clears an inherited production socket before resolving the dev socket.


figgyster (dismissed):
There’s one deletion-safety blocker here.

`inspectPlanAttachment` checks that a Notebook-owned source is untracked, but `retireNotebookPlanSource` later deletes it after only comparing bytes and inode identity. Those steps straddle the daemon attachment request, so another agent/process can `git add` or commit the source in between. If the bytes still match the copied artifact, line 412 removes a now-tracked file despite the command’s explicit “never delete a tracked source” contract.

Please re-check Git ownership immediately before `os.Remove` (and fail closed when Git status cannot be established), with a regression test that makes the source tracked between inspection and retirement. The copy can remain canonical on failure; the original must stay put. Multi-agent tools need their deletion promises enforced at the blade, not only at the front desk.

I also traced the repository-reference/legacy-copy path and the profile-routing cleanup; those look coherent. `git diff --check` is clean. I could not run the Go suite locally because the isolated runner currently provides Go 1.24 while this head requires Go 1.25.3.

figgyster (dismissed):
The deletion promise is enforced at the blade now. `retireNotebookPlanSource` rechecks Git ownership immediately before removal, treats only `ls-files` exit 1 as safely untracked, and fails closed on every ambiguous Git error. The new regression stages the plan after inspection and proves the source survives, while the broken-repository case covers the fail-closed branch.

That closes the blocker. This is a useful continuation of #571’s ticket-artifact work: tickets are graduating from a copy bin into an authority-aware handoff surface, and this follow-up keeps that new lifecycle from eating a file during concurrent agent work.

`git diff --check` is clean. I still could not execute the Go suite locally because the isolated runner provides Go 1.24 while this repository requires Go 1.25.3; the focused changes and tests are coherent on inspection. No blockers. Case closed.

figgyster (approved):
The force-update is a clean rebase onto current `main`. The PR remains the same 17-file ticket lifecycle patch; the reviewed implementation and regression tests are byte-identical to the approved fix, with only upstream changelog entries joining the branch.

The deletion-edge ownership check still closes the blocker, `git diff --check` is clean, and the Daemon plus both Linux release-preflight jobs pass on this head. The trend-only benchmark is still running, but the principal checks are green. Re-approved. GitHub made the case change coats at the door, but it is the same suspect.

PR #593, opened 2026-07-18
fix: implement `attn daemon stop`

victorarias:
## Summary

- `attn daemon stop` was unimplemented: it silently fell through to `runDaemon()`, which either accidentally started a foreground daemon (if none was running) or errored "daemon already running" (if one was). `attn db restore`'s own error text tells users to run `attn daemon stop`, so the missing subcommand was actively misleading.
- Extracts the flock-gated stop logic out of `cmd/attn/profile.go`'s `stopProfileDaemon` into a new, reusable `internal/daemonctl.Stop`, used by both the new `attn daemon stop` subcommand and the existing profile-clean path. `stopProfileDaemon` becomes a thin adapter that preserves its existing note strings.
- `attn daemon <subcommand>` now fails loudly on an unrecognized subcommand instead of silently starting a daemon.
- Preserves the safety property from `stopProfileDaemon` exactly: the pid file's exclusive `flock` is the liveness+ownership gate. If the lock can be acquired, the on-disk pid is stale and is never signaled — this prevents accidentally killing an unrelated process that happens to reuse a stale pid.

## Test plan

- [x] `go build ./...`, `go vet ./cmd/attn/... ./internal/daemonctl/...`, `gofmt -l` clean
- [x] `go test ./internal/daemonctl/... ./cmd/attn/...` — includes new real-process tests in `internal/daemonctl/stop_test.go` covering not-running, stale-pid, SIGTERM, and SIGKILL-escalation paths
- [x] Live smoke on a throwaway profile: started a daemon, ran `attn daemon stop`, confirmed clean shutdown and correct pid-file cleanup

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-18]
 
 ### Fixed
+- **`attn daemon stop` now actually stops the daemon.** Previously the
+  subcommand didn't exist: running it would try to *start* a daemon instead —
+  silently launching one in the foreground if none was running, or erroring
+  "daemon already running" if one was. It now stops the current profile's
+  daemon, matching the guidance `attn db restore` gives. Unknown `attn daemon`
+  subcommands now fail with an error instead of starting a daemon.
 - **Orphaned session workers now clean themselves up.** A per-session PTY worker
   whose daemon is gone for good (e.g. a torn-down test profile, or a daemon that
   was killed and never restarted) previously kept running forever, accumulating
--- a/cmd/attn/db.go
+++ b/cmd/attn/db.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/victorarias/attn/internal/config"
+	"github.com/victorarias/attn/internal/daemonctl"
 	"github.com/victorarias/attn/internal/store"
 )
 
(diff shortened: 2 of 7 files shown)


figgyster (changes requested):
There’s one safety blocker in the new stop path.

`Stop` treats every failed nonblocking `flock` as proof that a live daemon owns the PID file, then trusts and signals the PID stored there. That ownership inference is not true in this repo: `attn db restore` deliberately holds the same flock for the whole restore, while the file can still contain the stale PID left by the previous daemon. If that PID has been recycled, running `attn daemon stop` during a restore can send SIGTERM and then SIGKILL to an unrelated process—the exact failure this patch is trying to prevent.

The error handling has the same root problem: any `flock` failure is treated as contention. `acquireDaemonLock` already documents and tests the safer contract: only `EWOULDBLOCK` means another holder exists; errors such as `ENOLCK` are indeterminate and must fail closed.

Please make the lock-holder identity unambiguous before signaling (for example, have non-daemon holders replace the stale PID with a sentinel while they own the lock, or add another daemon-ownership proof), distinguish contention from other flock errors, and add regressions for (1) a restore-style holder with a stale live PID and (2) a non-contention flock failure. The function said the lock had one owner. The repository knew better.

Validation: `git diff --check` is clean and the visible Daemon/Changes/preflight CI is green. I could not rerun the focused Go suite locally because Figgy’s isolated test image is still Go 1.24 while this checkout requires Go 1.25.3.

figgyster (changes requested):
Thanks — the `ENOLCK`/non-contention branch now fails closed, and the sentinel protects the steady state once it has been written. One blocker remains in the transition between those states.

`acquireDaemonLock` acquires the flock at `cmd/attn/db.go:397`, but it does not truncate/write the sentinel until lines 415–425. During that window, a concurrent `Stop` gets `EWOULDBLOCK`, concludes that some holder owns the lock, closes its probe fd, and reads the old numeric PID that is still on disk. It can therefore signal the same recycled stale PID before the restore process reaches the sentinel write. The comment says the current holder is known to have written the content, but the ordering does not establish that fact.

The new regression misses this window because `TestStop_NonDaemonHolderSentinel` waits for the sentinel at line 302 before calling `Stop` at line 305. Please make daemon ownership/signalability provable without a lock-acquired-before-marker TOCTOU gap, and add a deterministic regression that pauses a non-daemon holder after flock acquisition but before sentinel publication while `Stop` inspects the stale live PID. The sentinel arrived at the scene. It just arrived one instruction too late.

Validation: the incremental diff and full PR diff are whitespace-clean. CI was still running at the final pre-review fetch; Figgy’s isolated Go image remains 1.24 while this checkout requires Go 1.25.3, so I could not rerun the focused suite locally.

figgyster (approved):
This closes both blocker waves cleanly. The stop path no longer infers ownership from flock contention or marker-write ordering: it fails closed on indeterminate flock errors, requires the numeric PID itself to have the PID file open, and keeps the sentinel for the precise non-daemon-holder case. The new `lock-pause` regression finally exercises the exact acquire-before-publication window instead of waiting past it.

The incremental diff is whitespace-clean, and Changes, Daemon, Frontend E2E, plus both Linux preflight builds are green on `8ddc2a51`. I couldn’t rerun the focused Go suite in Figgy’s isolated image because it still carries Go 1.24 while this checkout requires 1.25.3; the principal CI lane supplies that evidence.

One non-blocking portability note: `daemon stop` now depends on `lsof`; missing/inconclusive `lsof` fails closed rather than risking a signal, which is the right safety posture. If standalone Linux hosts are expected to be dependency-free, documenting or replacing that dependency can be a follow-up.

Approved. The lock finally has an alibi that survives cross-examination.

PR #594, opened 2026-07-18
feat(tickets): buffer observer notifications

victorarias:
## Intent / Why

Ticket bursts currently interrupt chiefs and subscribers repeatedly, while ticket mutations can overwrite intervening activity. This keeps updates immediate for the assignee doing the work, but gives every other observer one coherent interruption window and makes writers catch up before changing a stale ticket.

Related: attn ticket `chief-buffer`.

## Summary

- buffer non-assignee ticket delivery behind one durable 30-minute observer-wide attention window while preserving immediate assignee delivery and explicit inbox reads
- serialize nudge and watch consumption with polling-aware leases, durable restart recovery, provenance-preserving event bundles, and chief-role continuity
- make CLI status, comment, and attach mutations atomically consume target-ticket unread activity and reject the mutation until retry; report unread history on take and subscribe
- add app optimistic concurrency using `latest_event_seq` and `expected_event_seq`, including safe description-draft conflict behavior
- bump the protocol to 172, regenerate Go and TypeScript types, and document the contract

## Test plan

- [x] `go test ./...`
- [x] `pnpm --dir app test` (1,770 tests)
- [x] structured autoreview clean
- [x] signed isolated `attn-dev.app` build and install
- [x] packaged nudge-trigger scenario: paused unread indicator, manual delivery, inbox consumption, and busy-agent queueing
- [x] isolated-dev live CLI proof: an unread update rejected the first mutation without changing state; retry applied it
- [x] post-rebase focused Go and ticket-panel regression tests

No production app was installed or restarted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,8 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   ritual.** Agents may complete work that Victor accepted, whose requested PR
   merged, or that has an equivalent objective terminal signal; finished work still
   awaiting review remains in review.
+- **Ticket notifications now protect attention without delaying the agent doing the work.** The current assignee keeps immediate delivery, while chiefs, subscribers, and other participants coalesce unread activity into one 30-minute observer-wide window. Explicit inbox reads remain immediate, and inbox watches share delivery with nudges without duplicate wake-ups.
+- **Ticket updates now make agents read intervening activity before writing.** Status, comment, and attachment commands show unread updates and require a deliberate retry; app edits reject stale ticket details and refresh them before retry. Taking or subscribing to a ticket reports unread history without consuming it.
 
 ### Fixed
 - **Switching profiles from inside attn no longer keeps talking to the inherited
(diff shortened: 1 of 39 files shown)


figgyster (changes requested):
You built the right shape here: the event log stays the batch, cursors stay acknowledgement, and the new transaction seam closes the read-before-write race without inventing a second notification system. It also continues the ticket-as-durable-handoff arc from #571, but now the handoff has an interruption budget instead of ringing every bell in the building. The signed dev-app and live CLI evidence are the right witnesses.

One concurrency hole still breaks the observer-window contract:

- **`internal/daemon/ticket_notify.go` (`notifyUnreadTicketSession`) — serialize deadline reconstruction with delivery/catch-up.** This method reads unread events and `last_attention_at`, computes a deadline, then arms the earliest countdown without holding `deliveryMu`. A concurrent mutation catch-up can consume one ticket, advance `last_attention_at`, cancel the old timer, and rebuild the remaining unread work against the fresh 30-minute window; afterward, an already-running reconstruction that observed the old attention value can resume and pull that fresh timer back to its stale earlier deadline because `armNudgeCountdownAt` keeps the minimum. Explicit inbox consumption has the same stale-reconstruction race. That can interrupt a chief/subscriber before the newly established observer-wide window expires. Please put reconstruction under the same delivery serialization (using a locked helper for callers that already hold `deliveryMu`, so we don't trade the race for a deadlock) and add a deterministic interleaving regression test where the stale computation resumes after catch-up/inbox advances attention.

CI is green across daemon, frontend, E2E, and release preflight; `git diff --check` is clean. My local Go runner was too old for this repo's Go 1.25.3 requirement, so I treated CI plus your recorded `go test ./...` and signed live-app proof as the executable evidence. The lock race is the one suspect still walking.

figgyster (approved):
The stale deadline has nowhere left to hide. Reconstruction now holds the same `deliveryMu` as inbox consumption, catch-up, and fire-time delivery; callers already inside the critical section use the locked helper, so the fix closes the race without manufacturing a self-deadlock.

The deterministic interleaving test is the evidence I wanted: it pauses the old calculation before arm, proves catch-up cannot cross it, then verifies the remaining unread work is rebuilt against the fresh attention window. That turns yesterday's timing theory into a guarded contract.

Daemon, frontend, release preflight, and all E2E shards are green on this head. Approved. Tickets keep becoming the durable coordination surface, and this gives that surface some mercy for the humans and chiefs listening to it.

PR #595, opened 2026-07-18
test: seam external tool-home resolution behind ATTN_TOOL_HOME

victorarias:
## Why

HOME redirection in tests is the same footgun class as the 2026-07-18 db-loss
incident (a test overwrote the real `~/.attn` because HOME-based state
resolution wasn't seamed for tests), one step removed. This applies the same
mitigation pattern used there (`config.requireExplicitDataDirUnderTest`) to
tool-dotfile resolution: `internal/toolhome` adds an `ATTN_TOOL_HOME` env seam
plus a panic backstop that fires if `os.UserHomeDir()` is reached under
`go test` without the override set, making accidental real-HOME resolution in
tests structurally impossible for tool-dotfile paths.

## What changed

- New `internal/toolhome` package (`toolhome.Dir()` + go-test panic backstop,
  mirroring `config.requireExplicitDataDirUnderTest`)
- Converts every `os.UserHomeDir()` call in `internal/transcript` and
  `internal/agent` to `toolhome.Dir()`
- Migrates all ~19 HOME-redirecting test sites off `HOME`, including **two
  latent leaks the new backstop caught** (test sites that were silently
  reading the real HOME before this change)
- One deliberate exemption: `internal/client/client_test.go` keeps
  `t.Setenv("HOME")` because `DataDirForProfile` is HOME-based by design for
  cross-profile probing (documented inline)

No CHANGELOG entry — internal-only, no user-visible behavior change.

## Verification

- `go build ./...` and `go vet ./...` clean
- `go test -count=1 ./...` green (fresh run, no cache)
- Live daemon smoke on a throwaway profile with `ATTN_TOOL_HOME` unset,
  confirming the production fallback path never hits the backstop
- Test-only, internal seam change with no daemon-process, protocol, PTY, or
  UI surface — no additional live-app verification needed beyond the above

Known pre-existing flake unrelated to this change: `internal/daemon`
git-status scheduler race under bare `-race` (scope with `-run` if hit).

### DIFF
--- a/cmd/attn/main_test.go
+++ b/cmd/attn/main_test.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/victorarias/attn/internal/buildinfo"
 	"github.com/victorarias/attn/internal/protocol"
+	"github.com/victorarias/attn/internal/toolhome"
 	"github.com/victorarias/attn/internal/transcript"
 )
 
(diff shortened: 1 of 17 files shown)


figgyster (approved):
This closes the adjacent test-isolation hole exposed by the database-loss incident: attn-owned state and external tool homes now both fail closed under `go test`, instead of trusting a redirected `HOME` with the keys to the whole apartment.

The boundary looks right. Agent/transcript consumers use the new seam, broad packages establish package-wide throwaway defaults, fixture-specific tests narrow them further, and the remaining `HOME` override is the documented read-only `DataDirForProfile` exception. `git diff --check`, Daemon CI, and both release-preflight jobs are green. My focused local run could not start because the reviewer sandbox still carries Go 1.24.13 while this repo requires Go 1.25.3; the trend-only benchmark was still running and is not a merge gate.

Approved. This is the kind of durability work that earns its paranoia.

PR #597, opened 2026-07-18
feat(automations): deliver manual run walking skeleton

victorarias:
## Intent / Why

Automations need to create ordinary, visible attn work without bypassing the ticket and session model users already understand. This first slice establishes the durable delivery foundation with a manual CLI trigger, so later schedule and provider slices can build on a recoverable execution contract instead of adding parallel infrastructure.

## Summary

- Add `attn automation apply|list|show|run|runs` for versioned `v1alpha1` manual definitions and explicit reusable request IDs.
- Persist immutable definition revisions, occurrences, and runs before delivery, with stable ticket, session, workspace, and pane IDs for idempotent retry and startup recovery.
- Deliver a chief-owned ticket before creating its visible agent session, with fixed automatic approval, model and effort pins, canonical directory provenance, and structurally separate untrusted occurrence payloads.
- Close session-launch crash windows by persisting the complete session row before spawn, adopting surviving workers by stable ID, restoring prior session metadata after failed respawn, and retaining retryable runs when post-spawn verification is not yet available.
- Keep automatic approval and directory trust behind a daemon-internal launch seam, scrub one-shot capability variables from child environments, and preserve the launch contract on worker and embedded PTY backends.

## Test plan

- [x] `make generate-types`
- [x] `go test ./...`
- [x] `pnpm --dir app test -- --run` (173 files, 1,848 tests)
- [x] `git diff --check`
- [x] `gpt-5.6-terra` high autoreview: no actionable findings
- [x] Live named-profile happy path on `automations`: definition applied, run delivered, chief-owned ticket completed, and visible Codex session launched with pinned `gpt-5.6-sol`, low effort, and automatic review
- [x] Live idempotency: repeating the same request ID returned the same run, occurrence, ticket, and session
- [x] Live recovery: terminating the isolated daemon during delivery recovered the pending run to the same stable ticket/session/workspace/pane set after restart
- [x] Live failure validation: invalid JSON was rejected before a run was claimed

## Out of scope

Schedules, GitHub and inbox providers, recurring continuity, fresh-worktree delivery, automated review sessions, and UI management land in later slices from the accepted implementation guide.

No production app or daemon was installed or restarted.

PR #599, opened 2026-07-19
test(store): guard migration ordering

victorarias:
## Intent / Why

The Automations merge exposed that two branches had independently selected migration version 70. Add a narrow registry invariant so duplicate or out-of-order versions fail clearly in the store suite before branches merge.

Follow-up to #597.

## Summary

- validate that the real store migration registry is strictly increasing
- report the conflicting version and indexes for duplicates or descending entries
- cover the collision shape `[68, 69, 70, 70]`, the valid sequence `[68, 69, 70, 71]`, and an out-of-order sequence

## Test plan

- [x] `go test ./internal/store -run "TestMigration(VersionsStrictlyIncrease|OrderError)$" -count=1`
- [x] `go test ./internal/store -count=1`

## Out of scope

Later Automations follow-up backlog items remain separate so each change is independently reviewable.

### DIFF
--- a/internal/store/sqlite_test.go
+++ b/internal/store/sqlite_test.go
@@ -2,6 +2,7 @@ package store
 
 import (
 	"database/sql"
+	"fmt"
 	"path/filepath"
 	"reflect"
 	"testing"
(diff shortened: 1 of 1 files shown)


figgyster (approved):
You took the migration-number collision from #597 and turned it into a tripwire instead of another cleanup ritual. Good. This is the kind of tiny durability follow-up that saves a very expensive afternoon later: the real registry is checked, and failures name the conflicting versions and positions instead of leaving the next branch merge to conduct the autopsy.

The helper covers the duplicate and descending shapes without pretending migration gaps are illegal, `git diff --check` is clean, and Daemon CI plus release preflight are green. Safe to merge. Case closed.

PR #600, opened 2026-07-19
feat(cli): add live session transcripts

victorarias:
## Intent

Automations need a provider-neutral way to read session progress without scraping native Codex, Claude, or Copilot transcript formats. This adds a daemon-owned transcript API and CLI with stable events, resumable cursors, and centralized secret redaction.

## Summary

- add `attn session transcript <id>` with JSON, cursor resume, and follow modes
- normalize Codex, Claude, and Copilot transcripts into stable user, assistant, tool-call, tool-result, and error events
- redact structured secrets and common plaintext credential formats before events leave the daemon
- use per-event opaque cursors so resuming cannot skip sibling events from one native record
- add the protocol command and response, regenerate Go and TypeScript types, and bump protocol version to 175
- document the user-visible command in the changelog

## Test plan

- [x] `go test ./internal/transcript ./internal/protocol ./internal/client`
- [x] `go test ./cmd/attn`
- [x] `pnpm --dir app test`
- [x] build and install isolated `transcript-smoke` profile
- [x] verify normalized snapshot output, redaction, cursor-only incremental resume, follow delivery, and clean SIGINT exit against the named-profile daemon
- [x] remove the temporary profile after verification

## Out of scope

Transcript waiting/expect helpers and later automation backlog items remain separate reviewable changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-07-19]
 
+### Added
+- **Live agent transcripts are inspectable from the CLI.** `attn session
+  transcript <session-id>` prints timestamped conversation, tool, and error
+  events across supported agents, with `--follow` for live updates and opaque
+  cursors for safe resumption. Provider-specific records stay behind the daemon
+  boundary, and secrets are redacted before output.
+
 ### Fixed
 - **Retried delegations no longer create duplicate agents or tickets.** `attn
   delegate` now prints a durable request and operation identity before slow
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -170,7 +170,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '174';
+export const PROTOCOL_VERSION = '175';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 interface PRActionResult {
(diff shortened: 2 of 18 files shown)


figgyster (dismissed):
You’re making attn the observation layer above the agent harnesses now—not just the dispatcher. That follows the OpenCode and pi direction cleanly, and the daemon-owned redaction/cursor boundary is the right place to put it. But one Codex transcript shape has slipped through the floorboards.

Required before merge:

- `internal/transcript/events.go:277-312` does not normalize Codex `response_item` records whose payload is `{type:"message", role:"assistant", content:[...]}`. The existing transcript parser already treats that as a supported Codex assistant-message shape (`ExtractAssistantContent`, with `TestExtractLastAssistantMessage_CodexResponseItemOutputText`), but this new parser only defines tool fields in the response-item payload and falls through for `message`. A response-item-only transcript therefore emits no assistant progress at all, despite the provider-neutral contract advertised here. Please normalize that message shape and add coverage for both a response-item-only transcript and the common paired `event_msg` + `response_item` case so the fix does not double-emit the same turn.

The cursor/redaction structure and the live smoke evidence otherwise look solid, and CI is green. Fix this one missing provider path and the case can close.

figgyster (dismissed):
That closes it. You added the missing Codex `response_item` assistant path and, more importantly, made the dedupe survive the follow-poll boundary—the place a simpler adjacent-line fix would have quietly failed.

The response-item-only and paired-record regressions cover the exact contract I blocked on. `git diff --check` is clean; Daemon and Frontend CI are green, with the E2E shards and trend benchmark still running at last look. Safe to merge from my side. The observation layer keeps its badge. Case closed.

figgyster (approved):
The latest cleanup keeps the fix intact. You pulled the assistant-echo predicate into a named helper and removed a redundant cursor-state assignment; the response-item-only path and cross-poll dedupe behavior are unchanged.

`git diff --check` is clean, and Daemon, Frontend, and both release-preflight lanes are green. The lone E2E failure is the existing split/terminal paint race (`OLD line 0` missing while lines 1–49 are present), unrelated to this Go-only refactor—the unreliable witness from the black-screen alley again.

Re-approved on the current head. The observation layer can keep its badge.

PR #601, opened 2026-07-19
feat(automations): materialize exact PR worktrees

victorarias:
## Intent / Why

Slice 1 proved that a manual automation can durably create an ordinary chief-owned ticket and a steerable agent session. PR review needs the same delivery contract with a stronger location boundary: every accepted occurrence must run from the exact snapshotted pull-request revision in its own recoverable worktree.

## Summary

- adds strict typed GitHub PR occurrences and `attn automation run <id> --pr-url <url>`, resolved through one read-only PR GET
- adds profile-managed repository caches plus canonical, explicitly configured local-clone overrides
- materializes one detached exact-SHA worktree per stable session and records repository, source, worktree, provider ref, and revision provenance
- keeps provider payloads in a separate occurrence artifact instead of interpolating them into configured instructions
- preserves idempotency across request replay, moved PR refs, partial launch crashes, agent-modified worktrees, daemon restarts, and temporarily unavailable GitHub credentials
- passes Git credentials only through child-process environment, scopes HTTPS headers to the exact host, rejects authenticated plaintext HTTP, and never puts credentials in URLs, arguments, or logs

## Test plan

- [x] `make generate-types`
- [x] `go test ./...`
- [x] `pnpm --dir app test -- --run` (173 files, 1,848 tests)
- [x] `git diff --check`
- [x] final `gpt-5.6-terra` high autoreview: no actionable findings
- [x] packaged `automations` profile: managed-cache PR #404 run delivered ticket `auto-c9b2fa2f1e064e5c` and session `ddac6e6d-dac6-414a-89bb-63f02107ccdb` at exact SHA `67fc97c9974239492c9bfb6e7db4f47f0677e4e7`
- [x] same request ID replayed to the identical run, ticket, session, and worktree
- [x] packaged profile also exercised local-clone materialization and rejected a mismatched explicit override without falling back or persisting the definition
- [x] local-only large Bazel repository measurement: ~2.49 GiB exact worktree sharing a ~15.28 GiB Git object store, with a distinct Bazel output base

## Out of scope

Automatic review-request observation and continuity across later PR cycles remain in Slices 3 and 4. The private large-repository gate was measured locally because the environment prohibited sending private source to the external review model; both live agent materialization modes were verified against a public repository. No GitHub review, comment, approval, or other mutation is performed by an automated review session.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,13 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   producing output defers it.
 
 ### Added
+- **Manual PR-review automations now work from an exact isolated checkout.** Run
+  a repository-worktree definition with `attn automation run <id> --pr-url
+  <url>` to resolve the current PR head through a read-only GitHub request and
+  launch the local reviewer in a detached per-session worktree. Definitions can
+  use a profile-managed repository cache or explicit validated local-clone
+  overrides; the run records the source, main repository, worktree, and pinned
+  revision without posting or changing anything on GitHub.
 - **CLI-driven automations can prepare visible work before it needs attention.**
   Apply a durable manual definition with `attn automation apply`, run it on
   demand, and inspect its immutable run history. Each run creates an ordinary
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -170,7 +170,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '174';
+export const PROTOCOL_VERSION = '175';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 interface PRActionResult {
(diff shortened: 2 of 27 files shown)


figgyster (dismissed):
You’re making the checkout itself part of the durable automation contract now. That’s the right second slice after durable ticket/session delivery, and the exact-SHA pinning, provider-payload separation, recovery path, and credential transport are a strong piece of work. The packaged-profile proof is the sort of witness I trust. One path bug is still standing in the doorway.

Required before merge:

- `internal/git/materialize.go:95-102`: `ValidateLocalClone(staging, ...)` canonicalizes symlinks, so when the profile/socket root traverses a symlink, `mainRepo` is the real path while `staging` and `target` retain the symlinked spelling. After `os.Rename`, `strings.Replace(mainRepo, staging, target, 1)` does not match and returns the now-nonexistent staging path. The clone has been published successfully, but `PrepareLocation` immediately continues with a dead repository path and the run fails. `ATTN_DATA_DIR` and configured socket paths are cleaned, not symlink-resolved, so this is a real supported configuration shape. Revalidate/canonicalize the published `target` after the rename and return that path; please pin it with a managed-cache regression whose parent is reached through a symlink.

CI is green, but it had no witness for this filesystem shape. Fix that one and the rest of the case looks clean.

figgyster (approved):
You fixed the exact seam I blocked: the publish path now revalidates the target after the rename, and the symlink-parent regression proves the returned repository path is the live canonical one. That’s the right repair — evidence at the boundary, no special-case string surgery.

This repository-materialization slice now has the filesystem proof it was missing. Safe to merge. Case closed.

PR #602, opened 2026-07-19
feat(cli): add environment preflight

victorarias:
## Intent

Environment failures during Automations work were initially indistinguishable from product failures. This adds one read-only preflight command that fails early with the actual missing tool, unwritable path, routing conflict, or protocol mismatch before a build or live run begins.

Related: #600

## Summary

- add `attn preflight [--json]` with stable named checks and actionable `pass`, `warn`, or `fail` results
- report required source-build tools, writable checkout/profile/cache paths, and the active profile's resolved app, socket, and port
- compare the current CLI and installed app protocol with the live daemon, including the daemon's actual profile routing
- report the selected agent, model, and effort from explicit flags, active environment values, or agent-native defaults
- keep every check diagnostic: the command does not install tools, create directories, restart daemons, or change settings

## Test plan

- [x] `go test ./...`
- [x] `go vet ./internal/preflight ./cmd/attn`
- [x] `git diff --check`
- [x] Install the isolated `preflightqa` profile and verify all 16 live checks pass against its packaged app and daemon at protocol 175
- [x] Inject the production socket into `preflightqa` and verify JSON output reports a routing mismatch and exits 1
- [x] Remove the temporary `preflightqa` profile after verification

## Out of scope

Autoreview-specific model, reasoning-level, and bundle output remains the next independent backlog item.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -78,6 +78,15 @@ A large automated suite — even a race-clean concurrency suite — is not by it
 
 Use a **non-prod** profile for this (the `dev` sibling or a throwaway `ATTN_PROFILE`) — that is pre-authorized; never smoke on prod. See [Iterating On Attn Itself](#iterating-on-attn-itself-attn-on-attn-testing) and [docs/profiles.md](docs/profiles.md).
 
+After installing and selecting the non-production profile, run that profile's newly bundled CLI preflight before starting live verification:
+
+```bash
+profile_app="$(./attn profile resolve --field appPath)"
+"$profile_app/Contents/MacOS/attn" preflight
+```
+
+When the scenario launches an agent with pinned settings, pass the same values to preflight, for example `--agent codex --model <model> --effort high`. Use `--json` when a script or agent needs a stable result. A failing preflight is an environment/setup failure: fix its reported tool, writable-path, routing, daemon, or protocol cause before treating later behavior as product evidence. Preflight is diagnostic only and never repairs or restarts the environment. Use the installed profile's bundled CLI as shown above, not an unrelated `attn` earlier on `PATH`.
+
 If a PR needs live-app verification and you **cannot** run it (no ability to build/install/run a non-prod profile in this environment), **stop and ask the user for guidance** — do not merge on automated tests alone. State plainly in the PR and to the user what was and was not verified live.
 
 ## Packaged-App Harness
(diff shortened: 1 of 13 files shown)


figgyster (changes requested):
You’re putting a proper diagnostic front door in front of the Automations work, and I like the direction. Two contract gaps are still standing in it with suspicious alibis:

1. **Blocker — routing can pass while the CLI and daemon resolve different paths.** `resolveRouting` carries `config.DataDir()` and `config.SocketPath()` through as raw cleaned strings, and the daemon comparison later checks those strings for equality. Both config overrides may be relative. If the CLI and installed daemon start from different working directories with the same `ATTN_DATA_DIR=.attn-qa` (or relative socket override), both health reports spell the path identically while the processes touch different directories/sockets. Preflight then certifies the split-brain environment it was built to catch. Please have each side expose/compare canonical absolute paths (resolving symlinks where the path exists), and add a regression proving that the same relative override under different process working directories fails routing. This is the same canonical-path isolation rake we’ve met before; it has fresh paint, but it is still a rake.

2. **The stable named-check set changes when the app protocol cannot be read.** When `appProtocol` returns an error but daemon health succeeds, the code emits `protocol.app` and then skips `protocol.app_daemon` entirely. That makes the JSON schema-by-convention depend on the failure mode and contradicts the PR’s stable named checks / 16-check live contract. Please emit a failing `protocol.app_daemon` check in that branch as well and cover it in the table tests.

GitHub’s daemon and release-preflight checks are green, and `git diff --check` is clean. I couldn’t rerun the Go suite locally because the isolated review image has Go 1.24 while this module now requires Go 1.25.3; that is tooling weather, not a patch failure.

figgyster (approved):
Both blockers are fixed on `cf74437e`.

The routing comparison now uses one shared canonical-path contract on both sides of `/health`, resolves relative paths and symlinked existing ancestors, fails closed when canonicalization itself fails, and includes the exact different-CWD regression that was missing. The protocol report also keeps the same `protocol.cli_app` / `protocol.app_daemon` names when the installed app is unavailable, with coverage for that failure branch.

That closes the split-brain hole cleanly—and makes the new “run preflight before live verification” guidance trustworthy instead of ceremonial. Daemon CI and both release-preflight jobs are green; the Bench A/B job is still running. `git diff --check` is clean. Approved.

PR #604, opened 2026-07-19
fix(agents): instruct agents to call bare attn, not $ATTN_WRAPPER_PATH

victorarias:
## Why

PR #572 made every attn-launched process put the active attn binary's directory first on `PATH`, so bare `attn` always resolves to the correct app/profile binary. But agents kept calling `"$ATTN_WRAPPER_PATH"` anyway — because every instruction surface they read still showed it as the canonical form: ~40 command examples across the embedded skill references, plus the delegated-ticket prompt composed in `delegate.go`.

## What

- All command examples in `internal/agent/attn_skill/references/*.md` now use bare `attn`.
- The ticket-reporting prompt in `internal/daemon/delegate.go` now says `attn ticket status ...` / `attn ticket attach-plan`.
- Updated the one test pinning the old form.

**Deliberately kept:** the wrapper env plumbing (`launchenv`, `pty/manager`, `ptybackend`, `codex.go`) — it powers the PATH prioritization — and the two stale-binary *recovery* mentions (SKILL.md bootstrap, `versioncheck.go`), which are its only intended agent-facing uses.

## Verification

- `go build ./...` clean; `internal/agent`, `internal/daemon`, `internal/hooks` tests pass.
- Fresh-built binary embeds zero wrapper ticket examples and the new bare-`attn` text (skill ships via `//go:embed`).
- Live check in an attn-managed shell: first `PATH` entry is the app bundle dir, ahead of `~/.local/bin`/homebrew, and `which attn` resolves to the active app binary.

Note: existing delegated sessions keep the old materialized skill; new delegations pick this up once a rebuilt daemon is installed.

### DIFF
--- a/internal/agent/attn_skill/references/browser.md
+++ b/internal/agent/attn_skill/references/browser.md
@@ -13,14 +13,14 @@ separate browser surface for an explicit attn-browser request.
 5. Collect the cheapest fresh evidence that proves the result.
 
 ```sh
-"$ATTN_WRAPPER_PATH" browser open http://localhost:3000
-"$ATTN_WRAPPER_PATH" browser snapshot
-"$ATTN_WRAPPER_PATH" browser find --using role --value textbox --name Search
-"$ATTN_WRAPPER_PATH" browser type --element attn-element-1 --text "query"
-"$ATTN_WRAPPER_PATH" browser click --element attn-element-2
-"$ATTN_WRAPPER_PATH" browser wait --using text --value Results --state visible
-"$ATTN_WRAPPER_PATH" browser reload
-"$ATTN_WRAPPER_PATH" browser screenshot ./attn-browser.png
+attn browser open http://localhost:3000
+attn browser snapshot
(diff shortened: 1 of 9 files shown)


figgyster (approved):
You finished the other half of #572 instead of leaving the abstraction with two names, which is exactly the kind of deletion that keeps paying rent. Bare `attn` is now the normal agent contract; `ATTN_WRAPPER_PATH` goes back to being the recovery rope it was meant to be, not a ritual incantation copied through forty examples.

I traced the remaining wrapper references: the embedded skill keeps the deliberate stale-binary recovery note, while the runtime plumbing and hook tests still exercise the implementation detail. The changed prompt/reference surfaces are clean, the Daemon CI witness passed, and `git diff --check` is quiet. Figgy's local Go sandbox is still on 1.24 while this repo requires 1.25.3, so I couldn't independently rerun the focused packages here; that's tooling weather, not a defect in this patch.

Safe to merge. Case closed.

PR #605, opened 2026-07-19
feat(cli): add exact-head PR readiness waiter

victorarias:
## Intent

Agents currently poll CI and reviews separately, which can combine green checks or approval from an obsolete head. This adds one read-only gate that completes only when the same current commit is both green and approved by the required reviewer.

## Summary

- add `attn pr wait-ready <number-or-url> --reviewer <login>`
- read head SHA, check rollup, and review commit IDs in one GitHub snapshot per poll
- report meaningful transitions and reset readiness when the head changes
- fail closed for missing checks, unknown check states, and GitHub CLI collection truncation
- document when agents should use the gate and add a user-facing changelog entry

## Test plan

- [x] `go test ./cmd/attn`
- [x] `go vet ./cmd/attn`
- [x] `go test ./...`
- [x] Live read-only success against PR #404 at exact head `67fc97c9974239492c9bfb6e7db4f47f0677e4e7`
- [x] Live wrong-reviewer gate exits `124` on timeout instead of accepting another review

## Out of scope

- autoreview configuration output, which lives outside this repository
- unattended launch-contract and automation lifecycle changes from later backlog items

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -73,6 +73,9 @@ Use the bundled CLI, not an unrelated `attn` on `PATH`. Preflight is diagnostic;
 fix reported tool/path/routing/daemon/protocol failures before treating scenario
 output as product evidence.
 
+For a GitHub merge gate, run `attn pr wait-ready <pr> --repo <owner/repo>
+--reviewer <login>` once; do not poll checks and reviews separately.
+
 ### Packaged-app harness
 
 - Single-tenant: never run packaged-app scenarios in parallel.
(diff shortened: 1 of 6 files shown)


figgyster (approved):
You’re giving the automation arc a referee instead of another polling loop, and that’s the right kind of boring. After #601 made the checkout SHA part of the durable contract, binding checks and reviewer approval to that same head closes the stale-signal gap cleanly; the fail-closed handling at GitHub’s collection boundary is the bit that earns my trust.

Focused `cmd/attn` tests, `go vet ./cmd/attn`, `git diff --check`, and the live CI/preflight checks are clean. This is an isolated read-only CLI surface, so the unit tests plus the PR’s live exact-head/wrong-reviewer exercises satisfy the repo’s live-verification exemption. Safe to merge. Case closed.

PR #608, opened 2026-07-19
feat(automations): prepare requested GitHub reviews

victorarias:
## Intent / Why

Requested GitHub reviews should become prepared local work without requiring a second polling loop or creating duplicate reviewers. This slice connects the existing PR refresh to the durable automation engine so a request produces one exact-SHA, chief-owned ticket and steerable reviewer session while keeping provider data separate from configured instructions.

## Summary

- observe `github_review_requested` demand from both background and manual PR refreshes, with all/include/exclude repository filters
- persist provider cursors, review-request edge cycles, stable per-PR continuity bindings, and exact-once ticket occurrence events
- pin each accepted occurrence with a focused read-only PR snapshot and reuse the same ticket/session/workspace/pane for later same-contract cycles
- recover pending provisioning failures, fence stale/disabled observations, and fail closed for unsafe changed-head, stopped, swept-ticket, or changed-contract continuations
- expose structured occurrence paths on the shared ticket while keeping untrusted provider payload out of configured instructions
- add a local GitHub-shaped fixture and `wsctl refresh-prs` support for repeatable non-production verification

## Test plan

- [x] `go test ./...`
- [x] `pnpm --dir app test -- --run` — 1,848 tests
- [x] `gpt-5.6-terra` high branch autoreview — no actionable findings
- [x] signed `automations` profile build/install and packaged `attn preflight --agent codex --model gpt-5.6-terra --effort high --json` — 16/16 checks passed
- [x] live requested-review happy path created run `3cc3f93f-ead9-4c0d-a1a8-81c78aea7354`, chief-owned ticket `auto-3cc3f93fead94c0d`, and visible session `e7777b76-1eef-433c-8c16-01bdc78e1bec` in a clean detached worktree at `0bda5161dee8f01a1f53f4013d3c13459b1affe4`
- [x] live remove/re-request created run `0cc9bb5b-ae2a-418f-81c6-1e3588e7e0ad`, reused the exact ticket/session/workspace/pane, published its structured input path, reopened the ticket after delivery, and completed the second local review
- [x] live changed-contract cycle failed closed as run `e02adbf7-2a5a-490f-aae4-1f2286394862`, preserved the successful origin ticket as done, and added visible failure activity
- [x] fixture request log contained only `GET` on GitHub-shaped routes; the worktree remained clean

## Out of scope

Slice 4 owns changed-head resume/fallback behavior, stopped-session recovery, and the deeper dirty/missing-worktree continuity policy. This slice fails those cases closed rather than silently reviewing the wrong checkout or using stale instructions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,13 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   new comment from a person, or approval on a green exact head. Each outcome has
   its own exit code, with `--json` for the full detail. Comments from bots are
   ignored, as are comments already on the pull request when the wait begins.
+- **Requested GitHub reviews can prepare a local reviewer automatically.** An
+  enabled `github_review_requested` automation consumes attn's existing PR
+  refresh, pins the requested PR's exact head, and creates one durable
+  chief-owned ticket and steerable reviewer session. Repository filters,
+  restart-safe latest catch-up, review-request cycle deduplication, and
+  per-PR coalescing prevent polling or concurrent refreshes from launching a
+  duplicate reviewer; the automation never posts or changes anything on GitHub.
 - **Live agent transcripts are inspectable from the CLI.** `attn session
   transcript <session-id>` prints timestamped conversation, tool, and error
   events across supported agents, with `--follow` for live updates and opaque
--- a/internal/automation/automation.go
+++ b/internal/automation/automation.go
@@ -32,7 +32,13 @@ type DefinitionSpec struct {
 	Policy     PolicySpec   `yaml:"policy" json:"policy"`
 }
 type TriggerSpec struct {
-	Type string `yaml:"type" json:"type"`
+	Type         string               `yaml:"type" json:"type"`
+	Repositories RepositoryFilterSpec `yaml:"repositories,omitempty" json:"repositories,omitempty"`
+}
+type RepositoryFilterSpec struct {
+	Mode    string   `yaml:"mode,omitempty" json:"mode,omitempty"`
+	Include []string `yaml:"include,omitempty" json:"include,omitempty"`
+	Exclude []string `yaml:"exclude,omitempty" json:"exclude,omitempty"`
 }
 type LaunchSpec struct {
 	Driver     string `yaml:"driver" json:"driver"`
@@ -97,8 +103,17 @@ func ValidateDefinition(s *DefinitionSpec) error {
 	if strings.TrimSpace(s.Name) == "" {
 		return errors.New("name is required")
 	}
-	if s.Trigger.Type != "manual" {
-		return errors.New("Slice 1 supports only trigger.type manual")
+	switch s.Trigger.Type {
+	case "manual":
+		if s.Trigger.Repositories.Mode != "" || len(s.Trigger.Repositories.Include) > 0 || len(s.Trigger.Repositories.Exclude) > 0 {
+			return errors.New("manual trigger cannot configure repositories")
+		}
+	case "github_review_requested":
+		if err := validateRepositoryFilter(&s.Trigger.Repositories); err != nil {
+			return err
+		}
+	default:
+		return errors.New("trigger.type must be manual or github_review_requested")
 	}
 	if strings.TrimSpace(s.Prompt) == "" {
 		return errors.New("prompt is required")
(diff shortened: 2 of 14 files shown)


figgyster (dismissed):
You’ve turned the PR refresh into the intake desk for durable review work, and the overall shape is strong: the exact-head snapshot, provider-data boundary, continuity binding, and changed-contract/head fences all belong in the daemon. This is the automation arc growing a spine.

One blocker remains:

- `observeGitHubReviewRequests` only attempts delivery after `ClaimGitHubReviewAutomationRun` returns a candidate. Once that claim succeeds, `accepted_cycle` is advanced. If `PrepareLocation` or the unattended launch gate then returns `retryableAutomationDeliveryError`, `handleAutomationDeliveryError` deliberately leaves the run `pending`; however, the next PR refresh sees `accepted_cycle == cycle`, `ReconcileAutomationReviewRequests` returns no candidate, and this path never loads or retries that pending run. Manual execution is also rejected for provider-driven definitions, so despite the comment mentioning an “explicit retry,” the only remaining retry is `recoverAutomations()` on daemon restart. A transient clone/fetch/Codex-startup failure can therefore strand a requested review indefinitely while the daemon keeps running. Please make active observations (or a bounded retry worker) re-enter pending delivery, with a regression where the first delivery returns a retryable error and a later same-demand refresh delivers the same run exactly once.

The sandboxed Go suite and `git diff --check` pass on `2257a172`; Daemon CI and both Linux preflights are green. The witnesses showed up; this one lifecycle gap is still carrying a knife.

figgyster (dismissed):
You closed two real withdrawal races here: pending work now stops when the request disappears, and a request withdrawn before first launch can come back without inheriting an empty old binding. The fixed local-only instruction is the right trust boundary too. Good repairs. They don’t close the blocker I raised, though.

The still-active failure path is unchanged: after `ClaimGitHubReviewAutomationRun` advances `accepted_cycle`, a retryable delivery error leaves that run `pending`. A later refresh with the same still-requested PR gets no candidate from `ReconcileAutomationReviewRequests`, so `observeGitHubReviewRequests` never loads or re-delivers the pending run. The new withdrawal logic only acts when demand disappears; it cannot rescue demand that stays active after a transient clone/fetch/startup failure.

Required before merge: add a live retry path for that pending accepted run, with the regression from my prior review—first delivery returns a retryable error, a later same-demand refresh retries the same run, and it is delivered exactly once.

The sandboxed Go suite and `git diff --check` pass on `504c649a`; Daemon CI, Linux preflights, Changes, and E2E are green. The new witnesses are credible. They’re testifying about a different suspect.

figgyster (dismissed):
There it is. You brought the exact witness I asked for: the accepted cycle resurfaces only while its original run is pending, the next same-demand observation re-enters that same durable run, and the third observation proves delivery doesn’t duplicate. That closes the liveness hole without inventing another occurrence or weakening the exact-head contract.

This is the piece that makes the #597/#601 automation arc trustworthy under ordinary transient failure, not merely restart-safe. The sandboxed Go suite and `git diff --check` pass on `01b22c7e`; GitHub has not attached checks to this fresh head yet. Safe to merge once the precinct lab reports. Case closed.

figgyster (dismissed):
You found the second liveness seam hiding behind the first. Withdrawal is now a recoverable ownership transition instead of a database stamp: an undelivered launch stays automation-owned until the daemon can actually stop and forget it, failed termination keeps enough evidence to retry, reactivation finishes the old cancellation first, and delivered or continuation reviewers remain under the ordinary ticket/session lifecycle. That is the right boundary.

The new witnesses cover the ugly windows—failed kill, daemon exit between durable withdrawal and PTY teardown, re-request, delivered origin, and continuation withdrawal—and the implementation stays idempotent across them. The full sandboxed Go suite, `git diff --check`, and the new daemon/store regressions under `-race` pass on `aac6b7f2`. GitHub has not attached checks to this fresh head yet, so merge when the precinct lab reports. Approved.

figgyster (approved):
Clean restack. `range-diff` maps all twenty reviewed automation commits onto the new main; the only integration edits are the expected adaptation to #610’s unified unattended launch contract. The new tip commit merely extracts the session-existence check and tightens the ownership comment—same behavior, fewer shadows.

The full sandboxed Go suite and `git diff --check` pass on `387078e4`. Daemon CI, Changes, Frontend E2E, and both Linux preflights are green; only Bench A/B is still running. The provider-driven intake slice survived the neighborhood changing underneath it. Approved.

PR #607, opened 2026-07-19
test: synthetic PR for wait-ready live verification (do not merge)

victorarias:
Throwaway draft PR used to live-verify `attn pr wait-ready` comment detection and baseline behavior. Will be closed and the branch deleted.

PR #609, opened 2026-07-19
feat(cli): return pr wait-ready on first actionable update

victorarias:
## Intent

`attn pr wait-ready` was a binary merge gate: it returned only when the pull
request was mergeable, or when the whole timeout expired. That made a reviewer
requesting changes indistinguishable from slow CI — both surfaced as a 30-minute
`124`. The waiter should return as soon as something needs a human.

## Summary

- return on the first actionable update: failing check, changes requested, new
  human comment, or approval on a green exact head
- give each outcome its own exit code (`0` approved, `1` checks failed, `3`
  changes requested, `4` comment, `5` error, `124` timeout), print a plain-text
  result line by default, and add `--json` for detail
- collect checks, reviews, and every comment surface in one GraphQL snapshot,
  preserving #605's one-observation-per-poll invariant
- ignore bot comments using GraphQL `__typename`, and baseline comments already
  present when the wait starts so only new ones report
- add `--ignore-author` and accept `[host/]owner/repository` in `--repo`

`gh pr view --json comments` cannot support this: it strips the `[bot]` suffix
and omits the author type, so bots are indistinguishable from people. Getting
the same data over REST would cost four calls per poll; GraphQL costs one.

The token owner is deliberately **not** filtered. The operator and the agent
share one token, so filtering "self" would discard the operator's own comment —
the most actionable event there is. The baseline prevents self-waking instead.

## Test plan

- [x] `go test ./cmd/attn`, `go vet ./cmd/attn`, `go build ./...`
- [x] Live: `pending -> green` across a real CI run, duplicate polls suppressed
- [x] Live: comment posted mid-wait woke the waiter within one interval (exit 4)
- [x] Live: re-run on the same comment stayed quiet (exit 124) — baseline holds
- [x] Live: inline review-thread comment detected
- [x] Live: bot comment on #537 filtered out
- [x] Live: head change `c0bebb72 -> 39300b56` reset checks, comments unaffected
- [x] Live: `--json` keeps progress on stderr and stdout parseable

Live verification used a throwaway draft PR (#607), now closed and its branch
deleted. It caught two defects that the unit tests did not:

1. GitHub wraps a standalone inline comment in a bodyless `COMMENTED` review, so
   one remark was reported as two comments.
2. The JSON `comments` field carried the baseline on non-comment outcomes, so a
   timeout listed old comments where a consumer reads new ones.

Both now have regression tests built from the exact payloads GitHub returned.

## Not verified live

`approved` and `changes_requested` cannot be exercised on a synthetic PR:
GitHub blocks approving or requesting changes on your own pull request. They
rest on unit tests plus #605's live check against #404 at exact head. The
review-parsing logic is unchanged from that verified version, except that
`COMMENTED` reviews now divert to the comment path before the verdict branch.

## Out of scope

- resuming a wait from a prior baseline (`--since`), which needs a durable cursor
- daemon-side comment polling; this stays a standalone read-only CLI that runs
  without a daemon

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -73,8 +73,12 @@ Use the bundled CLI, not an unrelated `attn` on `PATH`. Preflight is diagnostic;
 fix reported tool/path/routing/daemon/protocol failures before treating scenario
 output as product evidence.
 
-For a GitHub merge gate, run `attn pr wait-ready <pr> --repo <owner/repo>
---reviewer <login>` once; do not poll checks and reviews separately.
+To wait on a GitHub PR, run `attn pr wait-ready <pr> --repo <owner/repo>
+--reviewer <login>` once; do not poll checks, reviews, and comments separately.
+It returns on the first actionable update and reports which one by exit code:
+`0` approved, `1` checks failed, `3` changes requested, `4` new human comment,
+`124` timeout. Bot comments are ignored; comments already present when the wait
+starts are the baseline.
 
 ### Packaged-app harness
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,12 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-19]
 
 ### Added
-- **PR readiness can be checked with one trustworthy gate.** `attn pr
-  wait-ready` reports check and review transitions, resets when the pull request
-  head changes, and succeeds only when one exact head is green and approved by
-  the required reviewer.
+- **Waiting on a pull request returns as soon as something needs you.** `attn pr
+  wait-ready` watches checks, reviews, and comments in one request and stops at
+  the first actionable update: a failing check, a review requesting changes, a
+  new comment from a person, or approval on a green exact head. Each outcome has
+  its own exit code, with `--json` for the full detail. Comments from bots are
+  ignored, as are comments already on the pull request when the wait begins.
 - **Live agent transcripts are inspectable from the CLI.** `attn session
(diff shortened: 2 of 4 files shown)


figgyster (dismissed):
One blocker before this waiter can be trusted unattended:

- `cmd/attn/pr.go:347` limits `reviewThreads` to `last:100`, then lines 481–490 deliberately discard bodyless `COMMENTED` review wrappers. GitHub's review-thread connection has cursor pagination but no update-time ordering option. On a PR with more than 100 threads, a human can reply to an older thread outside this slice; the corresponding empty wrapper is discarded, so `attn pr wait-ready` sees neither signal and can sleep until another outcome or the full timeout. That breaks the advertised “every comment surface” / first-actionable-update contract. Please paginate or fail closed on thread truncation, or keep a deduplicated wrapper fallback, and add a regression with more than 100 threads plus a fresh reply on an old one.

The rest of the shape is strong: one snapshot per poll, exact-head review state, bot typing, comment baselining, and distinct machine-readable outcomes all belong in this coordination layer. Full Go tests, `git diff --check`, and live CI are green; this pagination alley is the only body I found.

figgyster (approved):
The old-thread gap is closed. Sourcing inline comments from each freshly submitted review preserves the one-snapshot contract, makes thread age irrelevant, and fails closed if a single review's comment connection truncates. The live API probes earned their keep here, and the new dormant-thread regression pins the exact rake I blocked on.

This is the right extension of #605: `attn pr wait-ready` can now be the actionable coordination gate above the automation harnesses, not merely a green-light poller. Full Go tests, `git diff --check`, and current CI all pass. Approved.

PR #610, opened 2026-07-19
refactor(automation): unify unattended launch contract

victorarias:
## Intent

Automation launch intent currently crosses snapshots, daemon policy, PTY backends, worker processes, and child environments through several parallel fields. That makes model, effort, approval, trust, and recovery behavior vulnerable to drift—especially after daemon restart or when the worker inherits launch variables.

This change gives unattended launches one immutable internal contract that survives every handoff and remains the sole source for contract-owned policy.

## Summary

- add `UnattendedLaunchSpec` with explicit driver, model, effort, executable, approval, directory-trust, and recovery policy
- preserve the contract through automation snapshots, daemon spawning, embedded and worker backends, worker registry persistence, and reload recovery
- reject parallel generic launch fields when an unattended contract is present
- consume contract values only at the final wrapper boundary while preserving non-contract workflow guidance and chief context limits
- canonicalize configured driver names before snapshots are persisted
- add backend parity, registry round-trip, reload, environment-isolation, and validation regressions

## Test plan

- [x] `go test ./...`
- [x] `go vet ./cmd/attn ./internal/launchcontract ./internal/automation ./internal/daemon ./internal/pty ./internal/ptybackend ./internal/ptyworker`
- [x] repository pre-commit build and Go checks
- [x] structured Codex autoreview (`gpt-5.6-terra`, high) returned no actionable findings after fixes
- [x] isolated `launchcontractqa` app passed preflight at protocol 175
- [x] live automation canonicalized a spaced/case-varied Claude driver and delivered the exact model, effort, approval, trust, and recovery contract without leaking launch-only variables to the child
- [x] worker registry stored the complete contract without duplicate legacy model or effort fields
- [x] isolated profile and smoke-test data removed afterward

## Out of scope

Truthful durable automation-run lifecycle states remain the next backlog item.

### DIFF
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -429,7 +429,9 @@ func runPTYWorker() {
 	fs.StringVar(&cfg.CodexExecutable, "codex-executable", "", "codex executable override")
 	fs.StringVar(&cfg.CopilotExecutable, "copilot-executable", "", "copilot executable override")
 	var externalCommandJSON string
+	var unattendedLaunchJSON string
 	fs.StringVar(&externalCommandJSON, "external-command-json", "", "external plugin driver argv as JSON")
+	fs.StringVar(&unattendedLaunchJSON, "unattended-launch-json", "", "immutable unattended launch contract as JSON")
 	fs.StringVar(&cfg.ExternalCWD, "external-cwd", "", "external plugin driver working directory")
 	fs.StringVar(&cfg.RegistryPath, "registry-path", "", "registry path")
 	fs.StringVar(&cfg.SocketPath, "socket-path", "", "socket path")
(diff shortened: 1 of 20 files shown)


figgyster (approved):
You’re turning automations from a bundle of launch knobs into a durable execution contract. After exact-SHA checkout and exact-head readiness, this is the next layer that had to become boring before unattended runs could be trusted—and boring infrastructure is the good stuff.

I traced the contract through snapshot creation, daemon spawn policy, both PTY backends, worker registry persistence, reload recovery, and the final child-environment overlay. The parallel model/effort/executable/approval fields are rejected at the backend seam, legacy snapshots get only the missing trust/recovery defaults, and restart recovery keeps the same value instead of re-deriving policy from live settings. Focused Go regressions and `git diff --check` passed locally; the full visible CI set is green, and your isolated-profile live evidence covers the lifecycle seam the repository requires.

Safe to merge. The truthful automation-run lifecycle states you left for the next slice are still the right next problem; this PR gives them one less source of launch-policy ambiguity. Case closed.

PR #611, opened 2026-07-19
test: synthetic PR for inline sourcing (do not merge)

victorarias:
Throwaway draft PR to live-verify inline comments sourced from reviews. Will be closed.

### DIFF
--- a/docs/scratch/inline-probe.md
+++ b/docs/scratch/inline-probe.md
@@ -1,0 +1,1 @@
+Synthetic PR for inline-comment sourcing.


docs/scratch/inline-probe.md
```diff
@@ -0,0 +1 @@
+Synthetic PR for inline-comment sourcing.
```
victorarias, line 1: root inline comment

victorarias: reply to the existing thread — the case figgyster flagged

PR #612, opened 2026-07-19
fix(delegate): resolve worktree repo from workspace sessions, not the workspace record

victorarias:
## Diagnosis

The framing offered three possibilities (metadata drift / wrong resolution / correct-by-design needing a loud failure). The answer is a **combination of the first two**, and it rules the third out.

**`workspaces.directory` is a registration artifact with no invariant tying it to reality.** There is exactly one SQL statement in the codebase that writes it — the `AddWorkspace` upsert in `internal/store/workspace.go:25` — and its conflict clause overwrites `directory` unconditionally:

```sql
ON CONFLICT(id) DO UPDATE SET
    title = excluded.title,
    directory = excluded.directory
```

`rank`, `muted`, and `pinned` are INSERT-only, and `title` is explicitly guarded against re-derivation (`daemon.go:2231-2241`) precisely so a rename sticks. `directory` has no such guard. On top of that:

- it is **inherited wholesale** when a pane is dragged out into a new workspace (`workspacelayout.go:719`) — from the *source workspace*, not the moved session's own directory;
- it is **never recomputed from member sessions**. The only rollup derived from members is `Status`. Creating a worktree for a session inside an existing workspace updates only the session's directory and leaves the workspace record pointing at the old path.

So it can name a repository no member session has ever been in. That is exactly the observed state — both the "exo" and "attn" workspaces stored `victor-cloud` while their sessions lived in `~/exo` and `~/projects/victor/attn` respectively.

**The resolution then trusted that field.** `delegate.go` picked the base directory per placement:

```go
case delegationPlacementCurrent:
    directory = source.Directory     // real session dir
case delegationPlacementExisting:
    directory = workspace.Directory  // <-- the drifting field
case delegationPlacementNew:
    directory = msg.Cwd; if empty -> source.Directory
```

and `createDelegationWorktree` turned that into a repo via `GetRepoRoot` + `ResolveMainRepoPath`. Only the `--workspace` path consulted the workspace record — which is why `--repo` "fixed" it and every other placement was fine.

Telling detail: the code **already knew these can diverge**. The `--from` default at `delegate.go:328` guards on `ResolveMainRepoPath(baseRoot) == ResolveMainRepoPath(repo)` before inheriting a starting branch. The repo resolution itself had no equivalent check.

Sessions, by contrast, carry a `directory` re-derived from the real cwd on every register and spawn, plus `branch`/`is_worktree`/`main_repo` from `git.GetBranchInfo`. They are the reliable authority.

## Fix

Resolve the worktree's repo from the target workspace's **member sessions**, not the workspace record. When those sessions span more than one repository the choice is genuinely ambiguous, so fail and name `--repo` rather than guess — a confusing error beats a silent misplacement.

Deliberately scoped: only the worktree repo inference under existing-workspace placement changes. Current/new placements already base off a real session directory or `--cwd`. The workspace's `directory` keeps its meaning as the workspace's home (it is the user-picked folder for app-created workspaces and feeds the recent-locations picker) — this PR just stops treating it as a claim about repositories. An empty workspace still falls back to it, since nothing better exists.

I did **not** rewrite the `directory` write path. That is a broader behavioral change across the app's workspace-creation flow, and once the sessions are authoritative for delegation the drift no longer misplaces work. Worth a separate ticket if the drift bothers us elsewhere.

## Regression coverage

`TestDelegateWorktreeIgnoresStaleWorkspaceDirectory` builds two real git repos, puts the workspace's session in repo-a, then drifts the stored directory to repo-b via a re-registration (what production actually does), and asserts the worktree is created off **repo-a** with nothing beside repo-b. It fails on `main`:

```
worktree directory = ".../repo-b--feat-right-repo", want ".../repo-a--feat-right-repo" (off repoA)
```

Plus `...AmbiguousWorkspaceRepoRequiresRepo` (multi-repo workspace errors, creates nothing) and `...ExplicitRepoOverridesWorkspaceSessions` (`--repo` still wins).

## Live verification

Throwaway `wtrepo` profile, real daemon, real WS protocol — the fixture drift was produced by an actual `register_workspace` re-registration, not a hand-written DB row:

```
workspaces:  ws-target | Target | .../live/repo-b
sessions:    s-a       | .../live/repo-a | ws-target
```

Delegating `--workspace ws-target --worktree feat/live-check` with no `--repo`:

```json
{ "directory": ".../live/repo-a--feat-live-check", "worktree_created": true }
```

`git -C repo-a worktree list` shows it registered; repo-b has only `main`. Adding a second session in repo-b and retrying gives the loud failure:

```
delegate: workspace ws-target spans multiple repositories (.../repo-a, .../repo-b);
pass --repo to choose which one the worktree branches from
```

with no directory created. Profile cleaned afterwards.

## Docs

`internal/agent/attn_skill/references/delegation.md` said `--repo` "defaults to the workspace's repo" — updated to the real contract. The two CLI help strings also disagreed with each other (`main.go:711` said "workspace repository", `main.go:2317` said "source repository"); both now match.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,10 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   sessions move, so it could name a repository the workspace had nothing to do
   with. The repository now comes from the sessions already in that workspace,
   and a workspace whose sessions span several repositories asks for `--repo`
-  instead of silently picking one.
+  instead of silently picking one. Delegating into an existing workspace now
+  starts the branch from the repository's default branch: those sessions may sit
+  in several worktrees of one repository, and starting from whichever one the
+  daemon happened to list first was arbitrary. Pass `--from` to start elsewhere.
 - **Retried delegations no longer create duplicate agents or tickets.** `attn
   delegate` now prints a durable request and operation identity before slow
   worktree preparation, shows concise progress while it waits, and lets callers
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,9 +34,12 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   with. The repository now comes from the sessions already in that workspace,
   and a workspace whose sessions span several repositories asks for `--repo`
   instead of silently picking one. Delegating into an existing workspace now
-  starts the branch from the repository's default branch: those sessions may sit
-  in several worktrees of one repository, and starting from whichever one the
-  daemon happened to list first was arbitrary. Pass `--from` to start elsewhere.
+  starts the branch from the repository's default branch — `origin/<default>`
+  when it exists, so the work begins from what upstream has. Those sessions may
+  sit in several worktrees of one repository, so starting from whichever one the
+  daemon happened to list first was arbitrary, and starting from whatever the
+  main checkout had checked out was just as incidental. Pass `--from` to start
+  elsewhere.
 - **Retried delegations no longer create duplicate agents or tickets.** `attn
   delegate` now prints a durable request and operation identity before slow
   worktree preparation, shows concise progress while it waits, and lets callers
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,14 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   selected agent model and effort without changing the environment.
 
 ### Fixed
+- **Delegated worktrees no longer land in an unrelated repository.** When `attn
+  delegate --workspace <id> --worktree <branch>` ran without `--repo`, it chose
+  the repository from the workspace's recorded directory — a value that is
+  rewritten every time the workspace is registered and never updated when its
+  sessions move, so it could name a repository the workspace had nothing to do
+  with. The repository now comes from the sessions already in that workspace,
+  and a workspace whose sessions span several repositories asks for `--repo`
+  instead of silently picking one.
 - **Retried delegations no longer create duplicate agents or tickets.** `attn
   delegate` now prints a durable request and operation identity before slow
   worktree preparation, shows concise progress while it waits, and lets callers
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -708,7 +708,8 @@ placement:
 worktree options:
   combine with any placement (current, --workspace, or --new-workspace);
   combining with --cwd creates a worktree of the repo at that directory
-  --repo <path>              main repository (defaults to the workspace repository)
+  --repo <path>              main repository (defaults to the repository the
+                             target workspace's sessions are in)
   --from <ref>               branch or ref to start from
   --worktree-path <path>     override the generated sibling path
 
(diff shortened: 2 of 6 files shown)


figgyster (dismissed):
You found the right authority for the repository. `workspaces.directory` was a registration souvenir pretending to be live state, and the real-daemon proof plus the repo-a/repo-b witnesses make that case cleanly. But the replacement still guesses one level lower, where the wrong answer can send the delegated agent onto the wrong code just as quietly.

Required before merge:

- `delegationWorktreeBase` keeps the first session directory for a main repo (`internal/daemon/delegate.go:284-311`). When two member sessions use different worktrees/branches of that same repo, `len(repos) == 1`, so the choice is whichever session ID sorts first. `createDelegationWorktree` then converts that representative directory's branch into the implicit `StartingFrom` whenever `--from` is absent (`internal/daemon/delegate.go:372-382`). The new delegated branch can therefore start from branch A or branch B based only on session-ID ordering. Stable isn't the same as correct here. Please either fail loudly and require `--from` when same-repo members disagree on the starting point, or decouple repository inference from starting-ref selection with a deterministic, documented rule; pin the multi-worktree/same-main-repo case with a regression test.

The focused daemon and CLI tests pass locally, `git diff --check` is clean, CI is green, and the live verification is the right kind of evidence. This remaining ambiguity is the only thing keeping the file open.

figgyster (dismissed):
You killed the session-ordering guess, and the live reproduction was exactly the receipt I wanted. But the replacement rule still isn't the rule the code executes: "repository default branch" currently means "whatever the main checkout has checked out today." Ambient state changed hats; it didn't leave town.

Required before merge:

- Clearing `worktreeStartRefBase` at `internal/daemon/delegate.go:616-638` leaves `StartingFrom` empty. The shared creator then calls `git.CreateWorktree` (`internal/daemon/worktree.go:145-150`), which runs `git worktree add -b ...` from the main-repo checkout and therefore starts at that checkout's current `HEAD`. The protocol says so explicitly at `internal/protocol/schema/main.tsp:1361`. The new regression keeps the main checkout on `main`, so it cannot distinguish current HEAD from the claimed default branch. Put the main checkout on a divergent topic branch in the test: this head will start the delegated branch from that topic commit, contrary to the new CHANGELOG/skill contract and the comment at `delegate.go:620-622`. Please resolve and pass the actual default ref explicitly (the established targeted `origin/<default>` plus local fallback shape from the worktree flow is the boring answer), and make the regression prove a non-default main-checkout HEAD cannot steer delegation.

The updated focused daemon/CLI suites pass and the full patch is whitespace-clean. The repo/session decoupling itself is right; the final starting-ref source still needs to match the contract before this is safe.

figgyster (dismissed):
The inferred path finally does exactly what the prose says: upstream default beats stale local default, ambient checkout, and both member branches. That witness is excellent. But the same placement still changes semantics when the caller supplies `--repo`; one condition is keeping the old ambient-state path alive.

Required before merge:

- `worktreeStartRefBase` is cleared only inside the `msg.Worktree.Repo == ""` branch at `internal/daemon/delegate.go:663-674`. For `attn delegate --workspace <id> --repo <path> --worktree <branch>` without `--from`, the base remains `workspace.Directory`. `createDelegationWorktree` can then inherit that directory's branch when it belongs to the explicit repo, or fall through to the explicit repo checkout's current HEAD when it does not. That contradicts the new contract: `--repo` selects the repository; only `--from` should select a non-default starting ref. Clear the start-ref base for every existing-workspace worktree placement, then infer the repo only when `--repo` is absent. Add the matching regression: explicit `--repo`, no `--from`, main checkout parked on a divergent branch, and assert the delegated branch still starts from `origin/<default>` (the existing explicit-`--from` test already protects the override).

The latest focused daemon/CLI tests pass and the full patch is whitespace-clean. The resolver itself is right; this explicit-repo bypass is the remaining hole.

figgyster (approved):
Four passes later, the authority lines are finally clean. Existing-workspace delegation now takes repository identity from live sessions, or the stored directory only as a last-resort identity hint. It never takes a starting ref from that workspace record. And `--repo` selects the repository instead of acting like a disguised `--from`. The empty-directory guard also closes the daemon-CWD fallback instead of leaving an unrelated repository available by accident.

The new explicit-repo regression is the witness the contract needed. Legacy workspace branch, divergent ambient HEAD, stale local default, and advanced `origin/main` all have distinct commits, so the wrong implementation can't hide behind a friendly fixture anymore. The focused daemon and CLI tests passed. So did the full sandboxed validation and `git diff --check`. CI's green all the way down now, Bench A/B included.

Approved. This started as a stale-metadata fix and ended up removing four layers of ambient-state guessing. Expensive interrogation, but the code finally told the truth.

PR #613, opened 2026-07-19
feat(plugins): add attn-pi driver plugin (rock 1)

victorarias:
Rock 1 of `docs/vision/pi-attn-plugins.md`: pi launches, resumes, and lives as an attn session via a new bundled driver plugin, `plugins/attn-pi`.

## What

- **Pure driver, dumb state** (aligned with Victor): no `state_reporting`, no screen-scraping. Sessions ride the daemon's existing no-state path — created in `working`, `idle` on PTY child exit. Declared state arrives with the pi-side suite (rock 2).
- **Capabilities:** `resume`, `initial_prompt`, `model_pin` (`--model`, a fuzzy pattern per pi), `effort_pin` (`--thinking`, attn effort names are valid pi levels).
- **Version gate:** floor at pi 0.80.10 (the grounded version); resume refuses to run on an older pi than the session last used.
- **Resume token:** `PiMetadata {schema, pi_session_id, pi_version, model?, thinking?}` persisted daemon-side via `session.report_metadata`, handed back verbatim on `driver.resume`. The driver mints the pi session id at spawn (`--session-id` creates-if-missing).
- **Bundling:** `scripts/build-bundled-plugins.sh` refactored to a `stage_plugin` function staging both attn-opencode and attn-pi; attn-pi sources added to `scripts/source-fingerprint.sh` includes.
- No frontend changes needed: `driver.register` broadcasts settings, which publish `pi_available` + capability keys; the picker parses any `<agent>_available` key dynamically.

Also carries the phase-1 grounding docs (`docs/grounding/pi-plugins.md`, `plugins/attn-pi/AGENTS.md`) and the plan (`docs/plans/2026-07-19-pi-driver-plugin.md`).

## Testing

- `bun test` in `plugins/attn-pi`: 13/13 green (argv mapping, version floor, downgrade refusal, metadata validation, resume semantics).
- **Live verification** on throwaway profile `pismoke` (daemon built from this branch, headless over the daemon WS): smoke spike first (pi TUI under the worker PTY, resize re-flow, `--session-id` create+resume round-trip, pins, clean ctrl+d exit); then the full plugin path — `attn plugin install-bundled attn-pi` → driver registered (agent=pi) → spawn with effort pin → `working` + PiMetadata persisted → real turn → ctrl+d → `idle` → re-spawn same session id → `driver.resume` restored history with pins re-applied.

## Deferred (rock 2+)

State declaration, pi-side attn suite, delegation targets, safety envelope, kitty-graphics replay fidelity.

PR #615, opened 2026-07-19
feat(plugins): pi citizenship — declared state and in-band steering (rock 2)

victorarias:
Rock 2 of `docs/vision/pi-attn-plugins.md`: pi becomes a full attn citizen. One PR per Victor's call; plan doc at `docs/plans/2026-07-20-pi-citizenship.md`.

## What changed

**Suite (new, `plugins/attn-pi/suite/`)** — a pi extension bundled into the plugin and injected by the driver via `-e <suite.js>` at spawn/resume. It reports `session_start` (hello with pi session id — keeps the resume token current across fork/new/resume), `agent_start` (working), and `agent_settled` (stop with the last assistant text cached from `agent_end`). It answers `driver.deliver_message` by calling pi's steering API: `steer` mid-turn, plain send when idle. Fire-and-forget by design: any relay failure is swallowed, missing env makes it a no-op, and PTY child exit remains the authoritative liveness signal.

**Driver relay (`plugins/attn-pi/src/relay*.ts`, `driver.ts`)** — the suite talks only to the driver over a driver-owned unix socket (token-authenticated, ndjson JSON-RPC). The driver maps tokens to runs, owns the per-run report cursor (seq), forwards state/stop reports to the daemon, and relays `driver.deliver_message` to the suite. Empty settle text short-circuits to `idle` without a classifier call.

**Daemon** — new `message_delivery` capability + `driver.deliver_message` RPC; `attn.classify_stop` RPC (async classification of assistant stop text → idle/waiting_input/unknown); doorbells/nudges route through the plugin driver when the capability is declared — no PTY-typing fallback for such runs. Plugin API v4 → v5 (exact-match gate, all four sites).

**Build** — the suite is bundled as a single `suite.js` staged next to the driver binary (`--external @earendil-works/pi-coding-agent`; pi resolves it as a virtual module). Also pins **bun 1.3.14** (`.tool-versions` + fail-fast guard in `build-bundled-plugins.sh`): bun 1.3.13 emitted `--compile` binaries with payload past the code-signature extent, which macOS codesign refuses to sign — this broke `make` on main.

## Testing

- 34 bun tests (driver/relay/suite) + new daemon Go tests (`plugin_message_delivery_test.go`); full `make test` green; `make build-app` green with strict-valid signatures.
- Live-verified on a throwaway profile (`pirock2c`, packaged app): declared state flips working→idle on a real turn; question-shaped stop classifies `waiting_input`; a ticket nudge landed mid-turn via pi's Steering UI (in-band, not typed); a nudge to an idle session started a new turn; `/fork` inside pi refreshed the resume token and a kill+respawn resumed the forked conversation with a fresh run cursor.

Known follow-up (pre-existing): `attn preflight --agent <plugin agent>` can't see daemon-side plugin agents (CLI-local registry) — noted in the plan doc.


figgyster (dismissed):
Rock 2 is the right boundary. Pi declaring its own state and taking nudges through its native steering API is a much cleaner citizen than another PTY screen-scraper with forged papers. One ordering bug still makes the attention state lie, though.

**Blocker — reserve the stop report's sequence before awaiting classification** (`plugins/attn-pi/src/driver.ts:186`)

`suiteReportStop` waits for `attn.classify_stop` before calling `nextSeq(run)`. During that wait, an in-band message can start a new Pi turn and `suiteReportState` reports `working` with the next sequence. When the older classification eventually returns, its stop report receives an even newer sequence, so the daemon accepts the stale `idle`/`waiting_input` verdict over the live turn.

I reproduced this with a delayed classifier: metadata used seq 1, the newer `working` report used seq 2, and the older stop returned as seq 3. The regression should require the old stop to retain an earlier sequence than the subsequent working report (or otherwise generation/cancel-gate the classification result) so stale classification cannot overwrite current activity.

The existing evidence is otherwise strong: the clean Pi suite has 34 passing tests, the full sandboxed Go suite passes, `git diff --check` is clean, the PR's live-verification notes cover the real Pi paths, and all GitHub checks are green. This race is the one witness missing from the lineup.

figgyster (approved):
The race went straight. `ccbaa546` reserves the stop report's sequence before the classifier await, and the new deferred-classifier regression proves the old stop stays below a newer `working` report so the daemon discards it instead of lying about the live turn.

That closes the blocker cleanly and lets Rock 2 deliver what #613 deliberately deferred: Pi now participates in attn's real attention and steering contracts without screen-scraping or PTY-typing theater. I verified all 35 Pi tests, the full sandboxed Go suite, and a clean diff; Changes, Daemon, both Linux preflights, and Frontend E2E are green on this head. Bench A/B was still running at the final fetch.

Approved. Big swing, honest boundary, correct clock.

PR #616, opened 2026-07-20
docs: remove alignment briefs

victorarias:
## Intent / Why

`docs/alignment/` accumulated per-chunk alignment briefs as an unintentional side effect of a planning workflow — Victor never intended them to live in the repo.

## Summary

- Delete the `docs/alignment/` folder (10 files, 214 lines). Nothing else in the repo references it.

## Test plan

- Docs-only deletion; no behavior change. `grep -r docs/alignment` finds no remaining references.


figgyster (approved):
You’ve spent the last week making attn the durable authority over agents, tickets, and review continuity. Leaving these chunk briefs behind would create a second, increasingly stale authority in Markdown. One already points ticket artifact naming away from the contract that actually shipped. Good deletion.

The patch removes only the ten unreferenced briefs, the directory is absent from the resulting tree, and the docs-only CI path is green. Safe to merge. Case closed.

PR #617, opened 2026-07-20
feat(automations): scheduled trigger with a real maintenance case (Slice 5)

victorarias:
Slice 5 of the [Automations implementation guide](docs/plans/2026-07-18-attn-automations-implementation.md): automations can now fire on a cron schedule through the same engine, proven by a real recurring merged-worktree cleanup.

## What's in here

- **Scheduled trigger** (`trigger: {type: scheduled, schedule: {cron, time_zone}}`): robfig standard cron with a required IANA time zone (`TZ=`/`CRON_TZ=` prefixes rejected). Occurrence keys are the intended instant in UTC (`scheduled:<RFC3339>`), so identity is machine-independent.
- **Catch-up without storms**: a 1-minute observation loop fires at most the newest due instant per definition per tick. After downtime, `catch_up: latest` runs the newest missed instant once; `catch_up: skip` drops instants older than a 5-minute grace. The first observation of a new definition anchors its cursor without retroactive fires.
- **Singleton continuity**: `continuity: singleton` keeps every occurrence on one ticket/session (continuity key `"singleton"`), reusing the Slice 4 binding machinery; later occurrences nudge the live agent instead of spawning a new one.
- **Claim-failure safety** (from the pre-review verify pass): a rejected or failed claim leaves the schedule cursor behind the intended instant so the next tick retries instead of silently dropping the occurrence; scheduled occurrences record their continuity key as `subject_key` so the deleted-ticket continuity guard actually matches scheduled priors; `TriggerSpec.Schedule` is a pointer so non-scheduled definitions keep byte-identical canonical JSON (no spurious revision bump on re-apply after upgrade).
- **Provider-gated head check**: `WorkRequest.Provider` gates the PR head-SHA continuation check to github deliveries so scheduled continuations don't trip it.

## Documented boundaries (revisited in Slice 7)

- Editing a singleton scheduled definition's prompt/launch/location makes later occurrences fail visibly at delivery (same fail-visible-on-changed-contract stance as Slice 4's per-PR continuations). Recovery today: disable or recreate under a new id; proper edit semantics belong to Slice 7's lifecycle work.
- `catch_up` is now schedule-only: re-applying a manual/github definition carrying a stray `catch_up` fails validation (stored definitions unaffected).

## Verification

- Full Go suite green; new unit tests cover claim-rejection cursor retry, skip-grace boundary, disabled-definition claim rejection, the isRecovering observation gate, singleton subject_key recording, and canonical-JSON omission.
- Live packaged evidence on the isolated `automations` profile, serial scenario `real-app:scenario-automation-scheduled-cleanup` (fingerprint-guarded app built from `45c7684c`): anchor-without-fire, exactly one catch-up run after 135s downtime with a minute-aligned UTC occurrence key, a real codex agent removing a merged-clean worktree while preserving a dirty one (uncommitted file intact), all occurrences coalesced on one singleton ticket/session, and a storm-guard restart leg proving exactly one spawn. Run `automation-scheduled-cleanup-2026-07-20T12-22-03-841Z`, all legs green, verdict `ok:true`.

CHANGELOG and guide ledger updated in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-20]
 
 ### Added
+- **Automations can now run on a schedule.** A definition can declare a cron
+  schedule with a required time zone and fire an ordinary visible agent
+  session at each intended instant — for example, a recurring merged-worktree
+  cleanup. After daemon downtime, `catch_up: latest` runs the newest missed
+  instant once, while `catch_up: skip` quietly drops instants older than a few
+  minutes; a backlog never replays as a storm of runs. `continuity: singleton`
+  keeps every occurrence on the same ticket and session instead of opening a
+  new one each time.
 - **pi sessions now report their own state and take steering in-band.** A pi
   session launched by attn loads a bundled companion extension that declares
   when the agent is working, tells attn how a turn ended — done, or waiting
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   resume token current.
 
 ### Changed
+- **`catch_up` is now schedule-only.** Applying an automation whose manual or
+  GitHub trigger carries a `catch_up` policy fails validation with a clear
+  error; remove the field from the definition. Already-stored definitions are
+  unaffected.
 - **Later review requests return to a safe existing reviewer.** Per-PR
   automations now resume a stopped reviewer with its recorded transcript and
   pinned unattended policy, reopen archived tickets after successful delivery,
--- a/app/package.json
+++ b/app/package.json
@@ -33,6 +33,7 @@
     "real-app:scenario-tr301": "node scripts/real-app-harness/scenario-tr301-local-utility-session-switch.mjs",
     "real-app:scenario-codex-resume": "node scripts/real-app-harness/scenario-codex-resume-mapping.mjs",
     "real-app:scenario-automation-pr-continuity": "node scripts/real-app-harness/scenario-automation-pr-continuity.mjs",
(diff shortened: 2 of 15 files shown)


figgyster (changes requested):
You’ve taken the continuity machinery from #608/#614 and handed it a clock. That’s the right next turn for this arc. The intended-instant keys, immutable snapshots, and real merged-worktree cleanup are the pieces I wanted to distrust; they held up. One timing hole still drops the very occurrence this PR says it preserves.

Required before merge:

- `internal/daemon/automations_schedule.go:114`: keep a failed claim’s intended instant eligible ahead of the normal “newest due wins” catch-up rule. Right now a minutely schedule that rejects the 03:01 claim at the 03:01:30 observer tick leaves its cursor behind, but the next real tick at 03:02:30 computes both 03:01 and 03:02 and selects 03:02. The 03:01 occurrence is never claimed. I reproduced that exact cadence with a temporary regression: the retry delivered `scheduled:2026-07-20T03:02:00Z`, not the failed `03:01:00Z`. The existing claim-rejection test retries five seconds later, before another cron instant exists, so it misses the production path. Please preserve/retry the failed intended instant explicitly (or otherwise make the promised failure-safety true) and pin it with a one-minute-later regression.

The full Go suite, scheduled-focused regressions, diff check, and scenario syntax all passed. CI is green all the way down. The clock’s good. It just can’t skip the appointment when SQLite coughs.

figgyster (approved):
Fair. I booked the wrong suspect.

Once `catch_up: latest` is the public contract, preserving 03:01 across a 03:02 retry would make stale work outrank the fresher due instant, and doing exact retry honestly across restarts would need another durable queue. That’s a much larger machine than Slice 5 needs. `3baf2451` turns my reproducer into the right witness: one delayed execution, the newest due key, one claim, then the cursor advances. The original same-instant test still covers schedules where no newer slot has appeared.

This keeps #608/#614’s continuity machinery pointed in the right direction: attn owns recurring maintenance without replay storms or a second hidden retry system. The focused cadence regressions pass, the full PR diff check is quiet, and the follow-up only changes the stated contract plus its witness; the packaged run still exercises the same daemon behavior. Safe to merge. I was wrong about the blocker. Case closed.

PR #618, opened 2026-07-20
fix(terminal): recover from Ghostty model faults

victorarias:
## Intent / Why

A Ghostty terminal-model fault currently escapes React and crashes the entire attn UI. This change contains that fault to the affected pane, restores its daemon-owned PTY view, and leaves Victor an actionable diagnostic record for a proper native fix.

## Summary

- Catch Ghostty model faults from rendering, fit-resize, queued writes, and disposal; rebuild the canvas and WASM model instead of letting the error reach the app error boundary.
- Reattach the rebuilt pane through the existing same-app-remount path so the daemon replays its retained terminal state.
- Persist a model-fault record with operation, error, stack, model id, geometry, and UI snapshot in the UI diagnostics stream; also add a terminal lifecycle model_fault event.
- Show a 12-second notification after the replacement model is ready: Terminal issue recovered. We reloaded it for you. The toast names the exact profile-local ui-diagnostics.jsonl path and asks the user to send it to Victor.
- Keep long diagnostic paths readable in the toast and add focused coverage for the durable event and extended notice duration.

## Investigation findings

The two production failures on 2026-07-20 were not caused by a Notebook tile. The active screen was a Ghostty terminal; NotebookTile appeared only as a dynamically-loaded bundle ancestor.

- First failure: an out-of-range code point during rendering immediately after a fit from 82x48 to 153x76.
- Second failure: an out-of-bounds WASM memory access in terminal.update() immediately after a fit from 153x76 to 307x77.
- Both are faults in the Ghostty model/render path. The terminal component had no local fault boundary, so React unmounted the app after the exception.
- The shared fit path calls resizeGhosttyWithoutReflow. It temporarily injects DEC wraparound-off around resize to select Ghostty's no-reflow behavior. That policy protects attn's detachable, reconnectable UI from rewriting replayed terminal history when pane geometry changes.
- Replaying the retained terminal bytes and repeating the observed size transitions outside the live app did not reproduce the invalid model. The exact native trigger is therefore still unresolved; this PR captures the next occurrence rather than claiming an upstream root cause it cannot prove.

## Deferred proper fix

1. Replace the injected DEC wraparound resize trick with a native Ghostty WASM no-reflow resize API, if or when Ghostty exposes one. The API should make the preservation policy explicit instead of changing a terminal mode around resize.
2. Use the new fault-time records to build a minimal Ghostty-core reproduction, then fix or upgrade the vendored core rather than relying on UI containment.
3. Preserve the no-reflow user behavior until that native replacement is proven. Switching to ordinary reflow would be a behavioral regression for attached replay and geometry changes, not a neutral cleanup.
4. Consider a narrower terminal-runtime isolation boundary only if native model faults remain possible after the upstream fix.

## Test plan

- [x] Focused Vitest coverage for model-fault lifecycle diagnostics and the extended recovery notice.
- [x] Full frontend Vitest suite and Playwright suite via the repository pre-commit hook.
- [x] pnpm --dir app run build.
- [x] Installed and ran the final isolated attn-dev bundle.
- [x] Ran real-app:scenario-webgl-recovery: forced context loss, observed contextLost -> scheduled -> recovered, reattached the PTY, rendered fresh output, and confirmed no error overlay.
- [ ] Trigger a real model fault with the new diagnostics and use that artifact to land the native Ghostty fix.

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -75,7 +75,7 @@ import { useOpenPR, type OpenPRProgress } from './hooks/useOpenPR';
 import { useUiAutomationBridge } from './hooks/useUiAutomationBridge';
 import { ptySpawn } from './pty/bridge';
 import { clearBrowserHostFocus, controlBrowserHost, isBrowserHostOwnedTarget } from './browser/host';
-import { probeUiAfterSwitch } from './utils/uiDiagnosticsLog';
+import { probeUiAfterSwitch, UI_DIAGNOSTICS_FILE_DISPLAY } from './utils/uiDiagnosticsLog';
 import {
   agentLabel,
   getAgentAvailability,
@@ -1770,13 +1770,20 @@ sendFetchPRDetails,
   const [locationPickerSessionDirection, setLocationPickerSessionDirection] = useState<TerminalSplitDirection>('vertical');
 
   const [zoomModeBySessionId, setZoomModeBySessionId] = useState<Record<string, boolean>>({});
-  const { message: errorMessage, showError, clearError } = useErrorToast();
+  const { message: errorMessage, durationMs: errorDurationMs, showError, clearError } = useErrorToast();
   const [chiefTransferTarget, setChiefTransferTarget] = useState<{
     sessionId: string;
     targetLabel: string;
     currentLabel: string;
   } | null>(null);
   const [chiefTransferSaving, setChiefTransferSaving] = useState(false);
+
+  const handleTerminalModelRecovered = useCallback(() => {
+    showError(
+      `Terminal issue recovered. We reloaded it for you. Diagnostics were saved to ${UI_DIAGNOSTICS_FILE_DISPLAY}; please send this file to Victor so he can troubleshoot it.`,
+      { durationMs: 12_000 },
+    );
+  }, [showError]);
 
   const handleRebootstrapEndpoint = useCallback(async (endpointId: string) => {
     try {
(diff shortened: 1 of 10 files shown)


figgyster (dismissed):
The terminal-stability arc keeps digging downward. #472 stopped font-size churn, #473 taught dead WebGL contexts how to come back, and now you’ve got evidence that the Ghostty model itself can go bad. Using daemon replay as the authority while you hunt the native trigger is the right emergency move. A corrupt WASM model has already told enough lies for one night.

One blocker before merge:

- `app/src/components/GhosttyTerminal.tsx:643` keeps `modelFaultRef` set until the replacement model reaches line 2024. That suppresses duplicate failures from the old model, but it also suppresses a real failure from the replacement during its initial `fit()` at line 1984. `applyFitDimensions` catches that failure and calls `recoverFromModelFault`; the latch returns immediately, initialization keeps going, and this path can reattach plus show “recovered” for a replacement whose resize already failed. Split the old-model dedupe latch from the pending recovery/notice state, and add a regression that forces the replacement model’s initial fit to fault: it should schedule another epoch and withhold the recovered notice until a healthy model is ready.

The frontend suite and production build pass here. The new tests prove the diagnostic record and toast duration, but none drives the containment path above, which is why this one slipped past 1,845 green witnesses.

figgyster (approved):
You made the recovery state tell the truth about ownership. Duplicate faults are scoped to one renderer epoch now, while the pending notice survives across as many replacement attempts as it takes. The new witness reenacts the whole ugly sequence — bad original, bad replacement during initial fit, healthy third model — and proves only the healthy one becomes ready or rings the recovery bell.

That closes my blocker. This sits well on top of #472’s in-place remetricing and #473’s WebGL recovery: when the model itself is the unreliable witness, daemon replay remains the authority. Focused recovery tests and the production build pass here, and CI is green all the way down. Approved. Case closed.

PR #614, opened 2026-07-19
feat(automations): resume recurring PR reviews

victorarias:
## Intent / Why

Later review-request cycles for the same pull request should return to the reviewer attn already created, without losing the ordinary ticket, local review notes, or the exact launch policy. This deepens the durable subject binding introduced in Slice 3 while keeping each occurrence and run independently visible.

## Summary

- Resume a stopped per-PR reviewer only when attn has a recorded resume target, using the immutable automation launch snapshot and the new occurrence prompt.
- Verify the recorded Codex rollout actually exists before unattended continuation (`ResumeAvailable` via `transcript.FindCodexTranscriptForResume`); a pruned or moved rollout fails closed instead of spawning fresh.
- Reopen archived tickets after successful continuation, preserve automation-owned dirty or branch-changed worktrees, and fail visibly when a delivered worktree is missing or the PR head changed.
- Keep live reviewers on the existing nudge path and retain safe first-launch behavior when a review request was withdrawn before delivery.
- Normalize bundled Bun executables before app signing so non-production packaged builds pass strict macOS validation.
- Add a serial packaged-app scenario that uses local mock GitHub, a copied Codex rollout, and machine-readable assertions to prove continuity and failure behavior in seconds.

## Test plan

- [x] `go test ./internal/agent ./internal/daemon ./internal/store -count=1` (rerun after rebasing onto `main` post pi PRs #613/#615; two conflict resolutions: CHANGELOG.md and `build-bundled-plugins.sh`, both keep-both-sides)
- [x] Focused continuation, withdrawal, archived-ticket, dirty-worktree, changed-head, missing-worktree, and unavailable-transcript tests, plus present/missing Codex rollout regressions
- [x] Pre-commit at `b2a43445`: 1,848 frontend tests, frontend build, 189 Playwright tests passed (1 expected skip)
- [x] `make install PROFILE=automations` and packaged preflight at protocol 175 (rerun 2026-07-20 on a fresh machine/profile)
- [x] `ATTN_PROFILE=automations pnpm --dir app run real-app:scenario-automation-pr-continuity` — **rerun green on the exact-rollout fix** (2026-07-20, fresh `automations` profile, 15s)
  - same ticket `auto-1a7c07b4a1a04b0c`
  - same session `f2d4d83b-a407-44f2-8ffa-b769bb2a2b1e`
  - copied rollout `019f5cd5-3510-7781-b3b0-8ade5b2cef1c` passed to `codex resume`
[...]

PR #619, opened 2026-07-20
feat(automations): profile-level Automations surface (Slice 6)

victorarias:
# Automations: profile-level surface (Slice 6)

Slice 6 of the [Automations implementation plan](docs/plans/2026-07-18-attn-automations-implementation.md): the first operational UI for automations. A dock panel lists every definition with its trigger, enable/disable toggle, run-now for manual definitions, and run history with navigation to the run's ticket/session.

## What's in it

**Protocol (175 → 176).** Three new WS commands (`automation_definitions_get`, `automation_runs_get`, `automation_set_enabled`; the existing `automation_run` message is reused for run-now) answered by a typed `automation_action_result` correlated by `request_id`, plus a thin `automations_changed` broadcast carrying definition ids only. The broadcast is emitted from the daemon's core mutation/transition methods (apply, claim on all three trigger paths, fail, deliver, set-enabled), so CLI-socket and WS mutations both fan out; the frontend re-reads canonical SQLite state via the get commands instead of trusting event payloads. `automation_action_result` is deliberately not merged with the legacy unix-socket result shape (generic `data`) — they are not wire-compatible.

**Daemon.** `automationSetEnabled` flips the flag without touching spec/revision; re-enable shares the apply path's side effects (clears review-request edges, fences provider cursors), disable fails pending runs. Run-now is idempotent per `request_id` (persisted as the manual claim key). Runs-get is capped at 100 with an explicit `truncated` flag. WS `automation_run` mirrors the socket arm's `pr_url`/`input_json` mutual exclusion and routing.

**Frontend.** `AutomationsPanel` as a new dock panel, backed by a small Zustand store keyed by a `changedTick` that bumps on `automations_changed` and on every (re)connect (panel recovers across daemon restarts). No optimistic toggles — every mutation round-trips and errors render inline per definition. Run rows navigate to their ticket or session (empty-string pointer = absent). Wrapper timeouts are 30s: mutations can serialize behind an in-flight delivery holding `automationMu` (pre-existing lock structure, deliberately not restructured in this slice).

## Verification

- `go build ./...`, `go test ./internal/store ./internal/daemon`, `tsc --noEmit`, full frontend vitest suite green on `d6844d68` (the head minus the docs-only commit).
- Adversarial verify gate over the slice diff (4 lenses, 2 refuters per finding): 10 confirmed findings, all fixed in `d6844d68` (claim-time broadcasts on scheduled/GitHub paths, reconnect refetch, WS `pr_url` routing, timeout convention, dead label branch, plus the missing unit coverage for the hook pipe, store ordering/limit, and non-manual rejection).
- Live packaged evidence on the isolated `automations` profile (protocol 176 lockstep preflight PASS), scripted serial scenario `real-app:scenario-automation-surface`, run `automation-surface-2026-07-20T15-00-09-048Z`, all 4 legs green against the app built from `d6844d68` (this branch's last code commit; `3e9737d4` on top is docs-only): CLI-applied definition appears via broadcast; run-now yields exactly one delivered navigable run with no duplicate on reopen; disabled-definition run-now surfaces the daemon rejection inline without creating a run; full daemon restart preserves definitions and the same run id, still navigable.

Slice 7 (self-service editor, edit/delete lifecycle, retention/policy depth) follows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-20]
 
 ### Added
+- **Automations now have a panel in the app.** A new toolbar button opens an
+  Automations panel that lists every definition with its trigger and
+  enabled state, lets you enable/disable or run a manual automation now,
+  shows each definition's run history with failures called out inline, and
+  jumps straight to the ticket or session a run produced. Definitions are
+  still created from the CLI (`attn automation apply`); changes made there
+  appear in the panel immediately, and daemon rejections are shown in the
+  panel rather than silently swallowed.
 - **Automations can now run on a schedule.** A definition can declare a cron
   schedule with a required time zone and fire an ordinary visible agent
   session at each intended instant — for example, a recurring merged-worktree
--- a/app/scripts/real-app-harness/scenario-automation-surface.mjs
+++ b/app/scripts/real-app-harness/scenario-automation-surface.mjs
@@ -1,0 +1,432 @@
+#!/usr/bin/env node
+
+/**
+ * Packaged-app proof for Slice 6's profile-level Automations dock panel: a
+ * CLI-created definition appears in the panel via broadcast with no restart,
+ * "run now" driven through the real panel controls produces a navigable run
+ * and re-rendering the panel never duplicates it, a rejected run-now surfaces
+ * its error inline instead of failing silently, and both the definition list
+ * and run history survive a daemon+app restart.
+ *
+ * Two definitions are applied against one isolated profile daemon:
+ *   - `automations-surface-manual-<suffix>`: trigger.type manual, policy
+ *     continuity fresh (the only continuity manual trigger validation
+ *     allows), launched against a FAKE codex executable (argv logger, like
+ *     the Slice 5 storm-guard probe) so a run-now click reaches `delivered`
+ *     fast and deterministically.
+ *   - `automations-surface-scheduled-<suffix>`: trigger.type scheduled with
+ *     a once-a-year cron so it never fires during the scenario window;
+ *     exists only to prove the panel hides the run-now affordance for a
+ *     non-manual trigger.
+ *
+ * Run serially (packaged-app scenarios are single-tenant):
+ *   ATTN_HARNESS_PROFILE=<name> node scripts/real-app-harness/scenario-automation-surface.mjs
+ */
+
+import fs from 'node:fs';
+import path from 'node:path';
+import { execFileSync } from 'node:child_process';
+import { parseCommonArgs, printCommonHelp, launchFreshAppAndConnect } from './common.mjs';
+import { createScenarioRunner } from './scenarioRunner.mjs';
+import { currentHarnessProfile, resolveHarnessResources } from './harnessProfile.mjs';
+import { UiAutomationClient } from './uiAutomationClient.mjs';
+import { DaemonObserver } from './daemonObserver.mjs';
+import { captureFrontWindowScreenshot } from './nativeWindowCapture.mjs';
+
+const delay = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
+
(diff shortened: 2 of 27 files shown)


figgyster (changes requested):
You’ve spent six slices turning Automations from a daemon idea into something Victor can actually operate, and the canonical-state-plus-thin-broadcast shape is the right backbone. But the first Run now button has a delayed-action trap hiding under it.

Required before merge:

- `app/src/hooks/useDaemonSocket.ts:5374` forgets a run request after 30 seconds and the panel re-enables the button, while `internal/daemon/ws_automations.go:101` launches the server work with `context.Background()` and `internal/daemon/automations.go:166-170` durably claims the run *before* waiting on `automationMu`. The PR itself notes that another delivery can hold that lock for tens of seconds. Once it crosses the client deadline, the UI says “Run automation timed out” even though the first run is already pending and will still deliver later; clicking again generates a fresh UUID, so the retry claims a second occurrence and can launch duplicate work. `setAutomationEnabled` has the same late-success problem: it can flip the definition after the UI has reported failure. Make the deadline authoritative before any later mutation, or keep/reconcile the operation as pending and preserve a stable idempotency key across retry. Please pin it with a regression that holds `automationMu` past the client deadline and proves there is neither a late toggle nor a second manual run.

The full Go suite, the 85 focused frontend tests, TypeScript, and current CI are green. None of those witnesses exercise this slow-lock boundary. Green lamps, dark alley. I’m holding the merge here.

figgyster (changes requested):
You fixed both failure modes I called out on the last head. The 25-second daemon deadline now stops a late enable/disable write, and a retry during the same app lifetime reuses the original run-now key. Good evidence, too. The exact lock regressions saw the crime instead of interviewing a mock about it.

One restart boundary is still open before merge:

- `app/src/store/automations.ts:39-43,71` keeps `pendingRunRequests` only in the in-memory Zustand store and resets it to `{}`. After a run-now request times out, the daemon may still have the durable `manual:<request_id>` run pending in SQLite; if the packaged app exits or recovers before that run settles, the next launch forgets the key. `app/src/components/AutomationsPanel.tsx:43-45` then cannot adopt the pending occurrence from fetched history because reconciliation returns when no in-memory key exists, and `handleRunNow` at line 175 mints a new UUID. The next click can therefore claim a second durable run—the original duplicate-work bug, one app restart later. Make the retry identity recoverable across relaunch (for example, adopt the pending manual occurrence key from canonical run history, persist the key, or keep Run now unavailable while a pending manual run exists), and add a regression that resets/remounts the store with a pending manual run in fetched history before clicking again and proves only one occurrence survives.

I ran the full Go suite in the sandbox, the 93 focused frontend tests, and TypeScript. All passed. `git diff --check` passed too. Principal CI is green; Bench A/B is still working the late shift.

figgyster (commented):
You found the right end to pull. Canonical run history now reconstructs the pending manual request after a store reset, while a current-session key keeps precedence and scheduled occurrences stay out of the lineup. That closes the relaunch duplicate I blocked on without adding a second persistence ledger. The exact regression is an eyewitness: reset the store, fetch `manual:restart-key-1` as pending, click Run now, and the old identity is reused.

I ran the two focused frontend files here—36 tests passed—plus TypeScript and the diff check. The previous head’s full Go suite passed, and this follow-up doesn’t touch Go.

The code objection is gone. I’m holding the approval because the current head’s required Frontend E2E shard 3 and aggregate E2E checks are red. Once that witness comes back green, I’ve got nothing left on the file.

figgyster (approved):
The last witness came back sober. Frontend E2E shard 3 passed on rerun, the aggregate check passed with it, and the rest of CI is green all the way down.

This closes the two durability gaps that mattered: the daemon won’t mutate enablement after the UI’s deadline, and a manual run keeps one identity across both a slow delivery and an app relaunch by recovering it from canonical SQLite history. That’s the right ending for the arc from #608 through #614 and #617—attn finally has an operating surface that tells the same truth as the coordination engine underneath it.

Approved. Six slices, two trips back to the alley, one honest control surface. Case closed.

PR #620, opened 2026-07-20
fix(build): stop shipping a stale embedded frontend; de-flake split-blank-repro

victorarias:
Three fixes to the build-and-verify loop, found while investigating why the
automations epic (#614–#619) was expensive to run.

## 1. `build.rs` — cargo was never told `app/dist` is an input

`tauri::generate_context!()` embeds `build.frontendDist` when the proc macro
expands, and proc macros cannot emit `cargo:rerun-if-changed`.
`tauri_build::build()` emits that directive for `frontendDist` only on the
opt-in `CodegenContext` path, which we do not use — it resolves the config at
build-script time and so would not see the `--config` overlay named-profile
builds pass.

So on a frontend-only change the crate was **genuinely fresh** and cargo
correctly skipped it — shipping the previously embedded assets. Not a caching
quirk; a missing dependency edge. `build.rs` now reads `build.frontendDist` out
of `tauri.conf.json` and emits the directive itself.

A/B on a frontend-only change:

| `build.rs` | cargo |
|---|---|
| before | `Finished in 0.41s` — no recompile, ships stale assets |
| after | `Compiling app v0.11.1` |

Emitting this is purely additive: it can cause more rebuilds, never a wrong one.

This is what made `touch app/src-tauri/src/lib.rs` a necessary ritual — a
workaround that appears in no repo doc. A known-broken incremental path is why
the epic ran **13 full `make install` and zero `make install-daemon-dev`**.

## 2. `AGENTS.md` — install-tier guidance

Now that the cheap tiers are trustworthy, the guidance is an explicit rule keyed
on what changed (Go-only → daemon-only install; anything under `app/`, a
protocol change, or bundle metadata → full build), with "escalate when unsure."

## 3. `split-blank-repro.spec.ts` — de-flaked, not quarantined

`:68` and `:219` each failed 3× in CI during the epic, costing `gh run rerun`
cycles.

A single `__TEST_EMIT_PTY_DATA` can be dropped silently at three points: the
router's binding lookup (`paneRuntimeEventRouter.handleEvent`), the pane's
handle lookup (`useGhosttyPaneRuntime.deliverEvent`), and the fire-and-forget
`void terminal.write(...)`. `expect.poll` re-**read** the pane text but never
re-**sent** it, so one lost emit could only ever expire at the deadline.

That matches the CI failures exactly — **5.9s against a 5s poll**, i.e. ~0.9s of
setup then a full 5s of fruitless re-reads. The runner was responsive; an emit
was lost.

Setup frames now re-emit until the model shows them. Retry is applied **only to
scaffolding** — the post-split redraw stays a single emit, because its timing
against the resize is the behaviour under test and retrying it would mask the
very bug these specs exist to catch. The scrollback emit is also now asserted
rather than assumed: if it were dropped, wheel-up would yield a zero offset and
the test would pass while exercising nothing.

Also corrected a comment claiming `getPaneSize` was "the precise 'emit will be
delivered' signal" — it isn't: it misses the router binding, and reads a size
`GhosttyTerminalHandle` documents as provisional until a real fit lands.

## Verification

- **Injected-drop control/treatment**: dropping exactly one emit reproduces the
  CI error verbatim (`expect(received).toContain(expected) // indexOf`); the new
  helper survives the identical drop. Deterministic and load-independent.
- Spec **20/20** via `--repeat-each=5` on an idle machine.
- Full e2e suite: **189 passed**, 1 skipped.
- `make dev`: compiles, bundles, signs, installs, launches (exit 0) — the live
  verification for the `app/src-tauri` change.

I could not reproduce the flake *naturally* — CDP CPU throttling, load
saturation and repeat runs all stayed green. The fix rests on the failure
signature and timing plus the injected-drop proof, not on a natural repro.

No `CHANGELOG` entry: docs, build system, and test-only changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -39,7 +39,15 @@ permissions.
 - Non-production builds, installs, launches, and restarts are pre-authorized.
 - Production `make`, `make install`, and `make install-daemon` require Victor's
   explicit approval.
-- Default to `make dev`; use `make install-daemon-dev` for daemon-only changes.
+- Install the cheapest tier that covers the change:
+  - Go-only change (`cmd/attn`, `internal/**`) → `make install-daemon-dev`, or
+    `make install-daemon PROFILE=<name>`. Replaces and re-signs the sidecar and
+    restarts the daemon; no Tauri/Rust/frontend build.
+  - Anything under `app/` (frontend, `src-tauri`, plugins), a protocol change
+    (`generated.ts` moves with `generated.go`), or bundle metadata → `make dev`,
+    or `make install PROFILE=<name>`.
+  Escalate to the full build when unsure, or when a daemon-only install does not
+  show the change.
 - Named profile: select it with `eval "$(./attn profile-env <name>)"`, then run
   `make install PROFILE=<name>`. The shell's `ATTN_PROFILE` must match.
 - `profile-env` clears inherited routing overrides. Verify the emitted
--- a/app/e2e/split-blank-repro.spec.ts
+++ b/app/e2e/split-blank-repro.spec.ts
@@ -79,10 +79,7 @@ test('agent pane stays painted after opening a shell split', async ({ page, daem
   const terminal = await setupAgent(page, daemon, agentId);
 
   // 1) Paint a full frame and confirm it actually rendered (high quad count).
-  await emit(page, agentId, fullFrame('OLD'));
-  await expect
-    .poll(async () => page.evaluate((sid) => window.__TEST_GET_SESSION_PANE_TEXT?.(sid) ?? '', agentId), { timeout: 5000 })
-    .toContain('OLD line 0');
+  await emitUntilVisible(page, agentId, fullFrame('OLD'), 'OLD line 0');
   await expect
     .poll(async () => {
       const trace = await readTrace(page, agentId);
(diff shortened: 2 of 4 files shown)


figgyster (dismissed):
The Automations epic made you pay for this build lie thirteen times before it confessed. Good archaeology. Teaching Cargo that the embedded frontend is a real input closes the stale-asset hole at the dependency graph, and the named-profile overlay only changes bundle metadata, so the committed `frontendDist` remains the right edge to track.

I like the restraint in the split-blank fix too. You retry only the setup frames and leave the post-split redraw as a single shot, so the witness stops missing court without being coached on the behavior under test. The scrollback assertion also fixes a quiet false-green path.

The diff check passed, the Tauri/frontend/E2E precinct is green, and the signed `make dev` plus injected-drop run cover the two risks that matter here. Safe to merge. Case closed.

figgyster (approved):
You caught the one overclaim before it hardened into another ritual: tracking `dist` fixes stale embedded assets, but Vite rewrites the directory on every full app build, so this buys correctness rather than a faster `make dev`. That correction makes the install-tier guidance honest. Go-only work gets its speed from skipping Tauri through `make install-daemon-dev`.

The increment from the approved head is comment-only and its diff check passes. The implementation and focused evidence I approved are unchanged, so the fresh CI run is redundant weather rather than a reason to park the merge. Safe to merge on `6bb1de9c`. Case closed again. Codebases enjoy an encore.

PR #622, opened 2026-07-20
feat(automations): lifecycle semantics — edit, delete, retention, cleanup (Slice 7 PR A)

victorarias:
Slice 7 of the Automations epic is split into two PRs. This is **PR A: lifecycle
semantics**. PR B (the self-service YAML editor and validate-without-apply) follows
on top of this one.

Guide: `docs/plans/2026-07-18-attn-automations-implementation.md`.

## What changes

**A1 — edits rotate continuity, old snapshots stay pinned.**
`automation.ContinuationContract` (Prompt + Launch + Location) is the single
definition of "the same continuation". Applying an edit that changes the contract
rotates the definition's continuity bindings, so the next occurrence opens a fresh
ticket/session thread instead of resuming a thread that was created under different
instructions. The same type is used at delivery time by `validateAutomationContinuation`,
so apply-time and delivery-time notions cannot drift; this replaced an ad-hoc triple
check and deleted `sameAutomationLocation`. Policy and the per-occurrence GitHub head
SHA are deliberately excluded — neither changes what a continued thread *is*.

Snapshots themselves are untouched: `automation.Effective` is called once at claim
time and frozen into `automation_runs.snapshot_json`, and a run claimed before an
edit keeps delivering under its old snapshot forever.

**A1-fix — a revert no longer bricks delivery.** The pre-existing safety net refused
delivery whenever a prior run shared the contract and its ticket was missing. Keyed on
the contract alone, the sequence *edit A→B, then revert B→A* refused forever: after
rotation the old contract-A runs matched the reverted contract again, and there was no
recovery path (delete + re-apply rotates bindings but run history survives).
`AutomationContinuityRunHistory` now returns each prior run's own `snapshot_json`, its own
`ticket_id`, **and its own `session_id`**, and the daemon refuses only when the vanished
ticket belongs to the thread the request would actually be reusing — i.e. when the history
entry's session id is the request's own. A different session id means the request already
holds freshly reserved identifiers (no binding survived to hand it an old one), so nothing
is being reused and there is nothing to refuse. The directions are pinned by tests:
ordinary sweep refuses, rotation allows, revert-with-live-ticket allows, revert-with-swept-
ticket allows *on a fresh session*, and a request whose **own** session's ticket vanished
still refuses.

**A2 — delete retains provenance.** `attn automation delete <id>` soft-deletes the
definition: pending runs are failed, review-request edges and continuity bindings are
removed, provider cursors are fenced, and the row gets `deleted_at`. Runs, occurrences,
tickets, sessions, and artifacts are all retained, so finished work stays inspectable.
Re-applying the same YAML resurrects the definition (revision bumped, `deleted_at`
cleared, bindings rotated unconditionally) with its old run history intact.
`automationDelete` mirrors `automationSetEnabled`'s lock/deadline contract, so it
aborts rather than mutating if it is still waiting on `automationMu` at the 25s
daemon-side deadline.

**A3 — bounded retention.** A dedicated hourly sweep prunes terminal runs that fall
outside the keep window (newest 200 per definition) and are older than the age floor
(14 days), removing worktree, occurrence artifact, and run/occurrence rows. All three
knobs are env-overridable (`ATTN_AUTOMATION_RETENTION_KEEP` / `_MIN_AGE` /
`_SWEEP_INTERVAL`).

**A4 — explicit cleanup, dirty-safe.** `attn automation cleanup <id>` walks every
terminal run now, with no age or count gate, and is **disk-only**: it removes worktrees
and never touches rows or occurrence artifacts. It returns a three-way partition of the
runs it examined — `cleaned` / `kept_dirty` / `kept_active` — so every worktree the user
asked about is accounted for. A worktree with uncommitted changes (untracked files
included) is never removed. It reads the definition including soft-deleted ones, so a
retired automation's leftover worktrees can be reclaimed without waiting for the sweep.

A3 and A4 share one safety predicate (`automationRunCleanupSafety`) so the scheduled
and on-demand paths cannot disagree about what is safe to delete.

**The invariant both A3 and A4 turn on: a continuity thread's runs share identity.**
Every occurrence in a thread reuses the binding's session id, and the worktree path is
keyed on session id alone, so a thread's oldest terminal run and its live current run
resolve to the *same* directory. Its ticket is shared too, and `tickets.automation_run_id`
is written once at thread creation and never updated, so it points permanently at the
thread's *oldest* run — exactly the run a sweep prunes first. Both are load-bearing:
deleting an old run's worktree or its row destroys the live thread, and both failures are
permanent (the next occurrence fails closed with "reviewer continuity worktree is missing"
or "continuity origin run missing" and never recovers). So a still-bound thread's origin
run is never prunable, and `automationRunCleanupSafety` blocks on the binding — not just
on a live session row, since a thread routinely outlives its session row. The guide
records this as durable domain context.

**A5 — and a thread must eventually end, or that guard leaks forever.** The binding is what
keeps a thread's worktree alive, so something has to release it. Nothing did: bindings are
keyed per subject (for a reviewer automation, one row per pull request), the only per-subject
reaper fires once at edge-inactive time and no-ops while the ticket is still there, and
`Store.SweepExpiredTickets` — written for a 30-day production TTL — **had no callers anywhere
in the repo**. Left alone, the guard above would pin one worktree per reviewed PR forever and
make both A3 and A4 permanent no-ops for exactly the automation type this epic is built
around. Three changes close it:

- `SweepExpiredTickets` now cascades the continuity-binding delete inside its existing
  transaction (it already cascades to `automation_ticket_occurrence_events`), so a binding
  dies exactly when its ticket does. That is preferred over retrying the per-subject reap:
  a binding whose ticket is gone can resume nothing, so tying its lifetime to the ticket's
  removes the ordering hazard instead of adding a second reaper that has to run later and in
  the right order.
- A dedicated hourly ticket-retention sweep actually runs it (30-day TTL, both knobs
  env-overridable), mirroring the two sweep loops already in the daemon.
- The A1-fix safety net had to be scoped at the same time — see A1-fix above. It refused
  whenever *any* same-contract prior thread's ticket was missing, and a TTL sweep makes
  tickets missing by design, so wiring up the sweep without that change would have refused
  every re-request for a subject 30 days after its thread closed: a brand-new permanent
  brick of exactly the kind this PR exists to remove.

Net effect: the binding stops being a permanent pin — it now releases when its ticket ages
out. Hard-deleting closed tickets after 30 days is user-visible beyond automations, so it
gets its own CHANGELOG bullet.

**Scoping this honestly: the binding is one of two independent pins, and this PR bounds
one.** `automationRunCleanupSafety` checks a *live session row* before it checks the
binding, and that check is bare row existence (`store.Get(sessionID) != nil`), not PTY
liveness. Both pins have to lift for a worktree to be reclaimable, so ageing out the
ticket does nothing while the session row survives — and automation-delivered agent
sessions do not reliably end today. On the `automations` profile right now, 35 of 54
terminal runs are held by the session pin, 30 of them by sessions still sitting in
`launching`. Nothing here regresses that; it was already true before this PR and is
unchanged by it. But it means the accurate claim is "the binding no longer pins
forever", not "worktree growth is now bounded" — bounding it fully needs automation
session lifetime addressed too, which is a separate change and out of scope here. The
guide records both pins as domain context so the next reader does not have to
rediscover the second one.

**Protocol 176 → 179.** 177 adds `automation_delete`, 178 adds `automation_cleanup`,
179 adds the `cleaned` / `kept_dirty` / `kept_active` fields on
`AutomationActionResultMessage`. Both delete and cleanup are dual-surfaced over
WebSocket and the unix socket, with matching `internal/client` methods and CLI
subcommands.

## Live evidence

New packaged serial scenario `real-app:scenario-automation-lifecycle`, registered in
`scenarioCatalog.mjs` and the serial matrix. Run
`automation-lifecycle-2026-07-20T21-53-29-423Z`, `failureCount: 0`, against the app
built from `458455f8` on the isolated `automations` profile (preflight green, CLI/app/
daemon all on protocol 179). Three legs:

1. **edit-rebind** — a scheduled singleton delivers under prompt P1; editing to P2
   delivers on a fresh ticket *and* session with run 1's row unchanged; reverting to P1
   delivers on a third distinct thread. That third delivery is the live proof of the
   A1-fix.
2. **delete-resurrect** — run-now via the panel bridge, then delete via CLI: the panel
   drops the definition and `deleted_at` is set in the DB; re-applying the same YAML
   brings it back with its run history.
3. **cleanup-dirty-safe** — three independent worktrees from one definition, produced by
   delivering and then editing the prompt to rotate the binding twice: one clean and
   rotated away, one dirtied and rotated away, one still bound to the current thread with
   its session deliberately closed first. A single `attn automation cleanup` partitions
   them into `cleaned` / `kept_dirty` / `kept_active`; the clean one is gone from disk and
   from `git worktree list`, the dirty one survives with its file, the bound one survives
   on the strength of its continuity binding alone, all three run rows are unchanged, and
   a second invocation is a no-op for the already-cleaned run.

   The third worktree is the live regression proof for the shared-identity invariant: with
   its session row gone, the pre-fix code saw an ordinary terminal run and deleted the live
   thread's working directory.

Retention (A3) is deliberately not in the scenario — its policy is time-based and fully
unit-tested. That exclusion is stated in a comment at the top of the scenario file.

**The ticket TTL sweep (A5) is out of that scenario's reach too** — a 30-day TTL on an hourly
tick — so it was verified separately against a live daemon on the same build, on the same
isolated profile, with `ATTN_TICKET_RETENTION_SWEEP_INTERVAL=5s` and the TTL left at its 30-day
default. A 40-day-old closed ticket and the continuity binding pointing at it both disappeared
on the first tick, with `ticket retention sweep: removed 1 expired ticket(s)` in the profile
daemon log. That is the whole chain the fix depends on — ticket expires → binding released →
session no longer pinned — exercised end to end against real SQLite rather than a test double.

## Size

4325 added lines, but only **1185 are product code**: 2214 are tests, 748 the packaged
scenario, 100 generated types, 78 docs.

## Local verification

`go build ./...`, `go vet ./...`, `go test` across daemon/store/automation/protocol/
client/cmd, `pnpm --dir app exec tsc --noEmit`, and `pnpm --dir app test`
(175 files / 1889 tests) all pass.

## Notes for review

- **Only `repository_worktree` locations ever have worktrees.** `ResolvedLocation.Worktree`
  is never populated for `directory`-location automations, so retention and cleanup have
  nothing on disk to reclaim for those — by design, not a gap.
- Automation worktrees are never registered via `store.AddWorktree`, so removal goes
  through `git.DeleteWorktree` directly rather than `Daemon.doDeleteWorktree`'s
  registry-aware path, which would be a no-op discovery fallback for them.
- The guide's Slice 7 checklist line is split in this PR: the lifecycle half is checked
  with the citation above; the editor half stays unchecked for PR B.
- **Three tests are renamed with inverted assertions, deliberately.**
  `TestAutomationApplyRevertRefusesWhenOldTicketWasSwept` and two siblings asserted "refuse
  if *any* prior same-contract thread's ticket is missing". That invariant is over-broad, and
  once the ticket TTL sweep exists it would permanently brick every revert to a contract whose
  original thread had aged out — for no safety gain, since the reverted thread holds a freshly
  reserved session and reuses nothing of the old one. They are corrected forward rather than
  the fix being weakened to keep them green. The genuine hazard they were reaching for is now
  pinned by a stricter new test,
  `TestValidateAutomationContinuationRefusesOnlyForItsOwnVanishedTicket`, which drives the real
  claim path so the live binding supplies the inherited ids, calls `SweepExpiredTickets`, and
  asserts the refusal still fires when the vanished ticket is the request's *own*.
- An adversarial review pass late in this branch found two permanent-brick bugs, both
  from treating a run as privately owning its worktree and rows; the fixes and their
  regression tests are the last product commits here. Re-running the packaged scenario
  against the fixed code then surfaced a third, smaller gap — cleanup was skipping
  still-in-use worktrees silently — which is what `kept_active` and protocol 179 address.
  A second adversarial pass over that delta then found the leak in A5: the guard those
  fixes introduced was correct, but nothing ever released a binding, so it would have
  pinned every reviewer worktree forever. The evidence cited above is from the final head;
  only docs commits sit on top of the build it was captured from (`458455f8`) — the guide
  ledger, and a correction that scopes the lifetime claim to the one pin this PR actually
  bounds. No product, test, or scenario code changed after the evidence run.


figgyster (approved):
This is the lifecycle half of the Automations arc growing up: edits rotate only the continuation contract, deletion retires the durable edges without erasing forensic history, and retention/cleanup finally distinguish a genuinely dead worktree from a thread still using one.

The late correction on the TTL claim matters, and the code agrees with it. Sweeping an expired terminal ticket releases the continuity binding, but cleanup still protects any live session row; this fixes the forever-pin without pretending the remaining session-lifetime leak vanished in a fog bank.

I traced the shared-worktree guards through the sweep and on-demand cleanup, the soft-delete/resurrection path, and the contract snapshots. The regression coverage pins both the continuity origin and later shared-worktree runs. Current head is green across daemon, frontend, all E2E shards, release preflight, and the benchmark run. Approved.

PR #623, opened 2026-07-21
feat(automations): self-service YAML editor and validate-without-apply

victorarias:
Slice 7 PR B, the last implementation PR of the automations epic. Automations become writable from the app: the Automations panel's New and Edit buttons open the definition's YAML in a CodeMirror buffer, Validate checks a definition without storing anything, and Save is the only writer.

## What this adds

**One buffer for create and edit.** Create loads through the same `getDefinition('')` path as edit and gets the starter template back at revision 0, so there is no second code path to keep in step. A successful create adopts the id the daemon actually persisted, so a second Save in the same buffer is an edit rather than a stale-id create.

**Validate without apply.** `automation_validate` runs the exact same definition checks `apply` runs — they share one seam in `internal/daemon/automations.go` rather than drifting — and writes nothing. Also exposed as `attn automation validate --file <path>`.

**Three refusals instead of silent damage:**

- Changing the `id` of the definition you are editing. Apply is keyed on the id *inside* the YAML, so this would leave the original running and quietly create a second one. A rename has to be an explicit create.
- Creating an automation whose `id` already belongs to a live definition — that would replace it wholesale. A soft-deleted definition is still resurrectable by re-applying its id, so the guard checks `DeletedAt`, not mere existence.
- Saving over a definition whose revision changed elsewhere. This one offers Reload as the recovery path.

**Comments survive.** Definitions now store the applied `SpecYAML` verbatim (migration 75), so an edit round-trip preserves comments and formatting. Rows applied before this migration have no stored YAML and re-render comment-free from `SpecJSON` on first open — the editor says that in-line rather than hiding it.

**No stomping.** The buffer is populated only on mount or an explicit Reload. It does not subscribe to the automations store, so an `automations_changed` broadcast cannot replace text you are mid-typing.

## What an adversarial gate then found

A six-dimension adversarial review (find → refute, majority kills a finding) produced 13 raw findings; 2 were refuted 3/3 and the rest collapsed to six distinct defects. Five are fixed here, each with a mutation-verified regression test. Three of them were the same mistake wearing different clothes — **trusting a value another writer can move underneath you.**

**The revision counter did not cover `spec_yaml`** (`649a7e52`). Comments live only in `spec_yaml`; they are never re-derived into the canonical `spec_json`. So a comment-only save wrote new YAML and left `revision` untouched, and the stale-save guard was structurally blind to precisely the class of edit this editor exists to enable. Two people editing one definition silently lost one author's comments, with no "changed elsewhere" refusal.

That fix makes `revision` move more often, which means comment-only edits now also reach `rotateContinuityBindingsIfContractChanged`'s revision-differs branch. Worth stating explicitly because it would have been a real regression: that branch compares continuation *contracts*, and `ContinuationContract` is `{Prompt, Launch, Location}` with no revision in it, so editing a comment does not rotate a singleton thread.

**An edit-Save silently resurrected a definition deleted elsewhere** (`649a7e52`). `DeleteAutomationDefinition` also leaves `revision` alone, so a stale editor's `expected_revision` still matched a soft-deleted row: the guard passed, the upsert cleared `deleted_at`, and the definition came back live and enabled — after `automationDelete` had already failed its pending runs, fenced its provider cursors and purged its continuity bindings. For a cron trigger that means unattended sessions firing again, reported to the user as a successful save. The edit path now refuses and points at New; the create path still resurrects deliberately.

**A stale Validate response overwrote the cleared state**, rendering "Looks good." against text the daemon never saw — and because clearing validation on keystroke also re-opened the disabled Validate button, two requests could stack and resolve out of order with the older answer winning.

Plus two smaller ones: edits typed during an in-flight Save were discarded when the editor closed, and `attn automation validate` printed `null` on success because `json.Marshal` of a nil payload yields the literal `null`, which defeats `omitempty` and leaves the client decoding a non-nil four-byte `Data`.

The gate also caught a false claim in this PR's own guide ledger — it said all three of the earlier fixes carried regression tests, and the read-loop one does not. Corrected rather than backfilled with a test that would only assert Go's scheduler.

## Known gap, deliberately not fixed here

`enabled` is both a field in the definition spec and a column the panel's toggle writes independently. Disable an automation with the toggle, then open and Save it, and the YAML's `enabled: true` turns it back on. This is pre-existing in `attn automation apply`, but this PR puts Edit directly beside the toggle and makes it one click, so it is newly reachable. Every available fix changes product behavior — rewrite the user's YAML from the toggle, make the column authoritative and drop the spec field, or warn on divergence — so it wants a decision rather than a unilateral patch.

## Review pass on the product code

Reviewing the diff directly rather than trusting the implementation reports turned up three defects the tests had not caught. Each is a separate commit with a regression test:

- `d2ca16a7` — Reload was offered while creating. `loadedId` is null there, so it would have re-fetched the starter template *over the draft being typed*, for no reachable benefit.
- `2bc29c15` — validate ran inline on the client read loop. "Does not touch the store" is not the same as "is fast": `git.ValidateLocalClone` stats the path and shells out to git twice per override, head-of-line blocking that client's other messages.
- `14e0ab6e` — a create whose id collided with a live definition silently overwrote it. Mutation-verified: without the guard the create succeeds and bumps the victim to revision 2.

One test was also vacuous and is now not. `AutomationEditor.test.tsx`'s editor mock called `onChange` unconditionally, while the real `applyExternalContent` computes a minimal edit and **dispatches nothing when the text is unchanged**. With the mock that permissive, deliberately breaking `setText` left all 17 tests green. The mock now matches the real semantics, and there is a test for the unchanged-text case specifically; the same mutation now goes red.

## Evidence

Packaged scenario `real-app:scenario-automation-editor`, run `automation-editor-2026-07-21T00-46-50-361Z`, against the app built from `c5629930` on the isolated `autoedit` profile. All nine legs green, with the daemon's own state cross-checked through the bundled CLI at every leg:

```
step:ok leg1_starter_template
step:ok leg2_invalid_rejected_nothing_stored
step:ok leg3_create_with_comment
step:ok leg4_create_collision_refused
step:ok leg5_comments_survive_roundtrip
step:ok leg6_id_change_refusal
step:ok leg7_stale_revision_refusal_and_reload
step:ok leg8_comment_only_edit_bumps_revision
step:ok leg9_delete_elsewhere_then_save_refused
ATTN_VERDICT {"ok":true,"scenarioId":"AUTOMATION-EDITOR","runId":"automation-editor-2026-07-21T00-46-50-361Z","failureCount":0,"firstFailure":null,"durationMs":2838}
```

Legs 8 and 9 are the ones worth reading. The original seven all have a single writer, which is structurally why they could not catch any of the three stale-value defects above. Leg 8 builds its edit as a pure prefix of the live buffer, so "comment-only" is true by construction, and asserts `spec_json` is byte-identical before and after — otherwise a future refactor that folded comments into the canonical spec would leave the leg passing for the wrong reason. Leg 9 asserts the daemon's own state after the refusal, not just the error string, because a polite refusal sitting in front of a resurrected row is exactly the failure it exists to catch.

Comment survival is proven by the stored `SpecYAML` still carrying the run's `# harness-marker:` line after the round-trip, read back through the CLI.

The CLI fix was also confirmed against the running daemon rather than only in a unit test:

```
$ attn automation validate --file valid-auto.yaml
{
  "valid": true
}
```

Preflight PASS on `autoedit`, protocol 180 agreeing across CLI, app, and daemon. `go build ./...` clean, `go test ./...` exit 0 across 42 packages, `tsc --noEmit` clean, frontend 1917 tests passed.

The editor buffer is exposed to the UI automation bridge through a registered handle (`automationEditorAutomation.ts`) rather than DOM scraping, because CodeMirror virtualizes long documents — scraping `.cm-line` can silently truncate, which would be unacceptable in the seam that carries the comment-survival proof.

## Size

3921 insertions across 38 files, but the product surface is much smaller: the packaged scenario (593), `AutomationEditor.test.tsx` (553), `ws_automations_test.go` (452), and the e2e spec (134) are ~1.7k of it, and generated protocol output another ~120.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,14 +13,15 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   has New and Edit buttons that open the definition's YAML in an editor, so
   creating an automation no longer means dropping to a terminal. Validate
   checks a definition without storing anything and reports the first problem
-  it finds; Save stores it. Two mistakes are refused rather than silently
+  it finds; Save stores it. Three mistakes are refused rather than silently
   accepted: changing the `id` of a definition you are editing (which would
-  leave the original running and quietly create a second one), and saving over
-  a definition that changed elsewhere since you opened it — the latter offers a
-  reload. Your YAML is now stored as you wrote it, so comments and formatting
-  survive an edit round-trip; definitions applied before this release are
-  rendered from their stored form the first time and keep their new text
-  afterwards.
+  leave the original running and quietly create a second one), creating a new
+  automation whose `id` already belongs to one you have (which would replace
+  it wholesale), and saving over a definition that changed elsewhere since you
+  opened it — the last offers a reload. Your YAML is now stored as you wrote
+  it, so comments and formatting survive an edit round-trip; definitions
+  applied before this release are rendered from their stored form the first
+  time and keep their new text afterwards.
 - **Automations now have a panel in the app.** A new toolbar button opens an
   Automations panel that lists every definition with its trigger and
   enabled state, lets you enable/disable or run a manual automation now,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,14 +9,27 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-20]
 
 ### Added
+- **You can now write and edit automations in the app.** The Automations panel
+  has New and Edit buttons that open the definition's YAML in an editor, so
+  creating an automation no longer means dropping to a terminal. Validate
+  checks a definition without storing anything and reports the first problem
+  it finds; Save stores it. Two mistakes are refused rather than silently
+  accepted: changing the `id` of a definition you are editing (which would
+  leave the original running and quietly create a second one), and saving over
+  a definition that changed elsewhere since you opened it — the latter offers a
+  reload. Your YAML is now stored as you wrote it, so comments and formatting
+  survive an edit round-trip; definitions applied before this release are
+  rendered from their stored form the first time and keep their new text
+  afterwards.
 - **Automations now have a panel in the app.** A new toolbar button opens an
   Automations panel that lists every definition with its trigger and
   enabled state, lets you enable/disable or run a manual automation now,
   shows each definition's run history with failures called out inline, and
-  jumps straight to the ticket or session a run produced. Definitions are
-  still created from the CLI (`attn automation apply`); changes made there
-  appear in the panel immediately, and daemon rejections are shown in the
-  panel rather than silently swallowed.
+  jumps straight to the ticket or session a run produced. Changes made from
+  the CLI (`attn automation apply`) appear in the panel immediately, and
+  daemon rejections are shown in the panel rather than silently swallowed.
+- **`attn automation validate --file <path>`** checks a definition without
+  applying it, using the exact same checks `apply` runs.
 - **Automations can now run on a schedule.** A definition can declare a cron
   schedule with a required time zone and fire an ordinary visible agent
   session at each intended instant — for example, a recurring merged-worktree
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,12 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   leave the original running and quietly create a second one), creating a new
   automation whose `id` already belongs to one you have (which would replace
   it wholesale), and saving over a definition that changed elsewhere since you
-  opened it — the last offers a reload. Your YAML is now stored as you wrote
-  it, so comments and formatting survive an edit round-trip; definitions
+  opened it — the last offers a reload. If the definition was *deleted* while
+  you had it open, saving is refused rather than quietly bringing it back, so
+  an automation you turned off does not start running again behind you.
+  Your YAML is now stored as you wrote it, so comments and formatting survive
+  an edit round-trip — including against someone else editing the same
+  definition, since a comment-only change counts as a change; definitions
   applied before this release are rendered from their stored form the first
   time and keep their new text afterwards.
 - **Automations now have a panel in the app.** A new toolbar button opens an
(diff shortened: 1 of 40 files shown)


figgyster (changes requested):
You’ve spent this whole automations run making durable state tell the truth under retries, relaunches, deletions, and slow delivery. This editor is the operator’s hand on that machinery. I like the refusal work here. But it leaves one old split-brain wire live, and it can wake an unattended automation after the operator turned it off.

Required before merge:

- `internal/store/automations.go:132` writes `enabled` from the saved YAML, while the panel toggle changes only the database column at line 220. Disable an automation in the panel, then open it and save an unrelated prompt/comment edit: the stored YAML still says `enabled: true`, so this update re-enables it and clears/fences provider state as a real enable transition. The editor sits beside the toggle, so this is now a one-click path to an unexpected cron or review-triggered session. Pick and enforce one authority or add an explicit, tested reconciliation/confirmation flow; the save path must never silently reverse the toggle.

The earlier Daemon run also died in `gofmt` on the migration-75 comment in `internal/store/sqlite_test.go`; the later rerun is still working its way through the rain. Fix that formatter witness while you close the state split.

figgyster (changes requested):
That new selection-mapping E2E witness is the right repair for the runner-dependent scroll assertion. Good. But this push never touches the blocker: `enabled` is still split between the panel’s database-only toggle and the YAML the editor later saves.

Required before merge:

- `internal/store/automations.go:132` will still write the stale YAML `enabled` value over a panel disable. Disable an automation, open it, make an unrelated edit, and Save silently starts it again. This can launch scheduled or review-triggered work the operator explicitly stopped. Make the authority/reconciliation explicit and pin that exact sequence with a regression.

figgyster (commented):
The split-brain wire is gone. `SetAutomationEnabled` now treats the panel's current-management state as authoritative, writes it through to both stored representations, and bumps the revision so an editor already open cannot put the old value back. The new daemon-path regression runs the exact sequence that mattered: toggle off, reopen through the editor's own read path, save, and the automation stays off with the hand-written comment intact. That's the witness I wanted.

I’m dropping my code objection on b29f6d03. The Daemon, Frontend, every E2E shard, and both Linux preflight builds are green. Bench A/B is still running, so I’m leaving the final green light with the precinct lab for the moment.

figgyster (approved):
The last lab report is in: Bench A/B is green alongside Daemon, Frontend, all E2E shards, and both Linux preflight builds.

This closes the Automations editor chapter the right way. The panel now owns current enabled state, the YAML editor preserves the operator’s source while mirroring that state, and the revision fence stops stale drafts from reviving unattended work. I checked the exact toggle → reopen → save witness that blocked this PR. It holds. Approved.

PR #624, opened 2026-07-21
test(store): pin the pre-automations upgrade and record real recovery coverage

victorarias:
Closes out the automations epic's final verification matrix. Two files: one new test, one guide update.

## Upgrade — the gap this closes

Migration 73 does `ALTER TABLE tickets ADD COLUMN automation_run_id`. Until now that statement had only ever run against an empty or already-migrated `tickets` table — no test seeded a populated pre-automations database and migrated it forward. That's the statement most likely to fail on a real user's install.

`TestMigratesPopulatedPreAutomationsDatabaseToHead` seeds tickets, a session, and a PR, rewinds to the genuine pre-73 shape, migrates, and asserts every legacy row survives intact alongside the new column and the seven automation tables.

The rewind drops the automation tables, the `idx_tickets_automation_run` index, and the `automation_run_id` column — deliberately not just `DELETE FROM schema_migrations`, which would leave 73/74 hitting their `IF NOT EXISTS` no-op path and prove nothing.

**Evidence.** A database built by the *pre-epic binary* (`1e300bf7`, last commit before migration 73) rather than by rewinding head code:

```
pre-epic binary (1e300bf7) -> schema_head=72  automation_tables=[]
head binary     (310fbc8e) -> schema_head=75  automation_tables=[7 tables]
```

**Mutation-verified.** No-oped migration 73's `ALTER` and its dependent index so the migration still reports success. The test fails at the per-ticket integrity assertion — `sqlite_test.go:417`, `no such column: automation_run_id` — not at an upstream diagnostic. `GetSchemaVersion` still passed under the mutation, confirming the failing assertion is the one carrying the claim.

## Failure recovery — documented, not closed

The guide's nine-checkpoint partial-failure table was a specification that never became evidence. Surveyed on merged `main`: **one covered, four partial, four uncovered**. `EnsureWorkspace`, `EnsurePane`, `VerifyDelivery`, and `automationSessionIsLive` have zero test callers outside `fakePorts` stubs — verified directly, not just reported.

"Partial" here means the test re-invokes a function and asserts key idempotency; it never restarts a daemon or resumes from mid-sequence durable state, which is what that section actually asks for.

Victor's call was to ship and document rather than close the gap, so the guide now carries the honest per-checkpoint table with residual risk ranked. Highest: checkpoint 8 (a run can strand in `pending` forever, or mark delivered on mismatched links — `internal/daemon/automations.go` admits in a comment that this path is unreachable from unit tests) and checkpoint 7 (duplicate live agent session). Both fail silently.

No production code changed.

## Still outstanding

The packaged-app serial matrix has **not** run. macOS revoked the Accessibility grant for `app/scripts/real-app-harness/.build/attn-real-input-driver` after it was rebuilt, so every input-driven scenario dies at `AXIsProcessTrustedWithOptions`. An attempted run failed 8 of 14 scenarios entirely for that reason — seven with the explicit permission error, one timing out on an undelivered keystroke — **no product regression among them**. The guide records this and the final checklist item stays unchecked until the matrix actually runs.

### DIFF
--- a/docs/plans/2026-07-18-attn-automations-implementation.md
+++ b/docs/plans/2026-07-18-attn-automations-implementation.md
@@ -918,6 +918,37 @@ upgrade and packaged-app matrix pass from the integration branch.
 
 ## Verification strategy
 
+### Upgrade
+
+Run at epic close (2026-07-21) against merged `main` `310fbc8e`.
+
+A database was built by the **pre-automations binary** — `1e300bf7`, the last
+commit before migration 73 — rather than by rewinding head code, so the starting
+schema is what an installed pre-epic app actually produced:
+
+```
+pre-epic binary (1e300bf7) -> schema_head=72  automation_tables=[]
+head binary     (310fbc8e) -> schema_head=75  automation_tables=[7 tables]
+```
+
+Migration 72 → 75 applied cleanly and wrote its pre-migration backup first. A
+copy of the real production database was not used: it sits at version 73, and
+reading `~/.attn` is deliberately out of reach from this workflow.
+
+`TestMigratesPopulatedPreAutomationsDatabaseToHead`
+(`internal/store/sqlite_test.go`) pins this as a regression test, because none
+existed. It seeds tickets, a session, and a PR, rewinds to the pre-73 shape
+(dropping the automation tables, the `idx_tickets_automation_run` index, and the
+`tickets.automation_run_id` column — not merely deleting `schema_migrations`
+rows, which would leave 73/74 hitting their `IF NOT EXISTS` no-op path and prove
+nothing), then migrates and asserts every legacy row survives intact alongside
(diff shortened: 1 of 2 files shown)


figgyster (approved):
Victor, this is the kind of final matrix note I trust: the unbuilt bridge is marked on the map instead of getting painted over. The migration witness finally forces v73's ticket ALTER across populated legacy rows, and it proves those rows plus the new null provenance link make it through alive.

The ledger turns the #614/#619/#622/#623 durable-coordination arc into honest debt. Checkpoint 7's duplicate-session window and checkpoint 8's session-to-run-link gap are named for the next pass, not waved away as coverage. The focused regression, daemon CI, both Linux preflights, and the relevant E2E check are green. Safe to merge. Are 7 and 8 the first return trip, or is the Automations epic resting a while?

PR #628, opened 2026-07-21
refactor(automations): inline workdelivery and split the daemon god file

victorarias:
## Summary
This is PR1 of the [automations v2 simplification plan](docs/plans/2026-07-21-automations-v2-simplification.md). It's a mechanical refactor with **no behavior change**:

- Deletes `internal/workdelivery`: the 7-port `Ports` interface and 51-line `Service.Deliver` pass-through added indirection without depth. Its body is inlined verbatim as `Daemon.materializeAutomationRun`, with identical error wrapping. The only call-site change is `(workdelivery.Service{Ports: d}).Deliver` → `d.materializeAutomationRun`.
- The mock-only `workdelivery` step-order test is dropped deliberately — it only exercised a fake, not real behavior.
- Splits `internal/daemon/automations.go` (1,587 lines) by concern into `automations_definitions.go`, `automations_github.go`, `automations_deliver.go`, and `automations_recover.go`; `automations.go` keeps the command handler and shared helpers (176 lines).

Verbatim moves were verified by a scripted diff of function bodies against the pre-refactor file.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/daemon/... ./internal/store/...` — green
- [x] `gofmt -l` / `go vet` clean
- [x] Live-verified on the dev profile: manual automation applied and the run reached `delivered` through `materializeAutomationRun`; a real session spawned and stopped cleanly


figgyster (approved):
You finally took the knife to the Automations machinery instead of teaching everyone to live around it. This is the right first move for the v2 rebuild: `workdelivery` had seven ports and no judgment, while the actual delivery contract still lived in the daemon.

I traced the ticket-first materialization path, recovery, GitHub observation, and the continuity/withdrawal guards after the split. They still meet at the same seams, and the fake-only step-order test is a fair burial. The focused daemon/store suite, formatting check, and all relevant CI/preflight jobs are green. Bench A/B is still running, but this is a no-behavior refactor and its trend report is not holding the door.

Safe to merge. The v1 fixes taught us that Automations needs explicit state before more surface area; this plan starts by making the existing state machine readable enough to replace without losing a limb.

PR #629, opened 2026-07-21
refactor(automations): make enabled column-only and drop spec_yaml storage

victorarias:
PR2a of the [automations v2 simplification plan](docs/plans/2026-07-21-automations-v2-simplification.md).

## What changed

- **Clean-slate migration 76**: wipes every automation table and recreates
  `automation_definitions` without `spec_yaml`, nulling
  `tickets.automation_run_id`. This is safe because the feature has zero
  production usage — `~/.attn/attn.db` is pinned at migration 73; the v1
  automation migrations never shipped to a real production DB.
- **`enabled` gets a single authority: the column.** The spec loses the
  `enabled` field entirely; YAML carrying `enabled:` is rejected with a
  directed error pointing at the new `attn automation enable|disable` CLI
  verbs. Apply never toggles `enabled`; insert/resurrection default to
  enabled.
- `SetAutomationEnabled` no longer rewrites `spec_json`/`spec_yaml` and no
  longer bumps `revision` (revision guards spec content; `enabled` is no
  longer spec content). This deletes the #623 split-brain class by
  construction: `automationSpecWithEnabled` and `SetEnabledInYAML` are gone,
  along with the toggle's spec-rewrite path (net -171 lines).
- New `attn automation enable|disable` CLI verbs as the sole way to flip
  `enabled`.

## Testing

- `TestAutomationSetEnabledRoutesThroughSocketDispatch` pins the daemon's
  real unix-socket dispatch path (`handleConnection` → `ParseMessage` →
  `switch` → `handleAutomationCommand`), not just the handler function
  directly. This is mutation-checked: removing the
  `CmdAutomationSetEnabled` case from the dispatch switch reproduces the
  exact "daemon error: unknown command" regression caught live during
  manual verification — every other `automation_set_enabled` test calls
  `handleAutomationCommand` directly and would have stayed green through
  that regression.
- Live-verified on a throwaway dev profile: migration applied over a real
  v75 dev DB; disable survives re-apply with `revision` unchanged;
  run-while-disabled is refused; enable → run delivers.
- `go test ./...` green.

🤖 Posted on behalf of Victor.


figgyster (dismissed):
You took the #623 split-brain and buried the whole body this time. Good. The v2 rebuild is starting the way the previous Automations run kept begging it to: fewer representations, fewer reconciliation stories, one switch that means one thing. Migration 76 also tells the truth about unused v1 state instead of dragging its furniture into the new house.

The SQL migration coverage, the YAML rejection/reapply path, and the daemon/store suite all hold. I also ran the full Go suite here and CI plus both Linux daemon preflight jobs are green.

I can't close this from the Linux side, though. This changes daemon persistence and the real unix-socket command path, and the repository requires a non-production macOS live check for that class of work.

Required before merge:

- Post the dev-profile proof on this head: preflight, then apply an automation, disable it through `attn automation disable`, fetch/reapply its YAML, and show it remains disabled. That is the exact grave #623 left behind; let's watch the dirt stay put once in the real daemon.

figgyster (approved):
That was the witness I needed. Your dev-profile run hits the exact #623 failure shape — disable, fetch a spec with no `enabled` hitchhiker, reapply it, and watch the automation stay down. The old red flag is retired.

This is the right first cut of v2: migration 76 clears unused v1 machinery, the column owns the switch, and the spec stops pretending it has a vote. Full Go tests, CI, preflight, Bench A/B, and the current-head macOS proof all agree. Safe to merge. The bigger state-machine rebuild is still waiting in the alley, but this removes one of its favorite hiding places.

PR #630, opened 2026-07-21
refactor(automations): v2 state model — explicit run states, append-only continuity bindings, trigger-implied policy

victorarias:
Consolidated PR2b of the automations v2 plan (`docs/plans/2026-07-21-automations-v2-simplification.md`) — merges what was originally scoped as PR2b/PR3/PR4 into one migration and one review round. Follows #628 (mechanical split) and #629 (enabled column-only).

## What changes

**Migration 77 (clean slate — feature has zero production usage; prod DB is at 73, dev at 76):**
- `automation_runs`: adds `cancel_reason` and `attempts`; run states are now `pending | delivered | failed | cancelled` with `cancel_reason ∈ {review_withdrawn, definition_disabled, definition_deleted}`. The `AutomationReviewWithdrawnError` failure-sentinel string is deleted.
- `automation_continuity_bindings`: append-only — surrogate `id`, `status active|released`, `released_reason ∈ {contract_rotated, ticket_swept, definition_deleted}`, `released_at`, with a unique partial index enforcing one active row per `(definition_id, continuity_key)`. Rows are never deleted; `hasPriorAutomationContinuityRun` run-history replay is deleted — delivery reads binding status directly.
- `automation_review_request_edges`: drops `accepted_cycle`. Candidacy is now "no run exists for (subject, cycle)"; pending runs are re-delivered on each observation, which kills the class of bug where a retryable failure stranded a run until daemon restart (#608).

**Policy block deleted:** `trigger.type: scheduled` absorbs `continuity` (`fresh|singleton`) and `catch_up` (`skip|latest`) directly; manual is always fresh, github review triggers are always per-subject+latest. A spec containing `policy:` fails parse with a directed error.

**Engine seam:** `internal/automation` gains `ResolveContinuation` against a `BindingStore` interface (dependency-inverted; the daemon adapts `*store.Store`). It decides fresh/continue/self-heal from binding status alone.

## Two bugs caught in review (both mutation-checked)

1. **First-run binding self-release.** The initial `ResolveContinuation` treated a run's own just-created binding (whose ticket doesn't exist *yet* — ticket IDs are reserved at claim, tickets created during delivery) as dangling and released it at birth, silently breaking singleton continuity for every subsequent occurrence while all tests stayed green. Fixed with an `ownTicketID` parameter; regression-pinned by `TestScheduledSingletonSecondOccurrenceContinuesFirstOccurrencesThread` (real store, two full deliveries — reverting the fix makes it fail).
2. **Ticket sweep hard-deleted binding rows** (v1 leftover violating the append-only invariant). `SweepExpiredTickets` now releases (`status=released, reason=ticket_swept`) in the same transaction. `TestSweepExpiredTicketsReleasesActiveContinuityBindings` asserts via direct SQL that the row survives; reverting to DELETE fails it.

## Live verification (dev profile, macOS)

See the evidence comment below: migration 77 over the real dev DB, `policy:` rejection, manual run → `delivered`, and the load-bearing scheduled-singleton leg across two real cron ticks.

https://claude.ai/code/session_01APU1R83oaVeshfJM5W5QCM


figgyster (dismissed):
You took the lesson from #623 and #629 seriously: state that matters gets one owner, and state that ends gets a gravestone instead of a detective reconstructing it from old run history. The append-only binding row is the part I read twice. Ticket sweep releasing that row before the next singleton claim gives the next cron tick a fresh identity, while the two-tick dev-profile proof shows the live thread still holds when it should.

This is the right next cut in the v2 rebuild. Explicit cancellation reasons, run-existence candidacy, and trigger-implied policy take a lot of v1 ceremony off the books without smuggling it back in as SQL folklore. Full Go tests, the focused state-machine suites, diff check, Linux CI/preflight, Bench A/B, and Victor's current-head macOS migration/manual/singleton witness all came back green. Approved. Case closed.

figgyster (approved):
The last two lines put the old v1 implementation guide where it belongs: history, with a signpost to the v2 plan that actually governs the next work. Good housekeeping. The runtime patch is unchanged from the prior approval, and the new head is documentation-only. Approved.

PR #631, opened 2026-07-22
Fix real-app matrix scenarios; scenario runner framework and migrations

victorarias:
## What

Makes the real-app matrix scenarios from #626 pass and overhauls the harness so running and inspecting failures is cheap:

- **Scenario fixes** (commit 1): remote observer `client_hello` handshake, tr-family relaunch/close/resize anchor seeding and geometry assertions, utility-session switch focus.
- **Framework** (commit 2): `createScenarioRunner` (step tracing, `failure.json`, `failure-digest.txt`, one-line `ATTN_VERDICT`), fresh-world preflight (quit app, stop daemon, kill leaked pty-workers; production targets refused, unit-tested guard), stale-session sweep at launch with demote-then-unregister for chief sessions, matrix digest with each failed scenario's output tail.
- **Migrations** (commit 3): all 18 ad-hoc scenarios moved onto the runner, behavior preserved 1:1; includes the ticket-lifecycle chief-demote teardown fix and the osc8 probe-server keep-alive fix.
- **Build** (commit 4): bun codesign guard no longer aborts on binaries without `LC_CODE_SIGNATURE`.

## Evidence

- Every migrated scenario live-verified green individually on a throwaway profile (`fxm1`), packaged app rebuilt from this branch.
- Two full uncontended 35-scenario serial-matrix passes, 29/35 each with rotating failures: the constant 3 are credential-blocked remote legs (ai-sandbox claude OAuth, keychain passphrase wedge); the rotating ones are pre-existing intermittent classes (close/unregister timeout seen only in full-matrix context, nudge-trigger real-agent nondeterminism, tr205-codex remote width degradation), each passing solo and in the alternate pass.
- Unit tests: 1935/1935 (180 files), including new fresh-world guard and verdict/digest tests.

https://claude.ai/code/session_01LuiYi429D8eqxyPNPunS6C

### DIFF
--- a/app/scripts/real-app-harness/bridge-remote-hub.mjs
+++ b/app/scripts/real-app-harness/bridge-remote-hub.mjs
@@ -1069,11 +1069,14 @@ Remote hub options:
     saveJson(path.join(runDir, 'session-ui-selected.json'), selectedSession);
 
     setStep('remote-pty-interaction');
+    const existingRemotePaneIds = new Set(
+      (observer.getWorkspace(remoteSessionId)?.panes || []).map((pane) => pane.pane_id),
+    );
     const utilityPane = await client.request('split_pane', {
       sessionId: remoteSessionId,
       direction: 'vertical',
     });
-    remoteUtilityPane = await observer.waitForUtilityPane(remoteSessionId, 30_000);
+    remoteUtilityPane = await observer.waitForUtilityPane(remoteSessionId, 30_000, existingRemotePaneIds);
     saveJson(path.join(runDir, 'remote-utility-pane.json'), {
       splitResult: utilityPane,
       pane: remoteUtilityPane,
--- a/app/scripts/real-app-harness/common.mjs
+++ b/app/scripts/real-app-harness/common.mjs
@@ -124,7 +124,84 @@ export async function captureScreenshot(driver, outputPath) {
   }
 }
 
-export async function launchFreshAppAndConnect(client, observer) {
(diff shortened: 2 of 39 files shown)


figgyster (dismissed):
You took the 35-scenario machine apart while it was still running, then gave it a notebook and a crime-scene kit. That is exactly the right direction for this harness. But the new sweep has one bad address check, and it can evict a session that was never part of the run.

Required before merge:

- `app/scripts/real-app-harness/common.mjs:169` treats any `directory.startsWith(root)` as owned by the harness, then demotes and unregisters it. A sibling such as `/tmp/attn-real-app-sessions-old/keep` matches the default `/tmp/attn-real-app-sessions` root and gets removed from the daemon. Use a path-component-aware containment check (`path.relative`, or an equality/`root + path.sep` boundary) for both realpath candidates, and pin that sibling-prefix case in the sweep tests. This cleanup path has permission to tear down chief sessions, so it needs a fence that holds.

figgyster (approved):
You took the 35-scenario machine apart while it was still running, then gave it a notebook and a crime-scene kit. This follow-up gives the cleanup crew the one thing it was missing: an actual property line.

The root-or-separator check fixes the sibling-path eviction I flagged, and the regression covers the exact bad address plus the realpath form. That lets fresh-world cleanup do its job without mistaking a neighbor for evidence. Frontend E2E is green on this head. Safe to merge. Case closed.

PR #633, opened 2026-07-22
OrbStack remote endpoint, tr205 probe legs, and the shell-pane kill-path fix

victorarias:
## Summary

Continues the live-app matrix stabilization from #631. Four arcs:

- **OrbStack remote endpoint**: remote scenarios (tr205, tr402-remote, tr502, tr504, bridge-remote-hub) now default to a local OrbStack VM `attn-remote@orb`, rebuildable from the checked-in `provision-orb-remote.sh` (`pnpm --dir app run real-app:provision-remote`), replacing the retired ai-sandbox host. `ATTN_HARNESS_REMOTE_SSH_TARGET` retargets all remote scenarios at once.
- **tr205 deterministic probe legs**: new hidden `attn _probe-tui` command (`internal/probetui`) mimics codex/claude TUIs with styles pinned to recorded VT vocabulary and two-sided mirror tests; `cmd/agent-mirror` re-captures fixtures. The matrix runs `tr205-probe-codex`/`tr205-probe-claude` without a live model; the real-agent legs become `soakOnly` (need credentials seeded in the VM).
- **Real product fix — pane width recovery** (found by the probe legs): `fitShouldBailAsSuspicious` refused grows from an already-suspicious model, so panes shrunk by a neighboring split stayed stuck narrow after the split closed. Fixed in `GhosttyTerminal.tsx` + unit tests.
- **Real product fix — shell-pane close / remote worker leak** (found via tr504): interactive `bash -l` ignores SIGTERM, so closing a shell pane stalled 10s to SIGKILL, and on remote endpoints the worker's in-RPC kill blew the 5s control-conn deadline and leaked the worker. `Session.kill` now escalates to SIGHUP after a 2s grace (SIGKILL backstop intact); `WorkerBackend.Kill` gets a 15s deadline. Shell panes close in ~2s. Three new real-PTY kill tests. tr504 reworked to model the real ⌘W flow (close_session closes only that session's pane by design — the scenario now closes the split shell pane explicitly and verifies remote worker exit).

Also: automation-lifecycle aligned with the #629 enabled-outside-spec rule; tr401 codex header assertion tolerates the CLI update banner.

## Verification

- `go test ./...` green; new kill tests pass with meaningful timings (SIGHUP path 2.1s, SIGKILL fallback, cooperative fast-exit).
- Frontend unit tests green including new `GhosttyTerminal.resizePolicy` cases; probe mirror tests enforce vocabulary in both directions.
- Live-verified on throwaway profile `fxm1` (full `make install`): tr504 passes solo (shell-close step completes in ~2.1s — the SIGHUP grace — remote worker processes 9→4→0→0) and in-matrix (8.2s vs prior 71.3s timeout).
- Full 35-scenario serial matrix: 32/35; all three fails classified from digests as known rotating flakes (close/unregister matrix-only class, codex-resume live-agent discretion, one ghostty-scroll anchor one-off that passed solo immediately after). Every scenario now has a green steady state.

https://claude.ai/code/session_01LuiYi429D8eqxyPNPunS6C

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -97,6 +97,7 @@ starts are the baseline.
   `ATTN_HARNESS_PROFILE` overrides it.
 - Production requires both `ATTN_HARNESS_PROFILE=` and `--run-against-prod`.
 - On failure, inspect captured pane text and native screenshots before diagnosis.
+- Remote scenarios target the local OrbStack VM (`attn-remote@orb`); provision with `pnpm --dir app run real-app:provision-remote`.
 
 ## Test safety
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-07-22]
+
+### Fixed
+- **Terminal panes no longer get stuck narrower than their pane after closing a
+  neighboring split.** Content now grows to fill the pane again; previously it
+  could stay wrapped at the old, narrower width until an unrelated layout change
+  triggered a redraw.
+- Closing a shell pane now hangs the shell up promptly instead of stalling ~10 seconds and force-killing it; remote pane/session close reliably confirms worker teardown instead of timing out mid-kill.
+
 ## [2026-07-21]
 
 ### Fixed
--- a/app/package.json
+++ b/app/package.json
@@ -18,6 +18,7 @@
     "real-app:repro-panes": "node scripts/real-app-harness/repro-older-pane-writability.mjs",
     "real-app:bridge-smoke": "node scripts/real-app-harness/bridge-smoke.mjs",
     "real-app:bridge-remote-hub": "node scripts/real-app-harness/bridge-remote-hub.mjs",
+    "real-app:provision-remote": "bash scripts/real-app-harness/provision-orb-remote.sh",
     "real-app:bridge-repro-shells": "node scripts/real-app-harness/bridge-repro-shell-return.mjs",
     "real-app:bridge-repro-existing-shells": "node scripts/real-app-harness/bridge-repro-existing-shells.mjs",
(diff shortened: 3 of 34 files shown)


figgyster (dismissed):
You’re taking #631’s harness cleanup and turning it into a usable witness: one rebuildable remote endpoint, deterministic terminal probes where credentials used to make the evidence slippery, and two product bugs caught in the act. I’m glad the fix for a stranded narrow pane is small, and the TERM → HUP ladder keeps a shell close from leaving the remote worker connection to die in a timeout alley.

But this PR changes PTY geometry, worker kill timing, and the packaged-app matrix, and `AGENTS.md` requires a current-head non-production live verification for all three. The case and PR conversation contain no dev-profile/preflight or packaged-app run from this head. CI and the focused sandbox witnesses are good company; they cannot prove the installed app and remote daemon made it through the actual loop.

Required before merge:

- Rebuild/run the dev profile on `14acfec03181ae044a619565bbaaf6eb4c9ae5a4`, run its bundled `attn preflight`, then attach the current-head packaged-app evidence for both TR-205 probe legs and the remote shell-pane close/worker-cleanup path. The `ATTN_VERDICT` records and the preflight result are enough; I don’t need a novel about them.

figgyster (approved):
That’s the witness I asked for. The rebuilt `fxm1` install identifies this exact head, bundled preflight has the app, CLI, and daemon on protocol 180, and both deterministic TR-205 legs plus TR-504 made the trip through the OrbStack endpoint. The old fingerprint mismatch getting caught before the run is the kind of boring gate I trust; it keeps a green transcript from blessing yesterday’s app.

#631 started turning the matrix into a repeatable evidence trail. This gives it a durable remote home and lets terminal regressions show up without borrowing a model’s credentials or mood. The narrow-pane recovery and the TERM → HUP close path now have the current-head packaged-app witness they needed. Approved. Case closed.

PR #637, opened 2026-07-22
refactor(automations): unify the daemon protocol on one per-action surface

victorarias:
## What
- One automation command set (9 commands) shared by the Unix-socket CLI and the WebSocket app client, dispatched through a single action layer (`internal/daemon/automations_actions.go`); the daemon no longer has parallel socket/WS code paths.
- 9 typed per-action result messages replace the generic result envelope; dead wire fields dropped; `last_run` is embedded in definition summaries so the panel's list badge loads in one request.
- ProtocolVersion 180 → 181 (constants.go + useDaemonSocket.ts in lockstep).
- CLI: `attn automation show` prints the definition's canonical YAML to stdout; `delete`/`validate`/`cleanup` print purpose-built summaries; list/runs emit lowercase snake_case JSON.
- Frontend adapted to the unified protocol; fixes a regression found by the packaged-app harness where the selected definition's run history went stale on `automations_changed` broadcasts (now refetched on every broadcast; pinned by a panel unit test).
- Harness: automation scenarios adapted to v2 spec-canonical YAML semantics and the new wire shapes; scheduled-cleanup now runs `ensureFreshWorld` before its simulated daemon downtime (a running app respawned the daemon and invalidated the catch-up leg); sqlite3 CLI probes got a busy timeout ("database is locked (5)" flake).

## Evidence (macOS, dev profile, packaged app rebuilt from head b4d5fbf4)
- `attn preflight` (dev): all PASS; CLI/app/daemon all protocol 181.
- Packaged-app scenarios, all ATTN_VERDICT ok:true on this head, run serially:
  - AUTOMATION-EDITOR: 9/9 legs (canonical-YAML reload, collision/id-change/stale-revision/deleted-elsewhere refusals, comment-only save is a revision no-op, disable/enable never bumps revision).
  - AUTOMATION-SCHEDULED-CLEANUP: 4/4 legs — first full pass ever (restart catch-up fires exactly one run under latest policy, real merged-worktree cleanup, dirty-worktree preservation, singleton coalescing, storm guard).
  - AUTOMATION-PR-CONTINUITY: continuation reuses session/ticket, Codex resumes with the copied rollout id keeping pinned model/effort, missing continuity worktree fails visibly.
  - AUTOMATION-SURFACE: 4/4 legs incl. run-now → delivered visible in the panel (the regression fix, live).
  - AUTOMATION-LIFECYCLE: 3/3 legs (edit-rebind across prompt changes, delete-resurrect with history survival, cleanup-dirty-safe).
- `go test ./internal/...` green; frontend 1940/1940.

https://claude.ai/code/session_01APU1R83oaVeshfJM5W5QCM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,15 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   triggered a redraw.
 - Closing a shell pane now hangs the shell up promptly instead of stalling ~10 seconds and force-killing it; remote pane/session close reliably confirms worker teardown instead of timing out mid-kill.
 
+### Changed
+- **The automations CLI's `attn automation show` now prints the definition's
+  YAML directly to stdout** instead of a JSON-wrapped blob, and `delete`,
+  `validate`, and `cleanup` print small purpose-built summaries instead of raw
+  store rows. The Automations panel's list badge (which run failed, and when)
+  now loads with the rest of the list in one request instead of a separate
+  fetch per definition, so opening the panel with many automations feels
+  noticeably snappier.
+
 ## [2026-07-21]
 
 ### Fixed
--- a/app/scripts/real-app-harness/scenario-automation-editor.mjs
+++ b/app/scripts/real-app-harness/scenario-automation-editor.mjs
@@ -29,11 +29,13 @@
  * actually be created.
  *
  * No fake-agent probe is needed: this scenario never triggers a run. Every
- * definition it applies is manual-trigger, and all but leg10's are also
- * enabled: false, so `launch.executable` is never invoked and can stay unset
- * like the starter template's. leg10 must start from an enabled definition in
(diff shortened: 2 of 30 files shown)


figgyster (approved):
You paid #629 and #630's one-authority debt all the way down here. Socket and WebSocket automation calls now walk through the same action layer, so this stops being two almost-identical stories waiting for one of them to forget a detail at 2am. The typed results and the list's embedded `last_run` are the right kind of boring: less round-trip weather, less transport drift.

The piece I read twice was the guard split on apply. A socket caller still gets its intentional last-writer-wins path, while the editor's present `expected_revision` keeps its collision and stale-save teeth. Good boundary. The current-head dev-profile preflight and the serial packaged-app scenarios cover the places this refactor could have lied — especially the refreshed selected run history — and the whole CI board is green again.

Safe to merge. Keep the harness breathing on this surface; it already caught one stale-history ghost before a person had to.

PR #638, opened 2026-07-22
Fix Copilot mouse selection lifecycle

victorarias:
## Summary

- stop overriding Copilot's configured mouse mode at launch
- guarantee every forwarded terminal-mouse press receives a release when the pointer leaves the pane or the window loses focus
- encode DECSET 1003 passive motion as no-button motion instead of a left-button drag

## Root cause

The original missed-release workaround disabled Copilot mouse support globally. After restoring terminal-owned mouse input, the release itself reached the PTY, but subsequent DECSET 1003 motion with no physical button down was encoded as SGR 32. That code means left-button drag, so Copilot re-entered selection immediately after a correct release. The correct passive-motion code is SGR 35.

## Validation

- `make test`
- `pnpm --dir app test` — 1,957 tests
- `ATTN_E2E_BIN=/tmp/attn-copilotmouse-e2e pnpm --dir app exec playwright test e2e/terminal-interactions.spec.ts` — 20/20
- signed `attn-dev` build and bundled Copilot preflight
- packaged-app live probe: press 0 → drag 32 → release 3 → passive motion 35
- installed Copilot launch preserved the current user-configured mouse-off state without either CLI override

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-22]
 
 ### Fixed
+- **Copilot mouse selection no longer remains stuck after releasing the
+  button.** attn now respects Copilot's configured mouse mode, guarantees that
+  a tracked press receives a matching release outside the terminal or on lost
+  focus, and correctly reports passive pointer movement without pretending the
+  left button is still held.
 - **`attn pr wait-ready` no longer returns instantly on a stale review verdict
   while a re-review is pending.** When the reviewer has been re-requested, an
   approval or changes-requested verdict that was already present when the wait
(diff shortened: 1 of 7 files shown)


figgyster (approved):
You finally pulled the Copilot workaround out by its roots instead of teaching it to live in the walls. Good call. #631 and #633 made the packaged-app harness a real witness; using it here for the press, escape, release, then passive-motion sequence is exactly what that machinery was built to catch.

The document-level release watch has the right escape hatch: a tracked press gets one release even if the pointer exits the pane or the window drops focus, using the last actual terminal cell when there is nowhere safe to hit-test. And DECSET 1003 now emits SGR 35 after release, so Copilot sees hover rather than a ghost left-drag. The full CI board is green, and your current-head dev-app/preflight probe covers the bit browser tests never quite own.

Safe to merge. The old global `--no-mouse` dodge can stay dead in the alley.

PR #639, opened 2026-07-22
Server-authoritative terminal — Phase 0+1 (shadow mode)

victorarias:
First slice of the server-authoritative terminal initiative. **Targets the `epic/server-authoritative-terminal` integration branch, not `main`** — the epic stays open until the full migration (through Phase 3) is done.

Plan: `docs/plans/2026-07-22-server-authoritative-terminal.md`

## What this does
Adds a worker-side authoritative parsed terminal (libghostty-vt via cgo) in **shadow mode**: it's fed the same bytes as the existing replay stores, but **nothing reads it in production yet**, so app behavior is unchanged. This is pure risk-burn — it proves the emulator, the cgo integration, and the build/sign path before any restore logic depends on it.

### Phase 0 — native lib build infra
- `scripts/build-libghostty-vt.sh` + `ghostty-vt-native.pin` (pinned `ab0b9da`, deliberately **separate** from the frontend WASM pin — the Terminal C API doesn't exist at the WASM pin).
- Reproducible: a fresh build produced the identical `libghostty-vt.a` SHA as the spike. Output vendored to gitignored `third_party/ghostty-vt/`; the script applies `ghostty-vt-native.patch` when present (alt-screen patch lands in Phase 2).

### Phase 1 — emulator in the worker, shadow mode
- **`internal/ghosttyvt`**: new package owning all cgo (`New/Write/Resize/DrainResponses/PlainText/Serialize/Close`), a `cgo.Handle`-routed `write_pty` callback for query responses, finalizer backstop, and the trailing-CUP fix for the upstream cursor/tabstop ordering bug. Gated to `darwin/arm64`; a portable **no-op stub** keeps the package building on Linux CI.
- **`internal/pty`** feeds it under `replayMu` at both read-loop write sites, resizes it, and closes it in `closePTY`. Every use is nil-guarded so a construction failure can never break a session. `DrainResponses()` is discarded this phase (the scanner stays the sole query answerer).
- **Shadow divergence metric** (`ATTN_GHOSTTY_SHADOW_DIVERGENCE=1`): on session close, compares the ghostty viewport against the vt10x oracle and logs the first differing row.

## Tests
- `internal/ghosttyvt`: round-trip (plain text / cursor / reflow-vs-direct-resize), codex query trio incl. kitty `CSI ? u`, malformed-input safety, no-interrogative-sequences guard, Close idempotency.
- `internal/pty`: divergence helper unit tests.
- Full `go test ./...` green. Linux cross-build + vet + test-compile verified (stub path).

## Live verification (throwaway `vtshadow` profile, full install)
- cgo binary **codesigns cleanly** (no ad-hoc fallback) — retires the plan's cgo-signing worry.
- Preflight PASS (protocol 180).
- `real-app:scenario-ghostty-scroll` **PASS** — existing scroll-anchor behavior unchanged.
- Worker log: ghostty viewport **matched vt10x exactly (26 rows)** on a real streamed session; **zero panics / cgo faults / construction failures** across spawn→feed→teardown.

## Not in this slice
`Serialize()` production path, alt-screen patch, protocol `snapshot` field, daemon/frontend wiring, codex bootstrap (Phase 2), and flip-default + delete-raw-replay (Phase 3, gated on Victor's approval). Real recorded byte corpora are also still to come (tests use a synthetic corpus).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ app/src-tauri/*.gen.conf.json
 
 # Local env files
 **/.npmrc
+
+# Native libghostty-vt build output (source of truth: scripts/build-libghostty-vt.sh)
+/third_party/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -34,6 +34,21 @@ Run full app builds/installs outside the sandbox: code signing needs the macOS
 keychain; sandboxed identity lookup can cause ad-hoc signing and lose persistent
 permissions.
 
+### Native VT library
+
+`internal/ghosttyvt` links `libghostty-vt` (Ghostty's VT core) via cgo on
+Darwin/arm64; every other platform compiles a pure-Go stub. The static archive
+lives under gitignored `third_party/ghostty-vt/` and is built from pinned source
+by `scripts/build-libghostty-vt.sh` (source of truth; pin in
+`ghostty-vt-native.pin`). This needs **zig 0.16.x** on `PATH`.
+
+On a fresh checkout the archive is absent, so the `build` target depends on it
+and runs the script automatically on the first `make build`/`make dev`/
+`make install*` — a one-time clone + zig build of a few minutes. It rebuilds only
+when the pin or script changes; otherwise it is a no-op. If zig 0.16.x is missing
+the script fails early with the required version. See
+[docs/plans/2026-07-22-server-authoritative-terminal.md](docs/plans/2026-07-22-server-authoritative-terminal.md).
+
 ## Profiles and live verification
 
 - Non-production builds, installs, launches, and restarts are pre-authorized.
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,29 @@ ZIG ?= zig
 # fallback; callers may override this with a pinned identity or fingerprint.
 MACOS_CODESIGN_IDENTITY ?=
 
-build:
+# Native VT library (libghostty-vt) for the server-authoritative terminal.
+# Only Darwin/arm64 links it via cgo (internal/ghosttyvt); every other platform
+# compiles the pure-Go stub and needs nothing here. The archive lives under
+# gitignored third_party/ (scripts/build-libghostty-vt.sh is the source of
+# truth), so a fresh checkout has no archive and the first cgo link would fail.
+# `build` therefore depends on the archive target on Darwin/arm64: it runs the
+# build script — which requires zig 0.16.x and clones the pinned source — when
+# the archive is missing or the pin moved, and is a no-op once it is present.
+UNAME_S := $(shell uname -s)
+UNAME_M := $(shell uname -m)
+NATIVE_VT_LIB := third_party/ghostty-vt/lib/libghostty-vt.a
+ifeq ($(UNAME_S)/$(UNAME_M),Darwin/arm64)
+NATIVE_VT_DEP := $(NATIVE_VT_LIB)
+else
+NATIVE_VT_DEP :=
+endif
+
+# Rebuild the native archive when its pin or build script changes (or it is
+# absent). The script installs into third_party/ghostty-vt/{lib,include}.
+$(NATIVE_VT_LIB): ghostty-vt-native.pin scripts/build-libghostty-vt.sh
+	./scripts/build-libghostty-vt.sh
+
+build: $(NATIVE_VT_DEP)
 	go build -ldflags "$(GO_LDFLAGS)" -o $(OUTPUT) $(BUILD_DIR)
 
 build-linux-amd64:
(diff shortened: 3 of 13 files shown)


figgyster (changes requested):
This is the right first cut for the terminal migration. You kept the new parser in shadow mode, kept the existing responder in charge, and made the restore gamble wait its turn. That buys the project evidence instead of another replay-shaped ghost story. But two bits here leave a fresh checkout and a real worker holding the bag.

Required before merge:

- `internal/ghosttyvt/ghosttyvt.go:156` gives libghostty-vt `&t.handle` as persistent callback userdata. `ghostty_terminal_set` keeps that pointer after `New` returns, so C retains a Go pointer without pinning it. That breaks cgo's lifetime rule and can turn a later VT query callback into memory corruption or a runtime fault. Pin the handle field for the native terminal's full lifetime with `runtime.Pinner` (and unpin on `Close`), or use a C-owned callback context; keep the callback test exercising the retained path.
- A fresh macOS checkout cannot use the standard build/install routes. The new static archive and headers are deliberately ignored under `third_party/`, while `Makefile:36-37` runs `go build` directly and `install-daemon` depends only on that target. Nothing provisions `scripts/build-libghostty-vt.sh`, so the first `make build`, `make dev`, or `make install-daemon` fails at cgo link time unless the developer happens to know and run an undocumented extra command. Make the normal Darwin/arm64 build depend on a verified native-library target (or fail early with an explicit bootstrap instruction) and document the toolchain prerequisite.

The Phase 1 shape is worth landing once those seams are closed. The watcher needs a reliable build, and the native callback needs an alibi that survives the call that created it. Until then, this stays red.

figgyster (changes requested):
You fixed the build contract. A fresh Darwin/arm64 checkout now grows the pinned archive through the ordinary `make` path, and the AGENTS note tells the next poor soul why Zig is knocking at the door. Good. That’s one foundation shored up.

One blocker remains:

- `internal/ghosttyvt/ghosttyvt.go:36-39` still turns a `cgo.Handle` integer into a long-lived `void *`, only now the coercion happens in C. `runtime/cgo.Handle` explicitly calls out `void *` APIs: a handle integer must not be coerced to `unsafe.Pointer`; the supported retained-callback shape is the handle field's address plus a `runtime.Pinner` held until after `ghostty_terminal_free`, or a real C-owned context that stores the integer. This helper relies on an arbitrary integer such as `1` being a valid opaque C pointer and surviving a round trip through Ghostty. It may happen to on the current box, but it is outside the cgo contract you're trying to make this migration safe under.

Keep the build fix. Replace the disguised handle-to-pointer cast with the pinned-address or C-context form, then this first shadow slice has its alibi. Still red for that one.

figgyster (dismissed):
There it is. The callback now uses the sanctioned shape: C retains a pinned address, the pin survives until `ghostty_terminal_free`, and the handle belongs to the sink instead of accidentally keeping the whole terminal alive. That closes the actual memory-contract blocker. Good recovery.

One last required witness before merge: this new `runtime.Pinner` path only exists on Darwin/arm64, while the visible Linux CI compiles the stub. Please run the current `1c935e1` code through the native proof tier — a fresh dev-profile build/install, bundled preflight, the `internal/ghosttyvt` query-callback test, and the existing ghostty-scroll packaged-app scenario. The prior live run predates this cgo-lifetime repair; PTY lifecycle work needs evidence from the binary that contains it.

The architecture has finally stopped arguing with cgo. Now let the installed worker testify. Still red until that current-head witness lands.

figgyster (dismissed):
That native witness does the job. The throwaway-profile build, protocol-clean preflight, real callback test under `-race`, and two packaged-app scroll runs are exactly the proof this Darwin-only path needed. The Pinner repair is no longer a theory in a Linux mirror.

I’m still holding the merge here because the current-head **Daemon** check is red: `TestDaemon_ScheduledStateUpdate`, `TestDaemon_RegisterAndQuery`, and `TestDaemon_StateUpdate` all timed out waiting for recovery during setup. Those may be an old flaky alley, but a failed principal suite is still a failed principal suite until a rerun is green or the failure is fixed. Get that check back on its feet and send me the signal.

figgyster (approved):
You came at this one like it was a terminal migration, not a weekend refactor. Good instinct. The shadow parser now has a pinned callback lifetime, a build path a fresh checkout can actually follow, and the Darwin worker has testified from an installed app instead of a whiteboard.

The current-head witness is the whole story: native callback coverage under `-race`, a protocol-clean throwaway profile, two scroll-anchor runs, and the Daemon rerun all came back green. This carries the recent harness discipline into the oldest and meanest layer of the app — the byte stream itself. Safe to merge. Phase 2 can take the restore path only after this dull foundation has had time to keep being boring.

PR #641, opened 2026-07-22
feat(terminal): snapshot-first attach behind ATTN_ATTACH_SNAPSHOT (Phase 2)

victorarias:
## Phase 2 — Snapshot-first attach behind a flag (raw replay still exists)

Part of the server-authoritative terminal initiative (plan:
`docs/plans/2026-07-22-server-authoritative-terminal.md`). Bases on and merges
into `epic/server-authoritative-terminal` — **not** main, per the standing
epic-branch constraint.

With `ATTN_ATTACH_SNAPSHOT=1` every attach restores from a libghostty-vt
serialization of the worker's parsed terminal instead of replaying raw bytes.
**With the flag off, behavior is exactly today's** — the daemon skips the
snapshot branch entirely (verified below).

### What's in it
- **Serializer** (`internal/ghosttyvt`): `Serialize()` emits an alt-screen-aware
  VT dump via a carried Zig C-shim patch (`ghostty-vt-native.patch`, applied by
  `build-libghostty-vt.sh`, now a Makefile prerequisite). Upstream's formatter
  only serializes the ACTIVE screen; the shim composes `TerminalFormatter`
  (palette) + per-screen `ScreenFormatter` (primary, then `?1049h` + alt frame
  when alt is active), so leaving a TUI after restore shows primary content.
- **Worker → daemon**: worker always serializes into `AttachInfo.GhosttySnapshot`
  under `replayMu` (atomic with the seq watermark); the daemon decides whether to
  serve it. Plumbed pty → ptybackend → ptyworker → back. Protocol adds
  `AttachSnapshot` (ProtocolVersion 180 → 181).
- **Codex fresh-spawn** no longer needs raw replay: the worker forwards ghostty's
  query responses the scanner misses (kitty `CSI ? u`, DECRQM) via
  `drainGhosttyResponses` + `stripScannerOwnedResponses`, keeping the scanner the
  sole answerer for CPR/DA1/OSC-color. Env-gated; flag-off drains-and-discards as
  today.
- **Frontend**: classifies a present `data.snapshot` as a `ghostty_snapshot`
  replay kind that supersedes screen-snapshot/raw-scrollback, resets the model,
  resizes to the snapshot grid, writes the dump suppressed.

### Tests
- Go: `TestRoundTripAltScreen`, `TestStripScannerOwnedResponses` (9 cases incl.
  codex bootstrap trio), `TestAttachSnapshotSeqConsistency`. Full `make test`
  green.
- Frontend: 6 new `attachPlanning.test.ts` cases; full suite green (1963).

### Live verification — native tier (throwaway profile `satsnap`, arm64/cgo)
Preflight PASS (protocol 181 CLI/app/daemon). Flag ON:

| scenario | result | served by |
|---|---|---|
| tr204 relaunch formatting (claude) | PASS | `use_ghostty_snapshot` |
| tr401 window resize/reflow after restore | PASS | `use_ghostty_snapshot` |
| tr402 close-redraw (claude) | PASS | `use_ghostty_snapshot` |
| tr402 close-redraw (**codex** fresh-spawn) | PASS | `use_ghostty_snapshot` |
| tr201 split relaunch restore | PASS (retry¹) | `use_ghostty_snapshot` |

24 restore attaches served via `replay_decision=use_ghostty_snapshot`
(5.6–7.7 KB dumps), no panics/worker leaks in `daemon.log`.

**Gating proof** — flag OFF: tr204 still passes but via
`use_segmented_raw_replay` with `ghostty_snapshot_bytes=0` (zero snapshots
served). Legacy path byte-identical.

¹ tr201's first run timed out waiting for the live Claude agent to emit a
structured block (before the restore step) — known live-agent-discretion flake;
passed on retry.

**Pre-existing, not this change**: `tr401-codex-MAIN` fails identically with the
flag ON and OFF (3× each) — a pre-existing codex main-window narrow header-frame
issue, independent of Phase 2.


figgyster (changes requested):
You’re carrying the server-authoritative terminal one more block down the street: Phase 0 made the native seam survivable, and this finally lets a fresh client inherit a parsed terminal instead of reenacting byte history. The snapshot/grid handoff and the worker-side query gap are the right seams to harden.

One small blocker before merge: `ghostty-vt-native.patch` fails `git diff --check` on lines 8, 40, 61, and 168 because the added blank lines carry trailing whitespace. This patch is now build input for the native archive, so it deserves the same basic patch hygiene as the Go around it. Strip those spaces, then this particular witness can go home.

figgyster (commented):
The native patch swept the four dirty blank lines out of the doorway. `git diff --check` passes on this head, and the follow-up is whitespace-only, so the serializer and the compiled archive keep the same behavior we already inspected.

The daemon and frontend jobs are green. The E2E shards are still running, so I’m clearing the code blocker without pretending the last witnesses have testified. If those finish green, this snapshot-first slice is ready to keep walking toward the day raw replay can retire.

figgyster (approved):
The final E2E witnesses made it in, and the whole board is green. The whitespace repair was the only thing holding this head up; the native serializer, snapshot geometry, and query handoff keep the shape we already checked.

This is the next honest step after #639: the parser stops being a shadow on the wall and starts restoring a terminal for real, while the flag keeps the old replay road open for Phase 3’s reckoning. Safe to merge into the epic branch. Approved.

PR #642, opened 2026-07-23
feat(pty): flip snapshot restore on, delete raw replay (Phase 3)

victorarias:
## Phase 3 — flip snapshot restore on, delete raw replay

Lands the server-authoritative terminal restore as the **default and only** path,
now that Phase 3a (worker-owned OSC 133 block table) has removed the last
foundation gap. Rebased onto the epic to pick up 3a; every attach restores from
the worker's serialized libghostty-vt dump (+ structured blocks), and the raw
byte-replay machinery is deleted.

Plan: `docs/plans/2026-07-22-server-authoritative-terminal.md` (Phase 3).
Ships to the epic branch `epic/server-authoritative-terminal`, not `main`.

### What changes

- **Default on.** `ATTN_ATTACH_SNAPSHOT` is gone; `buildAttachReplayPayload`
  collapses to `omit_replay_for_policy` / `no_snapshot` / `use_ghostty_snapshot`.
  The worker's `drainGhosttyResponses` now unconditionally forwards the query
  gap, so the client always writes the dump with responses suppressed.
- **Raw replay deleted.** `ReplayLog` + whole-segment machinery
  (`replaylog.go`, `ringbuffer.go`), the replay-vs-snapshot decision tree and
  oracle verification in `ws_pty.go` (`rawReplaySegmentsMatchFreshSnapshot`,
  `LimitReplaySegmentsTail`, `maxAgentRawReplayBytes`, segment plumbing), the
  vt10x `screenSnapshot`-from-replay derivation, the frontend raw-replay
  classification in `attachPlanning.ts`, and the codex raw-replay carve-out.
- **Protocol swept.** Removed `scrollback` / `replay_segments` /
  `screen_*` from `AttachResultMessage` and deleted `model ReplaySegment`.
  `get_screen_snapshot` (grid tiles) keeps its own `screen_*` fields.
  `AttachSnapshot` carries `blocks` (Phase 3a) and `scrollback_truncated`.
  `ProtocolVersion` 182 (unchanged from 3a — same wire epoch).
- **Kept:** vt10x `renderedText()` for the state classifier and grid-tile
  seeding; the seq/watermark contract and its race tests
  (`attach_race_test.go`, representation-independent so the Linux stub passes);
  `stripScannerOwnedResponses` on the worker.

### Rebase reconciliation (3a × Phase 3)

Phase 3 was authored on Phase 2, before 3a. The rebase preserves 3a's block
plumbing while applying Phase 3's deletions:
- `GhosttyBlocks` is carried alongside the snapshot through every backend/worker
  hop and into `attachReplayPayload`; the client seeds `TerminalBlockStore` on
  restore (the `ghostty_snapshot` branch keeps its `seed_blocks` emit).
- 3a removed the vestigial `ghosttyvt.Snapshot.ScrollbackTruncated` (always
  false); Phase 3's `GhosttyScrollbackTruncated` diagnostic is preserved but
  reported `false` at the origin until libghostty-vt exposes a real signal.
- Dropped a stale `Session.scrollback: NewRingBuffer(...)` line from the 3a
  block-feed race test (the field and `NewRingBuffer` are gone with raw replay).

### Tests

- `go build ./...` (cgo/native) + `go vet` + `gofmt` clean.
- Full `go test ./...` green (incl. `-race` attach atomicity, native
  `TestBlockFeedRoundTrip`, daemon suite).
- Frontend: `tsc --noEmit` clean; `pnpm test` 1977 passed.

### Live native witness (current head `d8ce9b1d`, Darwin/arm64)

Throwaway profile `p3flip` (full `make install`, preflight PASS — protocol 182
across CLI/app/daemon). **No env flag** — snapshot is the default now.

- `real-app:scenario-terminal-block-resize` → `ok: true`. fish command blocks
  **4 → 4 survived** relaunch under the default-on snapshot restore; bash/zsh
  **0 → 0**.
- `real-app:scenario-ghostty-scroll` → `ok: true`.
- Daemon log across both runs: `replay_decision=use_ghostty_snapshot` (11) and
  `omit_replay_for_policy` (8, fresh spawns) — **zero raw-replay decisions**,
  confirming the ghostty snapshot is the sole restore path.

Ships to the epic branch `epic/server-authoritative-terminal`; does not touch
`main`.


figgyster (dismissed):
You’ve reached the part of the terminal arc where the old safety net gets cut away. I’m with you on that direction — #639 built the native footing and #641 made snapshot restore earn its keep. But the snapshot-less path is still a live contract, and this head breaks it.

Required before merge:

- `app/src/pty/attachPlanning.ts`: do not reset the client model or advance its dedup boundary to the server’s `last_seq` when `snapshot` is absent. The server has no serialized state in that case, so queued output at or below that watermark is not a duplicate from the client’s point of view. On `same_app_remount`, this clears the only rendered terminal and drops those queued bytes; an idle shell then stays blank until it happens to print again. That contradicts the documented construction-failure path in `AGENTS.md`, which says a snapshot-less attach keeps the client state and dedups live output. Preserve the existing model/watermark in this branch (or otherwise deliver all output it lacks), and pin the case with a regression test.

The rest of the deletion is the right kind of scary: one server-owned terminal instead of two competing histories. This fallback just needs to survive the rare night the native parser doesn’t wake up. Holding the line here.

figgyster (dismissed):
The Phase 3a rebase brings the OSC 133 block work under this roof, and that was the right prerequisite for killing replay. But this follow-up never touches the fallback I stopped yesterday. The same quiet little bug is still waiting behind the radiator.

Required before merge:

- `app/src/pty/attachPlanning.ts`: snapshot-less restore policies still reset the model, replace the client’s watermark with the server `last_seq`, and discard queued chunks at or below it. With no dump, those chunks were never rendered by the client. A same-app remount after Ghostty construction fails therefore clears the only visible state and can leave an idle terminal blank forever. Keep the existing model/watermark when `snapshot` is absent (or deliver every missing chunk), and add the regression witness. The documented fallback says exactly that.

#639 and #641 earned the right to retire raw replay. This one last fallback needs to survive the night first.

figgyster (commented):
You found the loose floorboard. The snapshot-less branch now keeps the model it actually has, uses that model’s watermark, and lets the unpainted queued tail through. That is the fallback contract I asked for, and the new same-app-remount witness catches the exact old blank-idle-session trap.

Phase 3a makes the dump carry the terminal’s command blocks; this fix makes sure the rare night with no dump does not erase the terminal that is already there. The raw-replay deletion is safe by me once the current Frontend and E2E checks finish green.

figgyster (approved):
Now it holds. The no-dump path keeps the only terminal it has, and the client-side watermark lets the queued tail arrive instead of vanishing into the drain. That's the small cruel edge this kind of cutover lives or dies on.

#639 built the server parser, #641 made it restore, and #644 gave the restore its command blocks. This lands the last code phase by deleting the replay maze without leaving the failure path behind it. The complete board is green. Approved. Send it into the epic branch.

PR #644, opened 2026-07-23
feat(pty): worker-owned OSC 133 block table + snapshot seed (Phase 3a)

victorarias:
## Phase 3a — worker-owned OSC 133 block table + snapshot seed

Unblocks the Phase 3 hold. Snapshot restore previously **lost** OSC 133 command
blocks (the VT dump carries rendered cells, not the `OSC 133;A/B/C/D` marks, so
the client rebuilt zero blocks on a snapshot attach; raw replay preserved them
only incidentally by re-feeding bytes through `parseOsc133`). This makes blocks
worker-owned state carried as a structured `blocks` field beside the dump, which
the client seeds its `TerminalBlockStore` from on restore.

Design + rationale: `docs/plans/2026-07-23-terminal-restore-fidelity.md`.
Implementation contract: Phase 3a in
`docs/plans/2026-07-22-server-authoritative-terminal.md`.

### What's here

- **`internal/pty/osc133.go`** — Go OSC 133 segmenter, semantics-identical to the
  client `parseOsc133`, but STRIPPING marker bytes (the client keeps them) so
  OSC 133 never re-enters the VT dump. Marker parity is enforced by a shared
  corpus consumed by both the Go test and a new `terminalOsc133.parity.test.ts`.
- **`internal/pty/blocktable.go`** — worker block table mirroring the client
  `TerminalBlockStore` (one pending block, self-heal on lost `133;D`, cap 200
  oldest-first). Positions are reference-counted tracked grid refs — a self-heal
  shares one native ref between two blocks and frees it exactly once. Alt-screen
  blocks are excluded at snapshot. Lifecycle proven identical to the client store
  by the shared `osc133_block_corpus.json`.
- **`internal/pty/blockfeed.go`** — `newBlockFeeder` wires the real segmenter +
  table (rails no-ops removed). Call sites, `replayMu` placement, and the atomic
  {dump, blocks, watermark} triple are unchanged from the rails.
- **`internal/ghosttyvt`** — the spike `TrackedRef` API graduated to production
  names (`trackedref*.go`); the prune probe became a real drop-detection
  assertion (`TestTrackedRefDropsWhenPruned`).
- **protocol** — `AttachSnapshot` gains `blocks?: AttachBlock[]`; `ProtocolVersion`
  → 182. Deleted the dormant `Snapshot`/`AttachSnapshot.scrollback_truncated`
  (the live `AttachResultMessage.scrollback_truncated` raw-replay signal is
  untouched).
- **frontend** — `TerminalBlockStore.seed()` lands completed blocks with a
  locally-computed `anchorText`, re-arms the pending block, and continues the id
  counter above the max seeded id; `useDaemonSocket` emits a `seed_blocks` PTY
  event after the dump write so anchors read the restored buffer.

### Tests

- **Go** — segmenter parity corpus (shared Go + TS), block-table lifecycle with
  ref-leak accounting (`LiveTrackedRefs` returns to baseline), `-race`
  atomicity, and a native end-to-end round-trip `TestBlockFeedRoundTrip`: real
  segmenter + real table → serialize → each block's rows index the restored dump
  at the right text, command text + exit codes captured at parse time,
  alt-screen block excluded. Linux stub build stays green (block-table paths
  inert without ghostty).
- **Frontend** — `seed()` unit tests + parity test; full vitest suite (1992
  passed), `tsc --noEmit` clean.

### Live native witness (current head `3b472e55`, Darwin/arm64)

Threw-away profile `p3ablk`, full `make install`, preflight clean (protocol 182
across CLI/app/daemon). Ran `real-app:scenario-terminal-block-resize` with
`ATTN_ATTACH_SNAPSHOT=1` propagated end-to-end (app → daemon → worker):

- Scenario `"ok": true` — all four phases green including `phase_b_relaunch_replay`.
- fish blocks **4 → 4 survived** the relaunch; bash/zsh **0 → 0** (correct-or-absent).
- Confirmed it went through the **snapshot** path, not raw replay: the daemon
  carried `ATTN_ATTACH_SNAPSHOT=1` and every attach logged
  `replay_decision=use_ghostty_snapshot replay_reason=attn_attach_snapshot_enabled`
  `ghostty_snapshot_bytes=5908 replay_bytes=0` — so `seed()` rebuilt the blocks
  from the structured field.

This meets the Phase 3a exit gate (Phase B green with snapshot attach on).

**Head note:** the witness above ran on `3b472e55`; current head `8da296f6`
differs by gofmt whitespace (struct-field alignment in `ws_pty.go` /
`blocktable_test.go`, to satisfy the CI "Check formatting" step) plus one plan
checkbox. `git diff -w 3b472e55 8da296f6 -- ':!docs'` is empty — the compiled
cgo behavior is byte-identical, so the native witness carries to the current
head.

Ships to the epic branch `epic/server-authoritative-terminal` like every phase;
does not touch `main`.


figgyster (approved):
At last, the snapshot is carrying the bits the grid forgets. #641 got the parser-owned restore across the river; it still left every OSC 133 block on the dock. This gives the worker ownership of command history, pins it through scroll and reflow, and ships it in the same lock hold as the dump and watermark. That's the missing seam.

I went looking for the native leak and the half-restored prompt. The shared corpus keeps Go and client lifecycle behavior tied together, tracked refs have retirement proof, and the real serialized round trip checks both the rows and the semantic stragglers: command text and exit status. The fresh snapshot-on harness witness is the right one after #642's hold.

Current Daemon, Frontend, and every E2E shard are green. Approved. Phase 3 can come back out of the rain.

PR #645, opened 2026-07-24
build(plugins): sweep orphaned Bun compile temps

victorarias:
## Summary
- remove orphaned `*.bun-build` temporary files after each bundled plugin compile
- keep staged plugin directories limited to the intended executable binaries

## Why
`bun build --compile` can occasionally leave its adjacent temporary Mach-O behind after a successful rename, cluttering the bundle output.

## Verification
- `bash -n scripts/build-bundled-plugins.sh`
- `bash scripts/build-bundled-plugins.sh` end-to-end: both plugins staged cleanly, no `*.bun-build` orphans remained, and the real binaries were intact.

## Live verification
Not applicable: this is build-script-only output hygiene with no app or daemon-observable behavior.

### DIFF
--- a/scripts/build-bundled-plugins.sh
+++ b/scripts/build-bundled-plugins.sh
@@ -84,6 +84,11 @@ stage_plugin() {
   rm -rf "${stage_dir}"
   mkdir -p "${stage_dir}/bin"
   bun build "${source_dir}/src/index.ts" --compile --minify --outfile "${stage_dir}/bin/${name}"
+  # `bun build --compile` writes to a `*.bun-build` temp next to the outfile and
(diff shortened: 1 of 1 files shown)


figgyster (approved):
This is the kind of small build-room cleanup I like: Bun leaves a body in the alley, and the next codesign run gets blamed for it. Rebuilding the stage directory first, then sweeping only `bin/*.bun-build` after a successful compile, keeps the shipped executable and every source artifact out of the blast radius.

`git diff --check` is quiet, the script syntax pass is good, and the current CI run is green. The plugin packaging path keeps getting more explicit about its generated-output boundaries. Good. Approved.

PR #646, opened 2026-07-24
build(ghostty-vt): download prebuilt native library by default

victorarias:
## What changed
- Make `libghostty-vt` provisioning download-first, using a public GitHub release asset keyed by the Ghostty pin and carried patch.
- Verify downloaded archives against the committed `ghostty-vt-native.lock`; add the maintainer publish workflow and document it.

## Why
Fresh Apple Silicon checkouts no longer need Zig 0.16.x just to build attn.

## Verification
- `bash -n scripts/build-libghostty-vt.sh scripts/lib/libghostty-vt.sh scripts/publish-libghostty-vt.sh`
- `make test`
- `make test-frontend` (183 files, 1,978 tests)
- Fresh clone: `make build`, which downloaded, SHA-verified, extracted, and linked the published prebuilt archive.

## Operational note
This relies on the live public `native-vt-prebuilts` GitHub release asset. If download or verification fails, the build safely falls back to a Zig 0.16.x source build. Linux and other non-Darwin/arm64 platforms continue to compile the pure-Go stub.


figgyster (approved):
You took the last expensive native seam in the server-authoritative terminal arc — the part that made every fresh Apple Silicon checkout build Ghostty with Zig — and gave it a real supply-chain path. #639 and #642 made the worker’s model the authority; this lets people actually get that authority built without spending the afternoon in a compiler alley.

The lock key matches the current pin plus carried patch, and the public asset hashes to the committed SHA. I ran the fresh provisioner: it downloaded, verified, extracted the expected archive, and left the keyed install marker behind. The source fallback still takes over when a maintainer changes that input. Daemon CI is green. Safe to merge.

PR #647, opened 2026-07-24
refactor(terminal): remove dead/degenerate replay-era code before epic lands

victorarias:
Cleanup pass over the epic before it merges to main: remove code left
vestigial once Phase 3 flipped the attach path to snapshot-only and deleted
raw replay. No production-path behavior change.

## Go — remove finished shadow-divergence scaffolding
`logShadowDivergence` + `ghosttyViewportTail`/`firstDivergentRow`/`rowAt`
helpers + the `ATTN_GHOSTTY_SHADOW_DIVERGENCE` flag + `shadow_divergence_test.go`.
This was Phase 0/1 rollout instrumentation (off by default) that compared the
ghostty viewport against the vt10x oracle to de-risk the migration. Ghostty is
now the authoritative restore emulator, so the scaffolding's job is done.

The vt10x screen itself stays — it still backs the approval-state classifier
and the `get_screen_snapshot` grid-tile/automation path.

## Frontend — collapse the replay-era classification
Phase 3 made the Ghostty VT dump the sole restore payload, but
`attachPlanning.ts` kept the old replay-vs-snapshot classification surface.
With one payload or none, most of it was degenerate:

- `hasReplayPayload` / `replayApplied` were aliases of `hasGhosttySnapshot`
- `replayKind` was a two-value union mirroring the same boolean
- `replaySkipped` (hardcoded `false`), `replayAllowedByPolicy`,
  `replayGeometryMismatch`, `attachedGeometryMismatch` were computed but read
  only by their own tests

Collapsed the aliases, dropped the four dead fields, and renamed the `replay*`
surface to `snapshot`/`restore` vocabulary across `attachPlanning.ts`, its
consumers in `useDaemonSocket.ts`, and the tests. The emitted events, seq-dedup
math, and the snapshot-less `same_app_remount` fallback are byte-identical and
still pinned by the existing regression tests.

Bridge event channel names (`source: 'attach_replay'`, `event: 'replay_complete'`)
are left as-is — live wire contracts, not dead vocabulary.

## Verification
- Go: `go build ./...`, `go vet`, `gofmt` clean, `go test ./internal/pty/` green.
- Frontend: `tsc --noEmit` clean, all 1977 unit tests pass (incl. 61
  `useDaemonSocket` + 15 `attachPlanning`).

Pure rename + dead-code removal with no production-path behavior delta; the
epic still owes its own final live pass before epic→main, which this rides into.


figgyster (commented):
You finally swept up the evidence tape after Phase 3 did its job. #639 built the shadow seam, #641 taught it to restore, and #642 made Ghostty the whole story; leaving the dead replay vocabulary behind would have been the sort of small debt that keeps collecting rent.

I traced the surviving attach path through the snapshot-less fallback and its client-owned sequence watermark. That contract remains intact, while the classification collapses to the only question left: is there a Ghostty dump to restore? `git diff --check` and the focused attach-planning witness pass; Daemon and all three E2E shards are green.

The Frontend check is currently red, so this stays a conditional review until that witness clears. No code blocker found in the cleanup itself. When Frontend goes green on this head, I’m ready to sign it off.

figgyster (approved):
The red Frontend light belonged to the ticket-overlay flake, and #649 put that ghost in the ground. The full board is green on the same head now.

You kept the important part of Phase 3 when you cleared its scaffolding: no snapshot still preserves the rendered model and its own sequence boundary, while a Ghostty dump is the sole restore story. That’s the terminal finally shedding its rollout husks before the epic walks into main. Approved.

PR #648, opened 2026-07-24
feat(ghosttyvt): build + run the native VT library on Linux (daemon-only)

victorarias:
## What

Makes `internal/ghosttyvt` (the worker's server-authoritative terminal restore, backed by `libghostty-vt` via cgo) **link and run on the headless Linux daemon** (linux/amd64 + linux/arm64), not just darwin/arm64. **No behavior change on macOS.** Enabling work only — retiring vt10x stays a later, separately-gated step.

### Why
The hub cross-compiles and runs a full daemon (PTY + classifier + restore) on Linux remotes, and Victor runs the daemon on Linux directly. There, ghosttyvt was the pure-Go stub, so restore always rode the snapshot-less fallback. With the library building on Linux, the worker can serialize the parsed grid there too.

## Changes
- **ghosttyvt**: widen `//go:build` to `(darwin && arm64) || (linux && amd64) || (linux && arm64)` across the package + tests; invert the stub tag. Split `#cgo` per tuple — the four macOS frameworks are darwin-only over-linking; the Linux tuples link only the self-contained `.a` (C++ runtime from simdutf/highway is statically baked in) + `-lm -lpthread`.
- **Prebuilt mechanism (#646) → per-platform**: assets `libghostty-vt-<key>-<goos>_<goarch>.tar.gz`, output `third_party/ghostty-vt/<goos>_<goarch>/`; lock keeps one shared source `key` + per-platform `sha256_<goos>_<goarch>`. `publish` cross-builds + uploads all three targets from one Mac. Download-first unchanged (no zig in CI/release).
- **Makefile**: resolve target platform from `GOOS`/`GOARCH` (either origin), gate the archive dep on the three tuples; `build-linux-*` recurse into `build` so the archive downloads before the cross cgo link; `test` gains the dep.
- **hub/bootstrap**: ensure the target's native archive before the cross-compile fallback.
- **ci.yml**: fetch the native archive in the Linux Go jobs. `release.yml` needs no change (`GOOS=… make build` already resolves + downloads the right prebuilt).
- **AGENTS.md**: app is macOS-only, daemon is supported on Linux; per-platform native VT section.

## Verification
- Full Mac Go suite green; `go vet ./...` + host build clean.
- ghosttyvt pkg + full daemon **cross-compile** for both Linux tuples.
- Fully cross-built **linux/arm64 daemon runs natively in the OrbStack VM** (`--version` ok; binary contains the real ghosttyvt — real-path error string + linked `ghostty_terminal_serialize_vt`, no stub marker).
- **Native-gcc** VT probe on the VM links with just `.a + -lm -lpthread` (no `-lstdc++`) and emits a real 5655-byte VT dump.
- Download path re-fetches + sha-verifies the published prebuilts for darwin + linux.


figgyster (approved):
You made the server-authoritative terminal stop being a Mac-only alibi. Phase 3 finally has the same parser on the Linux daemon where Victor actually runs it, while the snapshot-less fallback stays confined to targets that truly cannot carry the native library. That is the right kind of enabling work: boring artifact plumbing, but it closes a real hole in the authority story.

The per-tuple archive paths, cgo flags, Make targets, CI bootstrap, and hub cross-compile all agree on the target they mean to build. I pulled and checksum-verified both Linux artifacts, then ran the full Go suite against the amd64 library; the VT and hub witnesses held. The whole CI board is green too. Safe to merge into the terminal epic. The later vt10x retirement can wait for its own hearing.

PR #649, opened 2026-07-24
test(tickets): fix ticket-overlay status-select flake

victorarias:
The ticket detail panel container mounts before the async fetchTicket resolves. The test waited for the panel, then immediately fired a change event at ticket-status-select, which renders only after that async load completes — a race condition that flakes under CI load (manifested as a false failure on PR #647's Frontend check).

The fix: wait for the interactive select element itself, ensuring it's rendered and ready before firing the event.


figgyster (approved):
You found the right culprit from #647. The panel shell can arrive while its async ticket data is still trudging up the stairs, so poking the status select at that moment was a test accusing an empty chair. Waiting for the control itself makes this witness observe the behavior it claims to cover.

The focused ticket-overlay check returned successfully here, and Frontend plus every E2E shard are green. Approved at 41689dd8. Small repair. Real signal.

PR #650, opened 2026-07-24
feat(ghosttyvt): viewport snapshot API + vt10x parity corpus (retirement Phase 1)

victorarias:
Phase 1 of the vt10x retirement plan (`docs/plans/2026-07-24-vt10x-retirement.md`, included in this PR). Adds the ghosttyvt surface the three vt10x consumers need, plus the classifier-parity gate — no consumer is repointed yet (that's Phase 2).

## API additions (`internal/ghosttyvt`)

- `CursorPos()` — 0-indexed viewport cursor (backs CPR replies in Phase 2)
- `CursorVisible()` — DECTCEM via `ghostty_terminal_mode_get`
- `ViewportText()` — visible screen as plain text, one `\n`-terminated line per row, trailing blanks trimmed (the classifier's input shape)
- `SerializeViewport()` — self-contained styled VT stream of the viewport only (no scrollback), via selection-restricted formatting; replaying it reproduces content, styles, cursor position, and visibility

Stub mirrors all four. A blank-screen contract bug (`ViewportText` returning `""` instead of rows×`\n` — empty formatter output conflated with failure) was caught by the corpus below and fixed in `824c1be5`.

## Parity corpus (`internal/pty/ghostty_parity_test.go` + `testdata/corpus/`)

Raw `script(1)` captures of real sessions (claude approval dialog, codex approval dialog, fish resize traffic, vim alt-screen ×2 sizes, SGR-heavy ×2 sizes) replayed into both `virtualScreen` (vt10x) and ghosttyvt at two chunk boundaries, asserting identical viewport text + cursor.

**Finding: vt10x mis-parses the codex TUI's scroll-region traffic** (`CSI r` margins + `CSI S`): interleaved rows, cursor (0,3) vs the correct (0,17). A real tmux replay of the fixture agrees with ghostty exactly. So today's CPR replies and classifier text are subtly wrong during codex sessions — retirement fixes a live bug. That fixture asserts against a tmux-refereed golden, with a tripwire that fails if vt10x ever starts agreeing.

## Verification

- Native tier (darwin/arm64, head a8d3b346, macOS 26.5.2): `go test ./internal/ghosttyvt ./internal/pty -count=1` — all green, parity corpus 7/7 fixtures × 2 chunk sizes.
- `make build-linux-arm64` cross-link green.
- No behavior change: nothing outside `internal/ghosttyvt` production code; `internal/pty` gains only the test + fixtures.


figgyster (approved):
You found the thing vt10x was getting wrong in the Codex TUI, then put it in a jar where it can bite the next person who tries to resurrect it. Good. The retirement has been circling the block for days; this gives Phase 2 a parser contract instead of a prayer.

`ViewportText` keeps the classifier's row shape honest, including the blank-screen edge, and the viewport-only dump survives a fresh Ghostty replay with cursor state intact. The corpus handles the ordinary parity cases and makes the known Codex scroll-region divergence explicit rather than pretending the old model is a referee.

The current-head native darwin witness, Linux cross-link, and green Daemon run cover the native seam. Safe to merge into the epic. The consumers can move in the next room; this foundation holds.

PR #651, opened 2026-07-24
Daemon-owned session revive and reload

victorarias:
## Problem

After a machine restart, sessions whose PTY workers died showed a dead pane with `[Failed to attach PTY: session not found]` and no way to recover them. Recovery logic lived in the frontend, which could only act on state the client happened to hold — and after a restart it holds nothing.

## What this does

Moves session recovery entirely into the daemon. The client states intent; the daemon owns the strategy.

- **`recoverable` is a real session state.** On daemon startup, reconcile marks sessions with dead workers `recoverable` instead of leaving them in limbo. Transitions go through `applyState` like every other state.
- **Revive-on-attach.** Pane mounts attach with policy `revive`; when the target is a recoverable session, the daemon respawns it from its durable record before completing the attach. No user action needed — a restarted machine's sessions come back when their panes mount.
- **`store.LaunchIntent` persists the full launch contract** (executable, model, effort, yolo mode, and the complete unattended-launch spec) so the daemon can relaunch a session without a live worker registry and without refusing unattended sessions.
- **Daemon-owned reload.** New `reload_session {id, cols, rows}` command (protocol 183): live workers are killed and respawned in place under a reload lock with exit suppression; dead sessions are relaunched through the spawn pipeline from stored intent. The Reload button now sends this instead of orchestrating kill+spawn client-side.
- **Client recovery machinery deleted** (protocol 184: `kill_session.reload` removed). `spawnPtyRuntime` reduces to spawn + fresh attach; the daemon-known-attach, resume-after-attach-failure, and reload-kill paths are gone (net −600 lines). `relaunch_restore` stays client-side deliberately — it is a replay/rendering policy, not a recovery strategy.
- **Race fixes surfaced along the way:** commitSpawn re-reads current state before its upsert so mid-spawn transitions survive; spawn reservations prevent embedded-backend double launches; endpoint binding persists across respawns; Makefile rejects an explicitly-passed empty `PROFILE=` at parse time.

Every session create/relaunch now flows through one spawn pipeline (`validate → lock → normalize → resolve → execute → commit`); the only other PTY start is the in-place live-worker swap, which intentionally skips session-record commit because the session never went away.

Plan doc with per-phase details and follow-up resolutions: `docs/plans/2026-07-24-daemon-owned-session-revive.md`.

## Verification

- Go, frontend (1952 tests), and tsc suites green; protocol regenerated via `make generate-types`, version lockstep bumped in both `constants.go` and `useDaemonSocket.ts`.
- Live-verified at final head on throwaway profiles: protocol-184 preflight, packaged `recoverable-auto-revive` harness scenario (worker killed → auto-revived on pane mount, no error banner), and live Reload respawning a real session's worker in place.


figgyster (changes requested):
You moved recovery to the one layer that can actually keep the promise after the client and worker registry have both gone dark. That is the right turn. But the new durable contract still has one hole in the floor.

Required before merge:

- `internal/daemon/spawn_pipeline.go:384` persists `LaunchIntent` only after `ptyBackend.Spawn` succeeds. A daemon crash or machine restart in the interval after the worker starts but before this write leaves the already-persisted session with no launch contract; startup can mark it `recoverable`, but revive-on-attach then stops at `no stored launch intent`. That is exactly the restart recovery this PR is meant to own. Persist the resolved intent in the durable pre-spawn session write (and restore/clear it with the rest of the record on spawn failure), then add a regression around that commit boundary.

The focused daemon/store suite is green. The tests currently prove intent after the full successful pipeline, but not this new crash boundary.

figgyster (changes requested):
The merge brings the branch up to date, but it does not touch the launch-intent ordering I flagged. The worker can still start before its recovery contract is durable, so a restart in that interval leaves a `recoverable` session with nothing to revive from.

Please carry the same fix and crash-boundary regression before merge. The old red reviewed `05a73e8`; this is the one that applies to `b74fd7f`.

figgyster (changes requested):
The pre-spawn write and deterministic PTY reservation gate are both the right repair. The focused daemon, PTY, and store suites pass here. One rollback edge still writes history the session never earned.

Required before merge:

- `internal/daemon/spawn_pipeline.go:331` restores an old launch intent only when `hadPriorIntent` is true. If an existing legacy session had an empty `launch_intent`, the pre-spawn write survives a failed worker start. That leaves the restored session carrying the model/executable/approval contract of a launch that never happened, and a later automatic revive will act on it. Restore the empty state too (a clear API is probably the least ambiguous shape), on each failure path, and add the existing-session/no-prior-intent regression beside the new rollback test.

The previous reservation-test CI failure was an observation race; this head's gate removes that race. This is the remaining transactional boundary.

figgyster (commented):
That closes it. The launch contract is now written before the worker exists, and every surviving-row failure path restores the exact prior intent — including no intent at all. The deterministic PTY reservation witness and the no-prior rollback witness are the right kind of receipts.

The focused daemon, PTY, and store suites pass. The remaining Frontend/E2E CI lanes are still running, so I’m leaving this as a current-head conditional review rather than pinning an approval to unfinished weather. Once they land green, this is safe to merge.

figgyster (approved):
You took the recovery path all the way down to the ledger. The last version could restart a worker and leave a false launch contract behind. This one restores the exact prior truth, including none at all. That's the boring transaction boundary that saves a Sunday.

The whole board is green now, including the daemon, frontend, E2E, Linux, and benchmark runs. The server-authoritative terminal finally has a daemon-authoritative lifecycle beside it. Approved. Let auto-merge have the last word.

PR #652, opened 2026-07-24
feat(pty): repoint approval/CPR/snapshot readers to ghostty (retirement Phase 2)

victorarias:
Phase 2 of the vt10x retirement plan (`docs/plans/2026-07-24-vt10x-retirement.md`). The three live vt10x consumers now read from the session's ghostty terminal; vt10x stays constructed and fed, so revert is flipping the call sites back. No protocol change — AttachInfo fields keep their meaning.

## Repoints (`internal/pty/session.go`)

- **Approval classifier**: `evaluateApproval` samples `ghostty.ViewportText()` instead of `screen.renderedText()` (same row shape, proven by the Phase 1 parity corpus).
- **CPR replies**: `writeCursorPositionResponse` answers from `ghostty.CursorPos()` (0-indexed viewport → 1-indexed row;col).
- **Grid/automation snapshot**: `screenSnapshot()` serves `ghostty.SerializeViewport()` (self-contained styled VT stream, viewport-only) with cursor from `CursorPos`/`CursorVisible`. The payload replays into a fresh Ghostty model — same shape `seedTile` already consumes.

## Hardening (`internal/daemon/automations_deliver.go`)

`passUnattendedLaunchGate` matched the Codex directory-trust prompt with a raw `strings.Contains` over the snapshot payload. Under the styled ghostty stream the prompt text can be split by SGR/cursor runs, so matching now strips ANSI sequences (CSI, OSC with BEL/ST, ESC-single) first. Control flow unchanged; covered by a styled/split-prompt unit test.

## Tests

- `internal/pty` sessions under test now seed a real ghostty terminal (`newTestGhostty`, stub-skip probe) alongside vt10x; CPR/approval/snapshot/attach-race tests updated.
- Snapshot tests round-trip `ScreenSnapshot` into a fresh ghostty terminal and assert viewport text + cursor match the source.

## Verification

- Native tier (darwin/arm64, head 74032f80, macOS 26.5.2): `go test ./internal/pty ./internal/ghosttyvt ./internal/daemon -count=1` all green; `go vet ./...`, `go build ./...` clean.
- Live (isolated `p2vt` profile, full `make install`, preflight PASS, protocol 182):
  - fish pane responsive across window resizes — resize-triggered CPR answered from the ghostty cursor (worker log shows the replies).
  - claude session: `pending_approval` armed and cleared back to `waiting_input` in the daemon log via the pty-layer resolver now reading `ViewportText()`.
  - `get_screen_snapshot` against live claude and codex sessions: fresh styled payload, correct geometry/cursor, decoded content matches the screen — including the codex TUI, the screen vt10x mis-parsed (see the Phase 1 corpus finding).
  - Codex's own approval dialog did not appear live (environment auto-approves shell commands); the codex prompt shape stays covered by the parity corpus fixture.

### DIFF
--- a/internal/daemon/automations_deliver.go
+++ b/internal/daemon/automations_deliver.go
@@ -711,6 +711,50 @@ func automationSessionPrompt(configuredPrompt, inputPath string, localOnlyReview
 
 const codexDirectoryTrustPrompt = "Do you trust the contents of this directory?"
 
+// stripANSIForPromptMatch removes terminal control sequences from a rendered VT
+// stream so screen prompts can be matched independently of formatting and cursor
+// positioning.
+func stripANSIForPromptMatch(data []byte) string {
+	out := make([]byte, 0, len(data))
+	for i := 0; i < len(data); {
+		if data[i] != 0x1b {
+			out = append(out, data[i])
+			i++
+			continue
+		}
+		if i+1 >= len(data) {
+			break
+		}
+		switch data[i+1] {
+		case '[':
+			i += 2
+			for i < len(data) {
+				if data[i] >= 0x40 && data[i] <= 0x7e {
+					i++
+					break
+				}
+				i++
+			}
+		case ']':
+			i += 2
+			for i < len(data) {
+				if data[i] == 0x07 {
+					i++
+					break
+				}
+				if data[i] == 0x1b && i+1 < len(data) && data[i+1] == '\\' {
+					i += 2
+					break
+				}
+				i++
+			}
+		default:
+			i += 2
+		}
+	}
+	return string(out)
+}
+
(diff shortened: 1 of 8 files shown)


figgyster (approved):
You built the Ghostty model first, then made it answer the questions that matter. Good. The Codex scroll-region bug from the Phase 1 corpus was the tell: keeping approval, CPR, and tile snapshots on vt10x would have left two terminal truths collecting interest in the dark.

The viewport round-trip witness holds the new tile payload to the same replay contract, and the ANSI-aware trust-gate match closes the one seam styled output would have opened. Daemon, grid E2E, the change check, and the head-to-head benchmark are green on this head. Approved. Phase 3 can finally evict the old parser without leaving its important readers behind.

PR #653, opened 2026-07-24
refactor(app): extract the shared ⌘P palette shell

victorarias:
## Why

The app has three hand-rolled finder overlays — the global `ActionMenu`, the in-tile `NotebookFinder`, and `LocationPicker` — each re-implementing the same shell: dimmed backdrop, text input, keyboard-driven listbox, focus on mount, Arrow/Enter/Escape, highlight clamping, scroll-into-view, backdrop dismissal, and combobox/listbox ARIA wiring.

A global markdown-file opener (⌘P over recents + a fuzzy workspace index + absolute-path navigation) is coming next. This lands the shared shell first, so that opener is a second consumer rather than a fourth copy.

No behavior change.

## What

**`app/src/components/palette/`** — the extracted shell. It owns the interaction contract and knows nothing about what rows mean; callers own the query (so a later path mode can rewrite it to descend into a directory), the ranking, and row rendering. `NotebookFinder` drops from 125 to 46 lines.

Every element carries both a shared `palette-*` class (styled once, in `Palette.css`) and a caller-namespaced `<variant>-*` class. Styles live in one place while each caller keeps a hook for its own positioning — the Notebook finder is `absolute` within a tile, a window-level palette will override that — and the Notebook finder keeps the exact class names its e2e specs and packaged-app scenarios already select on.

**`palette/rank.ts`** — the fuzzy scorer, moved from `notebook/finderRank.ts`, now scoring a minimal `FileCandidate { path, title?, updated? }` instead of `NotebookEntry`. `App.tsx` already coerced arbitrary `fs_index` results into `NotebookEntry` via `fsIndexToNotebookEntries`, and an arbitrary root is not a notebook. `rankFiles` is generic, so callers keep their own row type through ranking. Its tests move with it, unchanged apart from the renames.

**`NotebookSurface`** — collapsed a duplicate mount: the tile and modal variants built identical `<NotebookFinder>` elements from the same state.

## Scenario repairs

Two packaged-app scenarios were stale since #588 (editor tile off-root gating, Editor labels), not broken by this change:

- **`scenario-notebook-link-nav`** seeded its probe notes in the Notebook root, but ⌘⌥N now roots a docked tile at the active workspace's directory (`resolveEditorTileRoot`), so the tile's finder indexed a tree that never held them. Notes now seed into the workspace directory, matching `scenario-notebook-editor-undo` — which also stops every run littering the shared Notebook root. Its title assertion expected `"Notebook"`; a note-less editor tile is titled `"Editor"`.
- **`scenario-notebook-tile-close`** selected a bare `.notebook-finder`. A notebook tile docked in an *inactive* workspace stays mounted in the DOM, so that selector matched a stale finder forever and the "Esc closed the finder" wait could never observe absence on a profile carrying any leftover tile. Now scoped to the active wrapper like its three sibling scenarios. Its presence probe also treated every error as absence, so an `html-to-image` failure inside a CodeMirror subtree would have reported a still-open finder as closed — it now treats only "not found in DOM" as absence.

## Verification

- `tsc --noEmit` clean; frontend suite green (1981 passed).
- `e2e/notebook-tile-finder.spec.ts` + `e2e/notebook-browser.spec.ts` green (21 tests) — a real browser, so the CSS move is exercised, not just the DOM.
- Packaged dev app: `notebook-tile-close`, `notebook-tile-finder`, `notebook-link-nav`, `notebook-editor-undo` all green. Screenshot confirms the overlay renders identically.
- `tile-close` initially failed on the dev profile. It was run against unmodified `HEAD` and against this branch under matching fresh throwaway profiles: both pass, confirming the failure was profile debris and motivating the selector fix above.

## Follow-ups (not in this PR)

`docs/plans/2026-07-24-markdown-file-opener.md` carries the alignment and remaining pre-work: one path-navigation backend (`browse_directory` gains a file filter, and its missing auth gate gets closed at that point), the file-activity store, and a note that `PostToolUse` hook payloads are currently discarded — relevant when agent-edited files become a ranking signal, since the naive version adds a second process spawn per tool call.

https://claude.ai/code/session_01JgB8qSfAyiuhhwTZqy5jVm


figgyster (approved):
You've spent the week teaching the daemon to remember terminals and revive workers; now you're taking a little more patrol duty away from the human at the keyboard. Good instinct. The palette owns the fussy focus/Escape/ARIA contract, while Notebook keeps its ranking and its old harness hooks, so the markdown opener arrives as a second customer instead of the fourth tiny kingdom with its own laws.

The new witnesses cover the shared interaction seam, the extraction leaves the existing tile selectors standing, and the Frontend plus every E2E shard are green on this head. Safe to merge. The opener plan also keeps path mode and filesystem authority on a short leash, which is how these things avoid becoming a late-night filesystem tour bus.

PR #654, opened 2026-07-24
feat(opener): global ⌘P markdown file opener with daemon-side recents

victorarias:
Adds a keyboard-driven way to open a markdown document when there is no link to ⌘+click and the agent is busy. ⌘P anywhere in the app opens a palette that starts on recently opened markdown files and fuzzy-filters the selected session's markdown as you type; picking a file docks it as a reader tile bound to that session, the same tile ⌘+click and `attn open` produce. A focused Editor tile keeps its own in-tile ⌘P.

This is the second PR of the opener work. The first (#653) extracted the shared palette shell; this one is the feature. Absolute-path navigation (typing `~/…` to browse) is the next PR.

## Recents

Recorded daemon-side in `openMarkdownTile` — the single chokepoint every route already passes through — so there is no client bookkeeping and no origin flag. They live in a new `file_activity` table keyed by `(path, source)` and ranked with the existing frecency scorer (`last_at` + `count`). `source` is `"opened"` today; surfacing files an agent edited becomes a new source value plus a ranking weight rather than a migration.

`openMarkdownTile` now refuses a file that no longer exists and forgets its entry. That failed open is the only place recents are pruned: summoning the opener never stats its list.

On an empty query the list is recents. Once you type, recents and index results rank in one list with a recency bonus — no pinned section, no mode change mid-typing.

## Index

`fs_index` gains a server-side extension filter applied **before** the 25k entry cap. Previously the cap counted every file and truncated in `WalkDir` lexicographic order, so in a large repository markdown under late-alphabet directories could silently vanish from the finder.

Inside a git repository the enumeration now comes from git (`ls-files --cached --others --exclude-standard`) with a `WalkDir` fallback outside one. That honors `.gitignore` instead of a hardcoded `node_modules` + dot-directory skip list, so build output and vendored trees cost nothing, and it reads an index git already maintains rather than stat-ing the tree. Accepted consequence: a gitignored markdown file is invisible to fuzzy mode — path navigation, next PR, still reaches it.

The palette opens on recents immediately and loads the corpus asynchronously, so a cold enumeration never delays ⌘P. When the walk reports `truncated`, the empty state says the index is capped rather than implying completeness.

## ⌘P ownership

`file.open` is a normal rebindable registry entry. The global dispatcher runs in the capture phase, so instead of racing it with the notebook surface's own container keydown, a surface registers a claim (`paletteClaim.ts`) and the global handler hands ⌘P back to whichever registered surface currently contains focus.

## Protocol

Version 185: `recent_files` / `recent_files_result`, and `extensions` on `fs_index`.

## Also

`TestMigration79` rewound only its own `schema_migrations` row. The runner resumes from `MAX(version)`, so that stops re-running the migration as soon as a later one holds the watermark — fixed to rewind past it.

## Verification

- Go suite, 2002 frontend unit tests, and the browser e2e suite green.
- New: store tests for file-activity recording/ranking/pruning, `fs_index` tests for the cap-after-filter rule and for git enumeration honoring `.gitignore` while still listing untracked files, daemon tests for open-recording and the forget-on-missing path, plus frontend tests for the opener's ranking, async load, and the ⌘P claim.
- Live-verified in the packaged app on a throwaway profile with a new packaged-app scenario (`real-app:scenario-markdown-opener`), green on consecutive runs: native ⌘P opens the palette, a gitignored file is invisible, an untracked file is findable, both picks dock session-bound markdown tiles, and a re-summon lists both opens as recents with the later one first while picking one reuses its tile.

https://claude.ai/code/session_01JgB8qSfAyiuhhwTZqy5jVm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-07-24]
 
+### Added
+- **⌘P opens a markdown file from anywhere in the app.** The palette starts on
+  the files you opened most recently and, as you type, fuzzy-searches the
+  markdown under the selected session's folder — no native file dialog, no file
+  tree. Picking a file docks it as a reader tile bound to that session, exactly
+  as ⌘+clicking a link does. A focused Editor tile keeps its own in-tile ⌘P.
+  The search covers what git tracks plus untracked files, so ignored build
+  output never crowds out your documents.
+
 ### Fixed
 - **Sessions interrupted by a daemon or machine restart now recover when their terminal pane opens.** attn preserves the session's launch settings, restarts the recoverable session using the pane's current terminal size, and reconnects it automatically instead of leaving a "Failed to attach PTY" message. Sessions that cannot safely be revived still show the normal attach error.
 
--- a/app/package.json
+++ b/app/package.json
@@ -54,6 +54,7 @@
     "real-app:scenario-workspace-close-last-session-switches-back": "node scripts/real-app-harness/scenario-workspace-close-last-session-switches-back.mjs",
     "real-app:scenario-workspace-close-one-session-keeps-selection": "node scripts/real-app-harness/scenario-workspace-close-one-session-keeps-selection.mjs",
     "real-app:scenario-tile-only-workspace-select": "node scripts/real-app-harness/scenario-tile-only-workspace-select.mjs",
+    "real-app:scenario-markdown-opener": "node scripts/real-app-harness/scenario-markdown-opener.mjs",
     "real-app:scenario-notebook-tile-finder": "node scripts/real-app-harness/scenario-notebook-tile-finder.mjs",
     "real-app:scenario-notebook-tile-close": "node scripts/real-app-harness/scenario-notebook-tile-close.mjs",
     "real-app:scenario-notebook-editor-undo": "node scripts/real-app-harness/scenario-notebook-editor-undo.mjs",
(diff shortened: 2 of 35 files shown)


figgyster (changes requested):
You're paying off #653's little shell heist with a real user-facing tool now. The daemon-owned recents and the one-chokepoint recording are the right kind of boring authority; I like this direction a lot.

But this v1 says remote endpoints are deferred, while `markdownOpenerRoot` still takes the selected session's `cwd` without checking `endpointId`. With a remote session selected, `sendFsIndex` asks the *local* daemon to enumerate that remote path. Usually it fails; on matching paths it can show the local machine's files, and picking one then tries to bind it to a remote session through that same local daemon before falling back to the OS. That's a broken local/remote ownership boundary, not a cosmetic empty state.

Required before merge:

- Gate the global opener to a local active session (fall back to the local notebook root/recents for remote selection), or make the full index-and-open route endpoint-aware. Please pin the remote-selected case in a regression test. The plan already puts remote browsing in a later slice, so keeping this PR honestly local is the small repair.

The rest held up: the full Go suite passed, the three opener suites passed (20 tests), and the CI board is green. Once the endpoint seam stops lying, this can go home.

figgyster (approved):
There it is. #653 gave you one palette shell; this makes it a real opener without teaching the local daemon to impersonate a remote machine.

`resolveMarkdownOpenerTarget` keeps the boundary where it belongs: a remote cwd never reaches local `fs_index`, and a remote selection forfeits the local tile binding instead of making up ownership. The two fallback witnesses are the right ones. Focused palette coverage is green, and the full CI board came home green too.

Safe to merge. The path-navigation sequel can take its own turn in the rain.

PR #655, opened 2026-07-24
refactor(pty): delete vt10x — ghostty is the sole terminal model (retirement Phase 3)

victorarias:
Phase 3 of the vt10x retirement plan (`docs/plans/2026-07-24-vt10x-retirement.md`): delete the vt10x virtual screen. After Phase 2 repointed every consumer (approval classifier, CPR replies, grid/automation snapshot) to the Ghostty terminal, vt10x was fed but never read. This PR removes it entirely — Ghostty (libghostty-vt) is now the sole parsed terminal model.

## What changed

- **Deleted `internal/pty/snapshot.go`** (364 lines): `virtualScreen`, `renderVisibleFrame`, and the SGR/color re-encoding machinery. Dropped the `github.com/hinshun/vt10x` dependency from `go.mod`/`go.sum`.
- **Spawn-fatal Ghostty construction**: `ghosttyvt.New` failure in `Manager.Spawn` now aborts the spawn with full cleanup (close ptmx, kill+wait the child, run the shell-pane cleanup) instead of logging and continuing. A session without the Ghostty terminal has no classifier, CPR, tiles, or attach restore — it is not viable.
- **Read loop simplified**: only `blockFeed` (which feeds Ghostty) runs under `replayMu`; the dual `screen.Observe` feed is gone. The `replayMu` invariant is now just {Ghostty feed, `lastReplaySeq`} atomicity for snapshots.
- **`ReplayScreenSnapshot` moved to `manager.go`** — it remains the observer-seed transport type used by the daemon's `snapshotSeedScreen`.
- **Corpus converted from parity to goldens**: `TestGhosttyCorpusGoldens` replays the 7 real-session fixtures and compares Ghostty viewport text against checked-in `.golden` files (regenerate with `-update`). Per-fixture golden cursor positions are asserted in the test table. The tmux-refereed `codex-approval-80x24.golden` is byte-identical to the Phase 1 version (sha256 `64d2c330f2fd6b0ba2baf4e899c39a7e85e2e00e91f19b735ae3640474b4b5df`); both chunk sizes (4096/1024) are retained.
- **Stub build tag widened** to `!cgo || !((darwin && arm64) || (linux && amd64) || (linux && arm64))`: plain `GOOS=linux go build` disables cgo by default and previously had no file satisfying the package. Product Linux builds are unaffected — the Makefile `build-linux-*` targets and hub bootstrap both set `CGO_ENABLED=1` explicitly; a fresh `make build-linux-arm64` binary contains the real ghostty symbols and no stub markers.
- Docs: AGENTS.md terminal section (single terminal model, spawn-fatal contract), plan doc Phase 3 marked complete, corpus README documents the golden regime.

## Verification

- `go build ./...`, `go vet ./...`, gofmt clean; full `go test ./internal/... ./cmd/... -count=1` green, including the converted corpus goldens on native Darwin/arm64.
- Zero remaining `vt10x` references in Go code, `go.mod`, or `go.sum`.
- Cross-build check: `make build-linux-arm64` produces an ELF with 24 `ghostty_terminal_*` symbol matches and 0 stub markers.
- **Live verification** on an isolated profile (`p2vt`, fresh `make install` from this head, preflight PASS, protocol 182):
  - Session spawn works under the new spawn-fatal construction (fish shell pane + claude agent session both spawned normally).
  - fish CPR: prompt responsive across window resizes; post-resize command output rendered.
  - claude approval flow: `pending_approval` armed on a real permission dialog and cleared on approve (pty-layer resolver over `ViewportText`).
  - `get_screen_snapshot` against the live session: `fresh=true`, correct geometry (62x27), cursor position and visibility populated, decoded payload text matches the on-screen pane.


figgyster (changes requested):
You’ve got the Ghostty cutover where it ought to end up: one parsed terminal, one authority, and no vt10x ghost still taking notes in the back room. Making a failed Ghostty construction tear the spawn down instead of limping onward is the right hard line after #650 and #652 moved the real readers over.

But the new corpus goldens leave a blank line at EOF, so `git diff --check` is red. This repo calls that the first blocker, and the rule is right — generated witnesses need to be tidy before we trust them.

Required before merge:

- Remove the added blank EOF line from `claude-approval-80x24.golden:7`, `fish-resize-80x24.golden:22`, `sgr-heavy-120x40.golden:40`, `sgr-heavy-80x24.golden:24`, `vim-altscreen-120x40.golden:3`, and `vim-altscreen-80x24.golden:3`.

The Daemon and E2E checks are green. I stopped before heavier local tests because the static gate is already red; no sense interrogating the whole precinct while the first witness is still holding a muddy shoe.

figgyster (approved):
That stray tail of blank viewport rows has been folded back into the fixture generator, where it belongs. The corpus now compares the meaningful viewport and writes the same normalized shape, so the old whitespace red has an actual regression witness behind it.

Phase 3 can close its book now. #650 built the Ghostty proof and #652 moved the live readers; this deletes the second terminal model and makes a missing Ghostty terminal a failed spawn, not a session that lies about what it can see. `git diff --check` is quiet, the sandbox Go suite passed, and the full CI board is green. Approved.

PR #657, opened 2026-07-24
feat(opener): path-navigation mode in the markdown opener

victorarias:
PR3 of the markdown-opener plan (`docs/plans/2026-07-24-markdown-file-opener.md`). PR1 (#653) extracted the palette shell; PR2 (#654) shipped the ⌘P opener with recents and fuzzy search. This adds the second mode and the shared path-navigation backend (R3).

## What the user gets

A query that starts with `/`, `~/`, `./` or `../` switches the opener from fuzzy search to browsing: it lists that folder one level at a time, Enter on a folder descends into it, Enter on a file opens it as a reader tile. A leading slash is the only trigger, so `docs/plan` still fuzzy-searches. This is how a document outside the session's folder — or one `.gitignore` hides from the index — stays reachable, without a native file dialog.

Dot-folders like `.claude` are now searchable too. They hold real documents, path mode always listed them, and fuzzy mode hid them; both now show them, and `.git` is listed by neither.

## Shape

- **One directory-listing backend, not a second copy.** `browse_directory` takes an optional extension filter and reports `is_dir` per entry, so the session picker keeps its directories-only listing and the opener gets directories plus markdown. `useFilesystemSuggestions` is reused as-is; its trailing positional arguments became an options object.
- **The file-listing variant is gated.** Directory names leak only tree shape, which this command always exposed to any local client. File names are the user's documents, so asking for them now requires the authenticated app client — the same gate `resolveFsRoot` applies to an arbitrary `fs_*` root. This closes the gap the plan flagged. Plain directory browsing, including the hub's endpoint-forwarded browsing, is unchanged.
- **Relative queries expand client-side** against the opener's root; the daemon would otherwise resolve them against its own working directory.
- Protocol 186.

## Verification

- Go: new `listDirectoryEntries` tests (directories-only default, files behind the filter, symlink/`.git` exclusion, prefix filtering across kinds) and a `browse_directory` gate test covering all three cases — untrusted+files denied, untrusted+directories allowed, trusted+files allowed. Updated `fs_index` tests pin the new dot-directory rule.
- Frontend: `pathMode.test.ts` (mode detection, relative expansion, no climbing above `/`) and `MarkdownOpener` path-mode tests (lists a directory, descends without opening, opens by absolute path, ignores a slash in an ordinary query). Full suite 2021 tests green; Playwright e2e 190 green.
- Packaged app, throwaway profile `mdpath`: `real-app:scenario-markdown-opener` extended with a path-mode leg — native ⌘P, type the session directory, descend into the gitignored `build/`, open the file fuzzy mode cannot see, assert the third tile docks. Green twice in a row. `real-app:scenario-workspace-creation-shortcuts` re-run to confirm the session picker still browses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,16 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   as ⌘+clicking a link does. A focused Editor tile keeps its own in-tile ⌘P.
   The search covers what git tracks plus untracked files, so ignored build
   output never crowds out your documents.
+- **Type a path in ⌘P to browse the filesystem.** A query that starts with `/`,
+  `~/`, or `./` lists that folder one level at a time — Enter on a folder goes
+  into it, Enter on a file opens it — so documents outside the session's folder,
+  or ones `.gitignore` hides from the search, are still a few keystrokes away.
+  Ordinary queries like `docs/plan` keep searching as before.
+
+### Changed
+- **Dot-folders like `.claude` are now searchable in ⌘P.** Documents kept in
+  dot-prefixed folders were listed when browsing by path but invisible to the
+  search; both now show them. Git's own `.git` folder stays hidden.
 
 ### Fixed
 - **Sessions interrupted by a daemon or machine restart now recover when their terminal pane opens.** attn preserves the session's launch settings, restarts the recoverable session using the pane's current terminal size, and reconnects it automatically instead of leaving a "Failed to attach PTY" message. Sessions that cannot safely be revived still show the normal attach error.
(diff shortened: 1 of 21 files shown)


figgyster (approved):
You’re closing the loop #653 and #654 started: ⌘P now has a way out to the documents the index should leave alone. Good instinct keeping one `browse_directory` seam for both picker and opener, with the file-bearing route behind authenticated app identity. #654 left a chalk outline around endpoint ownership; this slice keeps the opener honestly local.

The witnesses cover the right corners: mode switching, relative-root expansion, directory descent, `.git` exclusion, the trusted-client gate, and a packaged-app scenario that reaches a gitignored Markdown file through the real native flow. Daemon, Frontend, all E2E shards, Bench A/B, and release preflight are green on this head. Safe to merge. Case closed.

PR #658, opened 2026-07-24
ci: add React Doctor scan for the app workspace

victorarias:
Adds the [React Doctor](https://www.react.doctor/ci) action, which reports security, performance, correctness, accessibility, bundle-size, and architecture findings in React code.

- **On pull requests** it scans the changed files and posts a sticky summary comment covering only what the PR introduces relative to the merge base.
- **On pushes to main** it publishes a commit status with the health score, so the trend is visible over time.
- **Advisory, not a gate.** `blocking` stays at its `none` default, so findings never fail a check. The commented reference block in the workflow lists how to graduate it to `warning` or `error` later.

`pnpm --dir app run doctor` runs the same scan locally via the new `doctor` script.

## Two adjustments to the original change

1. **Workflow path.** It was written to `app/.github/workflows/react-doctor.yml`. GitHub only reads workflows from `.github/workflows/` at the repository root, so it would never have run from there; this PR puts it at the root.
2. **`directory: app`.** The default is `"."`, and there is no package.json at the repo root — the React app is the `app/` workspace. The action's `project` input defaults to `"*"` and might have discovered it anyway, but pointing at the workspace is explicit.

## Worth a look before merging

`millionco/react-doctor@v2` is a third-party action, and this grants it `pull-requests: write`, `issues: write`, and `statuses: write` on every PR — enough to write comments and statuses on the repo. The tag is mutable; pinning to the commit SHA (`736abc183ac491b4e954fc6bedec3a9a1b73d38b`) would freeze exactly what runs. Left as `@v2` as originally written — say the word and I'll pin it.

The workflow is inert until it lands on main, so its first real run is on the next PR after merge.

### DIFF
--- a/.github/workflows/react-doctor.yml
+++ b/.github/workflows/react-doctor.yml
@@ -34,7 +34,11 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: millionco/react-doctor@v2
+      # Pinned to the commit the v2 tag pointed at on 2026-07-24. The tag is
+      # mutable and this action runs on every PR with write access to comments,
+      # issues, and statuses, so the exact code is frozen here. Bump the SHA
+      # (and the comment) deliberately to take a new version.
+      - uses: millionco/react-doctor@01820bb4fd4d0a4aebcd8df2b2a143a098649cb2 # v2
         with:
           # The React app is the `app/` workspace; there is no package.json at
           # the repo root for the default "." to find.
--- a/.github/workflows/react-doctor.yml
+++ b/.github/workflows/react-doctor.yml
@@ -1,0 +1,55 @@
+# React Doctor — finds security, performance, correctness, accessibility,
+# bundle-size, and architecture issues in React codebases.
+#
+# Docs: https://www.react.doctor/ci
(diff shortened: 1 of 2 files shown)


figgyster (commented):
You put the scanner where this repo can actually hear it: root workflow, `app` as the workspace boundary, advisory mode before anybody starts treating a new lint cop like a branch-protection sheriff. And you pinned the action to a real commit after catching the annotated-tag trap. That is the kind of small security arithmetic that keeps a convenience check from growing teeth later.

React Doctor's own first run is green, and the local launcher resolves and runs in the review sandbox. `Changes` and `Frontend` are green too. I'm holding the final approval while `Frontend E2E (1/3)` finishes its lap; the current code has no review finding from me.

figgyster (approved):
The last E2E witness came back green. This is a sensible first move for a new scanner: it sees the `app` workspace, compares PRs against their real merge base, stays advisory, and the action code is fixed to the commit you meant to trust.

The local launcher resolves too. Let it collect a little street evidence before anybody promotes it to a gate. Safe to merge. Case closed.

PR #659, opened 2026-07-24
Fix the daemon socket test flake and track flakes from CI history

victorarias:
Flaky tests in this repo surface almost only in CI, where the daemon shards, the packages job and the e2e shards compete for a 2-core runner. This fixes the top recurring one and adds a way to see the rest without relying on memory.

## The fix

`TestDaemon_StopCommand_PendingTodos_SetsWaitingInput` dialed the daemon socket after a fixed 100ms sleep — the last of 15 daemon starts in `daemon_test.go` still doing so. Under CI load the listener sometimes had not bound yet, failing with `dial unix ...: connect: no such file or directory`. Across the last 400 CI runs that was 5 failures on 4 branches including main, each on a commit that passed on a later attempt.

Its sibling `TestDaemon_StopCommand_CompletedTodos_ProceedsToClassification` already polled but with a 2s budget, and has failed CI with `socket not ready after 2s`. Both now use the 5s budget the other 13 sites use.

Verified by zeroing the sleep, which reproduces the exact CI error, then confirming the polled version survives 120 iterations under CPU contention at `GOMAXPROCS=1`.

## The ledger

`scripts/ci-flake-report.sh` harvests failing test names from CI job logs (Go, Playwright, Vitest) and classifies each failure by whether the same commit is known to pass:

- `rerun-green` — a later attempt of the same run succeeded. The code did not change, so this is the strongest flake evidence available, and it is precisely what a re-run erases from `gh run list`. Most of the ledger came from here.
- `sha-green` — another run on the same commit succeeded.
- `unresolved` — the commit never passed; likely real breakage.

Entries split on **CI runs elapsed since the last occurrence** rather than calendar days, so a landed fix shows up as dormant instead of being re-litigated. On the current window that distinction matters: `split-blank-repro` and two terminal specs account for the largest raw counts but have not recurred in 90+ runs, having been fixed by #620. Ranking by raw totals would send someone to fix tests that are already fixed.

Job logs and finished attempts are immutable and cached on disk, so re-scanning a window costs almost nothing.

`.github/workflows/flake-ledger.yml` runs it daily, keeps one `CI flake ledger` issue in sync (creating it on first run), and comments only when a test newly enters the active set, so the steady state is silent.

## What the current window shows

Still active after this PR: the `ticketOverlay` wiring-proof vitest (waits for the detail panel, then synchronously reads a select that only renders after the async fetch resolves) and the `keyboard-shortcuts` focus-trap e2e. The latter looks like it may be reporting a real product bug — `WorktreeCleanupPrompt.focusPrimary` defers `.focus()` into a `requestAnimationFrame` while its don't-steal-focus guard is evaluated a frame earlier — and is deliberately left alone here.

## Verification

`gofmt` and `go vet` clean. Report script exercised across all four output formats over a 400-run window; the workflow's ledger parsing, new-flake diff and steady-state silence were tested locally against real generated output. The workflow's write path (issue create/edit/comment) first executes on GitHub after merge; its failure mode is a red workflow run, not a corrupted issue.

https://claude.ai/code/session_01YXtaooDskfC9o5hvRZLx41

### DIFF
--- a/.github/workflows/flake-ledger.yml
+++ b/.github/workflows/flake-ledger.yml
@@ -1,0 +1,101 @@
+# Keeps a single tracking issue in sync with what CI history actually shows
+# about flaky tests. The report distinguishes a flake (the same commit is known
+# to pass) from real breakage, and marks an entry dormant once enough runs have
+# gone by without a recurrence, so a landed fix is visible rather than assumed.
+name: Flake ledger
+
+on:
+  schedule:
+    # Daily, after the overnight runs have settled.
+    - cron: '0 13 * * *'
+  workflow_dispatch:
+    inputs:
+      limit:
+        description: 'How many recent CI runs to scan'
+        required: false
+        default: '400'
+
+permissions:
+  actions: read
+  issues: write
+
+concurrency:
+  group: flake-ledger
+  cancel-in-progress: false
+
+jobs:
+  ledger:
+    name: Update flake ledger
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      # Job logs and finished attempts are immutable, so the report's on-disk
+      # cache carries across days and only new runs cost API calls.
+      - name: Restore flake-report cache
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/attn-ci-flakes
+          key: flake-ledger-${{ github.run_id }}
+          restore-keys: flake-ledger-
+
+      - name: Generate ledger
+        env:
+          GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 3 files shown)


figgyster (approved):
Approved. The daemon tests quit guessing when the listener exists and join the five-second polling contract already used around them, so a slow shared runner no longer gets to decide the verdict with a 100 ms coin toss.

This carries the evidence-first performance work in #656 into reliability: the ledger calls a test flaky only when the same SHA later passes, leaves unresolved failures as real suspects, and measures dormancy in completed runs rather than weekends. `Changes`, `Daemon`, and both Linux preflights are green; the unrelated, non-gating Bench A/B job is still running.

PR #660, opened 2026-07-24
refactor(pty): split snapshot and attach payload families (retirement Phase 4)

victorarias:
Phase 4 of the vt10x retirement plan (`docs/plans/2026-07-24-vt10x-retirement.md`): post-retirement payload cleanup. Since Phase 3, `AttachInfo`'s two snapshot families were perfectly disjoint per method — `Attach` fills only the Ghostty dump + blocks, `Snapshot` fills only the viewport frame — so every RPC dragged the other family through four layers as permanently-zero plumbing. This PR splits them and slims the wire.

## What changed

- **Per-method payload types**: `Snapshot` now returns `pty.SnapshotInfo` (LastSeq/Cols/Rows/Running + nullable `ViewportSnapshot{Payload, Cols, Rows}`) end to end — pty, worker RPC, ptybackend interface, daemon. `AttachInfo` keeps only the attach family (Ghostty dump, blocks, scrollback-truncated). `ReplayScreenSnapshot` is renamed/absorbed into `ViewportSnapshot`.
- **Worker RPC split**: `MethodSnapshot` gets its own lean `SnapshotResult` struct instead of sharing `AttachResult`; `AttachResult` drops all `screen_*` fields. Existing graceful degradation is preserved: an absent `screen_snapshot` leaves the observer unseeded (worker-version skew, or no frame yet).
- **Wire slimming (ProtocolVersion 182 → 183)**: deleted `screen_cursor_x`, `screen_cursor_y`, `screen_cursor_visible`, and `screen_snapshot_fresh` from `get_screen_snapshot_result` and `WebSocketEvent`, plus the vestigial raw-replay `scrollback` string field. The styled VT payload is self-contained — cursor state rides inside the dump, and freshness is payload presence. No consumer read any of the deleted fields (frontend `ScreenSnapshotResult` reads only payload/cols/rows/last_seq); the legacy daemon web fallback's reads of `scrollback`/`screen_snapshot_fresh` were updated to match.
- **Daemon**: `snapshotSeedScreen` helper deleted (collapsed to a nil-check on `SnapshotInfo.Screen`); its tests replaced by handler-level tests (`handleGetScreenSnapshot` seeds from a viewport payload / leaves the observer unseeded without one). The automations unattended-launch gate reads the new `Screen.Payload` shape.
- Three-file lockstep followed: `main.tsp` edit → `make generate-types` → `ProtocolVersion` 183 in `constants.go` and `PROTOCOL_VERSION` '183' in `useDaemonSocket.ts`.
- Net −73 lines.

## Verification

- gofmt clean; `go build ./...`, `go vet ./...`, full `go test ./internal/... ./cmd/... -count=1` green; `pnpm --dir app exec tsc --noEmit` green; frontend suite green (1977 tests). Added `SnapshotResult` JSON round-trip coverage.
- Repo-wide sweep: zero remaining references to `ScreenCursor*`, `ScreenSnapshotFresh`, `screen_cursor_*`, `screen_snapshot_fresh`, or the raw `scrollback` field.
- **Live verification** on an isolated profile (`p2vt`, fresh `make install` from this head; preflight PASS, CLI/app/daemon all protocol 183):
  - Session spawn + live typing render normally.
  - `get_screen_snapshot` against the live session returns a self-contained payload whose decoded text contains the on-screen marker, with correct geometry (62x27).
  - The result message carries none of the retired fields on the wire.

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -118,10 +118,6 @@ type WebSocketEvent = GeneratedWebSocketEvent & {
   screen_snapshot?: string;
   screen_rows?: number;
   screen_cols?: number;
-  screen_cursor_x?: number;
-  screen_cursor_y?: number;
-  screen_cursor_visible?: boolean;
-  screen_snapshot_fresh?: boolean;
   last_seq?: number;
   cols?: number;
   rows?: number;
@@ -171,7 +167,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '182';
+export const PROTOCOL_VERSION = '183';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 // AutomationActionTimeoutError distinguishes "the daemon never sent a
@@ -2219,7 +2215,7 @@ export function useDaemonSocket({
                 pendingActionsRef.current.delete(key);
                 if (data.success) {
                   const result: ScreenSnapshotResult = {
-                    screenSnapshot: data.screen_snapshot_fresh ? data.screen_snapshot : undefined,
+                    screenSnapshot: data.screen_snapshot ?? undefined,
                     screenCols: data.screen_cols,
                     screenRows: data.screen_rows,
                     lastSeq: typeof data.last_seq === 'number' ? data.last_seq : 0,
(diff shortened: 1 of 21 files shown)


figgyster (approved):
You took the last two snapshot families that had been sharing a trench coat since Phase 3 and gave each one its own badge. Good. Attach now carries restore truth, observer snapshots carry just the viewport and its watermark, and the empty-frame/skew path stays honest instead of inventing a payload.

The Phase 1 corpus, Phase 2 consumers, and Phase 3 deletion all pointed here. This is the payoff: Ghostty is the one terminal authority, and the wire no longer remembers a parser that is already buried. The handler-level no-frame witness and the live protocol-183 run are the bits I read twice. Daemon, frontend, every E2E shard, and Bench A/B are green. Approved.

This looks like the curtain call for the vt10x-retirement arc. Is the plan complete after this, or is another cleanup beat still coming? Either way, this one can go home.

PR #661, opened 2026-07-24
docs(plans): mark vt10x retirement Phase 4 complete

victorarias:
Records Phase 4 (PR #660, ProtocolVersion 183) as shipped in the retirement plan doc and marks the plan complete. All four phases are merged to the epic: #650, #652, #655, #660. Docs-only.

### DIFF
--- a/docs/plans/2026-07-24-vt10x-retirement.md
+++ b/docs/plans/2026-07-24-vt10x-retirement.md
@@ -137,13 +137,23 @@ viewport-only as a follow-up.
 - Documented the single parsed-terminal contract and spawn-fatal Ghostty
   construction on supported platforms.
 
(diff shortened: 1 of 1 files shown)


figgyster (approved):
You closed the last page of an actual four-PR terminal migration instead of leaving a planning document to rot behind the code. I checked the Phase 4 account against the epic: the lean observer snapshot and attach payload split are there, and the retired cursor/freshness fields really are gone. That makes the single Ghostty authority legible to whoever catches this case next winter.

One bit of tombstone dust remains in the code-smells list: the two snapshot families still say "Phase 4" as if the work were pending. Worth putting that sentence in the past tense whenever you next touch the plan. It doesn't hold this archival close-up. The docs-only checks are green, `git diff --check` is quiet, and the retirement arc #650 → #652 → #655 → #660 now has its ending. Approved.

PR #662, opened 2026-07-24
Server-authoritative terminal restore

victorarias:
Lands the server-authoritative terminal epic. Restore stops being a
best-effort byte replay and becomes a serialization of the terminal the
daemon already parsed.

## What changes

**Restore is server-authoritative.** The worker keeps a real parsed
terminal (libghostty-vt) and serializes it — grid, scrollback, cursor,
modes, both primary and alternate screens — as the sole attach payload
(`attach_result.snapshot`). The client resets a fresh model, resizes to the
snapshot grid, and writes the dump with responses suppressed. The raw
replay machinery it replaced is deleted: RingBuffer scrollback, ReplayLog
and its segment machinery, the replay-vs-snapshot decision tree, the
screen-snapshot derivation, and the agent raw-replay bounding.

There is no fallback. A snapshot-less attach (the pure-Go stub on
non-macOS, or a ghostty construction failure) keeps whatever the client
already has and dedups the live stream against `last_seq`, rather than
blanking the terminal.

**OSC 133 blocks survive a restore.** A marker-stripped dump rebuilds no
command blocks, so the worker owns the block table and carries it beside
the dump as structured `attach_result.snapshot.blocks`. The frontend seeds
`TerminalBlockStore` from them after the dump write.

**vt10x is gone.** ghostty is the sole terminal model, including for the
state classifier, CPR replies, and grid-tile seeding. Parity was
established against a corpus of recorded real sessions before the switch,
and ghostty construction is now spawn-fatal rather than silently degrading.

**Native VT library.** `internal/ghosttyvt` links libghostty-vt via cgo on
Darwin/arm64 and Linux; every other platform compiles a pure-Go stub. The
archive is downloaded prebuilt by default and built from pinned source when
needed.

Restores are text-only: sixel/kitty images render from the live stream but
are not preserved across a restore. That is accepted; image-preserving
restore is a follow-up.

## Protocol

ProtocolVersion 187. `AttachResultMessage` drops `scrollback`,
`scrollback_truncated`, `replay_segments`, and `screen_*`; it gains
`snapshot` (`AttachSnapshot` + `AttachBlock`). `get_screen_snapshot` keeps
its own `screen_*` fields — separate command, unchanged.

## Merge with main

main moved session recovery into the daemon while this epic was open, so
the sync resolved several overlapping seams. Both protocol additions are
kept (`reload_session_result` and `revived` alongside the snapshot family);
`pty.Manager` keeps main's spawn reservation but loses the scrollback
parameter the epic deleted; two frontend tests were dropped because each
side had already deleted the code path the other's test covered. Generated
files were regenerated from the merged schema rather than hand-merged, and
`make check-types` reproduces them byte-for-byte.

## Verification

`go build`, `go vet`, full `go test ./...`, `make check-types`, and the
frontend suite (2042 tests / 190 files) are green on the merged head.

Live verification on a throwaway `epicmerge` profile, native tier
(Darwin/arm64, real cgo libghostty-vt), all on this head:

- `preflight` — PASS, protocol 187 consistent across CLI, app, and daemon
- `scenario-tr402-local-claude` — split open/close redraw, main pane
  recovers 560→280→560 with content preserved
- `scenario-tr401` — window resize with main + utility panes restored
- `scenario-recoverable-auto-revive` — kill the worker, restart the daemon,
  reopen the app: the session auto-revives and reattaches with no
  attach-failure banner. This exercises main's daemon-owned revive and the
  epic's snapshot restore together, which is the seam the merge created.
- `scenario-terminal-block-copy` — OSC 133 block selection and copy


figgyster (changes requested):
You're cashing the whole server-authoritative-terminal cheque in one go. #655 and #660 got rid of the competing terminal models and duplicate observer payloads, so this is the moment that arc either earns its keep or starts charging interest. The restore contract, the snapshot-only client path, and the Linux preflight all look like the right shape. CI is green across the board.

But I can't sign this head yet. The new native block ownership has two holes, and `git diff --check` is now screaming into a rain barrel full of raw terminal fixtures.

Required before merge:

- `internal/pty/session.go`: `closePTY` frees the block table and then the Ghostty terminal without `replayMu`, while `info()` resolves `TrackedRef.ScreenPoint()` under that lock. `Manager.Remove` first hands an already-looked-up session to an in-flight attach, then can close it, so an attach snapshot can race `blockFeed.close()`/`ghostty.Close()` and read or free the same native ref. `resize()` can also reflow the terminal outside that session boundary. Put feed, snapshot/ref resolution, resize, and final block/terminal teardown behind one lifetime lock, then add an attach/remove and resize/snapshot regression. Native handles don't forgive a little optimism.

- `internal/pty/blocktable.go`: duplicate or redraw OSC 133 markers leak their old tracked refs. A second `A` replaces `bt.pending` without releasing it, while repeated `B` or `C` overwrites `inputRef`/`outputRef` without a release. The corpus has no duplicate-marker case, so `Close()` only proves cleanup of the last retained references. Retire the replaced state and add leak-count coverage for repeated `A`, `B`, and `C` markers.

- The new byte-exact corpus files intentionally contain terminal padding, but `git diff --check origin/main...HEAD` reports 489 whitespace errors across `internal/pty/testdata/corpus/*.bytes`. Give those raw-output paths a focused `-whitespace` attribute or another byte-safe representation so the normal diff check can see real whitespace defects again.

figgyster (approved):
Approved.

You took the red review seriously, then found the extra PTY-close race while you were in the alley. `replayMu` now owns the whole native lifetime that matters: feed, snapshot, reflow, query reads, and final teardown. Replaced OSC 133 refs are retired, and the one whitespace waiver is confined to byte-exact terminal transcripts. Good witnesses too: the focused block and attach/snapshot suite passed in the sandbox, `git diff --check` is quiet again, and CI is green through the native Linux preflights and every frontend lane.

#655 and #660 cleared out competing terminal truths. This is the payoff: the daemon owns the terminal and an attach gets one coherent account of it. Old red review withdrawn. Case closed.

PR #663, opened 2026-07-24
feat(opener): surface agent-edited markdown in the ⌘P opener

victorarias:
A markdown file that Claude Code or Codex creates or edits now appears in the ⌘P opener without having been opened first, so a document an agent produced mid-turn is one shortcut away instead of something to go find. This is the last deferred item from the markdown-opener plan (`docs/plans/2026-07-24-markdown-file-opener.md`), which reserved the `source` column for exactly this and left the hook seam (R5) as a note.

## One hook, not two

Both agents' catch-all `PostToolUse` matcher previously ran `_hook-state … working` and discarded the payload. It now runs `_hook-tool-use`, which sets the working state as before **and** records any markdown the call wrote from the same payload. A second hook entry would have added a process spawn to *every* tool call an agent makes; this pays nothing extra.

Coverage was derived from real captured payloads (Claude Code 2.x, codex-cli 0.145.0), not from docs:

- **Claude** names the tool (`Edit`/`Write`/`MultiEdit`/`NotebookEdit`) and carries the path in `tool_input.file_path` (`notebook_path` for notebooks).
- **Codex** routes every file write through `apply_patch`, so the paths come from the `*** Add File:` / `*** Update File:` / `*** Move to:` headers of `tool_input.command`. A rename is recorded at its destination; a deletion is not recorded at all.
- **Not covered, deliberately:** a file rewritten through the shell (`sed -i`, a heredoc) arrives as an ordinary `Bash` call with no attributable path. Parsing shell commands for file writes is guesswork.

## Ranking

Edits land in the existing file-activity log as `source = "edited"` — the value the table was shaped for, so no migration. `recent_files` now:

- **merges sources into one entry per file** (counts add, the most recent source names the entry, an open wins a same-second tie), so a file that was both opened and edited cannot appear twice;
- **weights an edit at 0.6 of an open** of the same age and frequency — the agent chose the file, the user did not;
- **takes a `root`** (the opener's fuzzy root) and gives files under it a 1.5× bonus, so the workspace in front of you ranks first without other projects disappearing.

## Protocol 187

`files_edited` (session-scoped socket command) and `root` on `recent_files`. The daemon re-checks that a reported path is absolute and markdown rather than trusting the hook — the hook already filters, but the socket is a public surface and the store is the daemon's to keep honest.

## Verification

- Go unit tests: payload extraction (both agents, renames, deletions, relative-path resolution, non-markdown, malformed input), merged/weighted/workspace-bonus ranking, the daemon's own filter, and both agents' generated hook configs.
- Frontend unit tests and the full Playwright e2e suite (190 passed, 1 skipped) after the protocol bump.
- **Packaged app**, throwaway profile: `real-app:scenario-markdown-opener` gained `agent_edits_surface_without_being_opened`, which drives the real hook binary with both agents' captured payloads and asserts the files appear in ⌘P having never been opened. The fixtures are gitignored, so fuzzy mode cannot be what surfaces them. A `.go` edit in the same step must not appear. Green twice.
- **Live agents**, same profile: real Claude and real Codex sessions each asked to write a markdown file; both landed an `edited` row in `file_activity`. The Codex leg is the load-bearing one — its hook only runs if the generated `trusted_hash` still matches the command, which changed here.

One incidental fix: the scenario now pins `ATTN_SOCKET_PATH` to the profile under test when invoking the hook. The harness selects a profile with `ATTN_HARNESS_PROFILE` while the CLI routes on `ATTN_PROFILE`, so an unpinned hook call reached the production socket.


figgyster (approved):
You've finished the opener run #653, #654, and #657 set up without starting a second hook kingdom. Making the existing PostToolUse process carry both working-state recovery and the file signal is the right bit of thrift. The parse boundary stays narrow, and an edited file remains a weaker clue than a human open.

The path holds end to end: captured Claude/Codex shapes, daemon-side Markdown filtering, per-path source merging, and workspace-aware ranking. The new packaged harness crosses the hook, socket, and opener seam, and the full CI board is green. Safe to merge. The remaining opener weather can wait outside.

figgyster (approved):
You've finished the opener run #653, #654, and #657 set up without starting a second hook kingdom. Making the existing PostToolUse process carry both working-state recovery and the file signal is the right bit of thrift. The parse boundary stays narrow, and an edited file remains a weaker clue than a human open.

The path holds end to end: captured Claude/Codex shapes, daemon-side Markdown filtering, per-path source merging, and workspace-aware ranking. The new packaged harness crosses the hook, socket, and opener seam, and the full CI board is green. Safe to merge. The remaining opener weather can wait outside.

PR #664, opened 2026-07-25
refactor(pty): carry state observations instead of bare state names

victorarias:
First of five enabling refactors from
`docs/plans/2026-07-25-agent-state-detection.md` (added here). Behavior-preserving
groundwork; no user-visible change.

## Why

The PTY layer reported state to the daemon as a bare string through
`session.onState`, so the daemon learned *what* a source claimed but never which
source claimed it, why, or when it was observed. That is enough to obey a claim
and not enough to arbitrate between claims — which is the structural reason a
session color can get stuck: every source is a last writer, and a dropped edge
leaves a state with nothing that will ever revisit it.

## What

- New `pty.Observation{Source, Claim, Detail, At}`. Sources name what actually
  produced the claim: `screen` (the scraper), `approval` (the approval resolver),
  `worker_info` (the worker's cached-state replay), and `unknown` for a claim that
  arrives unattributed.
- Threaded through `session.onState`, `Manager.SetStateHandler`, both
  `ptybackend` backends, and `handlePTYState`, which now logs the attribution.
- State crosses a socket, so widening the in-process callback is not enough.
  `EventStateChanged` gains optional `state_source` / `state_detail` /
  `state_observed_at`. **No RPC version bump**, following the `MethodSnapshot`
  precedent documented in `internal/ptyworker/protocol.go`: a worker predating the
  fields omits them and the daemon reads an unknown-source observation timestamped
  on arrival.

Also commits the throwaway spike that produced the plan's measurements (all three
files skip unless their env var is set; the plan lists them for deletion once
Phase 0 instrumentation supersedes them).

## Verification

- `internal/ptyworker/state_observation_test.go` covers the round trip through
  real JSON and the legacy-worker path. Both asserts mutation-checked: perturbing
  the observed-at format and hardcoding the source each fail on the intended line.
- `make test` green, `make test-frontend` green (2048 tests).
- **Live**, full install on throwaway profile `statedet`, real Claude session via
  `real-app:scenario-tr402-local-claude`. States still flow and now arrive
  attributed:
  ```
  pty state update: session=43d04c94… agent=claude state=waiting_input \
    source=screen detail="screen scrape" observed_at=2026-07-25T15:02:32.730678+02:00
  ```

## Debt this surfaced (documented in the plan, not fixed here)

The live run proved one item rather than inferring it: the worker fabricates
`working` when it does not know its state. The *first* observation the daemon logs
for every fresh session is `state=working source=worker_info detail="watch
subscribe replay"` — including for `agent=shell`, which has no detector and
therefore cannot have been observed working. Invisible today because a spawn is
green anyway, but every session's evidence table would open with a false row.

Three more are listed under "Found while implementing" in the plan: the same state
stream is deduped twice with different rules; that dedup is why
`approvalResolver` re-emits a `pending_approval` it never observed; and
`handlePTYState` discards `obs.At` in favour of the store's clock.

### DIFF
--- a/docs/plans/2026-07-25-agent-state-detection.md
+++ b/docs/plans/2026-07-25-agent-state-detection.md
@@ -1,0 +1,656 @@
+# Plan: near-perfect agent state detection
+
+## Goal
+
+Make session colors correct essentially all the time for Claude Code and Codex, so
+that a **settled / unsettled** attention mode can be built on top of them. Today's
+detection is right most of the time but has three reproducible failure classes:
+
+1. **Stuck states.** A state can persist forever with no live evidence behind it.
+2. **Background work.** A Claude turn that yields with `attn wait-pr` (or any
+   background Bash/Workflow) outstanding is governed by one binary rule.
+3. **Guardian flash.** Codex auto-approve routes every approval to its
+   `auto_review` guardian LLM, but attn renders the guardian's latency as a
+   user-facing `pending_approval` (yellow) for a second or two on every Bash call.
+
+Phases 0–3 are about the colors themselves and must stand on their own: they are
+worth shipping even if no attention mode ever exists.
+
+**Attention mode sits on top and is opt-in.** It is a working mode the user
+enables, off by default at release — a different projection of the same resolved
+state, never a second source of truth. The end state, with it enabled, is an agent
+that unsettles itself when it is stuck, done-and-stale, or has a question, and
+stays quiet otherwise. With it disabled, nothing about the sidebar changes except
+that the colors are right.
+
+## Why the current design produces these failures
+
+Every source calls `applyState` directly, and every source is **edge-triggered**.
+There is no level signal, no evidence TTL, and no periodic re-evaluation.
+
+```text
+Current (last-writer-wins, all edges):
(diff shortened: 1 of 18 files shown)


figgyster (changes requested):
You took the first shovel to the stuck-state graveyard and found the right thing: a bare state name was never enough evidence. The attribution survives the worker wire, preserves the old worker path honestly, and leaves the actual resolver fight for the next slice. That restraint matters.

One body is still on the pavement. The required Daemon check stops at `internal/pty/spike_state_replay_test.go:316`: the `spikeScore` fields are not gofmt-formatted. Please run `gofmt -w internal/pty/spike_state_replay_test.go` and push the result. Then this groundwork has its alibi.

figgyster (approved):
That was the right-sized return trip. The spike fixture is formatted, the Daemon lane ran all the way through, and this evidence-carrying seam now has a real witness instead of a promise.

You’re setting up the state resolver after #655, #660, and #662 finally gave the daemon one terminal truth. Source, reason, and observation time are the missing fingerprints for deciding which state claim gets believed. Safe to merge. Case closed.

PR #665, opened 2026-07-25
refactor(agent,pty): name the screen detector on the driver, collapse the two

victorarias:
Second of five enabling refactors from
`docs/plans/2026-07-25-agent-state-detection.md`. Stacked on #664.
Behavior-preserving; no user-visible change.

## The driver named its detector twice

`Capabilities.HasStateDetector` said "this agent gets PTY state detection" and a
`switch agent { case "copilot": ...; case "claude": ... }` in
`internal/pty/manager.go` said which implementation that meant. Nothing tied them
together: a driver that set the boolean without a matching case got **no detector
at all**, silently, with no error.

Replaced with a `ScreenDetector` kind the driver declares, which `manager.go`
builds exactly as named.

The env override keeps working. `STATE_DETECTOR=0` disables the detector;
`STATE_DETECTOR=1` keeps whichever kind the driver declares — which is what it
already did in every reachable case, since forcing the old boolean on for an agent
with no case in the switch produced no detector either.

## The two detectors were the same object with different opinions

`copilotStateDetector` and `claudeWorkingDetector` each held a tail, a last state,
and a working-pulse timestamp; each trimmed the tail to 2000 chars; each
suppressed a repeat claim and re-emitted a sustained `working` after an interval.
Only the classification differed.

The mechanics collapse into one `screenDetector`; the per-agent part stays as a
`screenPolicy` (`classifyClaudeScreen` / `classifyCopilotScreen`). This is what
makes Claude's classifier a single deletion later rather than an unpick of a
shared 560-line file.

`claim()` reproduces both previous dedup/pulse paths exactly, including copilot
stamping the pulse clock on a first transition into `working`.

## Verification

All pre-existing detector tests pass unchanged — they are the safety net for the
collapse. Two new tests cover what the collapse newly risks:

- **the typed-nil trap**: the session wires its state callback on `detector != nil`,
  so returning a nil `*screenDetector` as a non-nil `stateDetector` would hand a
  detector-less agent a silent one. Mutation-checked by returning
  `(*screenDetector)(nil)`.
- **the now-shared claim suppression**: a settled state reported once must not
  re-emit. Mutation-checked by forcing the dedup branch.

### DIFF
--- a/docs/plans/2026-07-25-agent-state-detection.md
+++ b/docs/plans/2026-07-25-agent-state-detection.md
@@ -525,10 +525,14 @@ being fixed as part of the current step.
       we care about — ~6s after a permission request, and exactly 60s after a Stop
       that settled idle — so it is a first-class source, not a backstop.
 - [ ] Persist `ReviewerInLoop` on the session at spawn.
-- [ ] Enabling refactors 2 and 3: driver-owned observers (drop the capability
-      booleans and the agent-name switch in `manager.go`), and one `screenDetector`
-      in place of the two identical structs. Needed here because this phase adds a
-      third observer per agent.
+- [x] Enabling refactors 2 and 3: driver-named observers (`Capabilities.
+      ScreenDetector` kind replaces `HasStateDetector` plus the agent-name switch
+      in `manager.go`), and one `screenDetector` + `screenPolicy` in place of the
+      two near-identical structs. The per-agent *meaning* stays in
+      `classifyClaudeScreen` / `classifyCopilotScreen`; only the mechanics (tail,
+      claim suppression, working pulse) are shared. `STATE_DETECTOR=0` still
+      disables; `=1` keeps the driver's kind, which is what it already did in every
+      reachable case.
 - [ ] Enabling refactor 4: the Stop hook reports `background_tasks` /
       `session_crons` as facts; `nonTerminalStopState` and `sessionIsChiefOfStaff`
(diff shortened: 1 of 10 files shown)


figgyster (commented):
You’re doing the unglamorous work the resolver phase needs: the driver now owns the detector fact, and the shared tail/claim/pulse machinery stops Claude and Copilot from growing two nearly identical little crimes in different alleyways. I checked the old and new wiring: Claude still names Claude, Copilot still names Copilot, Codex stays out of PTY screen detection, and the `STATE_DETECTOR=0` escape hatch still removes the detector.

I can’t give this an approval yet. The live board has React Doctor and the non-gating Bench A/B result, but no daemon/PTY principal witness, and this repo requires a non-production live verification for PTY state changes. Please attach that check — one Claude and one Copilot session exercising detector attach/pulse/disable is enough — then I’ll close the file quickly.

figgyster (approved):
#664 started carrying observations instead of naked state names. This makes the next move possible: the driver owns which screen observer exists, and the PTY manager no longer keeps a second little book of the same facts. That split-brain switch was a debt with a knife in it.

The full `statedet` app run proves Claude and Copilot still attach through their declared detectors, keep their working pulses, and honor the disable path. Daemon, Changes, React Doctor, and Bench A/B all came back green on `cfdbb5e`; the rebase itself is only the spike fixture's gofmt repair. Safe to merge. Case closed.

PR #666, opened 2026-07-25
refactor(daemon): decide stop terminality and classification in the daemon

victorarias:
Refactors 4 and 5 of five from `docs/plans/2026-07-25-agent-state-detection.md`.
Stacked on #665. Behavior-preserving; **protocol 189**.

## The Stop hook was deciding terminality

A Stop that yields with background work in flight (`attn wait-pr`, a Workflow) or
parks on a scheduled wakeup is not the end of a turn. The hook binary decided
that and reported a **state name**. To decide it, `runHookStop` opened a socket to
the daemon and queried the session list to learn whether the session held the
chief-of-staff role — asking the daemon for an input so it could send the daemon a
conclusion.

`StopMessage` now carries the facts: `background_task_statuses` (verbatim harness
strings) and `pending_session_crons`. The precedence rules move to
`internal/daemon/stop_terminality.go`; the chief lookup becomes the in-process
`d.isChiefOfStaffSession` call it should always have been.

This is the one refactor in the series that is not purely structural: the resolver
in Phase 2 **cannot see background work at all** while the CLI collapses it into a
state string before it crosses the socket.

Both fields are optional, protocol bumped to 189. A hook binary predating them
reports neither fact and the stop reads as terminal — exactly what such a hook
already caused by not sending the state update. The non-terminal branch sits at
the top of `handleStop`, before resume-id capture, narration enqueues, and
classification, preserving the old flow where a non-terminal stop never reached
`handleStop` at all.

## classifySessionState was a resolver written as control flow

~130 lines, six `applyState` exits interleaved with capability gates, todo
counting, transcript path resolution, file IO, and a 30-second LLM call. The
decision could only be tested by standing up a daemon.

Split into three pure rules in `classify_decision.go` (`classifyPreTranscript`,
`classifyPostTranscript`, `classifyVerdict`) plus an IO shell that performs the
reads between them and owns a single store write. Reason labels keep the names the
old inline log lines used, so a log search for a past incident still finds them.

Behavior-preserving throughout: `clearClassifyingTurn` still only defers on a
successful read, `setClassifiedTurnID` still only fires on the classifier path, and
a no-new-assistant-turn read still writes nothing rather than overwriting a good
state with `unknown`.

## Verification

`make test-all` green. Tests moved with the decisions: `cmd/attn` keeps the corpus
of real captured Claude Stop payloads (now asserting fact extraction), the daemon
gets the precedence table, a legacy-hook case, and two socket-level wiring tests.
All rules mutation-checked individually.

**Live**, profile `statedet`, full install:

1. Real Claude session via `real-app:scenario-tr402-local-claude` — protocol 189
   connects, and the new decision logging works end to end:
   ```
   handleStop: session=9bad6afc…
   classifySessionState: session=9bad6afc… no state applied reason=no_new_assistant_turn
   classifySessionState: session=9bad6afc… state=idle reason=classifier
   ```
2. Both non-terminal branches through the **real production path** — a real
   captured Claude Stop payload piped into the bundled `attn _hook-stop`, over the
   socket, to the daemon:
   ```
   handleStop: non-terminal stop session=live-bg state=working   background_tasks=1 pending_crons=0
   handleStop: non-terminal stop session=live-bg state=scheduled background_tasks=0 pending_crons=1
   ```
   with `query` confirming the session read `working` then `scheduled`.

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -169,7 +169,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '188';
+export const PROTOCOL_VERSION = '189';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 // AutomationActionTimeoutError distinguishes "the daemon never sent a
--- a/app/src/types/generated.ts
+++ b/app/src/types/generated.ts
@@ -4382,9 +4382,11 @@ export enum StateMessageCmd {
 }
 
 export interface StopMessage {
-    cmd:             StopMessageCmd;
-    id:              string;
-    transcript_path: string;
+    background_task_statuses?: string[];
+    cmd:                       StopMessageCmd;
+    id:                        string;
+    pending_session_crons?:    number;
+    transcript_path:           string;
     [property: string]: any;
 }
 
@@ -11237,8 +11239,10 @@ const typeMap: any = {
         { json: "state", js: "state", typ: "" },
     ], "any"),
     "StopMessage": o([
+        { json: "background_task_statuses", js: "background_task_statuses", typ: u(undefined, a("")) },
         { json: "cmd", js: "cmd", typ: r("StopMessageCmd") },
         { json: "id", js: "id", typ: "" },
+        { json: "pending_session_crons", js: "pending_session_crons", typ: u(undefined, 0) },
         { json: "transcript_path", js: "transcript_path", typ: "" },
     ], "any"),
     "SubscribeGitStatusMessage": o([
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -87,74 +87,6 @@ type sessionCron struct {
 	Prompt    string `json:"prompt"`
 }
 
-// hasActiveBackgroundTask reports whether the Stop payload still has background
-// work in flight. Such a Stop is a yield, not a terminal stop: the agent will
-// auto-resume when the work completes, so the session should stay "working"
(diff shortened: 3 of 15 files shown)


figgyster (approved):
You’re moving the Stop hook out of the business of declaring the turn over, and that’s the right jurisdictional eviction. #664 began carrying state observations instead of verdicts; this gives the daemon the background-work and cron facts it needs before Phase 2 tries to arbitrate them. The in-process chief lookup and the single classification write are the parts I read twice. They hold.

The decision tables keep the old precedence honest, including the quiet chief exception and version-skew terminal path. The protocol guard is at 189, React Doctor and the native Bench A/B lane are green, and the focused decision witnesses passed here. The full daemon suite needs the Ghostty/SQLite toolchain missing from this review box, but there’s no loose wire in this diff.

This looks like the last bit of scaffolding before the evidence resolver gets the keys. Phase 2 is still the next stop, yeah? Safe to merge.

figgyster (commented):
The follow-up is plan-only, and it names two things the resolver work needs to face without blinking: the default backend currently eats the working keepalive, and Copilot’s screen markers may have gone stale. Good excavation. Both are evidence problems, not reasons to bloat this refactor.

`git diff --check` is quiet and React Doctor is green on `ce3809f`. Bench A/B is still running, so I’m leaving this as the current-head signal until the native daemon witness arrives.

figgyster (approved):
The native Bench A/B witness is in now, and it backs the seam this refactor is cutting: Stop facts reach daemon arbitration without the hook playing judge. The late plan notes about a swallowed keepalive and stale Copilot markers belong to the resolver chapter ahead, where they can be fixed with evidence instead of folklore.

React Doctor and Bench A/B are green on `ce3809f`. Safe to merge.

PR #668, opened 2026-07-25
feat(state): record every state observation and explain it

victorarias:
Phase 0 of the state-detection plan (`docs/plans/2026-07-25-agent-state-detection.md`).
Instrumentation only — nothing here changes which state a session ends up in.

## The problem

A stuck session color is invisible in `daemon.log`. The daemon logs the
transitions it *accepts*, and a stuck session is precisely one where nothing is
being accepted. Three very different failures leave an identical log:

- the right observation never arrived;
- it arrived and a driver's transition filter refused it before the store door;
- it arrived and the store's own commit rule refused it.

"Sometimes the colors get stuck" has been unfalsifiable for exactly this reason.

## What this adds

`internal/statetrace`: a capped (256) in-memory ring per session holding every
state observation the daemon acts on, with what became of it.

| outcome | meaning |
|---|---|
| `applied` | committed to the store |
| `vetoed` | rejected before `applyState` — a driver's `ShouldApplyPTYState`, a plugin-owned session, an unknown session |
| `discarded` | reached `applyState`; the store's commit rule refused it |
| `skipped` | the source looked and had no claim to make |

`attn state explain <id> [--json]` replays it over a new `state_explain` command
(protocol 190). Every recorded observation is also mirrored to `daemon.log` as a
`state trace:` line, so a session that has since been removed still leaves its
trace behind.

`sessionStateChange` grows an optional `origin{source, detail, observedAt}`.
Several sources share one cause — every trusted PTY observation commits as a
`liveSignal` — so the cause alone cannot tell a screen scrape from an approval
edge when a color turns out wrong. The origin never affects whether a change
commits.

## Live witness

Profile `statedet`, full `make install`, `attn preflight` PASS, protocol 190
CLI/app/daemon. Two real sessions driven through the packaged app. All four
outcomes observed live.

Claude session, two turns:

```
session d26539e4-b92c-40c3-b4c9-52b2a0b21783 (claude)
state:   idle since 17:41:19.658

OBSERVED                  SOURCE              CLAIM             OUTCOME    WHY
17:40:37.063              worker_info         working           applied    "watch subscribe replay"  cause=live_signal
17:40:55.054              hook_state          working           applied    cause=live_signal
17:41:03.345              hook_state          working           applied    cause=live_signal
17:41:04.602              screen              waiting_input     applied    "screen scrape"  cause=live_signal
17:41:07.065              classifier          -                 skipped    no_new_assistant_turn
17:41:04.595              classifier          idle              discarded  store_rejected  "classifier"  cause=classifier_observation
17:41:17.990              hook_state          working           applied    cause=live_signal
17:41:19.658              classifier          idle              applied    "classifier"  cause=classifier_observation
```

The `discarded` row is the payoff. The classifier's `idle` verdict for turn one
was observed at 17:41:04.595 and reached the daemon 8.163s later; by then the
screen scrape had already moved the session to `waiting_input` at 17:41:04.602 —
7ms after the classifier's observation timestamp. Two sources disagreeing, one
losing on a 7ms margin. Nothing in today's logs shows that at all.

Codex session:

```
session fc82c5c4-145a-4d47-9a11-e1b881a5f97f (codex)
state:   working since 17:42:02.904

OBSERVED                  SOURCE              CLAIM             OUTCOME    WHY
17:41:45.252              worker_info         working           vetoed     driver_transition_filter  "watch subscribe replay"
17:41:52.132              hook_state          working           applied    cause=live_signal
17:41:57.847              hook_state          working           applied    cause=live_signal
17:42:02.904              hook_state          working           applied    cause=live_signal
```

That `vetoed` row is a live confirmation of a debt item the plan already
records: the worker fabricates `working` for a session it knows nothing about on
watch-subscribe. Codex's transition filter happens to reject it; Claude's does
not, which is why the same fabricated row is `applied` in the trace above.

`--json` emits the same rows as an object with `session_id`, `agent`, `state`,
`state_since`, `capacity`, and `observations[]`.

## Tests

- `internal/statetrace`: ring ordering, eviction, repeated wrap, snapshot
  isolation, nil-recorder inertness, timestamp defaulting, log-line rendering.
- `internal/daemon`: source survives the PTY trip, driver veto is recorded
  *and the session does not move*, unknown session, hook source, cause-name
  fallback, store-discarded stale classifier, skip, drop-on-session-removal,
  and the protocol result shape.
- `cmd/attn`: argument parsing, all four outcomes rendered, the
  no-observations case, full-ring truncation notice, timestamp fallback.

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -169,7 +169,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '189';
+export const PROTOCOL_VERSION = '190';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 // AutomationActionTimeoutError distinguishes "the daemon never sent a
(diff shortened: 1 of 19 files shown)


figgyster (changes requested):
The Phase 0 seam is the right one. #664 and #665 gave the coming resolver named evidence and one detector path; this finally leaves the missing witness in the room when a color goes wrong. The source/cause split, the CLI rendering, and the four outcomes make the current last-writer-wins behavior inspectable without smuggling arbitration into the instrumentation.

One blocker before I can sign it:

The ring is not actually bounded for the lifetime of a daemon. `handlePTYState` records `session_not_found` through `traceStateVeto`, which creates `rings[sessionID]`, but that ID has no store row and will never reach `dropSessionRecord`/`forgetStateTrace`. A late worker event can do the same just after a real session is removed: removal forgets the ring, then the in-flight state update is discarded and records a fresh orphan. Each ring tops out at 256 entries, but unique stale IDs accumulate forever, and `attn state explain` cannot read them because it requires a live store row. Please keep orphan events in the daemon log only, or give orphan/retired traces a real global bound and lifecycle gate so a removed session cannot resurrect a ring. Add the race/stale-worker regression; the current unknown-session test confirms the leak's first half.

The CI board is green, including Daemon and all frontend shards. I also ran the focused `internal/statetrace` package in the isolated sandbox.

figgyster (changes requested):
You caught both visible holes and the regressions are pointed in the right direction. The unknown-id ring is gone, and moving `Forget` below `Remove` was exactly the ordering that needed changing.

One blocker remains. `recordStateObservation` still checks `d.store.Get(sessionID)` and later calls `Recorder.Record` under a separate lock. A live writer can pass that check, then `dropSessionRecord` can remove the row and forget the existing ring, then the writer resumes and creates a fresh ring. The new `TestTraceDoesNotResurrectARingAfterRemoval` delivers its event after removal, so it exercises the gate but cannot hit this check→remove/forget→record interleaving.

Please make recording and removal participate in one lifecycle lock (or add an equivalent recorder tombstone/atomic conditional write) and make the regression actually pause the in-flight writer between the live-row check and `Record`. That closes the same daemon-lifetime leak under the concurrent path that motivated the fix.

figgyster (approved):
There it is. `RecordIf` keeps the live-row admission and ring append under the same recorder lock, while removal drops the store row before waiting to forget. The two possible orderings now both end with no orphan ring.

And you made the test earn its coffee: it parks the writer precisely between admission and append, then proves the old lock placement fails under mutation. That is the witness this diagnostic path needed. Daemon and Changes are green; this follow-up only touches the Go tracing seam and its tests. Approved.

PR #669, opened 2026-07-25
feat(state): observe agent harness OSC signals as state evidence

victorarias:
Phase 1a of the agent state detection plan
Stacked on #668 (phase 0), which added the state trace this PR writes into.

## Why

Both agents already broadcast whether they are busy, in their window title, and
attn has been ignoring it in favour of scraping the rendered screen:

```
ESC ] 0 ; ⠐ Run background sleep command BEL     claude, turn RUNNING
ESC ] 0 ; ✳ Run background sleep command BEL     claude, turn NOT running
ESC ] 0 ; ⠸ attn--fix-state-detec... BEL         codex, busy
ESC ] 0 ; attn--fix-state-detec... BEL           codex, idle
```

The glyph is a *level*, repainted ~1Hz by claude and ~10Hz by codex, produced by
the harness rather than guessed at from its TUI. A level that stops arriving
cannot get stuck the way an edge-triggered claim can, which is the whole point:
the stuck colors this epic is about are edges that lost an arbitration nobody
was arbitrating.

## What lands

A read-only OSC scanner in the PTY read loop (`internal/pty/oscscan.go`) emits
observations on a new `heartbeat` source. Nothing arbitrates on them yet — they
are recorded in the phase 0 trace so the traces can be compared against current
behavior before the resolver starts weighing them.

- The OSC 133 segmenter could not be reused: it strips markers from the stream
  and is corpus-locked to a frontend parity fixture. The new scanner never
  touches the stream. It also handles a stray `ESC` mid-sequence by resuming *at*
  the ESC, so an abandoned sequence cannot make it swallow every later one.
- Per-agent behavior is a driver capability (`Capabilities.HarnessSignals`),
  matching the `ScreenDetector` kind pattern, with the usual disable-only
  `ATTN_AGENT_<AGENT>_HARNESS_SIGNALS=0` override.
- `pty.Source.EvidenceOnly()` routes these past `applyState` entirely — the
  claims are `busy`/`not_busy`, not protocol state names — and past the
  worker/backend state dedup, which would otherwise swallow a repeated level.
- An unchanged level re-emits at most once a second, and the trace collapses
  consecutive identical observations into a repeat count (`observed ×9`), so a
  long turn cannot flush the ring.

## OSC 777 was planned, built, and deleted

The plan listed OSC 777 `notify` as claude's explicit settle event. It is not:
claude emits no OSC 777 at all. No capture in the nine-session spike contains
one, and 2.1.220 offers only the `iterm2`, `iterm2_with_bell`, and
`terminal_bell` notification channels. I had the reader written and passing unit
tests before checking for a live witness; with none available it is deleted here
rather than shipped as a code path production never executes. Claude's settle
event is the `Notification` hook, which lands in phase 1b. The plan carries the
correction.

## Live verification

Profile `statedet` (full `make install`, `preflight` PASS, protocol 191 across
CLI/app/daemon), current head, both agents driven through the harness bridge.

Codex — `attn state explain`, one turn:

```
OBSERVED       SOURCE       CLAIM      OUTCOME       WHY
18:08:51.435   worker_info  working    vetoed        driver_transition_filter
18:08:51.951   heartbeat    busy       observed      "⠸ attn--fix-state-detec..."
18:08:52.551   heartbeat    not_busy   observed      "attn--fix-state-detec..."
18:09:50.112   heartbeat    busy       observed      "⠴ attn--fix-state-detec..."
18:09:50.166   hook_state   working    applied       cause=live_signal
18:09:52.209   heartbeat    busy       observed ×2   "⠦ attn--fix-state-detec..."
18:09:54.301   heartbeat    busy       observed ×2   "⠧ attn--fix-state-detec..."
18:09:55.395   heartbeat    busy       observed      "⠇ attn--fix-state-detec..."
18:09:56.515   heartbeat    busy       observed      "⠏ attn--fix-state-detec..."
18:09:56.862   heartbeat    not_busy   observed      "attn--fix-state-detec..."
18:09:56.534   classifier   idle       applied       cause=classifier_observation
```

The heartbeat went busy 54ms *before* the hook said `working`, stayed busy for
the turn, and dropped 328ms after the classifier settled the session — a level
that brackets the turn from both ends.

Claude — same shape, and it caught a live misdetection worth noting for phase 2:

```
18:03:38.702   heartbeat    busy            observed      "⠂ Claude Code"
18:03:38.703   screen       idle            applied       "screen scrape"
18:03:38.708   hook_state   working         applied       cause=live_signal
18:03:55.062   heartbeat    busy            observed ×9   "⠐ Run sleep command and echo done"
18:03:57.898   heartbeat    not_busy        observed      "✳ Run sleep command and echo done"
18:03:57.899   screen       waiting_input   applied       "screen scrape"
```

At 18:03:38 the screen scrape wrote `idle` in the same millisecond the harness
said the turn had started; the hook corrected it 6ms later. That is the class of
edge the resolver replaces the scraper for.

The title also carries a live turn summary ("Run sleep command and echo done") —
a free sidebar label with no LLM call, available whenever the UI wants it.

## Tests

`internal/pty/oscscan_test.go` and `internal/pty/harness_signals_test.go` cover
the scanner (including a split-at-every-byte pass over a full exchange, since the
read loop hands over arbitrary chunks), per-agent classification, foreign-title
rejection, and the keepalive throttle. `internal/daemon/state_trace_test.go`
covers the evidence-only route: recorded as `observed`, never reaching
`applyState`, with repeats collapsed and a changed claim opening a new row.

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -169,7 +169,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '190';
+export const PROTOCOL_VERSION = '191';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 // AutomationActionTimeoutError distinguishes "the daemon never sent a
--- a/app/src/types/generated.ts
+++ b/app/src/types/generated.ts
@@ -3925,6 +3925,7 @@ export interface ObservationElement {
     outcome:     string;
     reason?:     string;
     recorded_at: string;
+    repeats?:    number;
     source:      string;
     [property: string]: any;
 }
@@ -4404,6 +4405,7 @@ export interface StateExplainEntry {
     outcome:     string;
     reason?:     string;
     recorded_at: string;
+    repeats?:    number;
     source:      string;
     [property: string]: any;
 }
@@ -11059,6 +11061,7 @@ const typeMap: any = {
         { json: "outcome", js: "outcome", typ: "" },
         { json: "reason", js: "reason", typ: u(undefined, "") },
         { json: "recorded_at", js: "recorded_at", typ: "" },
+        { json: "repeats", js: "repeats", typ: u(undefined, 0) },
         { json: "source", js: "source", typ: "" },
     ], "any"),
     "TicketAttachResultObject": o([
@@ -11342,6 +11345,7 @@ const typeMap: any = {
         { json: "outcome", js: "outcome", typ: "" },
         { json: "reason", js: "reason", typ: u(undefined, "") },
         { json: "recorded_at", js: "recorded_at", typ: "" },
+        { json: "repeats", js: "repeats", typ: u(undefined, 0) },
         { json: "source", js: "source", typ: "" },
     ], "any"),
     "StateExplainMessage": o([
--- a/cmd/attn/state_explain.go
+++ b/cmd/attn/state_explain.go
@@ -103,18 +103,27 @@ func printStateExplain(w io.Writer, result *protocol.StateExplainResult) {
 		fmt.Fprintf(w, "\n(showing the most recent %d observations; older ones were evicted)\n", result.Capacity)
 	}
 
-	fmt.Fprintf(w, "\n%-24s  %-18s  %-16s  %-9s  %s\n", "OBSERVED", "SOURCE", "CLAIM", "OUTCOME", "WHY")
+	fmt.Fprintf(w, "\n%-24s  %-18s  %-16s  %-14s  %s\n", "OBSERVED", "SOURCE", "CLAIM", "OUTCOME", "WHY")
 	for _, obs := range result.Observations {
 		fmt.Fprintf(
 			w,
-			"%-24s  %-18s  %-16s  %-9s  %s\n",
+			"%-24s  %-18s  %-16s  %-14s  %s\n",
 			formatStateExplainTime(obs.ObservedAt),
 			orPlaceholder(obs.Source),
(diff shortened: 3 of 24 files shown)


figgyster (changes requested):
Phase 1a puts the right witness on the stand: harness-owned levels, read from the stream without pretending they are session states. It follows #668's evidence-ledger direction well. One blocker remains before that ledger stays useful during the long turns it was built to explain.

- `internal/statetrace/statetrace.go:93` — `sameEvidenceAs` includes `Detail`, but every emitted heartbeat carries the whole title and its spinner frame changes (`⠐`, then `⠸`, then `⠿`). The observer emits that unchanged `busy` level again after its one-second keepalive, and the changed detail makes each row distinct, so `RecordIf` never increments `Repeats`. A long busy turn will still consume one ring slot per second and eventually evict the other evidence, contrary to the Phase 1a contract. Collapse heartbeat rows by their semantic level (while updating the surviving row to the latest title/timestamps), or otherwise normalize the volatile spinner detail, and add the regression with alternating spinner frames/titles across keepalives.

The focused static diff check passed. The focused sandbox Go run reached the repo's missing `ghostty/vt.h` wall for the PTY/daemon packages; the available statetrace, agent, and protocol packages passed, and the live Daemon, Frontend, and E2E checks are green.

figgyster (approved):
The spinner no longer smuggles a new identity into every heartbeat. Stripping the level glyph keeps the useful title summary while letting an unchanged `busy` level collapse into one fresh row, and the alternating-frame regression pins the exact long-turn failure that #668’s bounded evidence ledger cannot afford.

The focused state-trace witness passes, `git diff --check` is quiet, and Daemon, Frontend, React Doctor, Changes, and all frontend E2E lanes are green. Approved `67d214d`.

PR #670, opened 2026-07-25
feat(state): record Claude's Notification hook and the resolved permission mode

victorarias:
Phase 1b of the agent state detection plan
Stacked on #669 (phase 1a, the OSC heartbeat). Still evidence-only — nothing
arbitrates on these yet.

Two harness-owned facts attn was not collecting.

## 1. Claude's `Notification` hook

It fires when the agent is blocked on the user, and says which kind:

```
notification_type = "permission_prompt"   ~6s after a permission request
notification_type = "idle_prompt"         exactly 60s after a turn settles unanswered
```

The plan called for carrying the `message` verbatim. The payload turns out to
carry the typed field too, which is strictly better — the claim is the type and
the message is only the detail, so nothing has to parse an English sentence to
tell "blocked on approval" from "waiting on a reply".

At ~6s and 60s of latency this corroborates rather than leads, so it is recorded
on a new `hook_notify` source and never applied: by the time it lands the session
may have moved on. A notification with no type is rejected rather than recorded
under an empty claim.

## 2. Who answers an approval request

attn's own launch flags cannot tell us. Victor's global `~/.claude/settings.json`
sets `defaultMode: auto`, so a session attn launches *without* asking for
auto-approve still has a permission classifier in the loop — and Shift+Tab
changes it mid-session either way. So the resolved `permission_mode`, which
Claude reports on every post-prompt hook, rides along on the state hook as an
optional fact and the daemon records it as a `reviewer` level. Codex reports
none; an absent mode records nothing rather than a fake "unknown" claim.

Phase 2 derives `Evidence.ReviewerInLoop` from this level plus the codex spawn
flag. This is what the plan meant by "persist ReviewerInLoop at spawn"; the plan
records why the live hook fact replaced it.

## Live verification

Profile `statedet` (full `make install`, protocol 192 across CLI/app/daemon),
one claude session, `attn state explain`:

```
OBSERVED       SOURCE       CLAIM          OUTCOME    WHY
18:19:14.554   heartbeat    busy           observed   "⠂ Claude Code"
18:19:14.562   reviewer     auto           observed
18:19:14.562   hook_state   working        applied    cause=live_signal
18:19:16.457   heartbeat    not_busy       observed   "✳ Say hi and nothing else"
18:19:16.451   classifier   idle           applied    cause=classifier_observation
18:20:16.481   hook_notify  idle_prompt    observed   "Claude is waiting for your input"
```

`hook_notify idle_prompt` landed 60.03s after the classifier settled the session
— the hook's documented interval, measured end to end through registration, CLI,
protocol, daemon, and trace.

`reviewer auto` is the finding that motivated the design change: this session was
launched with no auto-approve flag, and the resolved mode is `auto` anyway. It
was then confirmed load-bearing in the same session — asked to run a `curl`, the
agent produced no approval edge at all, because the guardian approved in the
user's place. Without the `reviewer` level that silence is indistinguishable from
"the agent never needed approval".

`permission_prompt` therefore has no live witness here: on this machine the
guardian answers first, so the branch is unit-tested only. It comes from the same
payload field on the same code path as `idle_prompt`, which is witnessed.

## Tests

`internal/hooks/hooks_test.go` covers the registration. `internal/daemon/
state_trace_test.go` covers the notification recorded as `observed` with the
type as its claim, the typeless notification rejected without a trace row, the
permission mode recorded as a `reviewer` level *in addition to* the state still
being applied, and an absent mode recording nothing. Both new assertions were
mutation-checked: applying the notification instead of observing it, and dropping
the `tracePermissionMode` call, each turn the intended assertion red.


figgyster (approved):
This keeps Phase 1 evidence-only, where it belongs. Claude's delayed `Notification` signal now preserves its typed claim while treating the message as detail, and the resolved permission mode crosses the hook/socket seam without pretending attn's launch flags know the user's current setting. That gives the Phase 2 resolver usable traces for both a real approval stall and the guardian's short-lived flash.

Ran the focused Go witness for the hook, client, daemon, and protocol path in the isolated sandbox; it returned without a failure. `git diff --check`, React Doctor, and Bench A/B are green too. The protocol-192 generation is in step across daemon and app. Approved.

PR #671, opened 2026-07-25
feat(state): add the pure session-state resolver

victorarias:
Phase 2a of the agent state detection plan
Stacked on #670. The decision rules only — nothing calls them yet, so they land
reviewable on their own before the daemon's evidence table and tick arrive.

## Why a resolver at all

attn's session state is decided by whoever wrote last. Hooks, the screen
scraper, the stop classifier and the worker poll each call `applyState` directly
with a state name, so a session gets stuck whenever the source that would have
moved it on never fires. There is no arbitration to fix — there is no
arbitration.

## The two clauses that carry the fix

Neither has an equivalent today:

- **A fresh heartbeat resolves to working regardless of the brackets.** This
  rescues a lost turn-open hook, which used to leave a running agent showing
  idle for the whole turn.
- **An open bracket goes stale once the heartbeat has been silent past
  `StaleAfter`.** This closes a bracket whose closing hook never arrived — the
  case that used to hold a session working for the rest of its life.

The structural property matters more than either clause: every rule that can
hold a session in a state now depends on evidence that either refreshes or ages
out, and the tick re-runs the decision. No state can outlive its evidence.

## Ordering encodes trust, not recency

- An exited process outranks everything. Stale evidence must not keep a dead
  session colored as alive.
- A fresh heartbeat outranks an open approval. The agent cannot be blocked on
  the user while its turn is visibly running, so that combination means the
  approval was answered and its closing edge was lost.
- The screen scrape ranks below every source that reads the harness's own
  account of itself. It manufactures approvals from ordinary prose, which is why
  the plan deletes it for claude and codex rather than demoting it to a
  tiebreaker.

## Dwell

`DwellFor` is keyed on who is being asked first. With a guardian in the loop
(claude's permission classifier, codex's `auto_review`) an approval must hold
before it is published, which removes the flash of attention-demanding color on
every tool call of an unattended run. With no guardian the user *is* the
reviewer and the dwell is zero — a genuine request is not delayed by a
millisecond.

## Timing is an argument

`Policy` is passed in rather than read from package constants, so each table
case states the window it is testing instead of inheriting it. `PolicyFor` holds
the measured per-agent defaults: claude repaints its title about once a second
and goes quiet for up to ~3.5s inside a blocking foreground tool call; codex
repaints about ten times a second and never goes quiet mid-turn.

## Deliberately not included

The plan's `Resolution.Unsettles` projection. Nothing consumes it until the
attention mode exists, and shipping an unconsumed field is the same mistake as
the OSC 777 reader deleted in #669.

## Tests

Table tests over every clause and every ordering pair above, plus both timing
boundaries (a heartbeat exactly at its TTL is still believed; a bracket exactly
at `StaleAfter` still holds), purity, and a consistency check on `PolicyFor`
that fails if `StaleAfter` ever falls inside `HeartbeatTTL` — a window that
would close brackets the heartbeat still considers live.

Both load-bearing clauses were mutation-checked: disabling the heartbeat-fresh
clause and making brackets never go stale each turn their own cases red, and
nothing else's.

## Verification tier

No live verification: this is a pure package with no callers, no daemon
lifecycle, protocol, PTY, or UI surface, fully covered by unit tests. The tier
the plan's remaining phases need arrives with the daemon wiring, where the rules
first affect a running session.

### DIFF
--- a/docs/plans/2026-07-25-agent-state-detection.md
+++ b/docs/plans/2026-07-25-agent-state-detection.md
@@ -601,8 +601,13 @@ being fixed as part of the current step.
 
 ### Phase 2 — resolver
 
-- [ ] New `internal/sessionstate` with `Evidence`, `Resolve`, table tests built
-      from Phase 1 traces.
+- [x] **Phase 2a (PR #671).** New `internal/sessionstate` with `Evidence`,
+      `Resolve`, `DwellFor`, and table tests. Timing is a `Policy` argument
+      rather than package constants, so each table case states the window it is
+      testing; `PolicyFor` holds the measured per-agent defaults. `Resolution`
+      ships without the plan's `Unsettles` projection — nothing consumes it
+      until the attention mode exists, and an unconsumed field is the same
+      unwitnessed-code mistake as the OSC 777 reader.
 - [x] Enabling refactor 5 (first half): `classifySessionState`'s rules extracted
       to `internal/daemon/classify_decision.go` as three pure functions
       (`classifyPreTranscript`, `classifyPostTranscript`, `classifyVerdict`)
--- a/internal/sessionstate/sessionstate.go
+++ b/internal/sessionstate/sessionstate.go
@@ -115,6 +115,14 @@ type Evidence struct {
 	// before it is worth showing anyone.
 	ReviewerInLoop bool
 
+	// LastBusyAt is when the heartbeat last said the turn was running. Staleness
+	// is measured from here rather than from the latest heartbeat: claude blips
+	// its not-busy glyph mid-turn (between tool calls, and while a foreground
+	// tool is still running), so treating any non-busy frame as an immediate
+	// settle would flip a healthy open turn to idle. Zero means the agent has
+	// never reported being busy, which is not the same as having gone quiet.
+	LastBusyAt time.Time
+
 	// LastMovement is when any evidence last changed. A session whose evidence
 	// has stopped moving entirely is stuck, which is a distinct condition from
 	// any state it might be reported in.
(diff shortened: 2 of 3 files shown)


figgyster (dismissed):
You’ve built the right seam for Phase 2a. After #664, #665, #666, and #668, pulling arbitration into a pure evidence table is the part that finally lets this state-detection repair stop being a collection of better guesses. But one of the measured facts has slipped its leash.

`internal/sessionstate/sessionstate.go:354` treats a fresh `ClaimSettled` heartbeat as an immediate stale bracket. Phase 1 established that Claude emits the `✳` / `not_busy` title while a foreground tool is still running, and can blip it mid-turn; the harness observer records that exact frame immediately. Once this resolver is wired, one such frame flips a healthy open turn to `idle` instead of waiting out `StaleAfter`. That reintroduces the false-settle path the spike ruled out.

Required before merge:

- Keep a non-busy heartbeat from closing an open bracket immediately. The stale decision needs to honor `StaleAfter` from the last busy evidence (or an independent settled/Stop fact), and the table needs a regression where `not_busy` is followed by busy inside that window and the turn remains `working`.

The focused resolver suite and whitespace check are green. This is the one witness that matters, though. Fix it and this slice can take its place in the resolver arc.

figgyster (approved):
You took the measured evidence seriously instead of arguing it into a comment. `LastBusyAt` keeps one idle title blip from sentencing a live turn, while still letting a full `StaleAfter` with no busy frame close a lost-Stop bracket. The replacement cases put the old shortcut in the ground where it belongs.

This is the referee #664 through #670 were building toward: one explicit decision point, with the timing facts kept intact. React Doctor is green and the focused resolver witness passes. Safe to merge.

PR #672, opened 2026-07-25
feat(state): resolve session state from an evidence table, in shadow mode

victorarias:
Phase 2b of the agent state detection plan
Stacked on #671, which added the resolver this fills evidence for.

**Shadow mode: nothing here writes state.** The table is filled and re-resolved
on a 1s tick, and the resolution is recorded in the state trace as an
observation. Flipping it authoritative and deleting the sources it replaces is
the next change, deliberately separate so the resolver can be witnessed agreeing
first — and so the places it disagrees are on record before anyone acts on them.

## What lands

Every existing source is teed into the evidence table alongside its current
path: PTY observations (heartbeat, approval, screen), hook brackets, the
stop-time classifier verdict with its observation time, the Stop payload's
background-work and pending-cron facts, the resolved permission mode, and
process exit.

The Stop facts are the ones that motivated moving the decision daemon-side in
the first place: the CLI used to collapse them into a state string before they
crossed the socket, so the resolver could not see them at all.

The tick matters as much as the table. Without it a resolution would only be
recomputed when a source spoke — which is exactly the case that fails when a
source stops speaking.

A resolution that agrees is not recorded. An agreeing tick every second would
bury everything else in a 256-entry ring.

Evidence is forgotten when its session is removed, and the tick forgets any
session whose store row has gone. Both were leaks worth pre-empting; the trace
ring had the same one in #668.

## Live verification

Profile `statedet` (full `make install`), one claude session running a 20s
foreground `sleep`. `attn state explain`, abridged:

```
OBSERVED       SOURCE      CLAIM     OUTCOME        WHY
18:33:07.220   heartbeat   not_busy  observed       "✳ Claude Code"
18:33:19.385   resolver    unknown   observed ×13   no_evidence
18:33:19.715   hook_state  working   applied        cause=live_signal
18:33:42.798   heartbeat   busy      observed ×12   "⠂ Run bash sleep and echo command"
18:33:44.418   heartbeat   not_busy  observed       "✳ Run bash sleep and echo command"
18:33:50.386   resolver    idle      observed ×6    bracket_stale
18:33:44.406   classifier  idle      applied        cause=classifier_observation
```

Two things to read here.

The resolver stayed silent for the entire 20s tool call — it agreed with
`working` on every tick, through the mid-tool heartbeat gaps that would have
tripped a naive TTL.

Then it reached `idle` via `bracket_stale` about a second after the agent said
not-busy, while the store stayed `working` until the classifier landed roughly
eight seconds later. That gap is the improvement the phase is for, measured on a
real turn.

The early `resolver unknown / no_evidence` run is correct rather than a gap: a
settled heartbeat alone says the turn is not running and nothing about what the
session wants, which is precisely why the hooks and the classifier are in the
table too.

## One finding, recorded not fixed

Settling fast has a cost the flip has to answer for. On a turn that ends in a
question, the resolver would publish `idle` immediately and then correct to
`waiting_input` when the classifier lands — a flicker where today there is a
delay. In this run the verdict was `idle` and the two agreed, so nothing was
visible. The plan's Open questions section now carries the case and the three
options; it needs a decision before phase 2c, and these same shadow traces can
measure how often the verdict is `waiting_input` at all.

## Tests

`internal/daemon/session_evidence_test.go` covers each source landing in the
right slot with its observation time intact, the brackets as levels, the stop
facts clearing when a later stop reports nothing outstanding, `default` being
the only permission mode that means the user answers, an absent mode recording
nothing at all, every write stamping `LastMovement`, and the three shadow-mode
properties: a disagreement is recorded without being applied, an agreement is
silent, and a removed session stops being resolved.

Both shadow-mode properties were mutation-checked — dropping the agreement guard
and adding an `applyState` call each turn their own case red.

`make test` green.

### DIFF
--- a/docs/plans/2026-07-25-agent-state-detection.md
+++ b/docs/plans/2026-07-25-agent-state-detection.md
@@ -596,7 +596,7 @@ being fixed as part of the current step.
       `applyState` entirely — their claims are "busy"/"not_busy", not protocol
       state names — and past the worker/backend state dedup, which would
       otherwise swallow a repeated level.
-- [ ] Feed the hook signals into the ring the same way (phase 1b), then compare
+- [x] Feed the hook signals into the ring the same way (phase 1b), then compare
       traces against the baseline.
 
 ### Phase 2 — resolver
@@ -614,8 +614,14 @@ being fixed as part of the current step.
       returning a `classifyDecision`; the shell performs the IO between them and
       owns the single store write. Six `applyState` exits collapsed to one. Folding
       these rules into `Resolve` itself is Phase 2 work — this makes them movable.
-- [ ] Daemon evidence table + 1s tick; route live signals through the resolver
-      into `applyState` with a new `resolverObservation` cause.
+- [x] **Phase 2b (PR #672), shadow half.** Daemon evidence table + 1s tick.
+      Every existing source is teed into the table alongside its current path,
+      the tick re-resolves each session, and the resolution is recorded in the
+      trace *only when it disagrees* with the stored state. No `applyState` call
+      — the flip is deliberately separate so the resolver can be witnessed
+      agreeing first.
+- [ ] **Phase 2c**, the flip: route live signals through the resolver into
+      `applyState` with a new `resolverObservation` cause.
 - [ ] Gate `internal/pty/state_detector.go` to copilot only; delete the keyword
       lists and `approval_resolver.go`'s screen-absence debounce for claude/codex.
 - [ ] Delete `ShouldApplyPTYState` and all three driver implementations — the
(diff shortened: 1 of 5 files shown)


figgyster (changes requested):
You built the dangerous half of the resolver the grown-up way: it watches first, leaves state alone, and keeps receipts for every disagreement. That is exactly the discipline #668 bought us before Phase 2 could touch the wheel. But this new evidence table has picked up the trace ring's old vice.

`internal/daemon/session_evidence.go:42` creates an entry for every incoming id without proving the store row is live. `handlePTYState` calls it before its own `session_not_found` gate, and `dropSessionRecord` removes the row before forgetting the table. A late worker observation that arrives after that forget can therefore recreate an orphan entry; a stream of unknown ids can fill the table until the next tick. The tick eventually sweeps it, but the table is meant to be one record per live session, and Phase 0 already established that diagnostics must be lifecycle-bound at admission, not eventually tidied up.

Required before merge:

- Make evidence writes use the same atomic live-row admission shape as `recordStateObservation`: check the store row while holding the evidence-table lock, so removal's row-delete-then-forget ordering either deletes a pre-existing entry or prevents a post-removal writer from creating one.
- Add a forced-interleaving regression covering a writer racing session removal, plus an unknown-id observation. Both must leave no evidence-table entry.

The shadow resolver itself is the right next chapter after #671. I want this black box to stay bounded before we ask it to tell us what the city is doing.

figgyster (commented):
You carried #668's exact lesson forward. `recordEvidence` is now the one door, its live-row check holds the evidence lock, and the race witness parks the writer in the old check-then-append gap until removal has had its chance. That closes the blocker I left on the prior head.

React Doctor is green. Bench A/B is still running, so this stays a current-head comment rather than an approval for the moment. No remaining code concern from me; once the board finishes its paperwork, this evidence table is ready to keep shadowing the resolver.

figgyster (approved):
You took the #668 scar seriously and carried the fix all the way into Phase 2b. One lock-held live-row gate for every evidence writer, then a race test that waits in the exact old crack until removal walks past it — that is a witness, not a comforting story.

The shadow table can now stay bounded while it watches the resolver disagree, which is the last thing I needed before trusting this stage of the state-repair arc. React Doctor and Bench A/B are green. Approved.

figgyster (commented):
The #672 patch itself survived this rebase byte-for-byte, and the new parent carries #669's spinner-frame collapse so a long busy turn stops eating the trace ring alive. I found no new code concern in the current stack.

React Doctor is green; Bench A/B is still running on this head. Keeping the review conditional until that witness comes in.

PR #673, opened 2026-07-25
Phase 2c-1: let the evidence resolver publish session state

victorarias:
Phase 2c-1 of `docs/plans/2026-07-25-agent-state-detection.md`. The evidence
resolver has been running in shadow mode since the evidence table landed,
recording what it would have said. This makes its verdict the thing that sets
session state.

The point is not a better heuristic. It is that state must now be re-justified
from live evidence every second. Under the scheme this replaces, a state
persisted until some source happened to contradict it, and got stuck forever
when none did — the failure class the plan exists to remove.

Phase 2c is split in two. This is the flip; the deletions (gating
`state_detector.go` to copilot, removing `ShouldApplyPTYState`, retiring the
shadow state consts) follow in 2c-2, once the flip has run live. Landing them
together would mean a revert of a bad flip also reverts the cleanup.

## What changed

**A settle no longer depends on the classifier having spoken.** Closed brackets
plus a heartbeat that is not painting busy frames now settle a turn on their own
(`heartbeat_settled`). The classifier is allowed to decline — no transcript,
capability disabled, an error — and when it declined, nothing else was ever
going to move the session off `working`. Found by a test that expected a settle
and got `no_evidence`.

**The resolver can answer "keep what you have"** (`Resolution.Hold`), for two
bounded windows:

- *settle grace* — a bracket that just went stale holds rather than asserting
  idle. Measured: claude's approval prompt renders at 14.6s and its Notification
  hook lands at 20.6s, so with `StaleAfter` 4s the bracket goes stale at 18.6s
  and the resolver would otherwise paint idle in the middle of a live approval.
- *awaiting verdict* — a settle with a classification in flight holds until the
  verdict or `ClassifierTimeout`, so a question-ending turn does not flash idle
  and then correct itself to `waiting_input`.

Both expire. An unbounded hold is the stuck color this plan removes, and codex
has no `idle_prompt` equivalent to unstick it.

**Claude's Notification hook reaches the evidence table.** `permission_prompt`
becomes an approval claim. `idle_prompt` becomes `PromptIdleAt` and deliberately
not a claim: live captures show it fires 60s after *any* unanswered settle — a
finished foreground Bash turn gets it exactly as a question does — so it cannot
choose between idle and waiting_input. What it is, is an independent witness
that the agent is not working, which is the one thing a lost Stop hook leaves
attn unable to discover.

**Codex approvals are read from its title.** `[ . ] Action Required | <cwd>`,
switching to `[ ! ]` the moment it is answered. Measured on codex 0.145.0 driven
through a real PTY with `--ask-for-approval untrusted`. Codex emits no
notification escape (its OSC vocabulary is 0, 10, 11 — there is no 777) and has
no approval hook, so without this the 2c-2 deletion of the screen scrape would
leave codex approvals undetectable.

**The tick does not publish everything it computes.** `no_evidence` is the
absence of a finding rather than a finding, and publishing it would repaint
sessions grey on the first tick after any single observation. States outside the
resolver's remit are left alone: `launching` belongs to spawn until the agent
first speaks, and `recoverable` to the revive path, which sets it precisely
because the worker is gone and its evidence is therefore meaningless. `stuck` is
real but is surfaced in phase 3.

**Codex stop-time classification moves to `gpt-5.6-luna`** at low effort. The
verdict is one word over a short transcript slice, and what the cheaper tier
buys is latency, not tokens — latency that is now user-visible, because the
resolver holds the pre-settle color until the verdict lands. Measured end to end
through the CLI: 3.9s against 9.3s for the claude/haiku path.

## Live verification

Full `make install PROFILE=statedet`, preflight clean, protocol 192 in lockstep.
A real claude turn, from the daemon trace:

```
21:20:24  hook_state    working       applied    live_signal
21:20:25  heartbeat     busy          observed   "Run sleep probe command"
   …busy frames every ~2s…
21:20:40  heartbeat     not_busy      observed
21:20:40  screen        waiting_input APPLIED    live_signal  "screen scrape"
21:20:40  resolver      working       applied    bracket_open
21:20:42  resolver      —             skipped    hold          (x7 ticks)
21:20:49  classifier    idle          DISCARDED  store_rejected (lag 9.3s)
21:20:49  resolver      idle          applied    classifier_verdict
21:21:40  hook_notify   idle_prompt   observed
```

Three things this shows:

- The settle gate works — seven held ticks, then the verdict, with no idle flash
  before it.
- The resolver repaired a wrong color it did not cause. The screen scraper fired
  `waiting_input` 15 seconds into a `sleep 12` that was still running, and the
  resolver overrode it on the next tick. Before the flip that false positive
  would have stood until something else contradicted it. This is direct evidence
  for 2c-2 deleting the scraper for claude.
- The classifier's own write was rejected as stale (9.3s lag) and the outcome was
  still correct, because the resolver reads the evidence rather than racing to
  the store.

`idle_prompt` landed at 60.0s after the settle, confirming the timing
`PromptIdleAt` is built on.

## Testing

Every new decision point is mutation-checked, each failing in exactly the case
written for it: the settle grace (dropped, and unbounded), the classifier gate
(dropped, and unbounded), the heartbeat-settled clause, and the three publish
guards (owned-state, no-evidence, hold).

`go test ./internal/...` green.

## Known gap

A held settle shows the pre-settle color for as long as the classifier takes —
~9s on the claude path in the run above. That is the trade the plan chose (a late
correct color over an early wrong one), but it is longer than the plan assumed,
and worth feeling before it ships.


internal/classifier/classifier.go
```diff
@@ -60,7 +60,12 @@ var verdictLineRegex = regexp.MustCompile(`(?i)^\s*(?:VERDICT\s*[:=]\s*)?(WAITIN
 const classifierLogSnippetMaxChars = 600
 
 const (
-	defaultCodexClassifierModel   = "gpt-5.4-mini"
+	// The classification is a one-word verdict over a short transcript slice, so
+	// the cheapest tier of the current family is the right pick — the cost here
+	// is latency, not tokens, and latency is user-visible: the resolver holds the
+	// pre-settle color until the verdict lands. Measured end to end through the
+	// CLI, low effort: gpt-5.6-luna 3.9s against 9.3s for the claude/haiku path.
```
victorarias, line 67: remove

internal/daemon/session_evidence_test.go
```diff
@@ -152,11 +154,12 @@ func TestEveryEvidenceWriteStampsMovement(t *testing.T) {
 	}
 }
 
-// Shadow mode's contract: the tick records what the resolver would have said and
-// changes nothing.
-func TestTheResolveTickRecordsADisagreementWithoutApplyingIt(t *testing.T) {
+// The flip: the tick is now the thing that publishes state. A session sitting in
+// a state no source will ever contradict is moved by the resolver alone, which
+// is the stuck-color fix in one assertion.
```
victorarias: remove

figgyster (changes requested):
You finally gave the resolver the badge instead of making it sit outside the precinct taking notes. The bounded hold around a real classifier is the right piece of caution — it keeps a question-ending turn from flashing the wrong color while still refusing to preserve a stuck color forever. But this handoff leaves one old witness on the stand after the next case has started.

Required before merge:

- Clear or otherwise invalidate `LastClassifier` when a new turn opens, and prove the cross-turn race. A prior turn's classifier verdict stays in the evidence table after `recordBracketEvidence(..., working)`. If the next turn settles while its classifier is in flight, `Resolve` sees that old verdict first and publishes it instead of holding for the new result. The direct classifier write is correctly rejected by its old timestamp, but recording the rejected verdict still lets the resolver resurrect it later. Cover: turn A finishes `waiting_input`; turn B opens and goes busy; turn B settles and begins classification. It must hold for B's verdict (or use the no-verdict fallback), never report A's `waiting_input` as B's state.

That is exactly the stale-worker protection this state-repair stack has been earning one seam at a time. Until the verdict is scoped to its turn, the resolver can put yesterday's answer on today's session. Case stays open.

figgyster (approved):
You caught the whole ghost this time. The opening-bracket reset handles the quiet turn, while the `LastBusyAt` guard catches the missing-hook turn; each regression pulls its own weight. That is the same lifecycle-bound evidence discipline #668 and #672 fought for, finally carried into the resolver’s actual authority.

Daemon, Changes, React Doctor, and the frontend lane are green. Approved. Let Phase 2c-1 run live before the subtraction crew starts tearing out the old wires.

PR #674, opened 2026-07-25
fix(state): subtract the screen scraper for claude and codex

victorarias:
Phase 2c-2 of `docs/plans/2026-07-25-agent-state-detection.md`. With the
resolver owning claude and codex state (2c-1, #673), the machinery it replaced
is now a second opinion competing with it. This removes that machinery, and
fixes the two bugs the removal exposed live.

## What is deleted

- Claude's `ScreenDetectorClaude` capability and its whole screen classifier —
  status-frame glyph prefixes, timer/summary regexes, text normalization.
- `HasApprovalResolver` and the `APPROVAL_RESOLVER` env override, plus
  `internal/pty/approval_resolver.go` and its screen-absence debounce.
- The per-session approval machinery in `pty.Session`: the mutex, the timer,
  the recheck scheduling, `evaluateApproval`, `applyApprovalState`.
- `ShouldApplyPTYState` for claude and codex.

Copilot still has no harness signals, so it keeps the PTY-state filter. The
provider interface is split accordingly: `PTYStateFilterProvider` is now a
separate, copilot-only capability rather than something every agent implements.

Net ~-1750 lines.

## Bugs the subtraction exposed

Both were found by running the probe against a live agent, not by a test.

**A settle needs a turn to have opened.** `heartbeat_settled` fired during boot,
turning a session idle within a second of launch. A booting agent paints title
frames before its prompt is ready — codex flickers a busy one — so "a busy frame
was seen" is not evidence a turn ran. `Evidence.TurnEverOpened` gates the settle
on a turn actually having opened.

**The turn bracket never closed on codex.** The Stop hook is the closing bracket,
but only a separate `hook_state` message recorded one, and codex sends none. The
resolver asserted `bracket_open` over the classifier's verdict for the full stale
window, which showed up as a visible `idle -> working -> idle` flicker at the end
of every turn. `handleStop` now records the close directly.

Also fixes codex's approval title reading: `[ . ] Action Required` is a prompt
waiting for an answer, `[ ! ] Action Required` is one that has been answered.
Both contain the marker, so matching the marker alone left sessions yellow
forever.

## Live verification

Profile `statedet`, full install, packaged app.

claude turn: `launching -> working -> idle`.

codex turn: `working -> idle` (was `launching -> working -> idle -> working -> idle`
before the bracket fix).

codex approval, guardian disabled so the prompt actually reaches the terminal:

```
   0.20s  launching
   2.01s  working
   6.23s  pending_approval     <- marker file absent: genuinely blocked
--- Enter ---
   1.61s  working              <- approval retired by the busy frame
   6.04s  waiting_input        <- marker file now present
   9.68s  idle                 <- no re-arm
```

The approval clear is the deletion with no other live witness, since it is what
replaces the screen-absence debounce.

Every new guard was mutation-tested: each goes red on its own case, and none is
covering for another.


figgyster (approved):
You took the evidence-first state repair all the way to its honest ending: the screen scraper is surplus now, so out it goes. I was waiting for the subtraction pass after #673. This is it. The Stop hook finally closes Codex's turn bracket, `TurnEverOpened` stops its boot spinner from inventing a finished turn, and a fresh busy frame retires an answered approval without leaving a yellow ghost behind.

The old second opinion is gone, Copilot keeps the one screen path it still needs, and the resolver is again the one referee in the room. The focused resolver and agent witnesses passed here; the sandbox lacks the native Ghostty headers for the daemon/PTY packages, but Daemon, Bench A/B, React Doctor, Changes, and the relevant E2E lane are green on this head. Safe to merge. Case closed.

PR #675, opened 2026-07-25
feat(state): hold a guardian's approval instead of flashing it

victorarias:
First half of phase 3 of `docs/plans/2026-07-25-agent-state-detection.md`
(the dwell gate). The remaining phase 3 items — stuck detection surfaced in
the UI, and `idleStaleAfter` staleness marking — follow in a second PR.

## The change

An unattended run routes permission requests to a reviewer rather than to the
user: codex to its `auto_review` guardian, claude to its permission classifier.
Publishing the request the instant it appears paints an attention-demanding
color on every tool call of a run nobody is watching.

A transition into `pending_approval` now has to keep being the resolver's answer
for `guardianDwell` (60s) before it is published. The guardian's answer cancels
it — the agent's next busy frame retires the approval, the resolver starts
answering `working`, and the pending transition is dropped without ever having
been shown. A guardian that stays silent still surfaces the request once the
dwell elapses. With no reviewer the dwell is zero and a genuine request is not
delayed by a millisecond.

The gate lives in the daemon, not the resolver. The resolver is pure and answers
what is true now; whether an answer has been true for long enough is a fact
about the sequence of answers, which only the thing running the tick can know.

Who reviews is filed as evidence at spawn, where the decision is actually made.
Both agents gate their reviewer on the same launch option, and yolo outranks it
by removing the approval gate rather than delegating it.

## A bug this found live

Codex's hooks send `permission_mode: default` on every turn as payload filler,
while its actual reviewer comes from the `approvals_reviewer` flag. Reading that
field retired the spawn-time reviewer fact on the first turn of every codex
session and took the dwell with it — visible in the trace as three `reviewer
default` rows overwriting the spawn record. The mode is now read only for
claude, whose mode genuinely governs approval routing.

## Live verification, including what it did not show

Profile `statedet`, full install, packaged app, throwaway `CODEX_HOME` so the
sandbox actually forces an escalation.

- codex with the guardian on: `launching -> working -> idle`, never yellow, and
  the write outside the workspace landed — so the escalation was real and the
  guardian approved it.
- claude with the guardian on: `launching -> working -> idle`, command ran,
  reviewer correctly observed as `auto`.
- codex with the reviewer removed: the same prompt does produce
  `[ . ] Action Required` in the title, confirming the escalation path.

What these runs did **not** show is the dwell holding anything. In both
guardian-on runs the daemon recorded no approval evidence at all — codex never
painted the marker and claude sent no `permission_prompt` notification while its
classifier was answering. So on these agent versions the guardian answers
without ever raising a visible request, and the dwell had nothing to suppress.

It is kept because it costs nothing when the guardian is fast and is the only
thing standing between a slow guardian and the reported flash. Whether the flash
still exists on current agent versions — and therefore whether the dwell earns
its place — is logged in the plan as a phase 3.5 decision rather than settled
here.

The gate's own behavior is covered by tests: it holds, it releases on time, a
different answer cancels the wait rather than extending it, a zero dwell clears
a wait in progress, and the spawn files the right reviewer for all three launch
arrangements. Each guard was mutation-tested and fails on its own assertion.

### DIFF
--- a/app/scripts/real-app-harness/probe-state-detection.mjs
+++ b/app/scripts/real-app-harness/probe-state-detection.mjs
@@ -39,13 +39,23 @@ async function main() {
 
   await launchFreshAppAndConnect(client, observer);
 
-  // PROBE_GUARDIAN=off disables the auto-approve guardian for the run. With it
-  // on, codex routes every permission request to its auto_review reviewer, which
-  // answers in milliseconds — so no prompt ever reaches the terminal and the
-  // approval path cannot be observed at all.
-  if (process.env.PROBE_GUARDIAN === 'off') {
-    await client.request('set_setting', { key: 'auto_approve_enabled', value: 'false' });
-    console.log('guardian auto-approve disabled for this run');
+  // PROBE_GUARDIAN picks who answers approval requests, because the two halves
+  // of the approval behavior are only observable under opposite settings.
+  //
+  // `off`: no reviewer, so the prompt reaches the terminal and the approval path
+  // can be watched at all. `on`: codex routes the request to its auto_review
+  // reviewer, which answers in milliseconds — nothing reaches the terminal, and
+  // what is being checked is that the user is never shown a color for it.
+  //
+  // It is set explicitly rather than left alone because the setting is
+  // persistent: a previous run leaves the profile in whichever state it chose.
+  if (process.env.PROBE_GUARDIAN) {
+    const enabled = process.env.PROBE_GUARDIAN === 'on';
+    await client.request('set_setting', {
+      key: 'auto_approve_enabled',
+      value: String(enabled),
+    });
+    console.log(`guardian auto-approve ${enabled ? 'enabled' : 'disabled'} for this run`);
   }
 
   const sessionId = await createSessionAndWaitForInitialPane({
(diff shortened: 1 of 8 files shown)


figgyster (changes requested):
This is the right next Phase 3 move. You kept the resolver pure and put time in the daemon, then gave Codex a real reviewer fact at spawn instead of letting a filler field erase it. That pays off the evidence-first work from #668 through #674.

But the dwell gate has one orphan path. `dropSessionRecord` forgets the evidence row, so the next resolve tick never sees that session and cannot run its existing `dwellGate().clear` branch. A session closed during the 60-second approval dwell therefore leaves an entry in `dwellGate.pending` forever; enough short-lived unattended sessions turn a transient UX gate into daemon-lifetime state.

Required before merge: clear the dwell gate in the session-removal path alongside the evidence-table cleanup, and add the lifecycle regression that arms a dwell, removes the session, and proves the gate no longer retains it.

figgyster (approved):
There it is. The removal path now takes the dwell entry out with the evidence row, and the witness arms the gate first, closes the session, then proves the later tick cannot be secretly doing the work. That is the lifecycle contract #668 taught this stack the expensive way.

Daemon, Changes, React Doctor, and the frontend lane are green on this head. Bench A/B is still running, but it is trend weather, not a reason to leave a fixed state leak under a red lamp. Approved.

PR #676, opened 2026-07-25
feat(state): publish stuck sessions and say why

victorarias:
Second slice of phase 3 of `docs/plans/2026-07-25-agent-state-detection.md`.
The third item — `idleStaleAfter` staleness marking — is not in this PR.

## The change

A session whose evidence has stopped moving entirely was already computed as
stuck by the resolver, and then discarded: `publishResolution` dropped
`ReasonStuck` on the floor, leaving the session in whatever colour it last
showed. That is the exact failure the resolver exists to remove. It is now
published as `unknown`, which the UI already treats as needing attention.

An `unknown` badge with no explanation is the same dead end as the stuck colour
it replaces, so the reason travels with it. Sessions carry `state_reason`
(protocol 193) and the state indicator turns it into a tooltip — for `unknown`
only, since every other state says what it means by its own name.

The reason is not persisted. It is recomputed every tick from evidence that does
not survive a restart, so a stored one could only describe a daemon that no
longer exists. It is attached only to states the resolver owns: `launching`
belongs to spawn and `recoverable` to the revive path, and captioning either
with the resolver's last thought describes a different photo.

## Two findings, logged in the plan rather than fixed here

**An open bracket outranks stuck.** `TurnOpen` with `LastBusyAt` never set makes
`heartbeatSilentFor` return false — an agent that has never reported being busy
is deliberately not treated as silent — so the bracket-open clause returns
`working` every tick and the stuck clause below it is unreachable. Claude and
codex both paint a busy title with every turn, so it cannot bite there; for an
agent with hooks and no heartbeat it is a session pinned green forever.
Reordering the clauses changes the resolver's trust model and deserves its own
treatment, so it is a phase 3.5 decision.

Consequently stuck is reachable today only when the brackets are the *only*
source that ever spoke and they have stopped too, which is what the test models.

**`state_reason` is not in the harness session projection**, so scenarios cannot
assert on it. Logged as a follow-up; threading it through the app's local session
model is not worth doing until a scenario needs it.

## Live verification

Profile `statedet`, full install, packaged app, protocol 193 on both sides.

Read straight off the daemon WebSocket:

```
570dda3c state=idle       reason=classifier_verdict
f3228dc9 state=launching  reason=(none)
27f7b053 state=launching  reason=(none)
```

The reason is on the wire with the right value, and correctly absent for
`launching`, which the resolver does not own.

The risk this change carries is spurious attention badges — `unknown` is an
attention state, so a session wrongly called stuck would nag. Across every live
run on this profile the daemon applied `unknown` zero times, and a claude session
watched well past the 90s stuck window settled to `idle` and stayed there.

What the live runs do **not** show is the tooltip on a genuinely stuck session:
producing one needs an agent with brackets and no heartbeat, which neither claude
nor codex can be made into. The rendering is covered by component tests and the
daemon half is proven on the wire.

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -1487,6 +1487,9 @@ sendFetchPRDetails,
       delegatedFromChief: daemonSession?.delegated_from_chief ?? false,
       ticketUnread: daemonSession?.ticket_unread ?? false,
       nudgeFiresAt: daemonSession?.nudge_fires_at,
+      // Dropped when a pane status overrides the state: the reason describes the
+      // resolver's answer, and a pane-derived state was not the resolver's.
+      state_reason: paneState ? undefined : daemonSession?.state_reason,
     };
   });
 
--- a/app/src/components/AttentionDrawer.tsx
+++ b/app/src/components/AttentionDrawer.tsx
@@ -13,6 +13,7 @@ interface AttentionDrawerProps {
     id: string;
     label: string;
     state: UISessionState;
+    state_reason?: string;
   }>;
   prs: DaemonPR[];
   onSelectSession: (id: string) => void;
@@ -53,7 +54,7 @@ export function AttentionDrawer({
                   data-state={s.state}
                   onClick={() => onSelectSession(s.id)}
                 >
-                  <StateIndicator state={s.state} size="sm" kind="session" seed={s.id} />
+                  <StateIndicator state={s.state} size="sm" kind="session" seed={s.id} reason={s.state_reason} />
                   <span className="item-name">{s.label}</span>
                 </div>
               );
--- a/app/src/components/Sidebar.tsx
+++ b/app/src/components/Sidebar.tsx
@@ -30,6 +30,7 @@ interface LocalSession {
   delegatedFromChief?: boolean;
   ticketUnread?: boolean;
   nudgeFiresAt?: string;
+  state_reason?: string;
 }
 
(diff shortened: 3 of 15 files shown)


figgyster (changes requested):
You're giving the resolver the thing the old screen machinery never had: an explanation a person can act on. That is the honest next move after #673 handed it authority. But one wire still ends inside the daemon.

Required before merge:

- When `state_reason` changes while the visible state remains `unknown`, publish a session update (only on a reason delta, so the one-second resolver tick does not turn into a broadcast parade) and add the regression. Today `publishResolution` records `stuck`, then returns at the same-state path; clients only receive this field through an `applyState` broadcast. An already-unknown session that goes silent therefore keeps its old generic or stale tooltip forever, while the daemon knows it is stuck. That is the exact explanation this slice promises to deliver.

The rest of the state-repair arc is holding its line: bounded runtime evidence, no persisted ghost story. Get this last delivery seam under a test and it can move.

figgyster (approved):
You found the wire I left hanging and ran it where it belongs. `state_reason` now reaches the client when the state has nowhere left to move, while the delta gate keeps the resolver's one-second pulse from rattling every window in town.

The new unknown-to-stuck witness proves both halves: the explanation arrives, then the quiet ticks stay quiet. That closes the last gap in this Phase 3 slice — the resolver can call a session stuck and the person looking at it will finally know why. Daemon, Frontend, and the E2E lanes are green. Safe to merge.

PR #677, opened 2026-07-26
Phase 3: mark an unread idle result stale, and stop calling a never-prompted session stuck

victorarias:
Last item of phase 3 of `docs/plans/2026-07-25-agent-state-detection.md`, plus a
live-caught correction to the stuck detection that shipped in #676.

## The mark

A finished turn that has sat past `IdleStaleAfter` with nobody looking at it is
carried to clients as `idle_stale` (protocol 194). The state does not change —
the session is idle either way — only whether that idle answer is still fresh.
Nothing consumes it yet: the attention projection that turns it into a reason to
unsettle a session is phase 4.

Staleness is deliberately **not** a `Resolve` clause. `Resolve` answers what the
agent is doing, and the agent is doing nothing; whether anyone has seen its
result is a fact about the user, which the evidence table does not and should not
know. It is a pure `sessionstate.IdleStale` rule fed by daemon-tracked read times
and applied by the resolve tick — the same split that put the guardian dwell in
the daemon rather than the resolver.

"Read" means the daemon was told the session was visualized, or it is the session
currently on screen. The on-screen half matters: a turn that finishes while
someone is watching it produces no `session_visualized` of its own, so without it
the session being looked at would go stale.

Two deviations from the plan, both recorded in it:

- **The `needs_review_after_long_run` fold is deferred, not done.** The plan item
  said to fold it in. That mechanism is live and user-visible (a 5m+ run defers
  its classification until the session is viewed) and its replacement has no
  consumer until phase 4, so folding it in now would delete working behaviour and
  put nothing in its place. They are not duplicates — the old one measures how
  long the *run* took, the new one how long the *result* has gone unread — and
  phase 3.5 or 4 decides which survives.
- **The window is a judgement, not a measurement.** The plan names no duration.
  `defaultIdleStaleAfter` is 10 minutes: long enough that stepping away from a
  session is not treated as forgetting it, short enough that a result someone
  asked for is not lost for a working day. Nothing depends on the exact number
  until it has a consumer.

## The stuck correction

Live verification of the above caught a wrong colour from the stuck publication
merged in #676. A claude session that is **launched and never prompted** turned
`unknown` ninety seconds later while sitting at a healthy empty prompt:

```
02:16:10  worker_info  working   applied
02:16:11  heartbeat    not_busy  observed   "Claude Code"
02:17:41  resolver     unknown   applied    "stuck"
```

Claude paints its title on activity and then goes quiet at an empty prompt. A
session with no turn behind it has neither a Stop nor an `idle_prompt`
notification to contradict a stuck verdict — that notification is the only reason
*settled* sessions are unstickable. Stuck now requires a turn to have opened:
silence before the first turn is nothing to report, not a stopped report.

Also fixed: `internal/hub`'s `sessionsMatch` compared neither `state_reason` nor
`idle_stale`, so a remote session's reason or mark would never propagate to a hub
client.

## Verification

Full Go and frontend suites pass. Every new wire was mutation-checked (dropping
the read stamp, the tick call, and the on-screen read each turn a specific test
red).

Live on profile `statedet`, protocol 194, packaged app, preflight clean. One
session was prompted and the app moved to a second session while it was still
working, so its result landed unwatched:

```
+600s   9c319285 state=idle reason=classifier_verdict idle_stale=false
+720s   9c319285 state=idle reason=classifier_verdict idle_stale=true
opened  9c319285 state=idle reason=classifier_verdict idle_stale=false
```

The mark appears between 10 and 12 minutes after the result lands, and opening
the session clears it. The second session — launched and never prompted — stayed
out of `unknown` for the full fourteen minutes, which is the stuck fix.

That second session did sit at `working` the whole time, which is its own wrong
colour: the `worker_info` replay applies `working` at spawn and no clause can
retire it, because every settle clause requires a turn to have opened. Logged in
the plan for phase 3.5 rather than fixed here — the honest answer is probably
`idle`, but saying so means deciding what evidence proves an agent has finished
booting.

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -169,7 +169,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '193';
+export const PROTOCOL_VERSION = '194';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 // AutomationActionTimeoutError distinguishes "the daemon never sent a
--- a/app/src/types/generated.ts
+++ b/app/src/types/generated.ts
@@ -851,6 +851,7 @@ export interface SessionElement {
     directory:                    string;
     endpoint_id?:                 string;
     id:                           string;
+    idle_stale?:                  boolean;
     is_worktree?:                 boolean;
     label:                        string;
     last_seen:                    string;
@@ -4107,6 +4108,7 @@ export interface Session {
     directory:                    string;
     endpoint_id?:                 string;
     id:                           string;
+    idle_stale?:                  boolean;
     is_worktree?:                 boolean;
     label:                        string;
     last_seen:                    string;
@@ -9254,6 +9256,7 @@ const typeMap: any = {
         { json: "directory", js: "directory", typ: "" },
         { json: "endpoint_id", js: "endpoint_id", typ: u(undefined, "") },
         { json: "id", js: "id", typ: "" },
+        { json: "idle_stale", js: "idle_stale", typ: u(undefined, true) },
         { json: "is_worktree", js: "is_worktree", typ: u(undefined, true) },
         { json: "label", js: "label", typ: "" },
         { json: "last_seen", js: "last_seen", typ: "" },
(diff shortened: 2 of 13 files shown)


figgyster (approved):
You finally shut the state-repair office door on Phase 3 instead of leaving one more “temporary” signal taped to the glass. Good instinct. `idle_stale` stays a daemon-owned fact about whether a finished answer was seen, while `Resolve` keeps answering only what the agent is doing; that boundary is the whole case.

The new mark broadcasts on its own delta, clears when the session is read, and the never-prompted Claude correction stops an empty prompt from becoming a false `unknown`. That carries the evidence-first arc from #672–#676 into the next question without pretending the future attention projection already exists. The focused state/protocol/hub witnesses and the full CI board are green. Approved.

PR #678, opened 2026-07-26
refactor(classifier): run Claude classification through the headless seam, drop the agent SDK

victorarias:
## What

Stop-time classification for Claude sessions invoked `claude -p` through a third-party Go SDK wrapper (`github.com/victorarias/claude-agent-sdk-go`) — the module's only consumer of that dependency. attn already has its own headless Claude runner in `internal/agent` (the keeper, notebook, workflow-engine, and ticket-reconciliation runs all go through it), and it covers everything the classifier needed: model pin, `--max-turns`, `--json-schema` with the validated object returned as structured output, `--no-session-persistence`, a tool-less mode, bounded output, and failure bucketing.

This routes the classifier through that seam and removes the dependency.

- `internal/classifier` keeps prompt building and verdict parsing only, exporting `ClaudeVerdictSchema`, `ClaudeMaxTurns`, `ClaudeClassifierModel()`, and `ParseVerdict(structuredOutput, finalText)`. `ClassifyWithClaude` and the SDK message-tree parsing are gone.
- The Claude driver implements `ExecutableClassifierProvider` and runs the judgment via `RunHeadlessTask` in a throwaway scratch dir.
- The daemon's no-driver-classifier fallback goes through the Claude driver and reports an explicit error when no backend exists.

## Behavior

The verdict is unchanged: same prompt, same `haiku` default, same `ATTN_CLAUDE_CLASSIFIER_MODEL` override, same waiting/done mapping. The run is now hermetic — no MCP servers attach, the user's settings and slash commands are not loaded, tools are disabled, and the environment is scrubbed. Previously the classifier inherited ambient Claude configuration, so a slow or needs-auth MCP connector could delay or sink a classification and leave a session `unknown`. This is the posture the Codex classifier path already had via `--ignore-user-config`.

Argv the classifier now runs:

```
claude --print --setting-sources "" --strict-mcp-config --model <model> \
  --max-turns 2 --json-schema <verdict schema> --no-session-persistence \
  --disable-slash-commands --no-chrome --allowedTools "" \
  --permission-mode dontAsk --output-format json <prompt>
```

One deliberate loss: the old per-message `stream-json` debug dump. `--output-format json` returns a single result envelope, so the daemon log now records the result meta (structured output, final text, `num_turns`, `total_cost_usd`) and, on failure, the bounded stderr/stdout tail.

## Tests

- New `internal/agent/claude_classifier_test.go` drives a stub `claude` and asserts the argv (capped, tool-less, schema-enforced, MCP-isolated), the structured-output verdict, the final-text fallback, the failed-run path, and that empty text never spawns the CLI.
- The two SDK-typed classifier tests are re-expressed against `ParseVerdict`, plus coverage for a verdictless structured output falling through to the final text and for the model env override.

## Verification

- `go build ./...` and the full Go suite green; gofmt clean.
- Live against the real CLI: correct `waiting_input` / `idle` verdicts from structured output.
- Packaged app on a throwaway non-production profile (full `make install`, `preflight` PASS), real Claude session driven through two turns: the daemon log shows `classifier: calling claude CLI executable=…`, then `classified as idle` for the statement turn and `classified as waiting_input` for the question turn.

Observed during verification and **pre-existing, not caused by this change**: both verdicts were then dropped with `state update discarded: … cause=classifier_observation`, because a live PTY state update lands inside the classifier window and the timestamp guard in `commitSessionState` rejects the later verdict. The subprocess latency is the same as the SDK's, so this race predates the PR; the applied state happened to be correct anyway (the PTY detector reached the same values). Left alone here.

https://claude.ai/code/session_01FYZ7b7fqCW1Kh7pJVr1tgR


figgyster (commented):
You pulled the classifier out of its little SDK side alley and put it on the same headless road the daemon already patrols. That’s the right debt to pay: one isolation boundary, bounded output, and a failure that says `unknown` instead of making up an alibi. The argument-level witness is good, and Daemon plus both Linux preflights are green on this head.

I need the one piece the repo guide reserves for this sort of state-lifecycle change before I sign: a non-production daemon run from this branch showing the headless classifier’s result reaches the live session state, with its failure path remaining `unknown`. This review box cannot run the macOS app tier. Add that evidence here and I’ll close the file.

figgyster (approved):
There it is. You ran the real packaged daemon through both verdicts and made the bad-model path confess to `unknown` in the live session state. That was the missing witness, and it holds.

This fits the state-repair work that just landed: one bounded Claude seam owns the judgment, structured output gets first crack at the verdict, and failure stays honest. The older PTY/classifier arbitration scar is still on the wall, but this refactor didn’t make it. Approved.

PR #679, opened 2026-07-26
refactor(state): make the resolver the single writer

victorarias:
Phase 3.5 of `docs/plans/2026-07-25-agent-state-detection.md`, against the epic
branch. The bar for this phase is not new behavior: it is whether the next
person can read `sessionstate.Resolve` top to bottom and predict the colors.

## The resolver is the single writer

`handleState` no longer applies a state for any of the three the hook reports
(`pending_approval`, `working`, `waiting_input`) — it files evidence and the
resolver's next tick decides. This is what makes the guardian dwell real: the
flash Victor reproduced was a hook applying `pending_approval` itself, ten
seconds before the resolver ever saw the evidence, so the dwell was never
consulted. `waiting_input` keeps working through its own evidence level rather
than a direct write. The dwell is 60s.

Two live witnesses on the `statedet` profile:

- A guardian approval opened at 14:39:51 and the session stayed `working /
  heartbeat_fresh` through 47 skipped publications (`outcome=skipped
  reason=dwell`), flipping to `pending_approval / approval_open` at 14:40:52. A
  guardian that answers inside the window produces no yellow at all.
- A claude background `sleep 45` held `working / background_work` for 43s and
  then resumed into a real turn, settling on `idle / classifier_verdict` with no
  flicker in between.

## The screen-scrape path is deleted

Confirmed dead, not merely suspect: a full copilot turn (1.0.73, gpt-5-mini,
prompt through to answer) produced **zero** `source=screen` claims — its whole
trace was the spawn-time `worker_info` replay plus a classifier error. Copilot
was the last consumer, so the path goes as a unit:
`internal/pty/state_detector.go`, the `ScreenDetector` capability,
`pty.SourceScreen`, the `Screen` evidence level and its clause in `Resolve`, and
`ShouldApplyPTYState` — the last per-agent PTY-state veto, with a test in
`internal/agent` guarding against a new one. `pty.Source` is left with one
predicate, `ClaimsProtocolState`.

Stated plainly: copilot now has no state source, so its sessions hold the state
the launch applied. That is already what happened while the scrape was silent.
Copilot regains detection when it gets harness signals, not by keeping
heuristics for a TUI that no longer renders them.

## Also removed

- The `idle_stale` field leaves the wire. The rule stays in
  `internal/sessionstate`, tested and unwired: a mark clients could read but
  nothing produced a decision from invited a client to invent the meaning.
- `store.UpdateStateWithTimestamp`, whose only callers were the writers this
  phase removed.
- The daemon-side dedup of the worker's state stream, which could only drop a
  genuine refresh (the worker already broadcasts on change).

## Two coherence bugs, both found by watching a live session

- `state_reason` was recorded *above* the dwell gate, so a session serving out
  its dwell published `working` next to `approval_open` — a pair contradicting
  itself. The reason now moves with the state.
- `pruneSessionsWithoutPTY` had no recency guard while the socket listener was
  already accepting registrations, so a session registered during startup had no
  PTY yet and was marked `recoverable` — a state the resolver does not own and
  therefore never takes back. It now takes the same `recoveryStartedAt` cutoff
  its worker-backend counterpart has always used. Surfaced as a load-dependent
  failure in the stop-terminality tests once they waited for a resolved state
  instead of reading the store immediately.

## Protocol

195 (`internal/protocol/constants.go` and `app/src/hooks/useDaemonSocket.ts`),
for the `idle_stale` removal.

## Verification

- `make test-all`: Go suite exit 0, frontend 190 files / 2051 tests passing.
- Live on the `statedet` profile (protocol 195, preflight PASS) for both
  witnesses above, driven through `scripts/wsctl` against real claude and
  copilot sessions.
- `pruneSessionsWithoutPTY` runs only on the embedded PTY backend, which is not
  the default, so its fix has no reachable surface in a normal profile; it is
  covered by a unit test that pins both verdicts (skip inside the cutoff,
  recover outside it).

The epic `CHANGELOG.md` entry is written once, when the epic merges to main.


figgyster (approved):
You finally made Phase 3’s dwell gate mean what it said. #675 put the policy in the right room, but `handleState` still had a side door into `pending_approval`; this shuts it, and the yellow-flash witness is the proof I wanted to see.

I traced the remaining evidence paths through the resolver, including the stale-bracket, classifier, long-run, and external-driver escapes. The clause-order tests give the next poor soul a map instead of a séance, and the live guardian/background-work runs cover the two races most likely to lie in a unit test. Daemon, Frontend, all three frontend E2E shards, Changes, and React Doctor are green. Bench A/B is still doing its slow walk, but it is trend evidence here, not a gate. Approved. The old scraper and staleness machinery can stay buried.

PR #680, opened 2026-07-26
fix(pr): wait-ready reports every event, its content, and remembers what it reported

victorarias:
`attn pr wait-ready` is the one call an agent makes to wait on a pull request: it exists so nobody writes a polling script, and so the answer carries enough to act on without a follow-up query. Three things kept it from doing that.

**It threw away events.** Precedence lived in two places — an early return for comments and a switch for everything else — so whichever condition the code happened to test first won and the rest were discarded. figgyster approving with prose exited 4 ("a new comment") instead of 0, because the approval body was collected as commentary and the comment branch returned before the verdict branch was ever reached. A closed PR or a failed check could likewise lose to a comment.

Now a poll collects every event it sees and one ordered list picks the winner: closed, checks failed, changes requested, human comment, approved, bot comment. A human comment outranks approval because someone is waiting for an answer; a bot comment ranks last because nobody is. Everything else that poll saw is still reported — `also <event>: …` on stdout, `events` in `--json`. The reviewer's own verdict is one event, not a verdict plus a comment. Bot comments are tagged rather than dropped and get their own exit code (6), so a caller can act on a human's remark and skip a doctor report.

**It reported that something was said, not what.** stdout and `--json` now carry comment bodies with `file:line` when inline, the verdict's own text, failing check names with their URLs, and the pull request URL.

**Every call re-baselined.** Comments present at wait start are correctly treated as context, but that baseline was rebuilt from scratch on each invocation — so a remark that landed while the caller was answering the previous one was absorbed into the new baseline and never reported, and the wait then blocked for its whole timeout as if the PR were quiet. Waits now resume: each records what it reported under `<data-dir>/pr-wait/<host>/<owner>/<repo>/<number>.json`, written atomically and pruned after 30 days.

The record keeps each event in the shape it comes in, because "already seen" means something different for each:

- comments are discrete, so their IDs are the record;
- a verdict supersedes itself, so its submission time is the record — used as the baseline a re-review is measured against, not as suppression, since a standing approval is the PR's current answer and every wait should say so;
- a failing check is a condition that stays true until someone pushes, so the record is which checks failed on which commit. The same failure on the same commit does not make the next wait return instantly with nothing new; a different failure, or the same one after a push, does.

It advances only to what was reported, plus the state a first-ever wait deliberately treats as pre-existing context — a wait that times out still carries its baseline forward, which is what closes the gap. `--reset` discards the record; `--since <RFC3339>` replays from an instant by arrival time, needing no comment IDs.

## Verification

Unit tests cover the ranking, the event set, the resume across calls (including the timeout-then-gap case), failure suppression and its three re-report cases, `--since`, and the on-disk round trip including a corrupt cursor and pruning. The two new gates were mutation-checked: removing the baseline-into-cursor line and removing the `sameFailure` gate each fail exactly one test.

`attn pr wait-ready` shells out to `gh` and touches no daemon or app surface, so the built binary was exercised against real pull requests with `ATTN_DATA_DIR` pointed at a scratch dir:

- #679 (merged; figgyster approved with prose, react-doctor commented) with `--since`: `closed` plus `also bot_comment`, the bot's body and the PR URL printed, exit 5.
- #679 without `--since`: `closed`, and the cursor written with the baseline comment ID.
- #404 (open, approved with a long verdict body): exit 0, one event, the verdict text printed, cursor recording `verdict_at`. A second call reports the standing approval again, as intended.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -137,10 +137,16 @@ output as product evidence.
 
 To wait on a GitHub PR, run `attn pr wait-ready <pr> --repo <owner/repo>
 --reviewer <login>` once; do not poll checks, reviews, and comments separately.
-It returns on the first actionable update and reports which one by exit code:
-`0` approved, `1` checks failed, `3` changes requested, `4` new human comment,
-`124` timeout. Bot comments are ignored; comments already present when the wait
-starts are the baseline.
+It returns on the first poll with an actionable update and reports it by exit
+code: `0` approved, `1` checks failed, `3` changes requested, `4` human comment,
+`5` error, `6` bot comment, `124` timeout. One poll can see several of those at
+once; the exit code names the highest ranked (closed, checks failed, changes
+requested, human comment, approved, bot comment) and the rest are still reported
+— `also <event>:` on stdout, `events` in `--json`. Do not treat the exit code as
+the whole answer when you need to know everything that happened. The reviewer's
+own verdict is one event, not a verdict plus a comment. Comments already present
+when the wait starts are the baseline; `--ignore-author` drops an author of
+either kind.
 
 ### Packaged-app harness
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -148,6 +148,14 @@ own verdict is one event, not a verdict plus a comment. Comments already present
 when the wait starts are the baseline; `--ignore-author` drops an author of
 either kind.
 
+The output carries what was said — comment bodies with `file:line` when inline,
+the verdict's text, failing check names with URLs, the PR URL — so act on it
+instead of querying GitHub again. Successive waits on the same PR resume from
+what the previous one reported (recorded under the data dir), so a comment that
+lands while you are answering the last one is still reported, and a failing check
+you were already told about does not return instantly a second time. `--reset`
+forgets that position and `--since <RFC3339>` replays from an instant.
+
 ### Packaged-app harness
 
 - Single-tenant: never run packaged-app scenarios in parallel.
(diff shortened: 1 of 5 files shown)


figgyster (dismissed):
You fixed the real sin in this waiter: one poll finally tells the whole story instead of letting an early return pick who gets heard. The ranking and cursor both fit the agent handback arc you've been building — one accountable answer, no scavenger hunt after it. But there is still one witness missing from the report.

Required before merge:

- `cmd/attn/pr.go:797` builds commentary for a non-verdict review without copying `review.BodyText` into the `prGraphQLComment` passed to `appendPRComment`. A reviewer can submit a prose `COMMENTED` review (the code correctly treats it as a human-comment event), `wait-ready` exits 4, then plain output prints only the reviewer's header and JSON returns an empty `body`. That breaks the PR's promise that an agent can act without fetching the review again. Carry `BodyText: review.BodyText` there and extend the existing commented-review parser test to assert the emitted body.

The daemon lane and both Linux preflights are green. I couldn't run the focused CLI suite in this review box because its Docker sandbox is unavailable, so that witness is CI's tonight. Once the prose survives the parser, this is ready for the next pass.

figgyster (approved):
There it is. The prose now rides with the `COMMENTED` review instead of vanishing between GitHub's payload and the agent's hand, and the regression puts a handcuff on the exact line that dropped it.

This was the missing seam in a good little coordination tool: one ranked answer, the words that make it actionable, and a cursor that does not lose the next thing while an agent is busy answering the last. Victor's live branch witness proves the shell path; Daemon, Changes, React Doctor, and the Linux preflights are all green. Safe to merge. Case closed.

PR #681, opened 2026-07-26
fix(pr): wait-ready no longer reports the caller's own comments

victorarias:
`attn pr wait-ready` reported the invoking GitHub account's own comments as
actionable events. Observed on a real pull request: an agent posted a reply to
its reviewer, started a wait, and was handed exit 4 with
`comment: 1 new comment from victorarias` and its own text echoed back.

## Why the existing baseline does not cover it

Comments present when a wait starts are baselined and never reported, but a wait
now resumes from the position the previous wait recorded. A comment posted
*between* two waits is genuinely new to that cursor — including one the caller
posted itself. `--ignore-author` worked as a manual workaround but had to be
passed on every call.

## What changed

The account `gh` is authenticated as is resolved once per run, and its comments
no longer produce an event. The caller of a wait is the actor; being told about
your own action is never the update you were waiting for. `--include-self`
restores the old behavior for a caller watching a pull request it does not act
on.

Design calls behind that:

- **Default on, with `--include-self` to opt back in.** Suppression by default is
  right for the only way this command is used — act, then wait. Watching a pull
  request you also comment on is the rarer case and now says so explicitly.
- **Lazy, once per run, inside the caller's deadline.** One
  `gh api user --jq .login`, issued after argument validation and immediately
  before the first poll, so `--help`, a usage error, and `--include-self` never
  pay for it. The polling loop cannot reach the lookup; it consumes a resolved
  value on `prWaitOptions`. The call is scoped to the pull request's host and
  runs under the wait's own timeout context, so its 15s ceiling can only cut the
  lookup short, never extend the budget `--timeout` promised.
- **Failure is not fatal.** A login that cannot be resolved leaves `SelfLogin`
  empty, which is the same state `--include-self` produces: every author is
  reported, exactly as before this change, with a one-line note on stderr. The
  login can only ever *remove* an event, so not knowing it costs one spurious
  wake-up, while erroring out would cost the caller the whole wait.
- **It composes with `--ignore-author`.** Both go through `prWaitOptions.ignored`,
  so suppressing yourself is the same operation as suppressing anyone else, and
  the two stack.

Unchanged: event ranking, exit codes, the cursor format and its location, and the
content payload. A verdict is read from the review record rather than the comment
surfaces, so a self-authored review still ends the wait as it did.

## Tests

`go test ./cmd/attn` green. Three new tests. The first drives the waiter through
the real snapshot parser rather than a hand-built readiness value, so it covers
the same path a live wait takes:

- `TestWaitForPRActionableDoesNotWakeOnTheCallersOwnComment` — with a non-empty
  cursor (a resumed wait, where the baseline cannot help), a self-authored
  comment produces no event, another author's comment on the same poll still
  does, and clearing the resolved login restores the old behavior.
- `TestResolvePRSelfLoginIsSkippableAndFailureTolerant` — a failing lookup yields
  no login and no error, the lookup is scoped to the pull request's host, and
  `--include-self` does not perform it at all.
- `TestResolvePRSelfLoginCannotOutlastTheWaitDeadline` — against a GitHub that
  never answers, the lookup ends with the caller's deadline rather than with its
  own ceiling.

Mutation-checked. Six mutations, each red on its own intended assertion:

| mutation | assertion that failed |
| --- | --- |
| suppression deleted | `the caller's own comment ended the wait: outcome=comment` |
| suppress every author once self is known | `want only figgyster's comment` |
| suppress self even under `--include-self` | `want the caller's own comment reported` |
| a failed lookup returns a login anyway | `want an unresolvable login to report everyone` |
| `--include-self` still performs the lookup | `--include-self asked GitHub who we are` |
| lookup gets its own clock again | `lookup took 15.00220875s with a 50ms budget: it borrowed its own clock` |

The second mutation also exposed a flaw in the test as first written: with an
unbounded context, a wait that stops firing an event polls a fixed snapshot
forever and hangs the package instead of failing. The two waits that are supposed
to return now carry a deadline.

The last row is the review finding from the first round, reproduced: with the
lookup on its own clock, a 50ms budget still took 15 seconds.

## Live verification

`attn pr wait-ready` shells out to `gh` and touches no daemon or app surface, so
per `AGENTS.md` the built binary run against a real pull request is the right
tier. Run against this PR at head `b947b5e2` with `ATTN_DATA_DIR` pointed at a
temp directory, so no cursor was written to `~/.attn`.

The two waits below see **identical pull request state and the identical stored
cursor** (`['IC_kwDOQoFRj88AAAABLwos7g', 'IC_kwDOQoFRj88AAAABLwpBiw']` before and
after A). The only difference is whether the running account is treated as self.

**A — post a comment, then wait.** The cursor already held earlier comments, so
this is a resumed wait and the new comment is unseen by it:

```
$ gh pr comment 681 --body "Third live-witness remark, posted by the account that runs the waits below (head b947b5e2)."
https://github.com/victorarias/attn/pull/681#issuecomment-5084194651

$ attn pr wait-ready 681 --repo victorarias/attn --reviewer figgyster --timeout 25s --interval 10s
pr=#681 head=b947b5e2e9a3 state=open draft=false checks=pending [...] review=waiting reviewer=figgyster
timeout: no actionable update after 25s (checks=pending review=waiting)
https://github.com/victorarias/attn/pull/681
exit=124
```

The stored cursor is unchanged afterwards — a suppressed comment is never
recorded as reported.

**B — control, same state, same cursor, `--include-self`:**

```
$ attn pr wait-ready 681 --repo victorarias/attn --reviewer figgyster --timeout 25s --interval 10s --include-self
pr=#681 head=b947b5e2e9a3 state=open draft=false checks=pending [...] review=waiting reviewer=figgyster
comment: 1 new comment from victorarias
  --- victorarias ---
  Third live-witness remark, posted by the account that runs the waits below (head b947b5e).
https://github.com/victorarias/attn/pull/681
exit=4
```

Same comment, same poll, same cursor: reported when the author is not treated as
self, silent when it is. That is the suppression being identity-scoped rather
than blanket.

**Another author still ends the wait**, witnessed on this PR: the babysitting
wait on head `89b9b9b9` — running as `victorarias` with suppression on — returned
on figgyster's review:

```
pr=#681 head=89b9b9b98109 ... review=changes_requested reviewer=figgyster
changes_requested: figgyster requested changes on 89b9b9b98109
  --- figgyster ---
  [review text]
exit=3
```

**Short timeout stays inside its budget** at head `b947b5e2`:

```
$ time attn pr wait-ready 681 --repo victorarias/attn --reviewer figgyster --timeout 1s --interval 20s
timeout: no actionable update after 1s (checks=none review=waiting)
elapsed_ms=1567
```

The witness comments on this PR are the fixture for the runs above.

`AGENTS.md` records the new default in the `wait-ready` paragraph.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -146,7 +146,10 @@ requested, human comment, approved, bot comment) and the rest are still reported
 the whole answer when you need to know everything that happened. The reviewer's
 own verdict is one event, not a verdict plus a comment. Comments already present
 when the wait starts are the baseline; `--ignore-author` drops an author of
-either kind.
+either kind. Your own comments are never events: the account `gh` is
+authenticated as is resolved once per run and its remarks are dropped, so posting
+a reply and then waiting does not wake you with your own text. Pass
+`--include-self` to watch a PR you also comment on.
 
 The output carries what was said — comment bodies with `file:line` when inline,
 the verdict's text, failing check names with URLs, the PR URL — so act on it
--- a/cmd/attn/pr.go
+++ b/cmd/attn/pr.go
@@ -246,7 +246,9 @@ type ghPRReadinessSource struct{}
 
 // prSelfLoginTimeout bounds the identity lookup. It is one small API call that
 // the wait cannot start without, so a hung network must degrade to "we do not
-// know who we are" rather than hold the wait open.
+// know who we are" rather than spend the caller's whole budget before the first
+// poll. It is a ceiling, never an extension: the lookup runs under the wait's
+// own deadline, so a --timeout shorter than this cuts it short instead.
 const prSelfLoginTimeout = 15 * time.Second
 
 // ghSelfLogin is the seam over `gh api user`. A variable so tests can drive both
@@ -267,11 +269,15 @@ var ghSelfLogin = func(ctx context.Context, host string) (string, error) {
 // fail. The login only ever removes an event, so not knowing it costs the caller
 // one spurious wake-up, while erroring out would cost them the wait entirely —
 // the command has to keep working on a machine whose `gh` cannot reach the API.
-func resolvePRSelfLogin(opts prWaitOptions, stderr io.Writer) string {
+//
+// It takes the wait's context rather than making its own, so the lookup spends
+// the caller's budget instead of adding to it: whichever of the two deadlines
+// comes first wins.
+func resolvePRSelfLogin(ctx context.Context, opts prWaitOptions, stderr io.Writer) string {
 	if opts.IncludeSelf {
 		return ""
 	}
-	ctx, cancel := context.WithTimeout(context.Background(), prSelfLoginTimeout)
+	ctx, cancel := context.WithTimeout(ctx, prSelfLoginTimeout)
 	defer cancel()
 	login, err := ghSelfLogin(ctx, opts.Host)
 	if err != nil || login == "" {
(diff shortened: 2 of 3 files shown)


figgyster (changes requested):
You closed the cursor gap that #680 left behind: after an agent speaks, `wait-ready` should be listening for the next voice in the room, not throwing its own reply back across the desk. The live witness and the parser-level regression both pin that down.

But `--timeout` promises a maximum wait, and the new `gh api user` lookup gets a separate 15-second clock before that timeout context exists. If GitHub stalls, `attn pr wait-ready --timeout 1s ...` can take roughly sixteen seconds before returning. Run the identity lookup under the command's overall deadline (or otherwise cap it by the remaining wait budget), then add the short-timeout regression. The fallback-to-report-everyone behavior is right; it just cannot borrow time the caller did not give it.

figgyster (approved):
There it is. The identity lookup now spends the same deadline the caller gave `wait-ready`, and the 50ms stalled-lookup witness nails the old fifteen-second escape hatch to the floor.

This finishes the coordination loop #680 opened: the durable cursor no longer hands an agent its own reply, and it no longer quietly cheats the wait budget while fixing that. The new-head live run and the green daemon/Linux/React Doctor witnesses hold. Approved.

PR #682, opened 2026-07-26
feat(state): resolve agent state from harness signals instead of the screen

victorarias:
Session colors were wrong often enough to be untrustworthy: a session stuck yellow after its approval was granted, a session left green after it had stopped, and `pending_approval` invented out of prose sitting in the terminal. The cause was the detection method. attn read the rendered TUI and guessed, while both Claude Code and Codex were already broadcasting what they were doing and being ignored.

This epic replaces the guess with the broadcast, and puts one ordered set of rules in charge of the answer.

## What it does

**The signals.** Both harnesses repaint a terminal-title glyph that says whether a turn is running — claude about once a second, codex about ten times. Claude's hooks bracket a turn (`UserPromptSubmit` / `Stop`) and announce permission requests (`Notification`). Measured against hook ground truth the title tracks the turn to within ~100ms in both directions. Screen scraping, on the same corpus, scored 1.0% on a 35s foreground tool call.

**The resolver.** `internal/sessionstate.Resolve` is a pure function over an evidence table: same evidence and same clock always give the same answer. Its clauses are ordered, first match wins, and the order encodes trust rather than recency — a fresh heartbeat outranks an open approval, because an agent cannot be blocked on the user while its turn is visibly running. It is the single writer for what an agent is doing; six subsystems used to write state directly, and a color could depend on which one wrote last.

**Bounded failure instead of stuck failure.** Hook brackets carry "work is outstanding"; the heartbeat bounds how long a bracket may lie. A lost closing hook settles once the agent has been quiet past `StaleAfter`, and total evidence silence for 90s reports the session stuck rather than leaving a color nobody should trust. With hooks perfect the heartbeat *costs* 1.5pp of accuracy; at 5% hook loss it cuts the worst stuck streak from 60.0s to 3.8s. That trade is the point — the reported bug was stuck colors, not average error.

**The guardian flash.** An unattended run with a guardian answering prompts used to flash yellow on every tool call. An approval now dwells 60s when a reviewer is in the loop — far past the guardian's measured round trip (90ms claude, low seconds codex) — and not at all when there is no reviewer.

**Diagnosis.** `attn state explain <session>` (`--json`) replays the observation ring and names the clause that produced the current color. Intermittent color bugs previously left no evidence trail at all.

Protocol 195. Copilot keeps screen reading until it has harness signals.

Plan and full decision log: `docs/plans/2026-07-25-agent-state-detection.md`. Phase 4 (the attention projection) is deliberately unplanned.

## On the heartbeat TTL

The per-agent TTLs were measured on an idle machine, leaving open whether a loaded PTY could batch reads, stretch the gap between title frames past the TTL, and settle a session that is still working. It cannot, and the reason is structural rather than a margin: expiring the TTL only stops the heartbeat from *overriding* the brackets, and an open bracket is then governed by `StaleAfter`, which at 60s is 40x claude's TTL. A wrong color needs a second independent fault — a lost turn-open hook, leaving the heartbeat as the only evidence — and the next title frame corrects it.

Raising the TTL to buy margin would be the wrong trade regardless: a stale-but-within-TTL busy glyph holds a session green for up to the TTL after a turn genuinely ends, so every millisecond added is settle latency paid on every turn to soften a blip that needs two simultaneous faults. `TestHeartbeatTTLExpiryCannotSettleAnOpenBracket` sweeps the whole span between the two windows, so a change that couples them fails there instead of quietly retiring the margin.

## Verification

Full Go suite green (45 packages, including `daemon` at 79s), 2051 frontend tests green.

Live on a throwaway `epicship` profile, freshly installed from this branch — preflight PASS with CLI, app, and daemon all at protocol 195; packaged-app smoke green; the `tr402-local-claude` scenario (a real claude session in the packaged app) green. `attn state explain` on that live session, mid-turn:

```
session 8c1a46c1-33e7-4a7c-8880-206f71e2ad98 (claude)
state:   working since 17:47:48.988

OBSERVED         SOURCE         CLAIM       OUTCOME     WHY
17:47:46.868     worker_info    working     applied     "watch subscribe replay"  cause=live_signal
17:47:47.569     heartbeat      not_busy    observed    "Claude Code"
17:47:47.988     resolver       idle        applied     "at_prompt"  cause=resolver_observation
17:47:48.232     heartbeat      busy        observed    "Claude Code"
17:47:48.239     reviewer       auto        observed
17:47:48.240     hook_state     working     observed
17:47:48.988     resolver       working     applied     "heartbeat_fresh: Claude Code"  cause=resolver_observation
17:47:51.116     heartbeat      busy        observed ×2 "Render width coverage verification for split stability"
17:47:51.822     heartbeat      not_busy    observed    "Render width coverage verification for split stability"
17:47:51.988     resolver       -           skipped     awaiting_verdict
```

Every source feeding the table, the resolver arbitrating between them, the title's turn summary riding along — and the last line is the anti-flicker gate holding the pre-settle color while a classification runs, instead of asserting idle and correcting itself a moment later.


figgyster (approved):
You finally pulled session state out of the terminal's theatre and gave it a chain of custody. The last week was the setup: #674 killed the screen scraper, #675 made guardian dwell real, #676 taught `unknown` to explain itself, and #679 put one writer at the door. This branch closes the case instead of hiring another bouncer.

The resolver's clause order is the piece I read twice. Fresh running evidence retires stale approval, brackets age into an honest settle or `unknown`, and removal clears evidence plus dwell so departed sessions do not haunt the daemon. The recorded live harness work and the green Daemon, Linux, Frontend, E2E, Bench, Changes, and React Doctor witnesses make a very large diff believable.

Approved. Ship it before the terminal learns another lie.

PR #683, opened 2026-07-26
fix(doorbell): send Enter as its own keypress after the paste

victorarias:
## The bug

Sending markdown annotations to a Claude session wrote the feedback into Claude's composer but never submitted it — the user had to press Enter themselves.

## Root cause

`typeDoorbell` delivered the payload and its Enter as a single PTY write: `ESC[200~ … ESC[201~\r`. Claude Code finalizes a paste on a timing boundary, so a CR arriving in the same read as the paste terminator is folded into the pasted text. Splitting the write alone does not help — an immediate second write is coalesced into that same read.

This shipped with the July 13 "submit Codex nudges atomically" change, which introduced the bracketed paste to give Codex a semantic boundary and kept the CR in the same write.

Every doorbell shares this primitive, so the ticket nudge, the notebook inbox doorbell, and the Present review notice had the same defect against Claude.

## Fix

Enter follows the paste as its own write, after a settle delay (150ms). Both writes stay under `doorbellMu`, so a `pending_approval` report still cannot commit between the prompt and its Enter.

## Evidence

Measured against a real `claude` driven over a PTY:

| delivery | result |
| --- | --- |
| CR in the same write (before) | `[Pasted text #1 +8 lines]` stranded in the composer, 0 tokens |
| CR as a separate write, no delay | same failure |
| CR after 50ms | submits |

Codex still submits with the delayed Enter, so the boundary the July 13 fix added is preserved.

## Live verification

Branch daemon on a throwaway profile, a real Claude session spawned through it, then the actual `markdown_annotations_save` + `markdown_annotations_submit` commands over the websocket: result `delivered`, and the session ran a turn on the feedback (`transcript watcher: assistant event … "Shortened the heading to # Probe."`). The reader UI is unchanged, so the verification drives the same command the app sends.

## Tests

`TestTypeDoorbellDelaysEnterAfterThePaste` pins that Enter is a separate write arriving a real interval after the paste. Existing doorbell tests move to the paste-then-Enter shape; the state-fence test now holds the paste write open to prove the fence spans the whole pair. Daemon `TestMain` zeroes the delay so no fake PTY waits on it.

https://claude.ai/code/session_01179jS4eGexTy5NeLo6JRJF

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-07-26]
 
+### Fixed
+- **Sending markdown annotations to a Claude session now actually submits.** The
+  feedback landed in Claude's composer and sat there unsent until you pressed
+  Enter yourself. Claude folds an Enter that arrives together with a paste into
+  the pasted text, and attn was sending both at once. Enter now follows the paste
+  as its own keypress. Same fix for the ticket nudge, the notebook inbox
+  doorbell, and the Present review notice, which are delivered the same way.
+
 ### Added
 - **`attn state explain <session>` (`--json`)** replays a session's recent state
   observations and shows which rule produced the color it is showing, and why.
--- a/internal/daemon/automations_deliver.go
+++ b/internal/daemon/automations_deliver.go
@@ -783,7 +783,7 @@ func (d *Daemon) passUnattendedLaunchGate(req automation.WorkRequest) error {
 			screen := stripANSIForPromptMatch(payload)
 			if strings.Contains(screen, codexDirectoryTrustPrompt) {
 				if !acknowledged {
-					if err := d.ptyBackend.Input(context.Background(), req.IDs.SessionID, []byte("\r")); err != nil {
+					if err := d.writePTY(req.IDs.SessionID, []byte("\r")); err != nil {
 						return fmt.Errorf("accept Codex directory trust: %w", err)
 					}
 					acknowledged = true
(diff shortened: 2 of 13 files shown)


figgyster (dismissed):
I want this Claude fix. The composer needs a real keypress, and the measured delay is the right kind of ugly truth to write down. But this crosses the same PTY boundary that #679 worked hard to make accountable, and the old one-write shape was also protecting a user from getting their keystroke drafted into an automated submission.

Required before merge:

- `internal/daemon/chief_of_staff.go:173` leaves a 150ms interval after the paste where `handlePtyInput` can call `ptyBackend.Input` without taking `doorbellMu`. A user who starts typing in that interval produces `[paste][their bytes][Enter]`; the delayed Enter then submits the doorbell plus their partial line. The existing recent-input check only looks before the paste, so it cannot close this race. Keep ordinary PTY input out of the gap (or use a shared per-session write fence) and add a regression that races user input after the paste, proving the observed order is paste, Enter, then user input.

The daemon, Linux preflights, benchmark, and focused timing story are green. Once the splice fence covers the new gap, this gets us the Claude behavior without teaching the doorbell to borrow somebody's keyboard. Right now it still has that bad old alley in it.

figgyster (approved):
That closes the gap I found. You kept the old state fence for its job, gave the PTY its own per-session lock, and made the test walk straight into the 150ms rain instead of waving from the curb. The ordering witness is the good kind: it proves the keystroke arrived before Enter, then proves it could not cut the line.

This follows the state-repair work from #679 and #682 in the right spirit. One writer owns a boundary, lifecycle cleanup takes the fence with the session, and the user’s half-typed line remains theirs. Daemon and Linux preflights are green. Safe to merge. Case closed.

PR #684, opened 2026-07-26
The agent queue: a turn opens on a state and closes only when you settle it

victorarias:
Slice 1 of the agent-queue plan (`docs/plans/2026-07-26-agent-queue-turn-and-settle.md`), targeting the epic branch.

The sidebar today lists agents and highlights the ones that are stopped in a state that wants you. That means a row leaves the moment the agent moves — including when *you* moved it, by answering — so the thing you were in the middle of disappears from the list you are working through, and an agent that finished quietly never appears at all.

This makes the sidebar a list of turns you owe. A turn opens when an agent reaches a state that wants the user and closes only when you settle it. No state transition ever removes a row.

## What changes

- **The daemon owns whose turn it is.** Two persisted stamps per session, `turn_opened_at` and `turn_settled_at`; a turn is owed iff opened is after settled. Opening is conditional (a SQL `WHERE` guard), so a state re-reported while a turn is already open leaves the original age alone and the row never moves under you. Settling is unconditional. `applyState` is the one place a turn opens, so startup recovery counts too: an agent that comes back wanting you wants you regardless of what put it there.
- **`turn_owed` is derived per broadcast, never stored.** Shell sessions, the chief of staff, and pinned or muted workspaces are filtered at read, not at open — their stamps still accrue, so unmuting a workspace surfaces the turns that were owed all along.
- **The mode gates the band, not the daemon.** Stamps accrue and `turn_owed` broadcasts whether the arrangement is on or off, so turning it on mid-session shows the queue that was already true.
- **A new sidebar arrangement, off by default** (Settings → General, plus a ⌘K action): the chief in its own anchored slot, then "Your turn" — the owed turns, oldest first by `turn_opened_at`, each showing its live state, its workspace, and how long it has been owed. The workspace tree below is untouched; the queue is additive, never a filter.
- **Settle** from the row's ✓ or ⌘E on the selected agent.
- When the arrangement is on, the four surfaces that ask "what wants me" — the grid tile glow, the attention drawer, jump-to-waiting, and the collapsed rail badge — follow `turn_owed` instead of the session state.

Protocol 196: `Session.turn_owed`, `Session.turn_opened_at`, and a `settle_turn` command. `turn_settled_at` stays off the wire — membership is the daemon's answer, and a client recomputing it would get a different one because it cannot see the exclusions.

## Verification

`real-app:scenario-agent-queue` drives two live Claude agents in two workspaces through the whole contract and asserts against the rendered band via a new `queue_get_state` automation verb: both turns in the band oldest-first and both still in the tree; clicking a row hands the agent over with the keyboard in its terminal; steering it leaves the row in the same position showing that it is working; a finished run does not remove it and a settle does; a settled agent that wants you again returns at the bottom; the chief occupies its own slot and never the band; toggling the arrangement off and back on returns the same queue with the same agent selected; a settle survives a daemon restart. Run green against a throwaway profile install (11/11 steps).

Slice 2 — a finished run also opening a turn — is a separate PR.


figgyster (changes requested):
You are finally giving the sidebar the thing it has been pretending to be: a list of promises the user still owes, not a weather report that changes while they are answering it. #679 made one authority for agent state; this carries the same accountable shape into attention itself. The guarded stamps, broadcast decoration, and remote pass-through are the bits I read twice. They hold.

But the current head still has a red required `Frontend E2E` gate (shard 1/3). Daemon, Frontend, the other two E2E shards, Bench A/B, and the static checks made it home, but the full E2E gate did not. A queue that claims to survive steering and restart needs that witness in the room.

Required before merge:

- Get the failing Frontend E2E shard green on this head, with the underlying failure fixed or shown to be an infrastructure retry and rerun successfully.

figgyster (approved):
You found the missing witness yourself: Command-E was a live wire in the Notebook. Moving settle to Command-Shift-E and driving the packaged app through that native accelerator puts the test back on the path the user actually walks.

The fresh Daemon, Frontend, and all three Frontend E2E shards are green, including the queue's settle-and-restart story that held my earlier review open. The guarded turn ledger carries #679's single-authority instinct into user attention without reopening the resolver. This is the first real slab of Phase 4. It holds.

PR #685, opened 2026-07-27
feat(queue): a finished run opens a turn

victorarias:
Slice 2 of the agent queue (`docs/plans/2026-07-26-agent-queue-turn-and-settle.md`), on top of slice 1. Targets `epic/agent-state-detection`.

## What changes

**`idle` opens a turn.** A run you asked for that finished without asking anything is a result nobody has read; it lands at the bottom of the queue and stays there until you settle it. This is the vision's largest consequence: nothing that ever ran leaves your plate by itself.

**A session sitting at its prompt does not.** `internal/sessionstate` resolves a never-run session at its prompt to `idle` too — indistinguishable at the state level from a finished run. Unguarded, every freshly launched agent would queue from the moment it booted. `sessionStateChange.atPrompt` carries the distinction through the state door: the state commits identically, it only skips the open, because there is no result behind it to read. This was not in the plan; it is recorded under `Decisions`.

**The long-run review gate is deleted whole.** It approximated the same thing badly — holding a 5+ minute run's final classification back until someone looked at the session. Removed: `needs_review_after_long_run` from the schema, `Session`, and the hub's equality check; the `session_visualized` command and the app's 5s dwell timer; `longRun` tracking with `markRunStartedIfNeeded` / `clearLongRunTracking` / `longRunReviewThreshold` and every call site; `classifyOrDeferAfterStop`'s deferral branch and `handleSessionVisualized`; `sessionstate.IdleStale`, `Policy.IdleStaleAfter`, and `defaultIdleStaleAfter`. A finished run now publishes its true color immediately.

Protocol 197 (`constants.go`, `generated.go`/`generated.ts`, `useDaemonSocket.ts`).

Selected-session tracking is unaffected: the daemon has always had a separate `session_selected` command, which the frontend still sends.

## Verification

`go test ./...` and `pnpm test` (2070 tests) green.

New Go tests: a finished run owes a turn; a session at its prompt owes nothing; a long run publishes its verdict immediately rather than on view.

Live: full `make install PROFILE=queue1`, preflight PASS (protocol 197 CLI/app/daemon), then `real-app:scenario-agent-queue` green across all 13 steps against two live Claude agents, including two new ones —

- `a_finished_run_returns_the_agent_to_the_queue`: alpha is settled, prompted with a task that ends without a question, stays out of the band while working, resolves to `idle`, and returns at the bottom behind the older turn.
- `a_shell_pane_never_queues`: a split shell pane registers as a real session in `idle` — the same state — and changes nothing in the band.

The pre-existing empty-band assertion on a freshly booted agent is what covers the `atPrompt` guard end to end.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -197,8 +197,10 @@ See [docs/plans/2026-07-18-db-loss-mitigation.md](docs/plans/2026-07-18-db-loss-
 - `app`: Tauri frontend; WebSocket `ws://localhost:9849`
 
 States: `launching`, `working`, `pending_approval`, `waiting_input`, `idle`,
-`unknown`. `needs_review_after_long_run` is a separate flag: a 5+ minute run
-defers final classification until viewed, then clears the flag.
+`unknown`, `scheduled`, `recoverable`. A turn opens when a session reaches a
+state that wants the user (`internal/attention`) and closes only when the user
+settles it; `turn_owed` is derived at broadcast from the persisted
+`turn_opened_at`/`turn_settled_at` stamps, never stored.
 
 IPC: `~/.attn/attn.sock`. WebSocket clients buffer 256 messages; sustained
 over-send may drop messages or disconnect slow clients.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,17 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   comes back the next time it wants you. Everything you have not settled is still
   there after a restart.
 
+### Changed
+- **An agent that finishes its run joins the queue too.** A run you asked for
+  that ended without a question is still a result nobody has read, so it lands at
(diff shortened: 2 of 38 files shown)


figgyster (approved):
You took the queue promise from #684 all the way to its uncomfortable conclusion: a finished run owes the user a look, even when it ends quietly. Good. The `at_prompt` reason is the bit I read twice. It lets a newborn agent become honestly idle without manufacturing a debt from thin air, while a real finished run comes back through the same state with a turn open.

And you buried the five-minute visualisation gate properly — daemon, app, wire contract, generated types, and the real-app witness. No orphaned flag waiting behind a wallpapered wall. The queue now carries the result instead of holding the result hostage until somebody happens to click the pane.

The formatting follow-up repaired the only red witness. Daemon, Frontend, all E2E shards, Bench A/B, and the focused queue harness are green on `f18227f`. Approved. Let the queue remember what people asked for.

PR #686, opened 2026-07-27
refactor(daemon): move the classify shell beside its rules

victorarias:
Pure code move, no behavior change.

`internal/daemon/classify_decision.go` holds the pure rules that decide what a settled turn means, and its header already described `classifySessionState` as "the shell that performs the IO between them". That shell lived ~1100 lines away in `daemon.go`, along with `runClassifier`. This moves both next to the rules they interleave with and adjusts the file header to say that both halves are now there.

`daemon.go` goes from 3807 to 3650 lines. No signature, call-site, or logic change — `git show --stat` is one function pair leaving one file and arriving in another, plus a `time` import.

Live verification is not applicable: the change has no app-observable surface at all — no protocol, lifecycle, PTY, timing, or UI — and the moved functions are covered by the existing daemon suite (`go test ./internal/daemon/` green).


figgyster (approved):
This is the right little cleanup after the state-repair run. The resolver gets to stay the authority, while the slow transcript/classifier shell now lives beside the decisions it feeds instead of lurking eleven hundred lines away in the old daemon alley.

I traced the move: signatures, call sites, capability gates, timestamp guard, deferred classifier cleanup, and fallback classifier path all came across intact. The plan now records the remaining simplifications honestly, too. Daemon, Changes, frontend E2E, and React Doctor are green. Safe to merge. The Bench run is still chewing its coffee, but this patch gives it no new behavior to bite.

PR #687, opened 2026-07-27
fix(queue): an agent you launched and have not spoken to owes a turn

victorarias:
An agent launched into a fresh workspace and never prompted sat outside the
queue. Reported live against a build of the epic, and confirmed in the daemon
log: the session resolved `idle` with reason `at_prompt` one second after spawn
and no turn was opened.

That was the guard slice 2 shipped with, working as written. `sessionstate`
resolves a never-run session at its prompt to `idle` — the same state a finished
run resolves to — and `sessionStateChange.atPrompt` carried the distinction
through the state door so the launch would not queue.

The guard read the queue as results waiting to be read. It is turns you owe, and
an agent you launched and have not typed into is the purest one there is:
nothing will ever happen in it until you do. The guard is gone, so `idle`
opening a turn needs no exception.

Removing it also retires a sharper edge. The guard's evidence, `everTookATurn`,
is in-memory and scoped to the daemon's lifetime rather than the session's, so
after a daemon restart every quiet session resolved `at_prompt` — including a
conversation hundreds of exchanges deep — and a run finishing inside the restart
gap could be swallowed instead of queued.

## Tests

- `TestASessionAtItsPromptOwesATurn` replaces `TestASessionAtItsPromptOwesNothing`,
  asserting the launch case now queues.
- Two harness changes, both fixing tests that were reading the wrong thing:
  the shell step waited for the session to exist and asserted its state on first
  sight, catching the spawn-time `working` color rather than the `idle` the
  exclusion is about; and the drive-to-a-turn budget was tight enough that a live
  agent choosing to explore looked like a product failure.

## Live verification

`real-app:scenario-agent-queue` against `attn-queue1.app`, protocol 197 matching
across CLI, app, and daemon (`attn preflight` PASS). All 13 steps green,
including the new assertion that an agent joins the band from the moment it
boots to its prompt, and the shell pane still never queuing from the same `idle`
state.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,8 +27,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ### Changed
 - **An agent that finishes its run joins the queue too.** A run you asked for
   that ended without a question is still a result nobody has read, so it lands at
-  the bottom of the list. A freshly launched agent sitting at its prompt is
-  unaffected — it has nothing behind it to read yet.
+  the bottom of the list. So does an agent you just launched and have not spoken
+  to yet: nothing will happen in it until you type, which makes it a turn you owe
+  like any other.
 
 ### Removed
 - **The long-run review gate is gone.** Runs over five minutes used to hold their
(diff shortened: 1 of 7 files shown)


figgyster (approved):
You corrected the premise, not just the predicate. #685 taught the queue to carry finished work; this catches the quieter debt from #684’s turn ledger: an untouched launch is still a promise hanging in the room. Dropping `atPrompt` is right. Its lifetime-only `everTookATurn` escape hatch was a restart-shaped hole in a durable record.

The state-door regression pins the fresh launch, the shell remains excluded, and the real-app scenario now watches the actual boot-to-idle path. Daemon, Changes, frontend E2E, and React Doctor are green; Bench A/B is still running, but it is not a gate here. Safe to merge. Case closed.

PR #688, opened 2026-07-27
feat(queue): the sidebar in queue mode is the standing order

victorarias:
## What

With the agent queue on, the sidebar is now the standing order itself rather than a band sitting on top of the unchanged workspace tree:

- **Your turn** — the turns you owe, oldest first.
- **Settled** — everything else you could go and look at, in the tree's own order (workspace rank, then session order), which is deliberately not a state order.
- The chief keeps its anchored slot above both and never queues.

An agent is drawn in exactly **one** place. Before this, every queued agent appeared twice — once in the band and once in its workspace group — so a row looked like it moved when only one of its two copies did, and there was nowhere for a settled agent to go.

Pinned and muted workspaces keep their groups below the bands. A pinned workspace is where you go and get work rather than a list handed to you, and pinning is how an agent leaves the queue for good. Turning the queue off restores the full tree unchanged.

## Affordances that the tree used to own

Removing the groups removes the only place two things could be asked for, so both move onto the rows:

- **Pin** — a 📍 on every queue row pins its workspace; without it the queue would be a one-way door. ⌘K also gains *Pin workspace* and *Mute workspace* for the workspace of the agent you are in.
- **The per-session menu** (chief of staff, close, reload) — the ••• that the workspace-tree session row owned. The packaged-app run caught this one: with the row gone, the whole menu was out of reach for anything in a band.

## Shape chosen

Settled rows are flat, with no workspace grouping; only pinned and muted workspaces still render as groups.

## Verification

`ATTN_PROFILE=queue1 pnpm run real-app:scenario-agent-queue` against a packaged build of this branch — 14/14 steps green, including the two new ones (`pinning_from_a_row_takes_the_agent_out_of_the_queue`, and `the_chief_never_queues` now opening the actions menu from a band row). Frontend suite: 2077 passing. `tsc --noEmit` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,8 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 - **Pin and mute a workspace from anywhere.** Every queue row has a pin button
   that takes its workspace out of the queue, and ⌘K can pin or mute the workspace
   of the agent you are in. Pinning used to be reachable only from a workspace
-  group header, which queue mode no longer draws.
+  group header, which queue mode no longer draws. The per-session menu — chief of
+  staff, close, reload — is on the queue rows for the same reason.
 
 ### Removed
 - **The long-run review gate is gone.** Runs over five minutes used to hold their
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,19 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   to yet: nothing will happen in it until you type, which makes it a turn you owe
   like any other.
 
+- **The sidebar in queue mode is now the whole standing order.** *Your turn* is
+  followed by *Settled*, and an agent is drawn in exactly one of them — one row
+  per agent, no workspace groups. Previously the band sat on top of the unchanged
+  workspace tree, so every queued agent appeared twice and looked like it moved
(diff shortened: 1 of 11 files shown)


figgyster (changes requested):
This is the right next cut after #684 through #687. The queue finally stops being a second sidebar with an alibi and becomes the order of business, while pinned work stays a place you choose to visit. The end-to-end witness has teeth, and the green frontend gates back up the main path.

But there is one hole in the one-agent/one-place promise.

Required before merge:

- `app/src/utils/queueBands.ts:68` chooses the chief before the pinned/muted exclusion at line 74. Pin or mute the chief's workspace through the still-reachable session action menu, then queue mode renders that chief in its anchored slot and again in the surviving workspace group. That is the duplicate-row problem this PR is here to bury. Apply the workspace exclusion before choosing the chief, or keep the chief out of the tree by another explicit rule, and add the pinned/muted-chief regression.

figgyster (approved):
That closes the only hole I found. You kept the chief in the one seat that should ignore the queue’s filing rules, then taught the surviving pinned and muted groups to leave that row alone. That’s the right split. The regression goes through both flags and the real-app scenario walks the pinned path, which is the witness this sidebar contract needed.

Phase 4 has stopped being a state ledger hiding behind a tree. #684 through #687 made the user’s debts durable; this makes them readable without making the same agent haunt two rows. Frontend and all three E2E shards are green on `df5f4a6`. Approved.

PR #689, opened 2026-07-27
feat(queue): the sidebar is the list of turns you owe

victorarias:
Second and final batch of the agent-state-detection epic. The first batch (#682) replaced screen-scraping with harness-owned signals so a session's state is trustworthy. This batch spends that trust: the sidebar becomes a list of the turns you owe.

## What changes for the user

**A turn is owned by the user, not by the agent.** An agent opens a turn when it reaches a state that wants you — waiting for input, waiting for approval, a state attn cannot explain, or a run that simply finished. The turn closes only when *you* settle it (⌘⇧E). Prompting the agent does not close it: you sent it back to work, and only you know whether that discharged what you owed it. Settling an agent that is still running is the ordinary move — it means "done with this for now", and the agent returns to the list the next time it wants you. Membership survives a restart; it is two persisted stamps, and `turn_owed` is derived at broadcast from them rather than stored.

**The sidebar in queue mode is the standing order**, off by default, toggled in Settings > General or from ⌘K:

```
chief of staff        (anchored, never queues)
Your turn             (oldest owed first, across every workspace)
Settled               (everything else, in tree order)
pinned workspaces     (kept as groups — a pinned agent lives here, not in the queue)
muted workspaces      (unchanged)
```

Every agent is drawn in exactly one place, which is what makes a row's position mean something. Ordering is by how long the turn has been owed, so a row does not move when the agent changes state under you; a new arrival lands at the bottom and leaves the rows above untouched.

Because queue mode stops drawing most workspace group rows, the affordances those rows owned moved onto the queue rows: pin, and the per-session menu (chief of staff, close, reload). ⌘K can pin or mute the workspace of the agent you are in. Turning the queue off restores the full tree exactly as before.

**The long-run review gate is gone.** Runs over five minutes used to hold their final color back until you looked at the session. A finished run now publishes what it actually is straight away, and the queue is what carries it to you.

## Shape

- `internal/attention` — reduced to `turn.go`: a pure statement of which states open a turn, importing neither the store nor the daemon. The old aggregator, adapters, and source are gone with the review gate they carried.
- `internal/store/turn.go` — the two stamps and their transitions; the store is the only writer.
- `internal/daemon/turn.go`, `session_state.go` — opening and settling on the existing `applyState` path.
- `app/src/utils/queueBands.ts` — pure derivation of chief / turns / settled from the workspace view models. It reads `turnOwed` rather than deriving it from state, because the daemon applies exclusions a client cannot see.
- `app/src/components/QueueBands.tsx`, `Sidebar.tsx` — rendering, and the reduction of the tree to what the bands exclude.

No protocol version bump beyond what the merged slices carried.

## Verification

Each slice was live-verified on its own PR and merged into the epic only after figgyster's approval. `app/scripts/real-app-harness/scenario-agent-queue.mjs` drives two real agents through open, reorder, settle, restart, and the pinned/muted/chief exclusions against the packaged app; it is re-run on this merged head and the result posted below.

## Deferred

Tracked in `docs/plans/2026-07-26-agent-queue-turn-and-settle.md` under Follow-ups: the Dashboard's own filter still reads `state === 'waiting_input'` alone rather than `turn_owed`; `store.UpdateState` restamps `state_since` on same-state writes; settle events are recorded in the state trace but nothing reads them yet. "Agent queue" is still a placeholder name — *Your turn* and *Settled* are the settled vocabulary.


figgyster (changes requested):
+You're closing the loop from #682 through #688: trustworthy state now becomes an honest ledger of what the user owes. I like that bet. The stamps, read-time exclusions, and the real-app queue witness all point in the same direction.

But the hub loses the one action that makes this ledger dischargeable. A remote session is shown in the queue, yet its `settle_turn` is not added to `remoteCommandSessionID`. The hub therefore does not forward it to the owning daemon, then handles it locally; that store cannot settle the remote turn, and the row remains owed when the next remote snapshot arrives.

Required before merge:

- Route `CmdSettleTurn` by `SettleTurnMessage.SessionID` in `remoteCommandSessionID`, with the matching remote-routing regression. The queue spans remote sessions too, so settling has to travel as far as displaying did.

The rest of this holds together. CI is green, and the focused Go sandbox got as far as the checkout's missing Ghostty VT headers before compilation — that environmental witness couldn't exercise the daemon packages.

figgyster (changes requested):
The harness wait is better evidence for the pinned-chief walk. But this push does not touch the remote settlement path, and the blocker is still alive on 73a3cbe.

`settle_turn` still falls through `remoteCommandSessionID` at the hub, so settling a remote queue row is handled locally instead of forwarded to the daemon that owns the turn. That leaves the turn owed after the remote snapshot comes back.

Please add the `CmdSettleTurn` / `SettleTurnMessage.SessionID` route and its regression. Everything else can keep moving once that wire reaches the other end.

figgyster (approved):
The remote route finally reaches the right room. `settle_turn` now follows `SettleTurnMessage.SessionID` to the daemon that owns the turn, and the regression pins the hub path that left remote rows haunting the queue.

The full CI board is green. Approved.

figgyster (approved):
The queue was getting haunted by a repaint gap. This closes that little graveyard: a busy frame has to stay gone long enough to mean a settle, while the prompt-idle witness retires spent background work and leaves scheduled wakes alone.

The focused session-state sandbox test passes, and the full CI board is green. Approved.

figgyster (approved):
The state machine has finally learned the difference between a pause and a corpse. Cron only hushes the queue until Claude proves prompt-idle, stop failures and hook evidence now get their day in court, and a subagent's tool noise cannot erase the main thread's bill.

The focused state, hooks, protocol, and client witnesses passed; the full CI board is green. Approved.

figgyster (approved):
The merge from `main` brought in work already reviewed on its own beat. The meaningful queue change remains the cron-policy reversal: `PendingCron` no longer settles or suppresses a turn, cron stops are classified, and pinning is the explicit hush button. The fresh merged-head board is green. Approved.

PR #690, opened 2026-07-27
Make docked tiles first-class focus targets

victorarias:
When an agent opens a markdown file into the workspace, the resulting tile could not be focused. Workspace focus was an agent-pane-only model: `activePaneId` is derived from the focused session, `findPaneInDirection` filtered tiles out of navigation, and CSS pinned tiles at full opacity so the app's only focus affordance never applied to them.

The consequences went well past the reported symptom:

- ⌘W acted on whichever terminal was focused before the tile — or, when the click landed DOM focus outside the workspace, fell through to `session.close` and killed the session.
- ⌘⌥arrows skipped tiles entirely, in and out.
- ⌘⇧Enter could not zoom or maximize a tile.
- ⌘Enter (send markdown annotations) is registration-gated on focus-within, so it was not even armed unless focus happened to land inside the tile.
- Keyboard scrolling of a tile body only worked in a workspace with no terminal at all.
- Nothing in the UI said which leaf had the keyboard.

## What changed

**Focus is now a property of a leaf, not of a session.** The workspace resolves an active leaf — the focused tile, else the session-derived active pane, else the first tile — and every leaf-scoped action targets it: close, zoom, maximize, directional navigation, and the focus chrome. Focusing a tile does not change which session is active (a tile has no session of its own; markdown tiles only carry a Send binding), so `activePaneId` stays session-derived and agent-pane-only and the two combine in the workspace view.

Splitting stays terminal-only: the daemon splits beside an agent pane, so ⌘D is inert while a tile is focused rather than acting on some other terminal.

Each tile now gets a body ref, so a plain click puts real DOM focus inside the tile. That is also what makes the shortcut dispatcher route ⌘W to the workspace rather than letting it fall through to `session.close`. The native ⌘W path in `App.tsx` additionally falls back to the workspace's `data-active-leaf-id` when focus is in the workspace but not inside the tile subtree.

`hasLeaf` and `findLeafInDirection` are new. `hasPane` and `findPaneInDirection` remain for the callers that must resolve to a session-bearing terminal (daemon active pane, close-pane focus fallback).

**Focus display.** The focused leaf is marked rather than merely undimmed: an inset accent ring on the leaf frame plus an accented header, applied identically to terminals and docked tiles so there is one focus vocabulary across leaf kinds. Both signals appear only when more than one leaf is on screen. Tile bodies stay at full opacity — a docked document is read while typing in the terminal, so dimming its prose would be a bad focus signal.

`docs/decisions/2026-05-31-docked-tiles.md` stated the old rule ("Terminal arrow-key navigation skips tiles. A tile is not a focus target") and is revised.

## Verification

Live, in the packaged `attn-dev` app (preflight clean), driving the daemon's real `open_markdown` — the command a ⌘-click on a terminal path issues:

- clicking the tile's prose gives its body DOM focus and moves the focus chrome off the terminal; clicking the terminal takes it back;
- ⌘⌥→ / ⌘⌥← cross into and out of the tile;
- ⌘⇧Enter renders the tile alone, with the focus bar naming the file;
- native ⌘W undocks the tile, with the terminal pane and session intact.

`real-app:scenario-notebook-tile-close` (native menu → `session.close` → undock) still passes. Full frontend unit suite green. The existing ⌘W regression test no longer forces focus with `tileBody.focus()` — it clicks, which is the gesture the old model could not survive; added coverage for the active-leaf chrome moving between pane and tile, and for leaf-vs-pane directional navigation.

https://claude.ai/code/session_01XzfHadX7TYFCtSFTUbjgmU

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,28 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-07-27]
+
+### Fixed
+- **A docked markdown file can now be focused like any terminal pane.** Focus in
+  a workspace used to belong to terminals only, so a markdown or notebook tile —
+  the surface an agent opens when it shows you a file — could never be the
+  focused thing. Clicking it did nothing you could see, ⌘W acted on whichever
+  terminal was focused before, ⌘⌥arrows skipped over it, ⌘⇧Enter could not
+  zoom it, and ⌘Enter (send annotations) only worked if focus had happened to
+  land inside. Docked tiles are now first-class focus targets: click one, or
+  navigate into it with ⌘⌥arrows, and it becomes the focused leaf — ⌘W closes
+  that tile, ⌘⇧Enter zooms it, and the annotation shortcut is armed.
+
+### Changed
+- **The focused pane or tile is now marked, not just brightened.** The focused
+  leaf carries an accent ring and an accented header, applied the same way to
+  terminals and docked tiles, so which one has the keyboard is readable at a
+  glance. Tiles stay at full brightness whether focused or not — a docked
+  document is there to be read while you type in the terminal.
+
+---
+
 ## [2026-07-26]
 
 ### Fixed
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -3247,10 +3247,18 @@ sendFetchPRDetails,
     // must close THAT tile here, not the active session's terminal pane. Without
     // this, Cmd+W while reading a note kills the previously-focused terminal/session.
     // (Native browser tiles are handled natively before session.close ever fires.)
-    const focused = document.activeElement;
-    const tileEl = focused instanceof HTMLElement ? focused.closest('[data-pane-kind="tile"]') : null;
-    const tileId = tileEl?.getAttribute('data-pane-id');
-    if (tileId && activeWorkspaceId) {
+    const focused = document.activeElement instanceof HTMLElement ? document.activeElement : null;
+    // Inside a tile's own subtree the tile is unambiguous. Otherwise, while focus
+    // is anywhere in the workspace, the workspace's own focus model decides —
+    // data-active-leaf-id names a tile whenever a tile is the focused leaf, so a
+    // click that landed on the workspace frame rather than the tile body still
+    // closes the leaf the user sees as focused.
+    const tileId = focused?.closest('[data-pane-kind="tile"]')?.getAttribute('data-pane-id')
+      ?? focused?.closest('.session-terminal-workspace')?.getAttribute('data-active-leaf-id')
+      ?? '';
+    const isTile = !!tileId
+      && !!document.querySelector(`[data-pane-kind="tile"][data-pane-id="${CSS.escape(tileId)}"]`);
+    if (isTile && activeWorkspaceId) {
       handleCloseTile(activeWorkspaceId, tileId);
       return;
     }
--- a/app/src/components/SessionTerminalWorkspace/SessionTerminalWorkspace.closeFocusedLeaf.test.tsx
+++ b/app/src/components/SessionTerminalWorkspace/SessionTerminalWorkspace.closeFocusedLeaf.test.tsx
@@ -59,6 +59,14 @@ function paneAndTileWorkspace(): TerminalWorkspaceState {
   };
 }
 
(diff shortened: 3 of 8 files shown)


figgyster (changes requested):
You’ve finally given tiles a seat at the same table as terminals. That is the right turn for the workspace: a document an agent opens is part of the work, not wallpaper beside it. But one old terminal-only escape hatch still walks around the new leaf model.

`terminal.open` (⌘`) calls `focusActivePane`, which moves DOM focus to the session-derived terminal without clearing `activeTileId`. The workspace still paints and routes the tile as `activeLeafId`; press ⌘W after using that shortcut and it undocks the tile you just left instead of closing the terminal now holding the keyboard. The same mismatch leaves the focus ring lying to you.

Required before merge:

- Make every path that focuses the active terminal restore pane focus in the leaf model (and add the tile → ⌘` → ⌘W regression). The Frontend check is also red on this head, so that witness needs clearing before this can go out.

figgyster (changes requested):
The render-derived reset is a better answer to the React warning. But it only notices a *different* `activePaneId`, and the shortcut I flagged does not change that id.

Click the tile, then press ⌘` (`terminal.open`). `focusActivePane` still calls `runtime.focusPane(activePaneId)` directly; the active pane is the very same pane recorded in `activeTile.whileActivePaneId`, so `focusedTileId` remains the tile. The terminal owns DOM focus, while `activeLeafId` still names the tile, and the following ⌘W undocks the tile. The new tests cover a mouse click back to a pane, but not this shortcut path.

Required before merge:

- Route `terminal.open` through the pane side of the leaf-focus transition, or otherwise clear `activeTile`, and add the tile → ⌘` → ⌘W regression. This current head still has the user-visible wrong-leaf close.

figgyster (changes requested):
The maximized-tile resurrection test earns its keep. Same-path tiles reuse an id, so clearing a vanished maximized leaf is state hygiene, not lint appeasement.

But this push does not touch the remaining focus handoff. `terminal.open` still calls `focusActivePane`, which directly focuses the unchanged `activePaneId`; `activeTile.whileActivePaneId` still matches, the tile remains `activeLeafId`, and ⌘W from the terminal undocks that old tile. The new test protects maximize cleanup, not tile → ⌘` → ⌘W.

Required before merge:

- Route Focus utility terminal through the pane side of leaf focus, or clear `activeTile` there, and add that exact shortcut regression. The current head still closes the wrong leaf.

figgyster (commented):
There it is. You took the focus handoff all the way through instead of teaching one shortcut a special lie: `terminal.open` releases the tile, and terminal-ready callbacks now respect an already-focused tile. The regression walks tile → Focus utility terminal → ⌘W and proves the pane closes while the document stays put. That is the leaf contract finally holding its own alibi.

The frontend and E2E checks are still queued on this head. The code blocker is closed; I’ll call this merge-safe when those witnesses come back green.

figgyster (changes requested):
The shortcut handoff itself is fixed. But the terminal-ready guard gets its answer from `activeLeafIsTileRef.current`, and line 368 writes that ref during render.

React can replay or discard render work. If a render that says “tile focused” is abandoned (or the reverse), the ready callback can observe that uncommitted value and either suppress or steal terminal focus against the committed leaf state. That is the same DOM-focus-versus-leaf-authority race this PR is closing, only timing-dependent now. React Doctor is right to call this an error.

Required before merge:

- Move the callback’s active-leaf snapshot to a commit-safe path, then keep the terminal-ready regression. The focus decision must come from UI React actually committed.

figgyster (commented):
You found the right seam. Terminal readiness now asks the committed leaf-focus effect to decide, instead of smuggling a focus verdict through a ref during render. The remount regression is the witness this needed: maximize a focused tile, restore the terminal, and prove it never steals the keyboard.

The fresh Frontend and E2E checks are still running. The code blocker is closed; this is merge-safe once that board comes back green.

figgyster (approved):
That was a long walk for one focus rule, but you took every turn that mattered. Phase 4 made user follow-through an accountable contract; this gives keyboard ownership the same honesty. Tiles now enter the workspace as real leaves, the utility-terminal handoff keeps DOM focus and leaf authority together, and the remount regression catches the late-arriving terminal that used to steal the scene.

Frontend, all three E2E shards, Changes, and React Doctor are green on `d8894226`. Safe to merge. The sidebar can stop lying about where the keyboard lives.

PR #691, opened 2026-07-27
fix: selection highlight border hidden on terminal panes

victorarias:
## Why

PR #690 introduced a selection highlight ring around the active pane when
multiple leaves are visible. The ring used an `inset box-shadow`, which is
painted *beneath* child elements in the CSS stacking order. Terminal panes
have opaque child backgrounds (`GhosttyTerminal` fills the body with
`--color-bg-editor`), so the shadow was covered on three sides — only the
header's accented `border-bottom` survived. Markdown tiles worked by
accident because `--color-bg-tile` is undefined and resolves to transparent.

## Summary

- Replace the `inset box-shadow` on `.workspace-pane.active` with a
  `::after` pseudo-element that paints the accent ring on top of all
  children
- The pseudo-element uses `position: absolute; inset: 0; z-index: 1;
  pointer-events: none` — no layout shift, no click interference

## Test plan

- [x] Frontend tests pass
- [ ] Open the app with multiple docked tiles (terminal + markdown), verify
  the orange accent ring is visible on all four edges of the active pane
- [ ] Switch focus between terminal and tile panes, confirm the ring moves
  correctly
- [ ] Verify no visual flash when opening a new agent or shell pane

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-27]
 
 ### Fixed
+- **The focus ring around a terminal pane is visible on all four sides again,
+  even with a ticket open.** It was painted as an inset box-shadow, which draws
+  beneath a pane's children — so the terminal's opaque background covered the
+  ring's left, right, and bottom edges, leaving only the header's accent border
+  visible. It is now a `::after` overlay layered above everything in the pane,
+  including the bound-ticket detail overlay, so focus always has one visible
+  owner regardless of what the pane is showing.
 - **A docked markdown file can now be focused like any terminal pane.** Focus in
   a workspace used to belong to terminals only, so a markdown or notebook tile —
   the surface an agent opens when it shows you a file — could never be the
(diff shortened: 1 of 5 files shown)


figgyster (dismissed):
You’re tightening the right seam. #690 made the active leaf matter, and a terminal that hides its own focus marker is the kind of small lie that makes keyboard work feel haunted.

But one opaque child still walks over this ring.

Required before merge:

- `app/src/components/SessionTerminalWorkspace/SessionTerminalWorkspace.css:114` puts the new `::after` ring at `z-index: 1`, while the existing ticket overlay at `app/src/components/SessionTerminalWorkspace/SessionTerminalWorkspace.css:340` is opaque and sits at `z-index: 5`. Clicking its chip first activates the terminal pane, then opens that overlay, so its body covers the ring’s left, right, and bottom edges again. Put the focus signal above pane-local overlays, or give that overlay the same frame, and pin the ticket-open state with a visual/regression witness. The promise here is that focus has one visible owner; the ticket view should not get a private exception.

figgyster (commented):
#690 made visible focus part of the workspace contract, and this follow-up finally carries it through the ticket view. The ring now clears the pane-local overlay, while the browser harness keeps that stacking relationship from quietly wandering off in the rain.

My blocker is cleared on `36fe040c`. The fresh Frontend and E2E jobs are still queued, so I’m leaving this conditional until the precinct lab finishes its paperwork.

PR #692, opened 2026-07-27
docs(plans): investigation plan for ghostty WASM terminal model crashes

victorarias:
## Intent

Today's production session hit the "Terminal issue recovered. We reloaded it for you." toast three times — all WASM traps inside the vendored ghostty VT core (pin `29d4aba`): two Zig `unreachable` traps in `ghostty_terminal_write` on the first write after single-column divider-drag resize bursts, plus one out-of-bounds render fault. This PR captures the investigation and the fix plan, plus the deterministic repro the investigation produced. No behavior changes.

## Summary

- `docs/plans/2026-07-27-ghostty-wasm-model-crashes.md` — evidence, upstream landscape, and a phased fix plan:
  - The daemon's native ghostty (`ab0b9da`) survived the identical byte+resize stream that trapped the app's WASM (`29d4aba`) — the bug lives in the old core pin.
  - Converging the WASM pin to the native pin is a shim rewrite, not a bump (upstream redesigned the Terminal C API; verified by trial build) — tracked as a follow-up.
  - Fix path: bisect ghostty within the range the old wasm-api patch still applies to, then cherry-pick the fixing commit onto the pin (same playbook as the June `startHyperlink` fix), with ranked fallbacks and capture-on-fault instrumentation for the trap fault modes.
- `app/scripts/repro-ghostty-vt-resize-hang.mjs` — spike artifact: fuzzing found a **fourth fault mode**, a deterministic infinite loop in `resize()` (widen ≥ +10 cols, then two more resizes, with OSC 8 + box-drawing + overflowing content; 153-char payload). It reproduces through both app resize call sites — plain `resize()` and the mode-7 no-reflow wrapper — so `resizeGhosttyWithoutReflow` is exonerated. Exit 0 is the pass criterion for a fixed wasm build; this script is the bisect test.

## Test plan

- [x] Repro script verified: exits 1 (hang at the second consecutive narrow) deterministically, 3/3 runs from the repo root.
- [x] Pre-commit frontend suite green.
- [ ] Bisect + cherry-pick per the plan's Phase 2/3 (follow-up PR); repro must exit 0 on the fixed wasm.

## Out of scope

- The fix itself (plan Phase 2–4), the capture-on-fault instrumentation, and the WASM/native pin convergence (follow-up plan).

### DIFF
--- a/app/scripts/repro-ghostty-vt-resize-hang.mjs
+++ b/app/scripts/repro-ghostty-vt-resize-hang.mjs
@@ -1,0 +1,132 @@
+// Minimal deterministic repro for an infinite-loop hang in the vendored
+// ghostty-vt wasm's resize path.
+//
+// Spike artifact of docs/plans/2026-07-27-ghostty-wasm-model-crashes.md.
+// Exit 0 is the pass criterion for a fixed wasm build: run this script
+// against a candidate ghostty-vt.wasm during the bisect/cherry-pick to
+// confirm the fix (exit 1 means the hang still reproduces).
+//
+// Sequence: create a 59x58 terminal, write one line containing a truncated
+// OSC 8 hyperlink + box-drawing + a long run of 'w' that overflows the
+// terminal width, then widen to 69 cols, narrow to 68, narrow to 67. The
+// THIRD call (the second consecutive narrow, 68->67) never returns: 100% CPU,
+// no exception, no trap signal -- a true infinite loop inside
+// ghostty_terminal_resize (or something it calls), distinct from the
+// `unreachable`/OOB traps seen in production from ghostty_terminal_write and
+// the renderer's update(). Confirmed root requirements (see accompanying
+// report/plan doc):
+//   - a widen of at least +10 cols (59->69; +9 does not reproduce)
+//   - followed by two separate narrow-by-1 resize() calls (a single narrow
+//     after the widen does not reproduce)
+//   - specific write content: plain ASCII alone does not reproduce; the
+//     OSC 8 fragment + box-drawing + overflowing 'w' run does
+//
+// Discrimination result: reproduces with BOTH plain term.resize() (as used
+// here) AND the app's mode-7 no-reflow wrapper from
+// app/src/utils/ghosttyResize.ts (write('\x1b[?7l') -> resize() ->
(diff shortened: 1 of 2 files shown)


figgyster (dismissed):
You’re doing the right kind of ugly work here: taking a crash that recovery can hide and refusing to call the toast a cure. The repro-first path, the no-reflow boundary, and the decision not to cosplay an upstream bump all hang together. This is the next chapter after the WebGL recovery work in #472–#473: the renderer can survive a broken model, but the model still has to answer for the crime.

I need one safety boundary written into the capture plan before this becomes the blueprint. Phase 1 proposes permanently dumping a per-pane ring of raw terminal bytes plus an attach snapshot into `terminal-diagnostics.jsonl` on fault. Those bytes can contain prompts, source, tokens printed by tools, or other terminal secrets; the snapshot repeats the same private evidence. A 512 KB cap limits the blast radius, not the disclosure. Specify a privacy-safe capture format before implementation: opt-in/user-controlled storage, or structured/redacted replay data that cannot persist arbitrary terminal contents, along with retention/access rules. Then the next production crash buys us a repro without quietly turning the debug directory into a terminal transcript archive.

`git diff --check` passes. The visible docs checks are green.

figgyster (changes requested):
That deterministic hang repro is a proper witness. It kills the tempting theory that mode 7 is the culprit and gives the bisect a real trail to follow. Good work. But this follow-up still leaves two bad exits open.

Required before merge:

- The capture plan at `docs/plans/2026-07-27-ghostty-wasm-model-crashes.md:124-142` and `:192-195` still calls for raw terminal bytes and attach snapshots in persistent diagnostics, with no opt-in, redaction, retention, or access boundary. Terminal output is private evidence by default. Put the privacy contract in the plan before this becomes the implementation blueprint.
- `app/scripts/repro-ghostty-vt-resize-hang.mjs:101-103` starts the 2.5-second hang watchdog before the worker has imported `ghostty-web`, compiled/instantiated the WASM, and posted its first `starting` message. On a slow machine, a healthy fixed build can exit 1 as an alleged resize hang before it reaches any resize. Start the per-step watchdog only after the worker announces the step, or give initialization a separate outcome/budget, so the bisect signal means what it says.

The visible checks are green. The previous Figgy red is stale; this one is for `027c92e`.

PR #693, opened 2026-07-27
fix(workspace): draw the focus ring above pane content

victorarias:
## Summary
- The focus ring around an active workspace pane used to be an inset `box-shadow` on the pane frame, which paints beneath descendant backgrounds. A terminal fills its body with an opaque background, so the ring was covered everywhere except behind the translucent header — visually collapsing to a single line.
- The ring is now a dedicated overlay pseudo-element stacked above pane content (below the bound-ticket overlay), so it draws fully on all four sides for both terminal panes and docked tiles.
- Also defines the previously-dead `--color-bg-tile` token as an alias for `--color-bg-editor`, since a docked tile shares the editor surface with a terminal pane.

## Test plan
- [x] `make test-frontend` (2058 tests) passes
- [x] Live-verified in the dev app: three-leaf workspace, ring renders correctly around both terminal panes and a markdown tile

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-27]
 
 ### Fixed
+- **The focus ring around a terminal pane is whole again.** On an agent or shell
+  pane the accent ring flashed on open and then collapsed to a single line above
+  the header, because the terminal's own background painted over it. The ring
+  now draws over pane content, so a focused terminal is outlined on all four
+  sides exactly like a focused markdown tile.
 - **A docked markdown file can now be focused like any terminal pane.** Focus in
   a workspace used to belong to terminals only, so a markdown or notebook tile —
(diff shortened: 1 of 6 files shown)


figgyster (approved):
You kept following #690's focus contract into the last place it could lie: the screen. The inset shadow lost its fight with the terminal canvas, and this little pane-local overlay gives visible ownership the same authority as logical leaf focus. It also keeps the focus ring above the bound-ticket panel, which is where #691 needed it to stay.

The CSS path holds. Changes, Frontend, all three browser shards, and React Doctor showed up and told the same story. Safe to merge. Case closed.

figgyster (approved):
You kept following #690's focus contract into the last place it could lie: the screen. The inset shadow lost its fight with the terminal canvas, and this little pane-local overlay gives visible ownership the same authority as logical leaf focus. It also keeps the focus ring above the bound-ticket panel, which is where #691 needed it to stay.

The CSS path holds. Changes, Frontend, all three browser shards, and React Doctor showed up and told the same story. Safe to merge. Case closed.

PR #694, opened 2026-07-28
feat(workspace): add selectable focus styles

victorarias:
## Intent / Why

The focused workspace leaf needs to be obvious without a full border competing with terminal, markdown, or browser content. This gives people two quieter focus treatments and keeps the choice consistent across launches.

## Summary

- replace the full active-leaf ring with a 3 px edge rail by default
- add an optional depth spotlight that dims sibling leaves and uses static depth and corner marks
- add Sidebar → Display → Tile focus controls and persist the selection locally
- keep spotlight selection changes compositor-friendly: opacity is the only animated property; shadows and corner marks are static, and reduced-motion disables the transition
- update the browser regression to verify both treatments remain visible over pane-local overlays

## Test plan

- [x] Frontend unit tests: 191 files, 2,062 tests
- [x] Production frontend build
- [x] Playwright E2E: 191 passed, 1 environment-specific case skipped
- [x] Packaged dev app build and bundled dev-profile preflight
- [x] Live packaged-app verification of rail and spotlight on a two-pane workspace
- [x] Relaunch verification that the selected style persists

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,17 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-07-28]
+
+### Changed
+- **Choose how the focused workspace tile is marked.** Sidebar display settings
+  now offer an edge rail or a depth spotlight for the selected terminal, markdown
+  file, browser, or other tile, and remember the choice across launches. The
+  spotlight fades only tile opacity during focus changes; its depth and corner
+  marks are static so it does not keep consuming animation work.
+
+---
+
 ## [2026-07-27]
 
 ### Fixed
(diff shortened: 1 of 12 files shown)


figgyster (commented):
You’ve taken the leaf-focus work from “the app knows” to “the user can choose how it speaks.” That’s the right next move after #690 and #693: the rail keeps the signal spare, while spotlight gives people who live in crowded workspaces a stronger answer without turning the terminal into an animated crime scene.

The preference is owned once in App and reaches every rendered workspace. The browser witness still checks that both markers survive the ticket overlay, and the follow-up puts an explicit button type on the new controls. I ran the focused Sidebar, selection-style, and focused-leaf suites in the sandbox (38 tests), plus TypeScript. Both passed.

The remaining Frontend and three E2E shards are still running, so I’m leaving this as a conditional review rather than handing out a green stamp before the precinct lab calls back.

figgyster (approved):
The board came back green, and this one earned it. #690 made a leaf’s authority real; #693 made it visible; now you let people choose the visual weight without letting it become a GPU tax paid every time focus moves.

The preference stays centralized, both markers still clear the ticket overlay, and the fallback rail keeps old storage and fresh launches boring in the good way. Frontend, all three E2E shards, their aggregate, Changes, and React Doctor passed on this head. Approved.

PR #695, opened 2026-07-28
fix(workspace): restore inactive tile dimming

victorarias:
## Intent / Why

The focus-style change in #694 left docked tiles at full opacity under Rail and dimmed them to a different level under Spotlight. That makes the selected-versus-unselected contrast depend on both the leaf kind and the chosen marker.

Restore one predictable rule: every unselected workspace leaf is dimmed to `0.58`, while the selected leaf remains fully opaque, under both focus styles.

## Summary

- remove the docked-tile full-opacity exception from `SessionTerminalWorkspace.css`
- remove Spotlight's separate `0.76` inactive opacity so both modes inherit `0.58`
- extend the real-browser focus harness to assert inactive docked-tile and active-pane opacity under Rail and Spotlight
- update the changelog to describe the resulting behavior

## Test plan

- [x] Frontend tests and production frontend build via the pre-commit hook
- [x] Playwright E2E: 191 passed, 1 platform-specific test skipped
- [x] Packaged `attn-dev` build and launch
- [x] Bundled dev-profile preflight
- [x] Browser regression verifies inactive docked tiles compute to `0.58` and active panes to `1` in both modes

## Related

- #694

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,9 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 - **Choose how the focused workspace tile is marked.** Sidebar display settings
   now offer an edge rail or a depth spotlight for the selected terminal, markdown
   file, browser, or other tile, and remember the choice across launches. The
-  spotlight fades only tile opacity during focus changes; its depth and corner
-  marks are static so it does not keep consuming animation work.
+  selected tile stays fully opaque while every unselected tile is dimmed to the
+  same level in both modes. Spotlight depth and corner marks are static so they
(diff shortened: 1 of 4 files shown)


figgyster (approved):
#695 pays off the little visual debt #694 left on the desk: a docked document no longer changes the meaning of “inactive” just because the user picked Rail instead of Spotlight. The real-browser witness catches the actual cascade under both styles — tile at `0.58`, selected leaf at `1` — and the whole frontend/E2E board is green.

This keeps the focus work from becoming a pile of nearly-identical exceptions, which is how CSS starts charging interest. Safe to merge. Case closed.

PR #696, opened 2026-07-28
feat(delegate): create worktrees by default

victorarias:
## Intent / Why

Delegated agents should not share a Git checkout by accident. A plain `attn
delegate` should provide isolation itself instead of relying on guidance to make
every caller assemble the right worktree flags.

## Summary

- Create a new worktree automatically when a delegation targets a Git
  repository.
- Start from the repository's default branch unless `--from` selects another
  ref; keep `--worktree`, `--repo`, and `--worktree-path` as explicit overrides.
- Add `--no-worktree` for deliberate continuation in the resolved checkout.
- Keep non-Git delegations in their resolved directory.
- Update the bundled delegation guidance to describe the CLI behavior and let
  more specific repository or agent guidance choose the opt-out.
- Remove tests that duplicated help or guidance prose while preserving behavior,
  routing, installation, and pruning checks.

## Test plan

- [x] `go test ./...`
- [x] Pre-commit Go format, build, tests, and Tauri daemon checks
- [x] Built and launched `attn-dev.app`
- [x] Bundled dev preflight passed for CLI, app, daemon, tools, and routing
- [x] Live plain delegation created an isolated worktree on an automatic branch
- [x] Live `--no-worktree` delegation reused the source checkout

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-28]
 
 ### Changed
+- **Delegated Git work starts isolated by default.** `attn delegate` now creates
+  a new worktree automatically when its target is in a Git repository, including
+  for read-only and discussion work. Use `--no-worktree` to deliberately continue
+  in the resolved checkout; explicit branch, base, repository, and path options
+  still customize the new worktree.
 - **Choose how the focused workspace tile is shown.** Sidebar display settings
   now offer dimming alone, an edge rail, or a depth spotlight for the selected
   terminal, markdown file, browser, or other tile, and remember the choice across
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -668,11 +668,13 @@ func writeDelegateHelp(w io.Writer) {
 	fmt.Fprint(w, `usage: attn delegate (--brief <text> | --brief-file <path>) [options]
 
 placement:
-  (no flags)                 add a pane to the source session's workspace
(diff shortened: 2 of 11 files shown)


figgyster (dismissed):
I’m with you on stopping delegated agents from sharing a checkout by accident. That is the right default, especially after attn has spent the last month making workspace and worktree ownership explicit. But this page is telling agents a behavior the live command does not have, and that turns the safety rule into a trapdoor.

Required before merge:

- Either implement the default worktree creation, or document the actual opt-in command. `parseDelegateArgs` only sets `Worktree` from `--worktree <branch>`; a plain `attn delegate --brief-file ...` still starts in the current checkout, while lines 82–87 promise a new worktree by default. The later no-placement example repeats that bare command, so an agent following this new guidance will do exactly the sharing this PR says it prevents.

figgyster (commented):
You took the objection all the way down to the daemon, which is where this promise belonged. A plain Git delegation now creates an isolated, default-branch worktree with a generated branch; `--no-worktree` names the deliberate continuation path instead of making shared state an accident.

The original blocker is cleared on this head. The focused delegation coverage and static diff check hold. Daemon and release checks are still running, so I’m leaving this as a current-head comment rather than spending an approval before the witnesses get back.

figgyster (approved):
The precinct lab came back green. You turned the safety rule into the default execution path: Git delegates get their own default-branch worktree and generated branch, while continuation has to say its name with `--no-worktree`.

That closes the gap between the guidance and the daemon, and it carries attn’s workspace-ownership work into the agent’s first command instead of leaving it as advice on a wall. Approved.

PR #697, opened 2026-07-28
feat(workspace): add dim-only focus style

victorarias:
## Intent / Why

Some workspace layouts only need a quiet indication of which leaf has keyboard focus. Rail and spotlight both add a selected-leaf marker; this adds an opacity-only choice for users who want less visual chrome.

## Summary

- add a persisted `dim` tile-focus option alongside `rail` and `spotlight`
- keep the selected leaf fully opaque and dim every unselected terminal, shell, agent, browser, markdown, or other tile without adding markers, shadows, or header treatment
- cover persistence, settings selection, and the real browser CSS cascade, including the absence of pseudo-element markers
- update the changelog to describe all three focus styles

## Test plan

- [x] `pnpm --dir app test` — 2,063 tests passed
- [x] frontend production build via the pre-commit gate
- [x] `pnpm --dir app run e2e` — 191 passed, 1 intentionally skipped
- [x] `make dev`
- [x] bundled `attn-dev` preflight — 0 failures, 0 warnings
- [x] selected `dim` in the running packaged dev app and verified it became the active option

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,12 +9,12 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-28]
 
 ### Changed
-- **Choose how the focused workspace tile is marked.** Sidebar display settings
-  now offer an edge rail or a depth spotlight for the selected terminal, markdown
-  file, browser, or other tile, and remember the choice across launches. The
-  selected tile stays fully opaque while every unselected tile is dimmed to the
-  same level in both modes. Spotlight depth and corner marks are static so they
-  do not keep consuming animation work.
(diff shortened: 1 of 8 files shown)


figgyster (approved):
You’re finishing the quiet end of the focus-ownership arc without loosening the contract #690 and #693 earned. `dim` makes the chosen leaf readable through contrast alone, and the preference still has one home feeding Sidebar and every workspace renderer.

The browser witness is the bit I wanted to see twice. The selected pane stays at full strength, the other leaf stays at 0.58, and no pseudo-element or shadow slips back in above the ticket overlay. Less chrome, but the same accountable focus. Frontend and the full E2E board are green. Approved.

PR #698, opened 2026-07-28
feat(queue): settling a turn hands over the next agent that owes one

victorarias:
Working through the agent queue took two keystrokes per agent: ⌘⇧E to settle the
one you are looking at, then ⌘J to go find the next. Settle now does both, so the
queue is worked front to back with one shortcut.

## What it does

Settling the selected agent selects the next row in **queue order** — turn age,
oldest first — wrapping to the top, and never the row just settled. With nothing
left to go to, selection stays where it is.

Unchanged: what Settle means, the fact that it is bound only while the queue
arrangement is on (off by default; Settings > General or ⌘K), and a queue row's
own settle button, which still settles that row without moving you — that is what
settling an agent you are not looking at should do.

## Two decisions worth naming

**Queue order, not attention order.** ⌘J jumps to the first session in a state
that wants attention; the queue is a different list, ordered by how long a turn
has been owed and with membership from the daemon's `turn_owed`. A settled agent
leaves the band while it may still be, say, `waiting_input`. So the target comes
from the queue band itself (`nextTurnAfterSettle`) rather than being re-derived.
⌘J's own behavior is untouched — whether it should become queue-aware is a
separate product question, not decided here.

Wrapping is the one addition to the queue's own model: settling the bottom row
moves to the top, since the rows above are still owed and falling off the end
would leave the user on a settled agent with work still queued.

**The target is read before the settle.** `sendSettleTurn` is a round trip to the
daemon and the band only drops the row once the broadcast lands. Reading the band
after the call would either find the settled row still in it or race the refresh —
intermittently landing on the agent you just settled, or doing nothing. Deciding
the target from the pre-settle list keeps the jump off that race entirely.

## Verification

`pnpm --dir app test` (2095 passed) and `pnpm --dir app exec tsc --noEmit` green.

Unit tests on `nextTurnAfterSettle` cover the next-in-order case, the wrap, the
session the band does not hold, and nothing-left-to-jump-to (single row, and an
empty band).

Live, in a packaged non-production profile (`settlejump`, preflight PASS,
protocol 198 across CLI/app/daemon), driving the real ⌘⇧E through the packaged
app's native menu path with two live Claude agents queued —
`real-app:scenario-agent-queue`, `ATTN_VERDICT ok:true`, 0 failures:

```
settling selected the next owed turn: d7ca4e2c-…   (pressed while beta was selected; alpha was the next turn)
selection stayed on the agent that was just settled: d7ca4e2c-…   (last turn in the band)
```

Both assertions are in the scenario, so the keystroke's behavior stays covered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ## [2026-07-28]
 
 ### Changed
+- **Settling a turn hands you the next agent that owes one.** ⌘⇧E now settles the
+  selected agent and selects the next turn in queue order — oldest owed first,
+  wrapping to the top — so working through the queue no longer takes two
+  keystrokes per agent. With nothing left to go to, the selection stays where it
+  is. The shortcut is still bound only while the queue arrangement is on, and a
+  row's own settle button still settles just that row without moving you.
 - **Delegated Git work starts isolated by default.** `attn delegate` now creates
   a new worktree automatically when its target is in a Git repository, including
   for read-only and discussion work. Use `--no-worktree` to deliberately continue
--- a/app/scripts/real-app-harness/scenario-agent-queue.mjs
+++ b/app/scripts/real-app-harness/scenario-agent-queue.mjs
@@ -30,7 +30,9 @@
  *   8. the chief of staff occupies its own slot and never the band,
  *   9. turning the arrangement off and back on mid-session restores the whole
  *      workspace tree, then returns the same queue with the same agent selected,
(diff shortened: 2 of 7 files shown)


figgyster (approved):
The queue is finally acting like a ledger instead of a scavenger hunt. #689 drew the hard line between a pause and a debt paid; this carries that line one useful move further. Capturing the next row before the settle broadcast is the bit I read twice. It keeps ⌘⇧E walking turn age when the daemon is still carrying the old row for a breath.

The queue-band witness covers order, wrap, and the one-turn stop, while the packaged-app run follows the keyboard handoff across a daemon restart. Frontend and every E2E shard are green. Safe to merge. Case closed.

PR #699, opened 2026-07-28
feat(sidebar): move the agent-queue switch into the display popover

victorarias:
The agent-queue on/off switch now lives in the sidebar's display popover — the
one the slider icon opens next to the "+" in the Workspaces header, alongside
Display (open / tight / boxed), Tile focus (dim / rail / spotlight), and
Tile-only workspaces. It is gone from the Settings modal.

The queue *is* a sidebar arrangement, and every other control that changes how
the sidebar looks was already one click away in that popover. Sending people to
a modal for the biggest of those choices was the odd one out.

## Where it sits, and why

**First in the popover, above Display.** Everything else there tunes how the
workspace tree is drawn; the queue changes what the sidebar *contains*, adding
the chief's slot and the "Your turn" band above the tree. Bottom-of-the-list
would have filed the biggest choice under the smallest ones. It carries its
spacing below rather than above (`--lead`) so it reads as its own line rather
than as part of the Display group.

## Wiring

The three existing popover controls are local UI preferences. Queue mode is
daemon-owned: written with `set_setting` and broadcast back to every client. So
`App.tsx` keeps ownership — it reads the value through `isQueueModeEnabled` and
hands `Sidebar` the resolved boolean plus a single writer, `handleToggleQueueMode`.
The sidebar never sees the setting key, and the ⌘K action now calls that same
writer instead of its own copy, so the popover and the command menu cannot
disagree.

## Verification

`pnpm --dir app test` (193 files, 2087 passed) and `pnpm --dir app exec tsc --noEmit`
are green. Sidebar gained a test for the switch's `aria-checked` tracking the prop
and firing the callback; the Settings-modal test now asserts the toggle is *not*
there.

Live, on a throwaway `qtoggle` profile install (preflight PASS, protocol 198):

- Popover switch ON → the "Your turn" / "Settled" band appears
  (`queue_get_state` → `present: true`, the session in `settled`), and
  `queue_mode_enabled` is `true` in the profile's `attn.db` — so it persisted
  through the daemon, not just locally.
- ⌘K then read **"Turn off the agent queue"**, agreeing with the switch. Running
  it removed the band, restored the tree, wrote `false` to the DB, and the popover
  switch was off when reopened — no reload needed.
- Popover switch ON again → band back, DB `true`.
- Settings → General: searching "queue" no longer surfaces an Agent queue section.

Screenshots taken during that run (popover with the switch on and off, the ⌘K
entry reading "Turn off the agent queue", and Settings after the removal) went to
the author directly — GitHub image upload is not available from the CLI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   keystrokes per agent. With nothing left to go to, the selection stays where it
   is. The shortcut is still bound only while the queue arrangement is on, and a
   row's own settle button still settles just that row without moving you.
+- **The agent queue is switched on from the sidebar itself.** The toggle moved
+  out of Settings into the sidebar's display popover — the slider icon beside the
+  "+" in the Workspaces header — where it sits above Display, Tile focus, and
+  Tile-only workspaces. ⌘K still turns it on and off and always agrees with the
+  switch.
 - **Delegated Git work starts isolated by default.** `attn delegate` now creates
   a new worktree automatically when its target is in a Git repository, including
   for read-only and discussion work. Use `--no-worktree` to deliberately continue
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -1937,6 +1937,13 @@ sendFetchPRDetails,
     if (claimPaletteFocus()) return;
     setMarkdownOpenerOpen(true);
   }, []);
+
+  // The single writer for the queue setting. The sidebar's display popover and
+  // the command menu both call it, so the two entry points can never disagree
+  // about what is in effect — and neither one has to know the setting's key.
+  const handleToggleQueueMode = useCallback(() => {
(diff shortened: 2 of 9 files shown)


figgyster (approved):
You finally put #689's turn-owed ledger where the person doing the owing can actually reach it. Good. Settings was a back room for a thing that changes the shape of the workday; this popover is the right street corner.

The part worth keeping is the ownership line: `App` still holds the daemon setting and both the sidebar switch and ⌘K go through the same writer. The sidebar only reflects the settled value, so another client cannot leave it telling a different story. The focused Sidebar/Settings witnesses and the full Frontend/E2E board agree. Safe to merge. Case closed.

figgyster (approved):
This rebase gave the case a new address, then fixed the last stale sign in `queueBands`: the switch lives in the sidebar display popover now, not Settings. Small thing. Worth making true.

The queue still has one daemon-owned writer in `App`, and the sidebar remains a reflection of the broadcast value rather than another little state machine waiting to lie. Frontend and the full E2E board are green on this head. Safe to merge.

PR #700, opened 2026-07-28
fix(app): clear React Doctor findings on the queue sidebar

victorarias:
React Doctor reported **19 warnings on the lines PR #689 changed** (score 60/100,
0 errors — [run 30397323031](https://github.com/victorarias/attn/actions/runs/30397323031)).
This clears all of them but one, with no suppressions, no rule or workflow edits,
and no behavior change beyond the accessibility fix.

Verified with `react-doctor --scope changed --base origin/main` — the same
diff-scoped question CI asks — which now reports **0 findings across the 14
changed files**.

## The one real defect: queue rows were mouse-only

A row in the agent queue carried its `onClick` on a plain `<div>`. It could not
be reached by Tab, no key press activated it, and a screen reader had no reason
to announce it as something you could do anything with.

Opening an agent is now a real `<button>` that fills the row, so Tab reaches it,
Return and Space press it, and it announces as "Open &lt;agent&gt;". It fills the
row from behind rather than wrapping its contents, which matters for two reasons:
clicking anywhere on the row still opens it, and every other child stays a direct
child of `.session-item`, so the row's flex layout and the `>` selectors that
style it are untouched. The settle, pin, and actions controls are lifted above it
so they stay independently clickable and no longer sit inside an interactive
ancestor.

The first attempt — `role="button"` plus key handlers on the row div — cleared the
two accessibility findings but introduced three `html-no-nested-interactive`
warnings, since the settle, pin, and actions buttons then sat inside an
interactive ancestor. The button-behind-the-row shape is what clears both.

## The structural findings

- **`no-barrel-import`** — direct imports instead of the `../shortcuts` barrel in
  `QueueBands`, `Sidebar`, `App`, `useKeyboardShortcuts`, and
  `useUiAutomationBridge`.
- **`only-export-components`** — `App.tsx` exported seven pure helpers alongside
  the component, which breaks Fast Refresh state preservation on every edit to a
  3600-line file. The presentation-notice helpers move to
  `src/utils/presentationNotices.ts` and the fs-signal helpers to
  `src/utils/fsChangeSignals.ts`, with their existing tests, which move and keep
  their coverage. `formatTurnAge` moves out of `QueueBands.tsx` to
  `utils/queueBands.ts` for the same reason.
- **`js-combine-iterations`** — `Sidebar`'s `map().filter()` and `map().map()`
  chains walk their lists once.
- **`button-has-type`** — the update banner's two buttons declare
  `type="button"`.

## Left deliberately

**`prefer-useReducer` on `SettingsModal.tsx:182`.** The rule fires on the count
of `useState` calls in one component, but its premise — state that changes
together — does not hold here. Those 23 values are independent per-field drafts,
each committed on its own blur or change; only the "reset the drafts when the
modal opens" effect touches them together, and React batches that into one render
anyway. Collapsing them into a single reducer would be a large rewrite of a
heavily interactive modal, with real regression risk, in exchange for a worse
fit. Left as-is rather than suppressed.

For context, the six files this PR touches still carry 53 findings of these same
rules on lines PR #689 did not change (mostly `button-has-type` and pre-existing
`Sidebar` rows with the same click-handler-on-a-div shape). Those are outside
this PR's contract and are not addressed here.

## Verification

- `pnpm --dir app exec tsc --noEmit`, `pnpm --dir app test` (2092 passed),
  `pnpm --dir app run e2e` (191 passed, 1 skipped) — all green.
- **Live, packaged app** on a throwaway `rdfix` profile (`make install
  PROFILE=rdfix`, preflight PASS): `real-app:scenario-agent-queue` green,
  `ATTN_VERDICT ok:true`, all 13 steps. That covers opening a row by click
  through the new control, settling, pinning, the chief slot, the arrangement
  toggle, and a settle surviving a daemon restart.
- **The keyboard fix specifically**, in that run: focus placed on a row's open
  control, which reports `{"tag":"BUTTON","label":"Open queue-alpha-…","focused":true}`,
  then a **real Return key through the packaged app** (CGEvent, not a synthesized
  DOM event) opens that agent and lands input focus in its pane.
- **Visually**: the row renders unchanged — state dot, label, workspace, actions
  in place — and the focus ring draws around the row when its open control has
  focus.

To support that evidence, `queue_get_state` now reports each row's open control,
and the automation bridge gains a `dom_focus` verb. Focus is placed directly
rather than walked to with Tab: where a blind Tab walk ends up depends on the
whole window and says nothing about this row, and an earlier version that walked
proved flaky across profiles. The key press itself stays real.

https://claude.ai/code/session_01RZL3Tpa2g9hXQ2BNQ3iqDG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-07-28]
 
+### Fixed
+- **Queue rows can be opened from the keyboard.** Each row in the agent queue
+  now offers a real, focusable control for opening its agent, with a visible
+  focus ring and a name screen readers announce. Previously the row responded
+  only to a mouse click. Clicking anywhere on the row still opens it, and the
+  settle, pin, and actions buttons are unchanged.
+
 ### Changed
 - **Delegated Git work starts isolated by default.** `attn delegate` now creates
   a new worktree automatically when its target is in a Git repository, including
--- a/app/scripts/real-app-harness/scenario-agent-queue.mjs
+++ b/app/scripts/real-app-harness/scenario-agent-queue.mjs
@@ -270,11 +270,41 @@ async function main() {
     });
 
     await runner.step('clicking_a_row_hands_the_agent_over', async () => {
-      await client.request('dom_click', { selector: `[data-testid="queue-turn-${alpha.sessionId}"]` });
+      await client.request('dom_click', { selector: `[data-testid="queue-select-${alpha.sessionId}"]` });
       await pollFor(async () => {
         const state = await client.request('get_state');
         return state.activeSessionId === alpha.sessionId ? state : null;
       }, 'the clicked row to select its agent', 15_000);
+      await waitForPaneInputFocus(client, alpha.sessionId, alpha.paneId, 15_000);
+    });
+
+    await runner.step('a_row_opens_from_the_keyboard', async () => {
+      // A queue row must be operable without a mouse. The focus is placed
+      // directly rather than walked to with Tab — where a blind Tab walk ends up
+      // depends on the whole window, which says nothing about this row — but the
+      // keypress itself is a real Return through the packaged app, so it proves
+      // the row responds to the keyboard and not only to a click handler.
+      await client.request('select_session', { sessionId: beta.sessionId });
+      await driver.activateApp();
+
+      const focused = await client.request('dom_focus', {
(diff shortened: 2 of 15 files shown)


figgyster (approved):
You took #689's queue ledger back into the shop and fixed the one thing an attention product cannot get away with: a row that looks selectable but only listens to a mouse. The full-row button leaves settle, pin, and actions above it, and the packaged-app harness has Return carry focus back to the right terminal. That's the real witness.

React Doctor is quiet now, and Frontend plus the whole E2E board came home green. This keeps the queue honest for keyboard users without turning a cleanup pass into an architecture heist. Approved.

PR #701, opened 2026-07-28
feat(settle): auto-settle a turn after you steer the agent

victorarias:
Settling a turn is how a turn ends, and today it is always manual. But when you steer an agent and it goes back to work, you have in fact dealt with that turn — you just never said so. This closes it for you, visibly and reversibly.

With the feature on, a session that owes a turn and transitions to `working` holds for 30 seconds, then its terminal tile runs a 15-second visual countdown labelled "Settling…". If it completes, the turn settles through the same operation ⌘⇧E performs. ⌘. (or clicking the chip) cancels and the turn stays yours. Auto-settle never changes which session is selected.

**Off by default** — it mutates state unasked, so it is opt-in like the queue. Settings > General, or ⌘K.

## Design

**The timers live in the daemon, not the frontend.** Settling is daemon state and a state transition is what starts and stops the countdown; a client-side timer would race the broadcast that feeds it and would vanish when the window closed. `nudge_countdown.go` is the precedent this follows: per-session `time.AfterFunc`, a timer-identity check at fire time so races are lost safely, and an absolute deadline on the wire.

**One settle path.** The fire handler calls the same `SettleTurn` the user's ⌘⇧E does. There is no second way to stamp the state.

**The countdown is visual, not numeric.** "14, 13, 12…" is something you read; a draining bar is something you notice from across the screen, which is the only way an indicator on a tile you are not focused on does its job. Only the deadline (`auto_settle_fires_at`) goes on the wire, present exactly while the countdown runs, and the frontend animates against it with a single CSS `width` transition — no `setInterval`, no rAF, so a hidden or backgrounded tile costs nothing.

**Two windows, one invisible.** The 30s arm is deliberately silent; only the 15s countdown is broadcast. The absence of a deadline is the whole contract for "nothing is pending".

## The two product calls the brief left open

**Flicker.** Anything that is not `working` cancels, with **no debounce of its own**. `internal/sessionstate` already holds a session `working` across every flicker it knows about — repaint gaps via `ReasonHeartbeatGap`, compaction, background work, the `SettleGrace` window after a stale bracket. By the time a transition reaches `applyState`, leaving `working` is a considered verdict rather than a stutter, and a second debounce would re-decide it with less evidence. Expressing it as "not working cancels" rather than as a list of states also makes any future state safe by default.

**Unwatched sessions.** They settle the same as watched ones — the turn is owed whether or not you are looking at it — which is why the timer is daemon-side and independent of rendering. A session with no tile on screen shows the countdown as a violet bar on its sidebar row, so a turn is never closed silently. It drains right-to-left in violet against the nudge countdown's left-to-right sky blue, so the two never read as the same event on a row that shows both.

**A cancel is remembered** until the session leaves `working`. Without that, the next re-reported `working` would re-arm and the cancel would only buy 30 seconds. Steering the agent again is a fresh decision and arms a fresh countdown.

## Live verification

On an isolated `autosettle` profile (never dev, never production), protocol 199 on both sides, `preflight: PASS (0 failed, 0 warnings)`.

**Full happy path, at the real 30/15 defaults**, driving a real `claude` agent:

```
21:37:49.419  state=working  turn_owed=true   auto_settle_fires_at=-                                  <- armed, silent
21:38:19.420  state=working  turn_owed=true   auto_settle_fires_at=2026-07-28T21:38:34.419444Z        <- +30.0s, countdown opens
21:38:34.421  state=working  turn_owed=false  auto_settle_fires_at=-                                  <- +15.0s, turn settled
```

**Cancel via the real ⌘. binding** (dispatched through the shortcut registry, not the raw command), mid-countdown:

```
+6.6s   state=working  turn_owed=true   auto_settle_fires_at=2026-07-28T21:52:49.637372Z
+20.1s  state=working  turn_owed=true   auto_settle_fires_at=-      <- cancelled; turn still owed
```

The session stayed `working` for another 24s in a separate run with no re-arm, confirming the cancel sticks.

**Abort on leaving `working`**: observed live — an agent that finished at +12s, before the 30s arm elapsed, cleared the pending timer and kept its turn.

**Settings validation, live**: `arm=1` and `countdown=900` both rejected (`auto-settle countdown must be between 3 and 600 seconds`), neither value stuck.

**Visual**: the "Settling…" chip on the terminal tile with its bar draining 93% → 80% over 8s of a 60s window, and the sidebar bar on an off-screen session while a different workspace was on screen (the on-screen session's row correctly shows nothing).

### A bug this caught

The countdown bar rendered at the correct width the entire time but **fully transparent**: `--settling-accent` was declared only on `.settling-header` (the chip button), while `.settling-header-track` is its *sibling*, not a descendant, so the track background and fill colour both resolved to an undefined `var()` — which CSS drops silently. The label was visible, the countdown was not, and every unit test passed. Fixed by declaring the palette on each root, and guarded by `SettlingIndicator.css.test.mjs`, which asserts every selector that *uses* the palette is covered by a rule that *declares* it. jsdom cannot resolve custom properties across elements, so only reading the stylesheet catches this class of bug. Mutation-checked: reverting the fix fails exactly the two track assertions.

## Tests

- `internal/daemon/auto_settle_test.go` — 12 tests: arm → count → settle, abort on leaving `working` (table over every non-working state), re-reported `working` keeps the same timer, cancel keeps the turn and does not re-arm, disabled never arms, disabling cancels a running countdown, no-turn-owed never arms, a hand settle cancels the countdown, the fire-time recheck refuses a session that has since moved, real timers, and settings validation. The abort test was mutation-checked: removing the cancel from `syncAutoSettle` fails it at the abort assertion itself, not at an upstream diagnostic.
- `pnpm --dir app test` — 195 files, 2105 passed.
- `pnpm --dir app exec tsc --noEmit` — clean.
- `go test ./internal/...` — green.

## Notes

- Protocol 199 (`auto_settle_fires_at` on `Session`, new `cancel_auto_settle` command), bumped in all three lockstep spots.
- `vite.config.ts` now also picks up `src/**/*.test.mjs`, for guards that must read a source file off disk — the app tsconfig has no node types and vitest stubs `?raw` CSS imports to empty.
- A concurrent branch (`feat/settle-jumps-to-next`) changes ⌘⇧E to jump to the next unsettled agent. Auto-settle deliberately does not move the selection, so the two should not conflict, but they touch the same settle path and are worth landing in a known order.


figgyster (dismissed):
You’re closing the loop #689 opened: a turn that really was dealt with can leave the ledger without somebody remembering one more keystroke. Putting the clock in the daemon, giving it an on-screen veto, and sharing the existing settle path is the right kind of ambition. But two seams still let the feature lie about what the user asked for.

Required before merge:

- `internal/daemon/auto_settle.go:279` checks `working` and `turnOwed` before `SettleTurn`, but that sequence is not serialized with `applyState`. The timer can observe a working, owed turn; a `waiting_input`/approval transition can commit immediately afterward; then the timer still settles the newly-opened turn. `syncAutoSettle` runs only after the state write, so its cancellation arrives too late. Serialize the fire-time check-and-settle with the state-transition gate, or make the store settle conditional on the session still being working, and pin the interleaving in a regression test.

- `app/src/components/SettingsModal.tsx:406` never copies `actualAutoSettleArm` or `actualAutoSettleCountdown` into the two draft states from lines 220–225. This modal is mounted before the daemon settings broadcast, so an existing 60/20 policy still displays the 30/15 defaults; blur then writes those defaults back over the saved policy. Sync those drafts with the incoming effective settings (and add the values to the effect dependencies), with a component regression for post-mount settings hydration.

figgyster (commented):
The two seams I stopped you for are shut now. `autoSettleFireMu` orders the fire-time decision against `commitSessionState`, and the new hook-held regression test catches the exact old handoff where a fresh `pending_approval` turn could disappear. Settings now hydrates both timer drafts from daemon state, with a witness that a late settings broadcast cannot turn a saved policy back into 30/15.

`f52a606` is the gofmt-only follow-up. The fresh CI board is still queued, so I’m keeping this review conditional until the precinct lab comes back, but there’s no remaining code finding from the earlier request.

figgyster (approved):
You took the right lesson from the queue work: a timer may retire a dealt-with turn, but it never gets to erase a fresh debt. The fire-time lock and its hook-held regression witness pin the one interleaving that mattered; the Settings fix also lets the daemon remain the owner of an already-saved policy.

The full board is green now. This carries #689 and #698 forward without blurring their ledger. Approved.

figgyster (commented):
The fresh head replays the same gofmt repair on a newer base. I checked the actual delta: no new settlement behavior slipped in through the back door, and the fire-time/state-commit contract I approved still holds.

Changes is green. Daemon, Frontend, release preflight, and the E2E board are still running, so this stays conditional until the precinct lab finishes its shift.

figgyster (approved):
You gave the queue enough judgment to automate the ceremony without ever making it a liar. The timer still sees each arm’s policy, the countdown survives being off-screen, and cancel keeps the debt where the user left it. That is exactly the sort of boring reliability this ledger needs.

The rebased head’s board is green, and the isolated packaged-app run caught the two places a mocked timer would have smiled and waved. Approved.

figgyster (commented):
This is the same four-space daemon-struct repair replayed onto yet another base. I checked the commit itself; the settlement and Settings contracts are unchanged.

Changes is green, while the fresh daemon, frontend, release, and E2E runs are still gathering their coats. Keeping this conditional until they report.

figgyster (approved):
You gave the queue enough judgment to automate the ceremony without ever making it a liar. The timer still sees each arm’s policy, the countdown survives being off-screen, and cancel keeps the debt where the user left it. That is exactly the sort of boring reliability this ledger needs.

The rebased head’s board is green, and the isolated packaged-app run caught the two places a mocked timer would have smiled and waved. Approved.

PR #291, opened 2026-06-09
feat(chief-of-staff): track delegated agents

victorarias:
## Why

A chief-of-staff session needs durable visibility into the agents it dispatches without changing ordinary attn delegation into a parent-child workflow. This adds the first reactive coordination loop: attn records only delegations initiated by the current chief, projects the target's live session state, and lets the worker report concise progress.

## What changed

- persist chief-originated dispatch relationships and reports in SQLite while keeping session state authoritative
- add `attn dispatch list` for chiefs and `attn dispatch report` with message/file input for delegated agents
- inject reporting instructions only into tracked delegated prompts; ordinary delegation remains unchanged
- broadcast dispatch snapshots and show current and previous chiefs' work in a dashboard card with session navigation
- teach agents the chief and worker workflows in the bundled attn skill
- bump the daemon protocol to 93 and update generated clients

## Test plan

- [x] `go test ./...`
- [x] `pnpm --dir app test` (83 files, 665 tests)
- [x] `pnpm --dir app exec tsc --noEmit`
- [x] `pnpm --dir app run build`
- [x] `git diff --check` and `gofmt -d` on changed Go files
- [x] packaged `attn-dev`: assigned a real Codex session as chief, delegated to a real Claude session in a new workspace without a worktree, and verified Claude independently reported `LIVE_DISPATCH_OK - Latest dated heading: ## [2026-06-09]`
- [x] verified `attn dispatch list` returned the durable record, latest report, and projected `idle` target state
- [x] verified the native dashboard rendered the chief, target Claude session, live state, report, and click-through row

## Out of scope

- proactive interruption or pinging of running agents
- chief verification policies or automatic follow-up
- rewriting dispatch ownership when the chief role transfers

PR #702, opened 2026-07-29
Every delegation binds and routes a ticket

victorarias:
## What

Ticket-bound delegation becomes the norm for **all** delegating sessions, not just the chief of staff.

Before this, only the chief's delegations created a ticket. Every other `attn delegate` was untracked: no ticket, no event log, no participation — so there was no way to reach the delegated agent with a note, and no record of what it did.

Now `attn delegate` from any session creates a ticket with the brief as its description and the delegated session as its assignee, and the delegated agent's prompt always carries the ticket self-report contract.

## Routing

Three identities land on every delegation ticket, each reached exactly once:

- **the assignee** — by the binding (assignee == session)
- **the creator** — by an explicit subscription
- **the chief of staff** — always, so its organizational overview covers every delegation in the system whoever started it

The chief reaches the ticket through its durable **role** identity, so its view survives the session filling the role changing. Which form that takes is the one place chief-initiated delegation still differs:

- **chief-initiated**: the chief role **owns** the ticket (unchanged).
- **ordinary**: the chief role **subscribes**. Same routing guarantee, no claim of ownership over work another session started.

That split is deliberate rather than incidental. Role ownership is also what marks a session `delegated_from_chief` in the sidebar (`TicketAssigneesOwnedByRole`), so owning every delegation ticket would have put the chief's badge on work it never started. A role-identity subscription is honored by all three participation queries (`UnreadTicketEventsFor`, `TicketParticipants`, `unreadTargetTicketEventsTx`), so it carries the same durability.

**No dedup guard was needed** when the creator *is* the chief: it then holds both its session subscription and the role identity, and `ticketnotify.ConsumeAll` already merges the two queues by event seq. That is pinned by a test that first asserts both identity queues really do hold the event, so the merge is load-bearing in the test rather than incidentally satisfied.

## Untangling the gate

`trackedByChief` was one boolean riding three behaviors. Each is now decided on its own merit:

| behavior | before | after |
| --- | --- | --- |
| ticket binding + report prompt | chief only | **every delegation** |
| chief role owns the ticket | chief only | chief only (unchanged) |
| unmute a hidden target workspace | chief only | chief only (unchanged) |

## Failure atomicity

The ticket is not incidental to a delegation — the agent's prompt tells it to report through `attn ticket status`, and the ticket is the only channel back to it. So a ticket failure still fails the whole delegation atomically (unregister the session, roll back workspace/worktree) rather than leaving a running session nobody can reach.

Subscriptions are written in the **same transaction** as the ticket (`store.CreateTicketWithSubscribers`), so a ticket never lands with a partial participant set.

## Slug allocation headroom

Ticket ids come from `ticketSlug(label)`, and the default label is the directory basename — so one popular base now accumulates ids fast. The allocator still walks `base`, `base-2`, `base-3`, … for readability, but falls back to a random suffix past that range instead of erroring. A full board can no longer fail a delegation whose session has already spawned.

## Docs

The shipped skill references that stated the old rule as fact are updated: `references/delegation.md`, `references/delegated-agent.md`, `references/tickets.md`, and the `SKILL.md` line about a tracked task still being a leaf task — which carries more weight now that tracking is universal, so it explicitly says a bound ticket says nothing about your role.

## Verification

- Go: `internal/daemon`, `internal/store`, `internal/ticketnotify`, plus the full `go test ./...`. Tests that asserted the old "ordinary delegation creates no ticket" rule were rewritten to the new contract, and there is now a test pinning that a plain delegation binds a ticket and routes it.
- Frontend: `pnpm --dir app test` (2096 passing).
- Live, on an isolated `plainticket` profile with `make install PROFILE=plainticket`: a new packaged-app scenario `real-app:scenario-ordinary-delegation-ticket` boots a chief and a separate ordinary delegator, delegates a real codex worker from the delegator, and asserts the panel renders the bound ticket, the worker's report reaches the delegator and the chief exactly once each, a note from the delegator reaches the worker, and the worker carries no delegated-from-chief badge. The existing chief `real-app:scenario-ticket-lifecycle` still passes against the same build.

https://claude.ai/code/session_01VAc4f5Fo3dWwS32JWZY15f


figgyster (approved):
This is the missing half of the delegation story. #291 gave the chief a ledger, then ordinary handoffs kept slipping out the side door with no way back to the worker. Now every delegation leaves a ticket behind.

The restraint is the part that holds: creator, assignee, and durable chief role all get the route, while chief ownership, the sidebar badge, and hidden-workspace unmuting still belong only to work the chief actually started. The routing tests and the real-app three-session witness put an eyewitness on the exact contract that could have gone fuzzy here.

Daemon, Frontend, every E2E shard, Linux preflight, React Doctor, and Bench A/B are green on `9aa9f0c`. Safe to merge. Case closed.

figgyster (commented):
You took the ticket ledger back to the foundation instead of stacking one more special case on top of it. The single participant relation is the right piece of furniture here: the chief’s durable cursor, an ordinary creator subscription, and a worker’s assignment now answer the same question everywhere.

The compensation stack is the other scar worth keeping. Ticket creation happens after a live session, pane, workspace, and sometimes a worktree; the new deepest-failure witness proves that this trail gets erased in reverse order instead of leaving a sad little workspace behind.

Daemon and Linux preflight are green on `44d86fc`. The Frontend E2E shards and Bench A/B are still running, so I’m holding the approval until the fresh integration board finishes.

PR #703, opened 2026-07-29
feat(new-session): make creating a worktree the default chooser action

victorarias:
## Summary

Creating a worktree is now the default, single-Enter action in the new-session repo chooser instead of a buried option.

- The new-worktree form is now always expanded at the top of the chooser (above existing destinations), pre-filled with a generated adjective-noun branch name (new `app/src/components/NewSessionDialog/worktreeNames.ts`) and defaulting to "start from origin/<default branch>". A fresh worktree off latest upstream now costs a single Enter.
- The name arrives text-selected so typing replaces it; a ⟳ button (or ⌃R) draws another name. Generated names are checked against existing branches so they never collide.
- `RepoOptions.tsx` replaced its single focus-index model with two focus zones (`create` / `destinations`). In the create zone only Enter/Tab/↓/⌃R/Esc are intercepted so letters keep reaching the text input; the destination shortcuts (D, R, 1-9) are inert there by design. ↓ enters the destination list, ↑ off the top returns to the form.
- The old "Create worktree..." action row is gone; destinations moved under an "Open existing" header.
- Deliberate scoping: focus starts on the create form ONLY when the chosen path is the repo root. If the typed path resolves to a specific worktree, focus stays on that row so Enter still OPENS it rather than creating a new one. Both cases are covered by tests.
- CSS: `.repo-options-destinations` gained a min-height floor and `.repo-options-content` a scroll valve — without them the always-open form collapsed the destination list to zero height in a small window (caught during live verification, not by tests).
- `app/scripts/real-app-harness/bridge-remote-hub.mjs` updated: it used to locate and click the now-deleted `new-worktree` option row.

## Test plan

- [x] `make test-frontend` — full frontend unit suite, 2121 tests passing (includes new `worktreeNames.test.ts` and reworked `RepoOptions.test.tsx` / `LocationPicker.test.tsx`), re-verified after rebasing onto latest `main`.
- [x] `pnpm --dir app run e2e -- e2e/location-picker.spec.ts` — 14/14 against a real daemon, including a new single-Enter creation test.
- [x] `npx tsc --noEmit` clean.
- [x] Live verification in the packaged dev app (`make dev`, preflight PASS on the dev profile): opened the chooser on the real attn repo (60 destinations, create form first, nothing armed), and drove a real end-to-end create on a throwaway repo producing `demo-repo--zesty-toucan` on branch `zesty-toucan` with a session opened in it. Artifacts cleaned up.

No Go changes in this branch.

https://claude.ai/code/session_01TqWwbHppd1pFyeD6vVKDFE

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,23 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   countdown starts (5–3600s) and the countdown's own length (3–600s)
   independently.
 
+### Changed
+- **Creating a worktree is the first thing the repo chooser offers.** After
+  picking a repository, the new-worktree form now sits at the top of the
+  chooser, already expanded, with a generated name such as `frosty-newt`
+  pre-filled and "start from origin/&lt;default branch&gt;" preselected — so a fresh
+  worktree off the latest upstream costs a single Enter. The name is selected on
+  arrival, so typing replaces it, and the ⟳ button (or ⌃R) draws another one.
+  Existing worktrees moved below the form under "Open existing"; ↓ steps into
+  that list, where Enter, D, R, and the number shortcuts behave as before.
+  Choosing a path that resolves to a specific worktree still lands on that
+  worktree, so Enter opens it rather than creating something new.
+
+---
+
+## [2026-07-28]
+>>>>>>> a11d4db3 (feat(new-session): make creating a worktree the default chooser action)
+
 ### Fixed
 - **Queue rows can be opened from the keyboard.** Each row in the agent queue
   now offers a real, focusable control for opening its agent, with a visible
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,6 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 ---
 
 ## [2026-07-28]
->>>>>>> a11d4db3 (feat(new-session): make creating a worktree the default chooser action)
 
 ### Fixed
 - **Queue rows can be opened from the keyboard.** Each row in the agent queue
--- a/app/e2e/location-picker.spec.ts
+++ b/app/e2e/location-picker.spec.ts
@@ -305,6 +305,8 @@ test.describe('LocationPicker', () => {
         await page.keyboard.press('Enter');
 
         await expect(page.locator('[data-testid="repo-options"]')).toBeVisible({ timeout: 5000 });
+        // The repo root lands on the create form; step into the destination list.
+        await page.keyboard.press('ArrowDown');
         await expect(page.locator('[data-testid="repo-option-0"]')).toHaveClass(/selected/);
 
         await page.locator('[data-testid="repo-option-1"]').hover();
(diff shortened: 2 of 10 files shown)


figgyster (dismissed):
You’re making the safe worktree path the path of least resistance, which is exactly where this chooser needed to go after #696. The two-zone keyboard model keeps the old “open this exact worktree” intent intact while giving a fresh branch the front door. Good bet. But this head still leaves two very ordinary traps in that front door.

Required before merge:

- `CHANGELOG.md:46` still contains `>>>>>>> a11d4db3 ...`. `git diff --check` fails on the head, and the release notes would ship a merge-conflict marker. Remove the marker and retain the intended section boundary.
- `app/src/components/NewSessionDialog/RepoOptions.tsx:127` only treats the current branch and branches with an attached worktree as taken. A normal local branch with no worktree is absent from `RepoInfo`, yet `git worktree add -b <generated-name>` rejects that branch name. That makes the single-Enter flow occasionally fail and contradicts the new “never collide” claim. Feed the generator the complete branch set (or validate/reroll against it daemon-side) and add the regression for an existing, unattached branch.

The CI witnesses are mostly standing up, but neither one sees that second branch shape and neither one can make a conflict marker safe. Fix those two, then this is the kind of friction removal that pays rent every time a new agent heads out into the rain.

figgyster (dismissed):
You closed the two holes from the last pass. The conflict marker is gone, and the generated-name collision now has a real retry witness. Good recovery. But the retry has one hand on the wheel when the user is driving.

Required before merge:

- `app/src/components/NewSessionDialog/RepoOptions.tsx:238-275` retries with a random generated branch after *any* “branch already exists” failure. The text field is editable, so a person can type a deliberate name, press Enter, and get an unrelated worktree instead of the expected conflict. Track whether the submitted value is still generator-owned (initial/reroll) and only auto-reroll that case; a user-entered name must surface the collision unchanged. Add the regression for a typed existing branch.

The one-Enter generated path should stay frictionless. A name someone typed is evidence, though. Don’t let the chooser throw it in the river and quietly build somewhere else.

figgyster (commented):
You put the steering wheel back where it belongs. Generated names can recover from the branch list’s blind spot; a name someone typed now survives the collision and gets its real error back. The handback test after an explicit reroll closes the little ownership loop properly. That’s the contract I wanted.

The functional blocker is closed on `18e9f20`. I’m leaving this conditional while the fresh Frontend/E2E board is still running; once it comes back green, this is safe to merge. No new code changes needed from me.

figgyster (commented):
The follow-up on `13986cc` adds `generatedName` to the create callback’s dependencies, so the generator-ownership decision stays current after rerolls. That was the last loose thread in the typed-name repair.

The functional blockers remain closed. The fresh Frontend and E2E jobs are still running, so this stays conditional until that board returns green. No further code change from me.

figgyster (approved):
You got the arithmetic right in the end. #696 made worktrees the safer default in the daemon; this makes that safety cheap enough that people will actually take it. The chooser now knows the difference between a generated suggestion and a name someone meant, and the reroll handback plus the dependency fix keep that distinction from going stale.

The full Frontend/E2E board is green on `13986cc`, React Doctor is quiet, and the daemon/Tauri skips fit this frontend-only patch. Safe to merge. The worktree path finally has the front door it deserved.

PR #704, opened 2026-07-30
feat(annotations): annotate any recent agent message, persisted by the daemon

victorarias:
Annotate what an agent said, from the terminal grid, and have the annotation still be there after the next turn and after the app restarts.

Alt-drag over an agent's text opens a small panel: pick a label, optionally write a comment, and Send types the collected notes into the session's prompt. Annotations render as a wash over the quoted text.

Two halves, described separately below: the feature, and the refactor of the frontend's daemon surface that it needed.

## The design

**The transcript span is the anchor; the grid is a projection.** An annotation stores a character range on the agent message's markdown, plus a `messageKey` identifying the message (transcript entry UUID, falling back to sha256[:16] of the content). Rows are re-derived every frame and never stored.

Anchoring to grid rows instead would not survive contact with a terminal: a new turn scrolls them, a resize rewraps them, a reset destroys them. An annotation would silently disappear, or wash text it never referred to. With a transcript anchor, scroll, resize, and reset change where an annotation is drawn, never whether it survives.

**Containment gate.** Before painting, the cells about to be covered are re-read and checked to still quote the anchored text. A projection that cannot be confirmed is not drawn for that frame. Being briefly absent is recoverable; washing the wrong text is a lie the user cannot detect.

**The daemon owns the annotations.** They persist to `session_annotation_drafts` on every mutation — no debounce, since the mutations are discrete clicks and committed comments rather than keystrokes, which leaves no window where an unmount loses a pending write. The pane hydrates from the daemon before anything is drawn. Frontend-only annotations would not outlive quitting the app.

**Ordering is the daemon's, by generation.** A save must strictly exceed both the stored generation and the tombstone; clearing raises the tombstone. A rejected save comes back `stale: true` and the client re-hydrates rather than insisting on a list the store already refused. Two panes on one session converge instead of clobbering each other.

The full rule set and the budget receipts are in `docs/decisions/2026-08-02-terminal-annotations-anchor-to-the-transcript.md`.

## Verification

A packaged-app harness scenario, `real-app:scenario-terminal-annotations`, drives a live claude agent: annotate a turn, read it back off the daemon over the wire, run a second turn, quit and relaunch the app, re-find the wash on the grid, then Send and check the tombstone. Every assertion compares against text captured from the grid at runtime, so it does not depend on what the agent chooses to say.

The scenario was checked for falsifiability rather than trusted because it was green. Mutating the hydrate-on-mount path to drop what the daemon returns turns it red at exactly the load-bearing step, and only there:

```
ok    daemon_holds_it            (8/8 assertions passed)
ok    survives_a_second_turn
error survives_app_relaunch      Timed out waiting for the relaunched app
                                 to hydrate the annotation from the daemon
```

The same mutation fails exactly two unit tests. Reverted, the scenario is green across repeated live runs.

Also covered: 23 component tests driving a stateful fake daemon (so the client's generation bookkeeping stays load-bearing rather than asserted against a spy), 8 store tests for the table's own wiring, and 8 unit tests for the harness's grid-parsing helpers — a bad row choice there produces a drag that resolves to nothing, which would read as a product regression rather than a parsing bug.

## The daemon command surface refactor

Adding one daemon command to the frontend meant touching four places that had to agree, and hand-copying a 20-line request-plumbing block. Three changes, each independently verified:

**One request helper.** 45 senders in `useDaemonSocket.ts` each spelled out the same offline guard, pending-key registration, send, and timeout. `sendKeyedRequest` does it once; `sendRequest` sits on top for the ordinary fresh-request-id case. −668 lines, socket guards 96 → 51. The key stays the caller's on purpose: a request-id key lets calls queue independently, while a fixed key like `refresh_prs` is deliberately last-writer-wins, and deriving it inside the helper would pick one for everybody. ~33 senders are left alone rather than bent into the helper — the ones that queue or resolve instead of rejecting when offline, that reject on a second guard before registering, or that wrap `resolve` to reshape the result. Each carries the reason at the call site.

**The daemon API arrives by context.** 104 of `AppContent`'s 124 props were daemon functions `App` never called; it forwarded them and nothing else. `App` now publishes the hook's return value through `DaemonApiProvider` and components read `useDaemonApi()`. `AppContentProps` drops to 20 entries, all of them App's own state. The four-place lockstep is deleted from `app/CLAUDE.md`, not documented. `settings` deliberately stays a prop: App holds it in `useState` and the hook returns a same-named ref-backed value, so they are not interchangeable.

**Routing guards.** `CommandMetadata.Scope` existed but was read by nothing except its own test — dead metadata beside four hand-maintained remote routers. `internal/daemon/command_routing_test.go` now parses the `Cmd*` constants out of `internal/protocol/constants.go` and fails when a new command is unclassified, or when a session-scoped one is answered locally instead of routed to the session's owner. The pre-existing unclassified list may only shrink, and every locally-answered session command names why. This catches a hub answering `session_messages_get` and the three `session_annotations_*` commands for sessions it does not own.

Mechanical refactors do not get to be trusted because the suite is green, so all three were mutated: dropping `request_id` from the helper turns 18 tests red; miskeying the pending entry, 80; handing an empty API through the context, 18 across 3 files. Each reverted and re-run green.

## Migration numbering

The store migration is numbered 86 rather than reusing 85. The runner skips any `version <= max(applied)`, so two migrations sharing a number means whichever lands second never runs on a database that saw the first, leaving a table that silently never exists. The comment at the call site states the rule.


figgyster (dismissed):
You went after the hard part first: proving the terminal can be treated as a projection instead of mistaking its rows for durable facts. That’s the right scar to chase after all the workspace and restore work. The Claude measurements, explicit refusal path, and the plan’s zero-wrong-word bar give this experiment teeth; the Codex gap is named instead of papered over.

Nothing here leaks into the app, and the focused spike package is intentionally fixture-free in a fresh checkout while its real-session corpus stays local. The static diff is sound, and the live board is green. Safe to merge. Then make the next spike earn its keep with the Q1/Q3 numbers before this grows production roots.

figgyster (dismissed):
You carried the transcript-span idea over the real boundary this time: the daemon owns the marks, the grid is only a projection, and the containment gate keeps an ambitious overlay from laundering someone else’s words. That is exactly where this feature needed to grow after the spike. But a remote pane is still left outside the room.

Request changes: `session_messages_get`, `session_annotations_get`, `session_annotations_save`, and `session_annotations_clear` are all session-owned commands, but `remoteCommandSessionID` in `internal/daemon/websocket.go` does not map any of them. A browser connected to the hub for a remote agent pane therefore handles the requests locally: message reads fail because the hub has no transcript/session row, while annotation reads, saves, and clears use the hub store rather than the daemon that owns the terminal. Add all four commands to the session router and cover the remote-owner path; the current unit table only exercises the older session commands. Until then the feature’s persistence promise ends at the first remote session.

Daemon, Frontend, the E2E shards, release preflight, Changes, Changelog, and React Doctor are green on this head. The local path has witnesses. It needs the remote one too.

figgyster (dismissed):
The harness cleanup is the right bit of skepticism: after a second long turn, an annotation leaving the grid can be the containment gate doing its job, not the daemon forgetting the mark. That keeps the local witness honest. The remote session is still outside the evidence.

Request changes: this head changes only `scenario-terminal-annotations.mjs`; `remoteCommandSessionID` still has no cases for `session_messages_get`, `session_annotations_get`, `session_annotations_save`, or `session_annotations_clear`, and `TestRemoteCommandSessionID` still has no coverage for them. A remote pane therefore sends all four to the hub, where transcript lookup fails or drafts land in the wrong store. Route those commands by `SessionID` and add the remote-owner regression. Until then, the annotations survive a local restart but not the trip to another daemon.

figgyster (commented):
That closes the real hole. The four commands now follow `SessionID` to the daemon that owns both transcript and draft row, and the regression table names the exact hub-versus-owner failure that brought us here. The annotation ledger can finally keep the same promise for a remote pane that it makes for a local one.

`git diff --check` passes on `9aacbe5`. The rebuilt Daemon, Frontend, E2E, release, and React Doctor board is still running, so I’m leaving this as a current-head conditional comment rather than signing a blank check. Once that board returns green, the remote-routing blocker is gone.

figgyster (commented):
This is the right little follow-up. The store now has a real component lifetime, and the dependency lists tell the same story instead of making the lint witness squint at stable state through a keyhole. The remote-owner repair is untouched; remote panes still send transcript and draft work to the daemon that owns the session.

`git diff --check` is quiet on `6c94911`. Daemon, Frontend, release, Changelog, and React Doctor are green. E2E shard 2 is red on a five-second daemon-start timeout before the unrelated LocationPicker case; it does not point at annotations, but it is still a dead witness. Rerun that shard, then this is ready for a signature.

figgyster (approved):
You took the annotation idea through the part where features usually lose their nerve: the grid can scroll away, the pane can reconnect somewhere else, and the record still belongs to the daemon that owns the session. The rebase also carries #732's command-scope guard into this case, so the four remote commands that once slipped past the hub cannot quietly become a fifth later. That is the project learning from its own crime scenes. I like it.

The harness follow-up is honest, too. A long alternate-screen answer can lose its visible marker while leaving valid prose on the grid; the selector now recognizes that shape and its unit witness pins the boundary before the prompt/footer can wander into the evidence.

The full Daemon, Frontend, all three Frontend E2E shards, release preflight, Changes, Changelog, Bench A/B, and React Doctor board is green on `a69e0c2`. Victor's isolated packaged-app annotations scenario also passed on this head. Approved.

PR #706, opened 2026-07-31
fix(delegate): start worktrees from default branch

victorarias:
## Intent / Why

Delegated worktrees could silently inherit unrelated commits from whichever branch happened to be checked out in their source directory. That made branch ancestry depend on placement flags and could carry another task into a pull request without either agent choosing it.

This completes the isolation rule introduced in #696: an omitted `--from` now means the repository default branch for every worktree-creating placement.

## Summary

- Resolve named and automatically generated delegation worktrees from `origin/<default>`, falling back to the local default branch, regardless of placement.
- Preserve `--from <ref>` as the explicit way to continue or stack on in-progress work.
- Cover the current-workspace, new-workspace with explicit repository, and `--cwd` paths against a checkout parked on an ahead topic branch.
- Update the bundled delegation reference and changelog to state the unified behavior.

## Verification

- Repository pre-commit gate: Go build and complete Go test suite passed.
- Focused delegation tests passed, including every affected placement, explicit `--from`, existing-workspace defaulting, and remote-default preference.
- Rebuilt and installed `attn-dev.app`; bundled preflight passed with the dev profile and protocol 200.
- Live scenario: created a repository with `topic/ambient` one commit ahead of `main`, delegated with `--cwd` and a named worktree, and confirmed the delegated branch matched `main` exactly (`main..HEAD` contained zero commits while `main..topic/ambient` contained one).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-07-31]
 
+### Changed
+- **Delegated worktrees always start from the repository default branch.** Every
+  `attn delegate` placement now uses `origin/<default>`, or the local default
+  branch when no matching remote exists, instead of inheriting the branch checked
+  out in its source directory. Use `--from <ref>` to deliberately continue or
+  stack on in-progress work.
+
 ### Added
 - **Delegate an existing ticket without duplicating it.** `attn delegate
   --ticket <id>` now starts an agent from that ticket's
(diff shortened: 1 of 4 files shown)


figgyster (approved):
You're closing a hole in delegation that was wearing a respectable hat: #702 made each handoff a durable piece of work, and this makes the code it starts from intentional too. I read the default-ref seam twice. `--from` still takes the wheel; every named worktree placement now reaches the existing fetch-then-create path with an explicit default ref instead of borrowing whatever branch happened to be lying around.

The new witness parks the checkout on an ahead topic branch across current-workspace, new-workspace, and `--cwd`, then proves the delegated branch still starts from the default. That is the old crime scene, finally taped off. Daemon CI and release preflight are green. Safe to merge.

PR #707, opened 2026-08-01
Dashboard becomes home; fix workspace undock leaving a dead sidebar row

victorarias:
## Summary

Two related pieces.

**1. The Dashboard becomes attn's turn-aware home.**

- It now renders beside the sidebar, the way an agent's terminal does,
  instead of taking over the whole window. Grid view still fills the
  window.
- The sidebar gets a fixed Home row at the very top (above the chief of
  staff and "Your turn"), bound to ⌘⇧H, highlighted while you're on it.
  The stale "WORKSPACES" heading and its dead ⌘G hint are gone.
- With the agent queue on, home leads with the turns you owe (oldest
  first, with each turn's age) and files everything else under
  *Settled*, still grouped by what it's doing. Previously it grouped
  purely by state, so agents you'd already settled kept being announced
  as waiting on you.
- Settling the last turn you owed used to leave you sitting on that same
  agent with nowhere further to go. It now lands you on home, which
  leads with an "All settled" banner plus a line on what's still
  running. Nothing jumps in on its own — the next turn waits in the
  queue until you go get it.

Plan doc: `docs/plans/2026-07-29-attn-home-dashboard.md`.

**2. Daemon fix: undocking the last tile of a workspace now tears it down.**

Closing the last notebook, markdown, or browser tile in a workspace with
no agents left the workspace stuck in the sidebar — every layout command
rejects a leafless layout, so the workspace's own × could never actually
close it. Undocking the last tile now tears the workspace down instead
of saving an unusable layout. Workspaces stranded by the old behavior
are cleaned up the next time the app starts.

New test: `TestUndockingLastTileTearsDownSessionlessWorkspace` in
`internal/daemon/workspace_tile_lifecycle_test.go`.

No protocol change.

## Test plan

- [x] `go build ./...`, `go vet ./...` — clean
- [x] `go test ./internal/daemon/...` — green (one unrelated pre-existing
      flake in `TestCreateDelegatedTicketFallsBackPastSequentialRange`,
      confirmed passing on rerun and untouched by this diff)
- [x] `pnpm test` — 2136/2136 passing
- [x] `npx tsc --noEmit` — clean
- [x] Both pieces live-verified on the dev profile prior to this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,18 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-08-01]
+
+### Fixed
+- **A workspace that holds only a tile can be closed again.** Closing the last
+  notebook, markdown, or browser tile in a workspace with no agents left the
+  workspace in the sidebar — and every later click on its × did nothing, so the
+  row could never be dismissed. Closing that last tile now closes the workspace
+  with it. Workspaces stranded by the old behaviour are cleaned up the next time
+  the app starts.
+
+---
+
 ## [2026-07-31]
 
 ### Changed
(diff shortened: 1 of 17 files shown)


figgyster (approved):
You took the queue ledger from #701 and gave it a front door. Home now tells one story: owed turns arrive first, settled agents quit posing as interruptions, and the last tile no longer leaves a dead workspace haunting the sidebar. That's a real improvement in how this thing feels to run.

The shell move keeps Grid a full-window takeover, while the daemon retires the leafless workspace and broadcasts the removal instead of marooning a useless row. The focused witnesses cover the turn ordering, settled split, Home affordance, and last-tile teardown. Daemon, Frontend, every E2E shard, Linux preflight, React Doctor, and the A/B bench are green.

The collapsed Home icon still lacks an explicit button type, per React Doctor. Tiny future lint sweep, nowhere near a merge gate. Approved.

PR #709, opened 2026-08-01
docs: extension platform vision and build roadmap

victorarias:
Adds the agreed direction and build roadmap for making attn self-extensible: agents running inside attn author real React/TypeScript extensions (no bespoke YAML/JSON data model) and register them live via `attn ext apply` — no daemon restart.

**docs/vision/extension-platform.md** — the north star: the authoring-time model and its rationale (including lessons from Present's strict-YAML manifest); six primitives (durable event bus with per-consumer cursors that becomes the internal spine, interceptable hook points with per-hook fail-open/fail-closed policy, River-style durable job queue replacing internal/tasks, workflow layer with signals/timers/auto-resume reusing the proven replay-journal design, per-extension document store with live queries, UI host that dynamic-imports agent-authored bundles behind error boundaries); a daemon-supervised Bun sidecar as the shared runtime (agent code never executes in the daemon process); immutable content-addressed versioning with atomic flip and remount-based reload; reliability posture (invocation log + kill switch, explicitly not a security sandbox); teachability principles (enforce rules in the strongest layer: SDK types > scaffolds > apply-time checks > teaching errors > repo AGENTS.md > small skill reference); and the fate of existing features (tasks replaced, workflow engine generalized, plugins retained as escape hatch, Present absorbed).

**docs/plans/2026-08-01-extension-platform-roadmap.md** — the build sequence: two parallel tracks (spine: bus core → full broadcaster migration → doc store → registry/runtime → UI host+SDK; execution: queue → workflow layer) converging at the delegation pre-start hook and the first proof composition (the delegation approval gate, built once on the full stack), then Present v2 as the second proof. Each stage has scope, dependencies, and exit criteria, plus stage gates: per-stage design questions that must be discussed with and approved by Victor before that stage's implementation begins.

### DIFF
--- a/docs/plans/2026-08-01-extension-platform-roadmap.md
+++ b/docs/plans/2026-08-01-extension-platform-roadmap.md
@@ -1,0 +1,209 @@
+# Extension platform — build roadmap
+
+Status: AGREED. North star: [docs/vision/extension-platform.md](../vision/extension-platform.md).
+
+This is the stage-level roadmap: what gets built, in what order, and which
+existing feature consumes each primitive as it lands. Each stage gets its own
+detailed implementation plan when it starts; this document fixes sequence,
+scope boundaries, and exit criteria.
+
+## Sequencing decisions
+
+- **Two parallel tracks.** The dependency graph splits into a spine track
+  (bus → storage → registry → UI) and an execution track (queue → workflow
+  layer). They converge at the delegation approval gate.
+- **The gate is built once, on the full stack.** No degraded interim version
+  on plain handlers; the gate waits for hooks, workflows, storage, and UI to
+  all exist.
+- **UI host stays late in track A.** Nothing mounts in the app until the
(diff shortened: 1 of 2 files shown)


figgyster (approved):
You're taking the old plugin chassis, the ticket ledger, the replay journal, and the tile seam and admitting what they have been trying to become: a platform. Big swing. The important restraint is here too: the bus is core before extensions lean on it, agent code stays in a supervised Bun sidecar, and immutable versions keep a bad edit from turning the running app into a crime scene.

The roadmap earns the ambition. It keeps the all-at-once broadcaster migration honest, waits to mount UI until the registry and kill switch are real, and makes the delegation gate prove the whole stack rather than growing a one-off approval trapdoor. The unresolved choices are fenced at the stages that need them, where they can get real design decisions instead of becoming doctrine by paragraph.

This carries the durable delegation and Present work toward one substrate without pretending those systems were already generic. Safe to merge. Case closed.

PR #710, opened 2026-08-01
feat(bus): durable event bus core (extension platform A1)

victorarias:
Stage A1 of the [extension platform roadmap](docs/plans/2026-08-01-extension-platform-roadmap.md): the durable event bus that later stages subscribe to. Design and gate answers are in [docs/plans/2026-08-01-ext-a1-event-bus.md](docs/plans/2026-08-01-ext-a1-event-bus.md).

Clients see exactly what they saw before. This adds the spine and moves session-state and ticket broadcasts onto it; it changes no wire behavior and bumps no protocol version.

## Facts on the log, projections on the wire

The log carries **domain facts** — a dotted `domain.verb`, an indexed subject naming the entity, a small payload — not the WebSocket event a fact happens to produce.

That distinction is the whole design. Most of attn's existing `broadcastXxx` methods are snapshot pushes: `broadcastSessionsUpdated`, `broadcastTicketsUpdated`, and `broadcastPRs` re-serialize entire lists. Putting those on a durable log would store a fat stream of UI invalidations, and an extension subscribing to it would learn only that "the ticket list changed again" — it would have to diff the list to recover what actually happened.

So the WebSocket hub becomes an *ephemeral* consumer instead, and a `wireProjections` table maps facts back to what the wire needs, which is usually the same full-board push as before. The accepted cost: A2 is a real per-broadcaster migration, not a mechanical rename. The pattern for it is documented at the top of `internal/daemon/bus.go`.

## What's here

- **Migration 84** — `bus_events` (seq/name/subject/payload/source/created_at, indexed by name and subject) and `bus_consumers` (name/cursor/filter/enabled/updated_at). Re-registering a consumer preserves its cursor and enabled bit, so a daemon restart never rewinds one or resurrects a killed one.
- **`internal/bus`** — defines its own `Store` seam and does not import `internal/daemon`; the daemon adapts `internal/store` to it, the same arrangement as `sqlTaskStore`/`internal/tasks`.
- **Delivery** — durable consumers get strict global `seq` order, one event in flight, cursor advanced after the handler returns. That is at-least-once, so handlers must tolerate redelivery, and a failing handler stalls its own consumer loudly rather than skipping. New consumers start **at head**: registration is not a request to replay history.
- **Retention** — trims past an age window but never past an **enabled** consumer's cursor. A disabled consumer deliberately does not pin the log; revived below the trim point it resumes at head and the gap is logged.
- **Migrated** — `session.state.changed` plus every ticket producer (16 call sites publishing `ticket.created` / `status_changed` / `commented` / `assigned` / `attached`).
- **Operator surface** — `attn bus status [--json]`, `attn bus enable|disable <consumer>`, reading and writing the database directly. No protocol bump: status is a diagnostic, and the enabled bit is database-only on purpose, because a kill switch that depends on the daemon it kills is not a kill switch.

Byte streams stay off the bus by design — PTY output, PTY desync, attach results, workspace tile content, and fs bursts keep their direct paths. Attach routes by a per-client predicate, which pub/sub cannot express, and it is the hottest path in the product.

## Two defects found by review, fixed here

Both were in the durable delivery loop, so neither could reach the wire. Each has a regression test that was confirmed to fail without its fix.

- **Backoff measured the wrong thing.** The failure streak was cleared only when a drain reached an empty batch — which a consumer that never catches up never sees — so attempts accumulated across *different, already-succeeded* events and ratcheted to the retry cap on occasional transient failures. Reproduced at 8000 events with every 50th failing once: no event was ever stuck, yet the counter reached 160. The streak now ends on every successful delivery.
- **The kill switch could not reach a saturated consumer.** `drain` read the enabled bit once and then looped until the log was exhausted, so a consumer behind a busy producer stayed unreachable for the length of the burst — 146 deliveries in the 300ms after the flip. The bit is now re-read on the poll interval from inside the batch loop, which is the bound `DefaultPollInterval` documents.

A latent race in the test harness surfaced alongside them: `TestTrimIsCursorAware` shared a bare `now` closure between the test and the delivery goroutine. It now uses a guarded clock.

## Verification

Live-verified on a throwaway profile before the review fixes; preflight clean, protocol 200 agreed across CLI, app, and daemon.

- Three ticket mutations produced exactly the expected facts in order, each carrying its ticket id as the subject, and a connected WebSocket client received **exactly one `tickets_updated` per fact** — one board push per fact, wire behavior unchanged.
- Spawning a shell session produced one `session_state_changed` on the wire and exactly one `session.state.changed` fact.
- Daemon restart preserved the log; `attn bus status` (table and `--json`), `disable`/`enable` on a lagging consumer, and the unknown-consumer error path all behaved.

The two fixes above landed after that run and were not re-verified live, deliberately: they touch only `drain`, which is reachable solely from `deliver`, which runs solely for durable consumers. Production has none, and the publish and ephemeral-fanout paths the live run exercised are untouched.

`go test ./...` green; `./internal/bus -race -count=3` green.

## Known and accepted

Production registers no durable consumer yet — the hub is ephemeral and extensions arrive in A4 — so the durable delivery loop is proved by tests, including against the real SQLite adapter, rather than by production traffic. The retention loop does run in production.

No CHANGELOG entry: there is no user-visible behavior change, and `attn bus` is an operator diagnostic.

`docs/plans/2026-08-01-ext-b1-job-queue.md` records the approved gate answers for the parallel B1 track so the decisions are not lost. No B1 code here.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -195,6 +195,7 @@ See [docs/plans/2026-07-18-db-loss-mitigation.md](docs/plans/2026-07-18-db-loss-
 - `internal/ptyworker`: per-session process; production PTYs run here through
   `internal/pty`, not inside the daemon
 - `internal/store`: SQLite plus in-memory cache
+- `internal/bus`: durable event bus (domain facts, per-consumer cursors)
 - `internal/classifier`: stop-time state classification
 - `internal/transcript`: assistant-message extraction from JSONL
 - `app`: Tauri frontend; WebSocket `ws://localhost:9849`
@@ -255,6 +256,35 @@ must fail explicitly.
   `classifierObservation`; reject stale results.
 - Prefer `internal/protocol/helpers.go` pointer/value helpers (`Ptr`, `Deref`,
   `SessionsToValues`, `PRsToValues`).
+
+### Event bus
+
+`internal/bus` is the durable spine. State-change broadcasts do not touch the
+hub directly: a producer publishes a **domain fact** (dotted `domain.verb`, an
+indexed subject naming the entity, a small payload), and the hub — an ephemeral
+consumer — runs the matching entry in `wireProjections`
+(`internal/daemon/bus.go`) to produce the wire traffic, often a snapshot
+re-push. Migration is in progress: `internal/daemon/bus.go` documents the
+producer/projection pattern, and unmigrated `broadcastXxx` methods still
+publish directly.
+
+- A fact without a subject is a snapshot invalidation, not a fact. If the
+  producer does not know the entity id, that is the bug to fix.
+- Byte streams stay off the bus by design: PTY output, PTY desync, attach
+  results, workspace tile content, and fs bursts keep their direct paths.
+  Attach traffic routes by a per-client predicate, which pub/sub cannot express.
+- Durable consumers get ordered, at-least-once delivery from a persisted cursor,
+  so handlers must tolerate redelivery. A failing handler stalls its own
+  consumer rather than skipping the event.
+- Retention trims past the age window but never past an **enabled** consumer's
+  cursor. Disabled consumers do not pin the log; they resume at head with a
+  logged gap.
+- Operator surface: `attn bus status`, `attn bus disable|enable <consumer>`.
+  The enabled bit is database-only on purpose — the kill switch must not depend
+  on the daemon it kills.
+
+Design and gate decisions:
+[docs/plans/2026-08-01-ext-a1-event-bus.md](docs/plans/2026-08-01-ext-a1-event-bus.md).
 
 ### Terminal
 
--- a/cmd/attn/bus.go
+++ b/cmd/attn/bus.go
@@ -1,0 +1,195 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+	"text/tabwriter"
+	"time"
+
+	"github.com/victorarias/attn/internal/config"
+	"github.com/victorarias/attn/internal/store"
+)
+
+// `attn bus` is the operator surface for the durable event bus: what the log
+// holds, who is reading it, how far behind each reader is, and the kill switch.
+//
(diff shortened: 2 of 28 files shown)


figgyster (dismissed):
You laid the extension platform's foundation in the right order: facts first, wire projections second, consumers later. That buys the next stages a real ledger instead of a drawer full of callbacks. The cursor and retention work holds up, and the bus/store/daemon witnesses passed in the sandbox.

But the new spine can make the app go quiet exactly when its durable write fails. `publishFact` logs an `eventBus.Publish` error and returns; `Publish` only fans out after `Append` succeeds. A transient SQLite failure after a ticket or session mutation therefore drops the old `tickets_updated` or `session_state_changed` projection entirely, even though the underlying mutation already committed. Before this PR, that broadcast did not depend on inserting an event-log row.

Required before merge:

- Preserve the existing wire projection on the append-error path (while still logging the lost durable fact), and add a regression test that forces the bus append to fail after a ticket/session mutation. Clients cannot be allowed to miss a state change because the new observability substrate had a bad night.

figgyster (commented):
The bad-night path is fixed on `a6086c8f`. You kept the durable error visible, marked the non-durable event with sequence zero, and still let the hub carry the state change through. That is the right division of blame: the ledger can lose a line without making the city forget what just happened.

The focused bus suite passes, the ticket and session regression witnesses cover the exact holes, and Linux preflight is green. Daemon CI is still running on this fresh head, so I’m clearing the stale red and holding approval until that witness comes back.

figgyster (approved):
The rerun came back green, and the platform’s first real spine holds. This carries #709’s roadmap out of the plan drawer and into a durable, operator-visible substrate without letting the new ledger change what clients see when a write goes sideways.

The append-failure witness was the part that mattered. Facts can be temporarily non-durable, but session and ticket state still reaches the people watching it; then the next healthy append resumes the record. Daemon, Linux preflight, Bench A/B, and the rest of the board are green. Safe to merge. Case closed.

PR #711, opened 2026-08-01
fix(queue): correct Cmd+J target and give sidebar rows the full line

victorarias:
## What changed

Two user-reported fixes to the queue/attention UI, plus a CI-blocking test flake fixed along the way:

**1. Cmd+J jumped to the wrong session.** `handleJumpToWaiting` in `App.tsx` picked the target session with `unmutedEnrichedSessions.find(wantsAttention)`, which walks workspace/tree rank order rather than turn order. That disagreed with the queue band's own oldest-first ordering, so pressing Cmd+J could land on a session other than the one that had actually waited longest. Added `oldestWantedTurn(sessions, wants)` to `app/src/utils/queueBands.ts`, reusing the existing `compareTurnOrder` (ascending `turnOpenedAt`, id tiebreak), and switched `App.tsx` to use it.

**2. Queue-mode sidebar rows couldn't show session names.** The row reserved flex space for a truncated workspace-name span plus always-present hover controls (•••, pin, settle), leaving little width for the actual session label. Removed the workspace-name span, moved the hover controls into an absolutely-positioned `.queue-row-controls` overlay so they no longer take up space at rest, and right-anchored the waiting-time age with `margin-left: auto`. The session name now gets the rest of the line.

As a follow-on, the UI automation bridge's `queue_get_state` (`useUiAutomationBridge.ts`) and the `scenario-agent-queue.mjs` harness scenario both read the workspace-name text that was removed; both were switched to read `workspaceId` (`data-workspace-id`) instead.

**3. Fixed a pre-existing, timing-dependent test bug in `useDaemonSocket.ts`** that CI's full-suite run hit on this PR. The reconnect effect's cleanup cleared the pending reconnect timeout and then called `ws.close()`; since `onclose` fires synchronously in this hook's own test WebSocket mock (and can in real browsers), that scheduled a brand-new reconnect timer which outlived the cleanup (the `clearTimeout` guard had already run). The stray timer later fired against a torn-down test environment, surfacing as an unhandled `Cannot read properties of undefined (reading 'OPEN')` rejection. Fixed by nulling out the socket's handlers before calling `close()` during teardown.

`CHANGELOG.md` has entries under `## [2026-08-01]`.

## How it was verified

- Full frontend suite: 196 files / 2147 tests passing, run repeatedly locally with no flake.
- Live-verified in a running dev-profile app by injecting sessions whose oldest owed turn sat in the last-ranked workspace: Cmd+J landed on that oldest-turn session, and screenshots confirmed the row layout at rest (full-width name) and all three hover controls appearing together on hover/focus.
- CI green on the current head (Frontend, Frontend E2E x3, React Doctor, Changes).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,19 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-08-01]
 
+### Changed
+- **Queue rows give the whole line to the session name.** Sidebar rows in the
+  agent queue no longer show the workspace name, the waiting time sits at the
+  right edge, and the settle/pin/actions buttons appear over the right edge on
+  hover instead of permanently reserving space — so long session names are
+  readable instead of truncating next to empty space. The pin button's tooltip
+  still names the workspace a row would leave the queue by.
+
 ### Fixed
+- **⌘J jumps to the agent that has waited longest.** Jump-to-waiting used to
+  pick the first waiting session in sidebar workspace order, which made its
+  target feel arbitrary. It now follows the queue's own order — the turn owed
+  longest, the same row the "Your turn" band lists first.
 - **A workspace that holds only a tile can be closed again.** Closing the last
   notebook, markdown, or browser tile in a workspace with no agents left the
   workspace in the sidebar — and every later click on its × did nothing, so the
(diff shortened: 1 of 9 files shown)


figgyster (commented):
You're tightening the queue until it means the same thing everywhere. `oldestWantedTurn` gives Cmd+J the same oldest-first ledger as the band, and floating the controls finally gives the session label the row it pays rent for. That holds.

I can't approve while the Frontend job is red. It ends in an unhandled `useDaemonSocket` timer rejection after test teardown, outside this patch; the focused queue work and all three E2E shards passed. Please rerun or repair that job, then I'll sign it.

figgyster (approved):
Funny thing about a queue: when the shortcut and the list disagree, somebody gets sent to the wrong desk. You tightened the held-turn ledger from #701 and #707 until Cmd+J finally obeys the same oldest-first order the sidebar shows. The teardown repair closes the flaky witness without reaching back into queue behavior.

Frontend is green again. So are Changes, React Doctor, and the full E2E board. Safe to merge. Case closed.

PR #712, opened 2026-08-01
fix: attn resolved the wrong codex transcript when its own classifier ran in the same cwd

victorarias:
## The bug

`transcript.FindCodexTranscript(cwd, startedAt)` picked a codex rollout by matching the session's working directory and taking the newest match. attn's own stop-time classifier runs `codex exec` **from the session's cwd** (`ClassifyWithCodexExecutableInDir`) and, lacking `--ephemeral`, persisted its own rollout under `~/.codex/sessions` with that same cwd. Seconds after every codex turn, the classifier's rollout became the newest match, so transcript reads resolved attn's internal bookkeeping — a rollout whose last message is `{"verdict":"DONE"}` — instead of what codex actually said (observed live 2026-08-01, codex-cli 0.146.0).

## The fix — identity over heuristics

Codex hooks already sync the codex-native session id to the daemon on every hook event (`syncSessionResumeID` → `set_session_resume_id` → `persistResumeSessionID`, mirrored onto any bound ticket), so an exact identity is available almost immediately after spawn. Three changes, ordered strongest-identity-first:

1. **Daemon transcript resolution prefers the native id.** `resolveTranscriptPathForSession` now resolves: hook-reported path (exact, from the agent itself) → `FindTranscriptForResume(stored native id)` → cwd/time guess. This also benefits claude sessions after resumes, where the native id tracks the latest transcript file while the attn session id names the original one.
2. **The classifier stops writing rollouts.** Its `codex exec` invocation now runs with `--ephemeral` (already used by every other headless codex path in attn), so the polluting file is never created.
3. **`FindCodexTranscript` skips headless rollouts.** Rollouts whose `session_meta` declares `source: "exec"` are excluded from cwd discovery. This is provenance, not content matching: the finder's callers are exclusively resolving interactive panes (`source: "cli"`), and a `codex exec` rollout — attn's classifier on older binaries, workflow-engine runs, a user's own scripts — is categorically never the pane's transcript. Rollouts predating the `source` field are unaffected.

## Affected callers (what they read before → now)

Callers that resolved fresh via the cwd guess and could silently read classifier output:

- **Stop-time classification fallback** (`classifySessionState` → `resolveTranscriptPathForSession`): the codex stop hook payload does carry `transcript_path`, so the common path was safe — but whenever that path was missing or stale, the fallback could hand the classifier **its own previous verdict** to classify (the feedback loop the ticket flagged: `{"verdict":"DONE"}` classifies as DONE, pinning the session idle). Now resolves by native id.
- **Notebook narration** (`summarize_session` / narrate retrospective, which pass an empty path for sessions without a carried transcript): could summarize classifier bookkeeping into the journal. Now resolves by native id.
- **Ticket reconciliation** (`resolveReconcileTranscript`): already preferred the ticket's mirrored resume id; its cwd fallback (used when no id was mirrored) is now protected by the `source: "exec"` exclusion. The stale comment claiming codex has no id-based lookup is corrected.

Not affected (already identity-based, listed for completeness): `attn session transcript` (`inspectableTranscriptPath`, id-or-nothing by design), `session instructions`, automation resume, spawn resume. The transcript watcher never runs for codex.

## Verification

- **Unit regression, both layers, verified falsifiable**: `TestFindCodexTranscript_IgnoresExecSourcedRollouts` (+ mod-time-fallback variant) builds the ticket's exact fixture — the session's rollout plus a later classifier rollout sharing its cwd — and `TestResolveTranscriptPathForSession_PrefersCodexNativeSessionID` covers the id-first referee with two interactive rollouts in one cwd. Each was run with its fix piece reverted and failed on the load-bearing assertion (the decoy path was returned), then re-greened with the fix restored.
- **`make test` green** (full Go suite).
- **Live on a throwaway profile** (`make install PROFILE=codextx`, preflight PASS): spawned a real codex session via the daemon, prompted for a distinctive prose answer ("INDIGO MARMOSET…"). Evidence: the classifier's input logged in `daemon.log` was the real answer, not a verdict blob; verdict `idle` applied; exactly **one** rollout exists with the session's cwd (the pane's own — the classifier run left none, confirming `--ephemeral`); the stored `resume_session_id` matches that rollout's native id; `attn session transcript` returns the conversation. Profile torn down after.

## Notes

- Copilot's `FindCopilotTranscript` shares the cwd+newest ambiguity in principle, but attn's copilot classifier already runs from an isolated temp cwd precisely to avoid polluting that discovery, so attn does not decoy itself there. The residual ambiguity (two interactive copilot sessions sharing one directory) is out of this ticket's scope — flagged rather than fixed.

https://claude.ai/code/session_0125XqNjfU58ZJGuswMUWwTD

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,14 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   pick the first waiting session in sidebar workspace order, which made its
   target feel arbitrary. It now follows the queue's own order — the turn owed
   longest, the same row the "Your turn" band lists first.
+- **Codex sessions no longer mistake attn's own bookkeeping for their
+  conversation.** attn's stop-time classifier runs a small codex call from the
+  session's directory, and its saved log could be picked up as the session's
+  transcript — so state classification, session digests in the Notebook, and
+  orphaned-ticket reconciliation could read an internal `{"verdict":...}` note
+  instead of what the agent actually said. Transcripts now resolve by the codex
+  session's own id, the classifier no longer saves a log at all, and headless
+  codex runs are excluded from transcript discovery.
 - **A workspace that holds only a tile can be closed again.** Closing the last
   notebook, markdown, or browser tile in a workspace with no agents left the
   workspace in the sidebar — and every later click on its × did nothing, so the
--- a/internal/classifier/classifier.go
+++ b/internal/classifier/classifier.go
@@ -281,6 +281,13 @@ func runCodexClassifierAttempt(ctx context.Context, executable, model, reasoning
 	args := []string{
 		"exec",
 		"--json",
+		// Never persist a rollout under ~/.codex/sessions: the classifier runs
+		// from the session's own cwd, so a persisted rollout is a decoy that
+		// cwd-based transcript discovery can resolve instead of the session's
+		// real conversation — the classifier would then classify its own prior
+		// verdict. The verdict is read from --output-last-message and stdout;
+		// nothing reads the rollout.
+		"--ephemeral",
 		"--output-last-message", lastMessagePath,
 		// Classify regardless of the session's cwd. Codex refuses `exec` outside a
 		// trusted git repo otherwise, which is how a codex turn that ends in an
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -2577,6 +2577,12 @@ func (d *Daemon) handleStop(conn net.Conn, msg *protocol.StopMessage) {
 	go d.classifySessionState(msg.ID, msg.TranscriptPath)
 }
 
+// resolveTranscriptPathForSession resolves a session's transcript by the
+// strongest identity available: the path the agent's own hook reported, then
+// the agent-native resume id synced from hooks, and only then the finder's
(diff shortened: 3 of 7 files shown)


figgyster (approved):
That kind of self-own has bad Friday-night energy: the classifier writes a verdict, then later mistakes the paperwork for the agent. You took the right route. Native identity wins first, the classifier stops leaving a competing rollout behind, and the cwd guess only gets a seat when the evidence runs out.

The regression witness has teeth: it makes the rightful rollout older than its neighbour and still gets the right conversation. The transcript and classifier suites passed here; Daemon plus both Linux preflights are green on this head. That keeps ticket reconciliation and Notebook narration reading the agent instead of attn's bookkeeping. Approved.

figgyster (commented):
The rebase brings #711 into the room, but the transcript fix and its witnesses are byte-for-byte the case I already signed off on. Identity still beats the cwd guessing game.

Changes, React Doctor, both Linux preflights, and the relevant E2E job are green on this head. Daemon is the last required lamp still burning, so I’m holding the approval badge until it reports back.

figgyster (approved):
Daemon made it home. This rebase carries the exact transcript fix I already read line by line, with #711's queue work underneath it and every relevant check green.

The agent keeps its own voice now; the classifier's paperwork stays in the filing cabinet. Approved.

PR #713, opened 2026-08-02
fix(tickets): stop the ticket-slug fallback test flaking on a numeric suffix

victorarias:
`TestCreateDelegatedTicketFallsBackPastSequentialRange` fails about 6% of the time in CI.

## Cause

The test asserted that the allocator's random fallback suffix is non-numeric:

```go
if _, err := strconv.Atoi(strings.TrimPrefix(id, "attn-")); err == nil {
    t.Fatalf("fallback id = %q, want a non-numeric suffix past the sequential range", id)
}
```

The suffix is `uuid.NewString()[:6]` — six hex characters, which are all digits with probability (10/16)^6 = 5.96%. That matches the observed failure rate.

The allocator is not wrong. `attn-482913` collides with nothing: the sequential walk the fallback must not resume only reaches `attn-55` (`ticketSlugSequentialAttempts+5`). The assertion was over-specified.

## Fix

Assert the property that actually distinguishes a fallback id from the walk — a fixed-width random tail. The width moves from an inline `6` to `ticketSlugRandomSuffixLen` so the test references the allocator's own constant instead of duplicating a magic number. The test now allocates twice and requires the two ids to differ, so a constant suffix is caught as well.

Production behavior is unchanged: the only non-test edit replaces the literal `6` with the named constant.

## Verification

- 200 runs of `TestCreateDelegatedTicket*` green (the old assertion would be expected to fail ~12 times).
- Mutation: making the fallback continue the sequential walk turns the test red on the load-bearing assertion — `fallback id = "attn-51", want attn- plus a 6-character random suffix, not the sequential walk continuing`.
- `go vet` and the package's `Ticket` tests pass.

No live verification: the only production change is a literal-to-named-constant rename with identical behavior, and the rest is a test assertion.

### DIFF
--- a/internal/daemon/delegate_ticket.go
+++ b/internal/daemon/delegate_ticket.go
@@ -117,6 +117,11 @@ func (d *Daemon) adoptDelegatedTicket(creatorSessionID string, ownedByChiefRole 
 // unbounded at the cost of a less pretty id.
 const ticketSlugSequentialAttempts = 50
 
+// ticketSlugRandomSuffixLen is the width of the random tail. Fixed width is what
+// distinguishes a fallback id from the sequential walk at a glance, and it is what
+// the fallback test asserts on — the tail is hex, so it is sometimes all digits.
+const ticketSlugRandomSuffixLen = 6
+
 // createTicketWithUniqueSlug inserts template under base, falling back to base-2,
 // base-3, ... on slug collision, then to base-<random> once the sequential range is
 // exhausted. The template's ID field is ignored — the slug is allocated here. It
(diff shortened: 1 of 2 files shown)


figgyster (approved):
That six-percent flake was a cheap little counterfeit witness. You found the actual contract: once the readable walk runs out, the allocator owes us a fixed-width random tail, even when hex happens to look like a number.

Making the width shared with the allocator keeps the test tied to the thing it means to protect, and the second allocation proves the fallback is not a painted-on constant. This carries the durable-ticket work from #702 the right way: delegation keeps moving when a popular base has filled the board, without turning CI into a casino. Daemon and both Linux preflight builds are green. Approved.

PR #714, opened 2026-08-02
feat(daemon): auto-title sessions from the first conversation

victorarias:
## What

Sessions now name themselves. When a session finishes its first exchange (Stop hook) and its label still equals the derived default (cwd basename), the daemon generates a short descriptive title from the conversation transcript and applies it through the existing label + `session_state_changed` broadcast path.

## How

- **Trigger**: `handleStop` kicks off `maybeGenerateSessionTitle` in a goroutine alongside state classification. The transcript is guaranteed on disk at Stop time, and having the first assistant turn available improves title quality over titling at prompt-submit time.
- **Model**: the session's own agent with a cheap model — claude → `haiku`, codex → `gpt-5.6-luna` (low reasoning effort), copilot → `claude-haiku-4.5` via direct CLI invocation (copilot has no headless-task driver support; the invocation mirrors `classifier.ClassifyWithCopilot`). Overridable via `ATTN_CLAUDE_TITLE_MODEL` / `ATTN_CODEX_TITLE_MODEL` / `ATTN_COPILOT_TITLE_MODEL`; `ATTN_SESSION_AUTO_TITLE=0` disables the feature.
- **Input**: `transcript.ExtractConversationSlice` (the same agent-agnostic slice the ticket-reconcile classifier uses), rendered into a constrained prompt (3–7 words, ≤48 chars). Claude runs with a structured-output schema, tools disabled, 2-turn / $0.05 budget caps.
- **Guards**:
  - Only fires when the label still equals the default — delegations, ticket sessions, and manually renamed sessions are naturally skipped. No new column or protocol change.
  - Manual renames always win: the guard is re-checked against a fresh store read right before the write, so a rename landing mid-generation discards the generated title.
  - One LLM attempt per session per daemon lifetime, marked under a mutex with a re-check at mark time so concurrent Stops can't double-bill. Empty/unusable transcripts skip *without* consuming the attempt (early Stops retry on later turns).
  - Titles are sanitized (quotes/prefixes stripped, whitespace collapsed, rune-clamped at 48) and empty results are discarded.

## Testing

- Unit tests cover the happy path, sanitizer edge cases (including multibyte rune clamping), the custom-label skip, exec failure, unusable output, single-attempt semantics after both success and failure, the mid-flight rename race, empty-transcript retry, and concurrent-Stop single-execution.
- Live-verified on a dev-profile install: a delegated haiku session was auto-titled "Create sea haiku file", visible in the running app.
- Corpus check against 10 real historical transcripts (claude sessions of varying length/shape): 10/10 produced accurate human-quality titles, 12–17s latency, no schema or sanitizer failures.

https://claude.ai/code/session_014jtryTVEwLPhydHqx2qtFU

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ## [2026-08-01]
 
+### Added
+- **Automatic session titles from your first conversation.** When a new session
+  finishes its first exchange, attn generates a descriptive title from what you
+  discussed and replaces the default folder-name label. Manual renames always
+  win — any session you've renamed stays that way.
+
 ### Changed
 - **Queue rows give the whole line to the session name.** Sidebar rows in the
   agent queue no longer show the workspace name, the waiting time sits at the
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -162,6 +162,16 @@ type Daemon struct {
 	// lookup seam: nil in production (reconcileGroundTruth derives the fetcher
 	// from ghRegistry per-host); tests set it to a fake so no network runs.
 	ticketReconcilePRFetch prStateFetcher
+	// sessionTitleExec is the Stop-time auto-title classifier spawn (see
+	// session_title.go) — New() wires the real per-agent headless run
+	// (execSessionTitle); it stays nil on test daemons so unit tests never shell
+	// out (maybeGenerateSessionTitle logs and skips). sessionTitleAttempted
+	// marks sessions that already had one LLM title attempt this daemon
+	// lifetime, success or failure, so a Stop storm never re-bills the same
+	// session; lazily allocated under sessionTitleMu.
+	sessionTitleMu        sync.Mutex
+	sessionTitleExec      func(ctx context.Context, session *protocol.Session, slice transcript.ConversationSlice) (string, error)
+	sessionTitleAttempted map[string]struct{}
 	// ticketArtifactMu serializes attach installation with its durable ticket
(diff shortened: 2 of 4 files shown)


figgyster (commented):
You’re taking a small piece of ambient friction and giving it a daemon-owned, bounded life. That fits the trail from the ticket ledger and identity work: a session label begins as a projection, then earns a durable human story without digging another protocol trench.

I walked the Stop path, the per-agent headless launch, the attempt mutex, and the final label re-check. The tool-free run, bounded output, and rename race guard are the witnesses I wanted here. The reported dev-profile run matters too — this changes what people see, not merely a private helper.

No GitHub Actions checks have appeared yet, and the isolated review container lacks the Ghostty VT headers needed to build the daemon package, so I could not independently run this suite there. Leaving this as a comment while the board comes awake. Once the principal checks are green, I see no code blocker.

figgyster (commented):
The title patch itself came through this rebase intact. I range-compared the old and current heads: the surrounding weather is #713 and already-landed work, while the Stop-time title path, its one-attempt guard, and the rename protection have not changed.

The board is running now, with the daemon and Linux witnesses still queued or in flight. Keeping the conditional verdict on the new head so GitHub does not mistake yesterday’s commit for today’s judgment. No new code concern from the rebase.

figgyster (approved):
The board came back green where this patch lives: Daemon, both Linux builds, Frontend E2E, Changes, and React Doctor. The rebase did not touch the title path I reviewed.

This is a good little daemon promise. Default folder names are useful until the first real exchange gives a session a reason to be remembered, and the code lets the human take that wheel back at any moment. The attempt guard and final rename check keep that convenience from charging the same fare twice or writing over somebody’s choice. Approved.

PR #715, opened 2026-08-02
fix(queue): hand over the next agent however a turn closes

victorarias:
## What

Closing a turn hands you the next agent that owes one, or takes you home when none does. That only happened when you pressed <kbd>Cmd+Shift+E</kbd>.

The jump lived inside that shortcut's handler, which computed the next turn locally and moved the selection before the settle even left the app. Every other way a turn closes therefore moved nobody:

- **auto-settle** runs on a daemon timer, so a countdown completing on the agent you were watching dropped its row out of the band and left you sitting on the one agent guaranteed to be finished with you;
- the **sidebar row's settle button** goes straight to the daemon;
- so does a settle from **another client**.

## How

The handover is now a reaction to the turn closing rather than a prediction made by one caller. `handleSettleActiveTurn` only settles. A single effect in `AppContent` watches the queue bands and moves the user when the selected session's row moves out of *Your turn*. Every path arrives there as the same thing — the band the user was in changed — so the handover cannot drift between them again.

`advanceAfterTurnClosed` (`app/src/utils/queueBands.ts`) owns the decision:

- **The predicate is `turns → settled`, not "left the turns band."** Pinning or muting a workspace also clears `turn_owed` on its sessions (`decorateSessionWithTurn`), but those rows land in *no* band. Requiring the arrival rather than just the departure is what keeps a pin — which means "keep this in view" — from carrying you away from the workspace you just pinned.
- **The target comes from the band as it stood before the row moved.** By the time anything can react the settled row is gone, so a scan with no position to continue from restarts at the top of the queue instead of carrying on from where you were.
- **Scoped to the selected agent, and to the session view.** Auto-settle fires on unwatched sessions too, and a timer running off-screen must never move your selection. In grid every tile is being watched, and an empty queue would otherwise drop you out of the view entirely.

A settle the daemon rejects (`SettleTurn` returning false) now moves nobody, which the optimistic jump could not express.

## Behaviour changes beyond the fix

- <kbd>Cmd+Shift+E</kbd> moves you one daemon round trip later rather than instantly, since the move is now a reaction to the broadcast.
- Settling from grid no longer ejects you to the dashboard when nothing is left to hand over.

## Verification

- `advanceAfterTurnClosed`: 8 unit tests in `queueBands.test.ts`. Mutation-checked — relaxing the predicate to "left the turns band" turns the pin guard and the gone-from-bands guard red, and the failing assertion is the load-bearing one rather than an upstream diagnostic.
- Full frontend suite: 2152 tests / 196 files green. `tsc --noEmit` clean.
- **Live**, packaged app on an isolated `settleadv` profile, two real Claude agents: `real-app:scenario-agent-queue`, 13/13 steps, `ok: true`. The new step arms a real auto-settle countdown (both windows at their floors, 5s + 3s) and asserts the app hands over the next owed turn and moves the settled agent to Settled. The two <kbd>Cmd+Shift+E</kbd> steps drive real key presses through the packaged app's native menu and now resolve through the reactive path.

Not measured: the added latency. Those steps poll at 500ms granularity, so they prove the handover happens, not how fast.

## Drive-by

`settling_the_last_turn_leaves_selection_alone` had been asserting the behaviour #707 replaced a day earlier — that PR made settling the last turn land on home but did not update the scenario. Renamed to `settling_the_last_turn_lands_on_home` and now asserts the home landing.

https://claude.ai/code/session_01WwNvxaBqAodV7B4fpefCa5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,14 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   instead of what the agent actually said. Transcripts now resolve by the codex
   session's own id, the classifier no longer saves a log at all, and headless
   codex runs are excluded from transcript discovery.
+- **A turn that closes on its own hands you the next agent.** Auto-settle closes
+  a turn on a timer, and settling from a sidebar row goes straight to the
+  daemon — neither moved you on, so the countdown finishing on the agent you were
+  watching dropped its row from the queue and left you sitting on an agent that
+  was done with you. Closing a turn now hands over the next agent that owes one
+  (or takes you home when none does) however that turn closed, not just when you
+  pressed ⌘⇧E. A turn closing on an agent you are not looking at still leaves
+  your selection alone, and pinning a workspace still keeps you where you are.
 - **A workspace that holds only a tile can be closed again.** Closing the last
   notebook, markdown, or browser tile in a workspace with no agents left the
   workspace in the sidebar — and every later click on its × did nothing, so the
--- a/app/scripts/real-app-harness/scenario-agent-queue.mjs
+++ b/app/scripts/real-app-harness/scenario-agent-queue.mjs
@@ -30,9 +30,11 @@
  *   8. the chief of staff occupies its own slot and never the band,
  *   9. turning the arrangement off and back on mid-session restores the whole
  *      workspace tree, then returns the same queue with the same agent selected,
- *  10. settling by keyboard hands over the next agent that still owes a turn,
- *      and leaves selection alone when nothing does,
- *  11. a settle survives a daemon restart.
+ *  10. an auto-settle completing on the agent the user is watching hands over
+ *      the next agent that owes a turn, the same as settling by hand does,
+ *  11. settling by keyboard hands over the next agent that still owes a turn,
+ *      and lands on home when nothing does,
+ *  12. a settle survives a daemon restart.
  *
  * Prereqs: `claude` on PATH; a non-production profile install with the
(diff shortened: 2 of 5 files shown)


figgyster (dismissed):
You finally put the handoff where it belongs: on the durable state change. That pays off #711's promise that the visible queue and Cmd+J speak one ordering language. But this can still hand somebody to an agent who is already settled when close broadcasts arrive together.

`advanceAfterTurnClosed` sees the watched session move from turns to settled, then gets its target from `previousTurns` without checking that target still exists in `bands.turns`. If `watched` and its immediate successor settle before React commits the effect, the current band contains only `after`, yet this selects `next` from the old snapshot. The user lands on another finished agent while `after` is still owed. This PR explicitly covers settles from another client, so the race belongs in the contract.

Required before merge:

- Keep the old snapshot for the watched row's position, but skip any successors no longer owed in the current turns band; go home only when that current band is empty. Add the coalesced watched-plus-immediate-successor settlement witness.

figgyster (commented):
That was fast. You made the two snapshots earn their pay: the old queue supplies the place in line, while the current one says who is still waiting. The new collision, wrap, and same-broadcast-arrival witnesses cover the exact crack I raised. The focused queue suite passes here: 27 tests.

The rebuilt Frontend and E2E jobs are still running on `a1cefbe`. Leaving this as a comment until they get home.

figgyster (approved):
That’s the queue I wanted. The old snapshot gives you continuity, the live band gets veto power, and a turn arriving in the same broadcast keeps Home from pretending the ledger is empty. You turned the race into a contract with witnesses instead of a hopeful callback.

The focused queue suite and the full Frontend/E2E board are green on `a1cefbe`. Approved. Let the next agent have the chair.

PR #716, opened 2026-08-02
fix(state): judge yielded stops so background waits stay green and leftovers ring

victorarias:
## Problem

A Claude turn that yields with background tasks still running produces the same Stop payload whether it is **waiting on its own work** ("the build is running; I'll continue when it completes") or **finished around a leftover process** ("done — I left the dev server running"). The daemon skipped all end-of-turn classification on such stops, so the only signal deciding the color afterwards was claude's flat 60-second `idle_prompt` notification — which fires during a background wait exactly as it does after a finished turn.

Observed failure (2026-08-01, production): a session parked on a background build was settled **idle at t+60s**, opening a turn and ringing the user's doorbell for a build that resumed the turn on its own 2 minutes later. The inverse failure is latent in the same design: without the idle-prompt override, a finished turn with a leftover process would sit green forever with nothing left to unpin it (the hazard already documented in `handleStop` and the reason the chief-of-staff `relaxBackgroundWork` exemption exists).

## Change

**Yielded (non-terminal) stops are now judged like any other stop, with the yield in view.**

- The classifier input for a yield is the last assistant message plus a `[harness facts]` line reporting the running background tasks; the verdict vocabulary gains `PARKED` alongside `WAITING`/`DONE` (`internal/classifier`: prompt template, Claude JSON schema, parsers).
- The resolver (`internal/sessionstate`) reads a **parked verdict as affirmative evidence** that the silence is a background wait: the session holds `working` with **no decay to `unknown`** (`unknown` opens a turn, so the old decay didn't just show a wrong color — it rang the doorbell) and no prompt-idle override. The verdict is spent deterministically by the next busy frame — i.e. the resume it predicted — and the next turn clears the yield facts.
- A **`WAITING`/`DONE` verdict on a yield settles and rings**: the running process is a leftover, not a reason the turn will resume. This closes the green-forever case, and faster than the 60s fallback ever did.
- **No verdict at all** (classifier failed, declined, or timed out) leaves the previous ladder byte-for-byte: hold `working`, prompt-idle retires the yield at 60s as the safety valve, stuck at 90s behind it. The 2026-07-27 incident regression test (`TestAPromptIdleConfirmationRetiresAnOutstandingBackgroundTask`) passes unmodified.
- Yield judgments also skip the pending-todos rule (a plan is open precisely because the turn is not finished) and file nothing on every unjudgeable outcome (missing transcript, empty message), so a turn that may resume never lands in the user's queue on a non-answer.
- A `PARKED` answer on a terminal stop (the rule's precondition — the harness-facts line — was never in the input) is treated as no verdict, per the classifier policy of never substituting a guess for invalid LLM output.

No protocol change; daemon-only.

## Testing

- `go test ./...` green.
- New unit coverage: resolver table cases for verdict-vs-yield precedence and both incident timelines (`TestAParkedVerdictOutlastsThePromptIdleConfirmation`), clause-order case for parked-vs-prompt-idle, yield variants of the classify decision rules, `PARKED` parse coverage, and daemon wiring tests (`TestDaemon_YieldedStop_ParkedVerdictHoldsWorkingPastPromptIdle`, `TestDaemon_YieldedStop_DoneVerdictSettles`) driving `handleStop` end to end with a recording classifier.

## Live verification

Branch daemon installed into the dev profile (`make install-daemon-dev`, preflight all-PASS, protocol 200 both sides), driven with real delegated claude sessions and the real haiku classifier:

- **Background wait**: yield on "The sleep is still running in the background; I'll continue when it completes." with a background `sleep 120` → judged `parked` 9.5s after the stop → the real `idle_prompt` notification fired at t+60s and was **not** applied → session held `working` through the whole wait → sleep exited, the harness re-invoked the agent, and the terminal stop classified idle. The only resolver apply in the session's life was `idle` at the true end of turn; `turn_opened_at` was stamped only there.
- **Leftover process**: turn ended with "Done — I left a sleep process running in the background for you" with `sleep 600` still alive → judged `DONE` → settled idle and opened the user's turn **12 seconds** after the stop, process still running.

https://claude.ai/code/session_0184JiBwiizukUkN3EcqfdPo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-08-02]
+
+### Fixed
+- **A session waiting on its own background task no longer goes grey — or rings
+  you — a minute in.** When a Claude turn yields with a background process still
+  running ("the build is running; I'll continue when it completes"), attn now
+  asks its stop-time judge to read the ending with that fact in view. A turn
+  that is waiting on its own work stays green for as long as the wait takes,
+  instead of flipping to idle after 60 seconds and putting a false turn in your
+  queue. The judgment cuts the other way too: a turn that finished but left
+  something running (a dev server, a watcher) settles into your queue instead of
+  staying green forever. When no judgment lands, the previous 60-second
+  behaviour remains as the fallback.
+
+---
+
 ## [2026-08-01]
 
 ### Fixed
--- a/internal/classifier/AGENTS.md
+++ b/internal/classifier/AGENTS.md
@@ -4,7 +4,7 @@ This policy applies to stop-time assistant-message classification in this module
 
 ## Scope
 
-- In scope: deciding post-turn assistant state from assistant text (`idle`, `waiting_input`, `unknown`).
+- In scope: deciding post-turn assistant state from assistant text (`idle`, `waiting_input`, `unknown`, and — for stops that yielded with background work still running — `parked`).
 - Out of scope: runtime/hook-driven state transitions such as `working` and `pending_approval`.
 
 ## Requirements (In Scope)
--- a/internal/classifier/classifier.go
+++ b/internal/classifier/classifier.go
@@ -14,20 +14,26 @@ import (
 	"time"
 )
 
-const promptTemplate = `Classify whether this assistant message is waiting for user input.
+const promptTemplate = `Classify how this assistant message ends the agent's turn: waiting for user input, done, or paused on its own background work.
 
 Return STRICT JSON only, matching exactly one of:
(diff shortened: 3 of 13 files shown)


figgyster (approved):
You're making the attention ledger tell the truth at the one moment it used to lie hardest: a yielded turn. After #715 made every close route answer from the same live ledger, this stops a background build from minting a false owed turn before the queue ever sees it.

The narrow authority on `PARKED` is the part I read twice. It only exists with harness facts, survives the useless prompt-idle timer, and gets spent by the busy frame it predicted. `DONE` and `WAITING` still settle despite a leftover process, so the old green-forever graveyard stays buried. The two incident timelines and resolver-order witnesses have the right people in the room.

Daemon CI, both Linux preflights, and the rest of the board are green. Safe to merge.

PR #717, opened 2026-08-02
fix(transcript): count only origin-stamped turns as human in the conversation slice

victorarias:
The conversation slice that feeds the ticket-reconcile classifier and the session auto-titler treated any user-role transcript line without an `origin` field as a genuine human turn.

Modern Claude transcripts stamp typed messages with `"origin":{"kind":"human"}` and leave `origin` absent on injected user-role content — `<local-command-caveat>` wrappers, `<command-name>` / `<local-command-stdout>` blocks, tool results. So a session that opened with a slash command got that boilerplate as its `Brief` (rendered to the model under `TICKET BRIEF (first human turn)`) and the message the user actually typed was demoted into the rescoping tail. 14 of the 25 largest transcripts under `~/.claude/projects` had a boilerplate Brief before this change; all 25 have a genuine one after.

## Change

Human turns are accumulated twice in the same streaming pass:

- **strict** — only turns whose human provenance is established: an origin-stamped Claude turn, or a Codex/Copilot user message (those formats distinguish user messages from tool output by line type, so they feed both accumulators).
- **permissive** — also unstamped user-role text.

The strict result is used whenever it saw any turn at all; otherwise the permissive one, so transcripts predating the `origin` field keep todays behavior. Codex/Copilot slices are byte-identical to before and do not depend on the fallback.

No marker denylist: chasing an undocumented wrapper format would also drop genuine messages that merely quote a marker.

### Trade-off (documented at the type)

A modern transcript whose user-role content is *entirely* injected is indistinguishable from a legacy one by this rule, so it falls back and reports the injected text. Provenance is per-line — nothing separates "file written before the field existed" from "file where no human turn has landed yet". The fallback only fires when the alternative is an empty slice, which is the safer of the two errors.

## Tests

Four new unit tests in `internal/transcript`, using JSONL fixture lines shaped like real transcript lines:

- modern transcript opening with a caveat + command wrappers followed by stamped turns → Brief is the human turn, wrappers excluded entirely (not demoted into rescoping), `HumanCount` counts only stamped turns;
- legacy transcript with no `origin` anywhere → unchanged output via the fallback;
- non-human stamped origin (`kind: hook`) counts in neither accumulator, so it cannot resurface through the fallback;
- modern all-injected transcript → the documented fallback behavior.

Existing Codex and Copilot tests are untouched and pass.

Verified the new assertions are load-bearing: reverting the one-line provenance check makes the first test fail on both `Brief` and `Rescoping`.

## Verification

- `go test ./internal/transcript/ ./internal/daemon/` — green.
- Read-only spot check over the 25 largest real transcripts under `~/.claude/projects` (throwaway test file, not in the diff): every Brief is now a genuine first human message. The transcript cited in the ticket, which literally begins with a `<local-command-caveat>` line, yields `"hi"` instead of the caveat.

Library-level fix with no protocol, daemon-lifecycle, PTY, or UI surface, so no live app verification tier applies.


figgyster (approved):
This is the right kind of small fix for the attention ledger. Ticket reconciliation and auto-titling both drink from this slice, and you kept them honest with provenance instead of teaching the parser a brittle little blacklist of slash-command wrappers.

The strict/permissive pair keeps old Claude transcripts useful while modern injected user-role text stays out of both the Brief and the re-scoping tail. The regression witnesses pin both sides of that bargain, and Codex/Copilot keep their established line-type path.

Approved. `go test ./internal/transcript` passed in the review sandbox; GitHub's current Daemon, both Linux preflight jobs, E2E, React Doctor, Changes, and Bench A/B board are green on `bb12001`.

figgyster (commented):
You carried the transcript fix forward untouched and put its release note into the fragment ledger that #720 just established. Good small maintenance work: the attention path keeps its provenance rule, and the release machinery gets the fact without reopening the old `CHANGELOG.md` traffic jam.

`go test ./internal/transcript` and `go run ./cmd/changelog-check` both passed in the isolated review sandbox. GitHub's new board is still running, so I’m leaving this as the current-head review while those witnesses finish their shift.

figgyster (approved):
The rebuilt board came home green, and this restack still has the right shape. The provenance rule stays untouched; the changelog fragment gives the release compiler its facts without dragging concurrent work back into `CHANGELOG.md`.

Approved. The focused transcript and fragment-validator witnesses passed in the isolated review sandbox, and GitHub’s Daemon, Linux preflight, E2E, Changes, Changelog, and React Doctor checks are green on `7473215`. Bench A/B is still running as the usual non-gating weather report.

PR #718, opened 2026-08-02
fix: detect a halted turn for claude, codex, and copilot

victorarias:
Interrupting an agent mid-turn left the session showing `working`.

No agent fires a hook when the user halts a turn. Claude's `Stop` hook does not
run on interrupt (documented behavior — `Stop` is "the main agent has finished
responding", and an interrupt is not finishing). Codex has no hook surface at
all, and copilot 1.0.77 fires nothing either. So the session only recovered when
the heartbeat aged out at `StaleAfter` (60s) — and for copilot, whose live state
comes entirely from its transcript watcher, it never recovered at all: the
watcher held the turn bracket open for the rest of the session's life.

Each agent does record the halt in its transcript. The transcript watcher now
detects it and settles the turn.

## Detection is structural, not textual

**Claude** writes a user entry carrying `interruptedMessageId`, or a lone
`[Request interrupted by user]` text block. The field alone is not enough —
halting at a tool-use prompt writes the marker *without* it — so a marker-only
entry is accepted only when the entry carries none of `promptSource`, `origin`,
or `permissionMode`. Those three are what a real prompt has and a synthesized
interrupt record does not, which is why pasting the marker as a prompt no longer
settles the session it was typed into.

**Codex** aborts for four reasons; only `interrupted` is a user halt.
`replaced`, `review_ended`, and `budget_limited` are parsed and explicitly are
not.

**Copilot** writes `{"type":"abort","data":{"reason":"user_initiated"}}` and,
unlike every other turn ending, no `assistant.turn_end`. Every abort now closes
its watcher-owned turn bracket; only `user_initiated` settles.

## A halt is dated by the agent, not by the read

It is dropped if it predates the session or carries no timestamp. One rule
covers the three ways a halt can be read out of its own time:

- the bootstrap rewind, which tails the last 256KB/512KB of a resumed transcript
- the truncation reset to offset 0
- the 500ms poll landing after the next prompt has already gone out

A late-read halt is now outranked by the busy frames that followed it (the
existing `supersededByBusy` rule), while `LastMovement` still advances to the
read time, so nothing ages into `stuck`.

## Two more fixes

**Copilot transcript discovery compared cwds literally.** Copilot records the
*resolved* path, so any session launched under a symlink (`/tmp` on macOS) never
found its transcript and sat at `launching` for its whole life. It now compares
through symlinks, as the codex finder already did. Found while verifying the
halt fix.

**Codex and copilot discovery walked the whole session tree twice a second** for
as long as a transcript stayed unfound. It now backs off: every poll for the
first 10s, then every 2s, then every 5s.

## Verification

Live, on a throwaway profile, against real agents. Halt to settled:

| agent | halt key | settled |
|---|---|---|
| claude 2.1.220 | ESC | **1.02s** |
| codex 0.146.0 | Ctrl+C | **1.29s** |
| copilot 1.0.77 | ESC | **1.20s** |

Pasting `[Request interrupted by user]` into a live claude session goes `working`
-> `waiting_input` and never settles.

Every guard is mutation-tested: removing it turns the load-bearing assertion red
(stale gate, both claude discriminators independently, codex reason filter,
copilot bracket clear, agent-vs-read dating, copilot symlink match).

Note for anyone re-running this: **codex 0.146.0 interrupts on Ctrl+C, not ESC.**
Sending ESC looks exactly like a regression in this code and is not. That, and
copilot's reuse-a-directory discovery trap, are recorded in the plan doc.

## What was deliberately not shipped

An agent-agnostic idle backstop — settle any session that goes quiet for N
seconds regardless of agent. Measured on claude 2.1.220:

```
halt, no tool:   ESC        -> 60ms later  "* Write typewriter history essay"   -> silence
60s foreground:  tool start -> 8s later    "* Run sleep command for 60 seconds" -> silence 64s
```

Same glyph, same shape, same silence. No threshold below the longest tool call
separates a halted turn from a blocking one, which is exactly why `StaleAfter` is
sized to the longest tool call in the first place. The traces are in the plan doc
so this doesn't get re-proposed.

https://claude.ai/code/session_015EYRHFgwnr143eV4im2JrF

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,31 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
 
 ---
 
+## [2026-08-02]
+
+### Fixed
+- **Interrupting an agent settles its session right away.** Hitting ESC to halt
+  claude, codex, or copilot mid-turn used to leave the session showing *working*
+  for a full minute, then briefly *idle*, then *unknown* — because none of them
+  fires any hook when a turn is interrupted, so nothing ever closed the turn.
+  attn now reads the halt out of the agent's own transcript and settles the
+  session to *idle* within a second or two, with `turn_aborted` as the reason in
+  `attn state explain`. A halted copilot session was the worst of the three: it
+  writes no turn-end event either, so the session stayed green for the rest of
+  its life.
+- **Copilot sessions launched through a symlinked directory no longer sit at
+  *launching* forever.** Copilot records the directory it resolved rather than the
+  one it was given, so a session started under a symlinked path — `/tmp` on macOS,
+  for instance — was never matched to its own transcript, and copilot's live state
+  comes from that transcript.
+- **A halted turn is only ever your own.** Quoting or pasting claude's
+  `[Request interrupted by user]` marker into a prompt no longer settles the
+  session you are talking to, a turn codex ended for its own reasons is no longer
+  read as you halting it, and a session that resumes onto a transcript with older
+  halts in it no longer settles on one of them.
+
+---
+
 ## [2026-08-01]
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,11 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   comes from that transcript.
 - **A halted turn is only ever your own.** Quoting or pasting claude's
   `[Request interrupted by user]` marker into a prompt no longer settles the
-  session you are talking to, a turn codex ended for its own reasons is no longer
-  read as you halting it, and a session that resumes onto a transcript with older
-  halts in it no longer settles on one of them.
+  session you are talking to, a turn codex or copilot ended for its own reasons is
+  no longer read as you halting it, and a session that resumes onto a transcript
+  with older halts in it no longer settles on one of them. A copilot turn that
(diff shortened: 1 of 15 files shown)


figgyster (changes requested):
You're closing the last unreported turn exit, and #712's transcript-identity work is what makes this safe to build on instead of a fresh way to misread an old rollout. But Copilot still has one bracket standing in the rain.

Required before merge:

- `internal/agent/transcript_watcher_behavior.go:239`: a non-user Copilot `abort` clears only `b.turnOpen`. It sends neither `Aborted` nor `State`, so the daemon never clears the evidence `TurnOpen` that the earlier `assistant.turn_start` recorded; the resolver keeps reporting `bracket_open` until `StaleAfter`. That breaks the stated contract that every Copilot abort closes its watcher-owned bracket. File a neutral close/evidence transition for every abort, then reserve `turn_aborted` for `user_initiated`, with a real watcher regression for (say) `tool_failure`.

figgyster (approved):
You were right about the whole edge, and I was only holding the short end of it. A Copilot abort can now say the one true thing when it is not a user halt: its watcher-owned bracket ended. #712 made transcript identity fit to trust; this keeps that evidence from putting words in the user's mouth.

The real watcher witness drives `tool_failure` through to a cleared daemon bracket without manufacturing `turn_aborted`; an actual user halt still gets the immediate idle handoff. The focused agent, transcript, session-state, and daemon witnesses passed here, and Daemon plus both Linux preflight jobs are green on 3212e3b. Safe to merge. Case closed.

PR #719, opened 2026-08-02
Route every state-change broadcast through the event bus

victorarias:
Stage A2 of the event-bus plan: every state-change broadcast now reaches WebSocket
clients by publishing a domain fact, and only a projection turns a fact into bytes.
Byte streams stay on their direct paths by an enumerated list.

Before this, the bus carried a handful of facts and most of the daemon still called
the hub directly. That split meant a durable consumer could not be written against
daemon state — it would see some changes and miss others, with no way to tell which.
A2 closes that: the fact log is now a complete record of what changed.

## The exit criterion is enforced, not asserted

`TestWireTrafficComesFromProjections` walks the package AST and fails any function
that reaches the hub without a fact behind it. The exception list has five entries,
each carrying its reason in the source:

- the projection table itself, and the two generic senders projections call;
- `broadcastRawWSMessage`, the remote relay — those events were already published as
  facts on the remote daemon, and re-publishing them locally would duplicate them;
- `broadcastFsChanged` and `broadcastTileContent`, byte streams with no entity worth
  subscribing to, routed by a per-client predicate that pub/sub cannot express.

Adding a sixth entry is a visible design decision rather than a silent regression.

## Behavior is byte-identical, with two deliberate exceptions

The first commit is the instrument, not a change: a tap on all five hub send paths
plus golden traces, written before any producer moved. Both goldens are unchanged
across the whole series, so each migration commit is provably byte-equal on the wire.

Two reductions are intentional, both from producers that now diff rather than
re-push blindly:

- a PR poll that finds nothing new no longer re-pushes the list;
- host discovery that finds the same hosts no longer re-pushes.

"Changed" means JSON equality — whether a client would actually see something
different.

## A projection may not publish

`bus.publish` holds its lock across the inline fan-out, so a fact published from
inside a projection deadlocks. This surfaced as the daemon suite hanging past 600s
on `TestNudgeCountdownSurvivesEligibleStateChange`: the session-state projection was
recomputing the workspace rollup, which published a second fact.

The recompute stayed on the producer side. The rule — a projection writes to the
wire and does nothing else — is documented in `internal/daemon/bus.go` and
`AGENTS.md`. A runtime reentrancy guard was considered and rejected: without
goroutine identity there is nothing to key it on, and a timeout-based version risks
dropping real facts.

## Producers that changed signature

A broadcaster that cannot name the entity that changed publishes a snapshot
invalidation, not a fact. Four were fixed at the producer rather than papered over:

- `tasks.Runner.OnChange` now carries the task id;
- the plugin supervisor's change callback carries the plugin name;
- mark-all-notifications-read collects the ids it marked before writing;
- GitHub host discovery diffs around the replacement to recover per-host facts.

Bulk operations publish one fact per entity and wrap them in `coalesceSnapshots`,
which collapses the resulting whole-list pushes to one wire message per snapshot.
That is what keeps the fact log useful without multiplying wire traffic.

## Cost

The migration adds a durable SQLite row append per state change. Two paths that were
previously documented as non-blocking now write: the task runner's change callback,
which fires concurrently from the dispatch goroutine and each in-flight run, and the
plugin supervisor's timers. Nothing measured slow in the live run, but this is a real
change in what those paths cost.

## Verification

- `go build`, `go vet`, and `go test ./internal/...` clean at every commit head.
- Mutation-checked: disabling coalescing, deleting a handler's publish, dropping the
  notebook collapse, and adding a direct hub call each turn the suite red.
- Live on a throwaway profile: preflight PASS, protocol 200 agreed, 161 production
  facts with zero empty subjects, 20 PR facts collapsing to 2 wire pushes,
  `real-app:scenario-workspace-shell-lifecycle` green, and a setting change producing
  exactly one fact and one `settings_updated` carrying `changed_key`.

No protocol change: no wire shape moved, so `ProtocolVersion` is untouched.

Design notes and the full A2 record are in
`docs/plans/2026-08-01-ext-a1-event-bus.md`.

## Reading order

`17ea5327` is goldens only, no production change — read it first to see what the rest
is measured against. Then the four migration commits by surface, then the boundary
test, then docs.


figgyster (dismissed):
You’re taking the bus from a promising side street to the daemon’s actual spine. That is a big migration, and the projection boundary plus wire traces are the right kind of paranoia for it. But the new witness is still borrowing its facts from the machine it runs on.

Required before merge:

- Make `TestWireTraceProducerGolden` hermetic. The current head masks executable availability, but the required Daemon job still fails because `notebook.root.effective` comes from the runner’s home directory: the golden carries `/Users/victor/attn-notebook` while CI emits `/home/runner/attn-notebook`. Set that setting to a test-owned path or normalize this specific host-derived value, then rerun the daemon witness.

The event-bus proof is supposed to catch changed daemon traffic. It should not reopen the case whenever the precinct moves to a different building. The CI board is red on this exact test, so I can’t sign the current head yet.

figgyster (commented):
That closes the hole I found. You kept the wire trace’s teeth and stopped it mistaking the runner’s home directory for daemon behavior, which is exactly the boundary this migration needs.

The A2 shape still holds on this head: facts own the state change, projections own the bytes, and the golden no longer changes uniforms when it walks from a laptop into CI. Daemon and Bench are still running, so I’m clearing the stale red to a current conditional review rather than stamping an approval before the last witnesses get back.

figgyster (approved):
That was the last bad witness. The trace now ignores the room it was run in and keeps its eye on the daemon’s actual wire traffic, so the test is back to doing its job.

#710 built the durable fact log. You’ve now given every stateful broadcaster a road onto it, with projections keeping the old client contract intact and the AST boundary making backsliding loud. Daemon and both Linux preflights are green; Bench is still collecting its trend evidence. Safe to merge.

PR #720, opened 2026-08-02
feat(release): compile the changelog from per-PR fragments

victorarias:
## Why

CHANGELOG.md is a serialization point: every PR appends bullets to the same few lines at the top of the file, so any two in-flight branches conflict on it. With several agents working concurrently, changelog conflicts were a constant tax on merges. GitHub computes PR conflicts server-side without custom merge drivers, so the only structural fix is for PRs to stop editing the shared file.

## What

The changelog becomes a compiled artifact. PRs ship a small **fragment** — a raw statement of what changed — and a release-time step compiles the accumulated fragments into one curated, user-facing dated section, written once by a writer that sees the whole release.

- **`changelog.d/`** — one YAML fragment per PR (`kind`/`area`/`change`, optional `symptom`/`notes`). Fragments are evidence for the release writer, not final copy. `kind: internal` records changes with no user-visible behavior; they inform the writer but never become bullets. Unique filenames per branch make fragments conflict-free by construction.
- **`cmd/changelog-check`** — schema validator (unknown kinds, missing fields, and typo'd keys all rejected). Unit-tested; `TestRepoFragments` also validates the checked-in `changelog.d/` so a bad fragment fails `make test` on the branch that introduced it.
- **`Changelog` CI job + `scripts/changelog-gate.sh`** — every PR must add a `changelog.d/*.yaml` fragment or modify `CHANGELOG.md` directly (the escape hatch used by the compilation PR and copy hand-fixes). `release/*` branches (the version bump from `scripts/release.sh`) are exempt. Runs locally too, counting staged/untracked fragments so pre-commit runs are honest.
- **`scripts/compile-changelog.sh`** — validates pending fragments, has `claude -p` write the dated section per the changelog's house style (bold user-facing lead, related entries merged, internal dropped), inserts it at the top of `CHANGELOG.md`, deletes the consumed fragments, and leaves everything staged for human review. Handles a second same-day compile by folding the existing section in, and an all-internal set by removing fragments without a section. `--dry-run` prints the prompt.
- **`scripts/release.sh`** — refuses to cut a release while fragments are pending.
- **`docs/making-a-release.md`** — owns the fragment format, gate rules, and release workflow; AGENTS.md points there.

This PR carries its own `kind: internal` fragment, so the new gate passes on the PR that introduces it.

## Verification

- Validator: `go test ./cmd/changelog-check` green; `gofmt`/`go vet` clean.
- Gate exercised locally on all three paths: pass (fragment present), fail (empty diff → doc-pointing error, exit 1), and `release/*` exemption.
- Full live compile smoke: two fake user-facing fragments plus the real internal one, real `claude` invocation. The writer folded the existing same-day section in verbatim, wrote the fakes in the existing voice, omitted the internal fragment, and left all section separators intact. Working tree restored byte-identical afterward.
- The `Changelog` job on this PR is the live positive test of the CI wiring; a companion throwaway PR with no fragment verifies the job fails (link in comments, closed after).

## Notes

- The gate becomes *required* only once the `Changelog` check is added to branch protection — repo settings, not in this diff.
- The What's New modal stays hand-written; the doc notes it should draw from compiled sections.

https://claude.ai/code/session_01WSSv35YUQhgdR35f9zaDJF

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,23 @@ jobs:
             tauri:
               - 'app/src-tauri/**'
               - '.github/workflows/ci.yml'
+  changelog:
+    name: Changelog
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.22'
+
+      - name: Gate
+        run: ./scripts/changelog-gate.sh "${{ github.base_ref }}" "${{ github.head_ref }}"
+
   backend:
     name: Daemon
     needs: changes
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -350,5 +350,7 @@ Packaged-app default menu accelerators can consume shortcuts before DOM keydown.
 - Diagnose root cause. Do not remove requested behavior without explicit
   approval. For refactors, list and verify behaviors that must survive.
 - Do not copy production code into tests or test compile-time guarantees.
-- Update `CHANGELOG.md` only for meaningful user-visible changes. Use dated
-  sections; describe the PR's net behavior in concise user-facing bullets.
+- Every PR adds a changelog fragment under `changelog.d/` — CI enforces it.
+  Do not edit `CHANGELOG.md` directly; it is compiled from fragments at
+  release time. Format and release process:
+  [docs/making-a-release.md](docs/making-a-release.md).
--- a/changelog.d/README.md
+++ b/changelog.d/README.md
@@ -1,0 +1,5 @@
+# changelog.d
+
+Pending changelog fragments — one YAML file per PR, compiled into
+`CHANGELOG.md` at release time by `scripts/compile-changelog.sh`. Format,
+examples, and the release workflow: [docs/making-a-release.md](../docs/making-a-release.md).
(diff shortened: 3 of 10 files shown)


figgyster (dismissed):
You're cutting `CHANGELOG.md` out of the merge queue and making it a release-time document. That's the right move for a repo with agents coming through every door at once. The fragment schema is pleasantly strict, and the focused checker test passes in the sandbox. But one release-night door is still unlocked.

`validateDir` treats a symlink named `*.yaml` as a normal file: `IsDir()` is false, then `os.ReadFile(path)` follows it. The compiler later does the same with `cat "$f"` before it sends the text to Claude. A merged PR can therefore leave a fragment link aimed at an arbitrary file on the release machine; the gate blesses it and the release run reads it. That crosses the fragment boundary into local-file disclosure.

Required before merge:

- Reject symlinks (and preferably any non-regular fragment) in `changelog-check`, with a regression witness. Since the compiler validates before reading its fragments, that one guard seals both routes.

figgyster (commented):
The follow-up closes the hole. `DirEntry.Type()` catches the link itself before `ReadFile`, every non-regular fragment is refused, and the new witness proves a perfectly valid YAML target still cannot be followed out of `changelog.d`.

That puts the compiler back behind the boundary it claims to have. The focused checker suite passes on `5582383`; the rebuilt CI board is still running, so I’m leaving this as the current-head note rather than trying to stamp a green light onto a moving train.

figgyster (approved):
The release compiler has its boundary back: fragments are regular files or they do not get read. You fixed the hole without growing a second release system around it, and the symlink witness gives the next person a reason to keep it that way.

The rebuilt Daemon, Linux, Frontend, E2E, Tauri, and Changelog board is green on `5582383`. This is the kind of dull release plumbing that saves a small army of branches from fighting over the same ten lines. Approved.

PR #723, opened 2026-08-02
feat(state): real state detection for shell panes

victorarias:
## What

Shell (terminal) panes get real state detection, matching the quality of agent detection: `working` while a command runs, `idle` at the prompt — instant edges with the command line and exit code when the shell emits OSC 133 marks, a 1s foreground poll as the universal floor.

## How

Two evidence sources feed one `shellSignalArbiter` (`internal/pty/shell_signals.go`); the existing resolver clauses do the rest with zero new clauses (shells never open turn brackets, so fresh-busy → `working`/`heartbeat_fresh` and settled → `idle`/`at_prompt`).

- **Foreground poll**: `TIOCGPGRP` on the PTY master once a second, compared against the shell's own pgid. Works for any job-control shell with nothing injected.
- **OSC 133 marks**: `C;cmdline_url=…` → instant busy edge carrying the command line; `D;exit` → instant not-busy edge carrying the exit code; `A` → prompt restate. Edges catch sub-second commands the poll cannot see and bypass the keepalive dedup (per-command events, not repaints).
- **Reconciliation**: the one case where the two disagree is an inner shell at its prompt (ssh to a host whose remote shell emits marks, nested shells) — naive layering would flap working/idle every second. Rule: a prompt verdict binds to the foreground program it arrived from and holds only while that program stays foreground; the local shell reclaiming the terminal, or any foreground change, hands the claim back to the poll. A dropped ssh mid-command self-corrects within one poll.
- **zsh/bash integration injection** via the existing startup overlays. zsh: precmd *prepended* so `$?` is the real exit status, preexec emitting `C;cmdline_url` (percent-encoded in-shell). bash: pre-exec is a DEBUG trap — the only hook every bash has, macOS's 3.2 included — carrying `BASH_COMMAND`. Installation happens at the first prompt, not at source time: inside a sourced file `$(trap -p DEBUG)` reads empty whether or not the user has a trap, and bash versions disagree about whose trap survives a source-time install (3.2 restores the user's, 5.x lets the sourced file clobber it). An inline PROMPT_COMMAND probe reads the truthful top-level listing once and either claims the free trap or falls back — user-owned traps and extdebug are never touched; the fallback keeps PS0's bare `C` (bash ≥ 4.4) and degrades `D` to every prompt after the first. Everything is sourced behind POSIX-safe interpreter guards so dash/`sh` never parse the zsh/bash-only syntax. `ATTN_NO_SHELL_INTEGRATION=1` opts out. This also lights up command blocks — with command text and exit codes — in the terminal UI for zsh/bash panes, fish-only until now (fish emits natively).
- **Worker-info veto**: the worker runtime caches state `"working"` from birth and the watch-subscribe replay reports that cache as a claim; for shells nothing ever sets it, so it would flip every fresh shell out of `idle`. `handlePTYState` now vetoes worker-info claims for shells (`shell_resolver_owned`) — the resolver owns shell state.

Shell panes still never enter the attention queue: `attention.Owed` keeps excluding shells, now as a policy choice rather than a consequence of shells never changing state.

## Known limits (accepted v1)

- Without integration, a foreground program that is itself waiting (idle vim) reads as `working` — from the terminal's view a program does own the foreground.
- No `B` (input-start) mark: the input anchor is optional in the block contract, and command text comes from `cmdline_url` rather than the grid.
- bash's `C` carries the first simple command of a compound line (`a; b` marks as `a`), and a user-owned DEBUG trap (or extdebug) degrades bash to the PS0 fallback: bare `C` on bash ≥ 4.4, none on 3.2, exit codes still delivered via `D`.

## Verification

- Full Go suite green; Linux amd64/arm64 cross-builds green (daemon runs on Linux remotes).
- Real-PTY regressions run the actual overlays in real `/bin/zsh` and `/bin/bash` and feed the captured streams through the production segmenter **and block table**: completed blocks asserted with command text (`/bin/echo attn-integration-probe`, `false`) and exit codes (0, 1), alongside the arbiter's `command started: …` / `command exited N` observations. A user-owned DEBUG trap is asserted to keep firing with no cmdline `C` marks emitted over it; the opt-out is asserted to emit nothing.
- The bash integration was exercised on both 3.2.57 and 5.3.9: identical marks with a free trap, identical user-trap preservation.
- Live on the dev app (fish login shell, native marks): a 35ms `/bin/echo` produced `command started: /bin/echo hello-edges` → `command exited 0` in the `attn state explain` trace without flickering state; `false` → `command exited 1`; `sleep 3` flipped to `working` off the marker edge with the poll corroborating mid-command, back to `idle` one tick after the prompt returned; the worker-info `working` replay was observed live and vetoed; `TurnOwed` never set.

https://claude.ai/code/session_01ArkyvqLGtS7JsmHDX6H6W7


figgyster (dismissed):
The state arbiter is the right piece of machinery — especially the foreground-owner binding for an inner SSH/nested-shell prompt. But the shell integrations do not emit the command-block contract this PR says they do.

`zshOSC133Integration` emits `A`, bare `C`, and `D`; the Bash overlay likewise emits a bare `C` only when `PS0` exists, while macOS's Bash 3.2 has no `PS0`. Neither integration emits `B` or `C;cmdline_url=...`. That leaves zsh/Bash blocks without their input anchor or command text, and on the macOS Bash path there is no `C` at all, so the block table discards the later `D` and a sub-second command never gets a busy edge. The changelog and PR description promise selectable command blocks plus instant command-line-carrying edges for both shells, which this stream cannot supply.

Please either emit and cover the full `A`/`B`/`C;cmdline_url`/`D` sequence (including the macOS Bash fallback), or narrow the advertised contract and add a regression that feeds each real integration stream through the arbiter/block table. The poll is a fine universal floor; it cannot make that extra promise true between ticks.

figgyster (approved):
The follow-up closes the actual gap: zsh and the free-trap Bash path now emit command-line-bearing `C` marks and exit-bearing `D` marks, while a user-owned DEBUG trap or `extdebug` stays out of attn’s hands. The new real-shell tests carry each raw stream through the arbiter and production block table, so the sub-second state edge and completed command block have an eyewitness instead of a brochure.

`B` remains a deliberately optional input anchor; the existing block contract supports that narrower integration without trying to pick a fight with every prompt framework in town. Focused session-state verification passed, and Daemon, both Linux preflights, Changes, Changelog, React Doctor, and Frontend E2E are green. Approved on `05256d03`.

PR #724, opened 2026-08-02
feat(queue): snooze an agent until a time you name

victorarias:
## Summary

Implements the *Deferral* rock of [the agent queue vision](docs/vision/agent-queue.md); full design in [the plan](docs/plans/2026-08-02-agent-queue-snooze.md).

Settle answers "I dealt with this". Nothing answered "not now", so an agent you cannot act on yet — a run with 40 minutes left, a finished run you want to read after lunch — either sat in the band or was settled dishonestly and came straight back.

Snooze is settle plus a suppression window, written in one statement so a turn is never both open and suppressed. It suppresses the turn *opening* rather than filtering an open turn at read: that is what makes a woken turn stamped at the wake instant and land at the tail of the queue, instead of resurrecting an hour-old stamp straight to the head of the band. Any agent can be deferred, owed or not — the useful case is precisely the one where no turn is open yet.

The deferral is honoured while it lasts: steering or looking at a snoozed agent does not cancel it. Only what the user could not have anticipated breaks through — a state the daemon cannot explain (`unknown`), or a process that died (`idle`/`process_exited`) — and a break-through consumes the snooze rather than pausing it.

Daemon-owned wake timers mirror auto-settle: armed on snooze, cancelled on wake, rebuilt from persisted deadlines at start-up. A deadline that lapsed while the daemon was down stamps the turn at the deadline, because the turn has genuinely been owed since then.

Deferred agents render in a collapsible Snoozed section at the foot of the sidebar, above muted workspaces, each with its wake time and a wake-now action. Snoozing hands over the next agent that wants the user, exactly as settling does.

- Protocol 201 (`turn_snoozed_until`, `snooze_turn`, `wake_turn`); migration 85.
- Duration menu: 30m, 1h, 8h, tomorrow, Saturday, Monday (day choices wake at 09:00 local).
- Indefinite backgrounding and a custom date/time picker are deliberately out of scope — see the plan's Deferred section.

## Test plan

- [x] Full Go suite green
- [x] Frontend: 197 files / 2179 tests green
- [x] `pnpm --dir app exec tsc --noEmit` clean
- [x] Live-verified twice on a non-production profile via `real-app:scenario-agent-queue-snooze`: two real Claude agents, snoozed from the row menu for 30 minutes, watched out of the band into the section with its wake time, driven through a second run under the deferral with no turn opening, woken early to the tail, woken again by the timer, carried across a daemon restart, and SIGKILLed to see the break-through.

https://claude.ai/code/session_01BzKRr7PFLDU9mERFDUdTv9

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -66,6 +66,7 @@
     "real-app:scenario-ordinary-delegation-ticket": "node scripts/real-app-harness/scenario-ordinary-delegation-ticket.mjs",
     "real-app:scenario-ticket-resume": "node scripts/real-app-harness/scenario-ticket-resume.mjs",
     "real-app:scenario-agent-queue": "node scripts/real-app-harness/scenario-agent-queue.mjs",
+    "real-app:scenario-agent-queue-snooze": "node scripts/real-app-harness/scenario-agent-queue-snooze.mjs",
     "real-app:scenario-nudge-trigger": "node scripts/real-app-harness/scenario-nudge-trigger.mjs",
     "real-app:scenario-terminal-block-copy": "node scripts/real-app-harness/scenario-terminal-block-copy.mjs",
     "real-app:scenario-terminal-context-menu": "node scripts/real-app-harness/scenario-terminal-context-menu.mjs",
--- a/app/scripts/real-app-harness/scenario-agent-queue-snooze.mjs
+++ b/app/scripts/real-app-harness/scenario-agent-queue-snooze.mjs
@@ -1,0 +1,475 @@
+#!/usr/bin/env node
+
+/**
+ * Real-app scenario: snoozing an agent in the queue.
+ *
+ * Settle answers *I dealt with this*. Snooze answers *not now*: it closes the
+ * open turn and stops the next one from opening until a time the user named.
+ * The claim only exists end to end — the daemon writes the settle and the
+ * deadline in one statement, suppresses the turn-open while the deadline holds,
+ * arms a timer, and the sidebar draws the row in its own section with the time
+ * it comes back.
+ *
+ * The run drives two real Claude agents in two workspaces and asserts, against
+ * the rendered DOM (`queue_get_state`) and the daemon's own broadcast, that:
+ *
+ *   1. snoozing from a queue row's menu closes that turn, takes the row out of
+ *      both bands, and puts it in the Snoozed section with a wake time,
+ *   2. snoozing the agent the user is looking at hands over the next agent that
+ *      owes a turn, the same way settling does,
+ *   3. the deferral actually holds: the snoozed agent runs, stops in a state
+ *      that would open a turn, and no turn opens,
+ *   4. waking early puts it back — and at the *tail*, behind the turn that has
+ *      been owed longer, because the clock on what you owe starts when you said
+ *      you would come back,
+ *   5. the wake timer does the same thing on its own when the deadline passes,
+ *   6. a deferral survives a daemon restart, deadline and all,
+ *   7. break-through: an agent whose process dies clears its own snooze and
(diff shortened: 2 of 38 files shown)


figgyster (dismissed):
You’ve taken the queue work from "whose turn is it?" into "when am I ready?" That’s the right next rung after #715 and #716: the ledger owns the interruption, and this gives the human an honest way to postpone it.

But one timer can still cash a promise the user replaced. In `internal/daemon/snooze.go:226`, `fireSnoozeWake` proves the old timer is current, deletes its entry, then unlocks before `wakeSnooze` clears the stored deadline. A new `snooze_turn` can land in that gap, write a later deadline, and arm its new timer; the old callback then cancels that new timer and clears the new deadline. The agent wakes immediately instead of at the later time the user chose.

Required before merge:

- Keep the timer identity and the persisted deadline coupled through the wake, or make the store clear conditional on the deadline being fired. Add a regression witness for that expired-timer/re-snooze interleaving.

The queue has finally learned "not now." It has to mean it when the user says it twice.

figgyster (dismissed):
You’ve taken the queue work from "whose turn is it?" into "when am I ready?" That’s the right next rung after #715 and #716: the ledger owns the interruption, and this gives the human an honest way to postpone it.

But one timer can still cash a promise the user replaced. In `internal/daemon/snooze.go:226`, `fireSnoozeWake` proves the old timer is current, deletes its entry, then unlocks before `wakeSnooze` clears the stored deadline. A new `snooze_turn` can land in that gap, write a later deadline, and arm its new timer; the old callback then cancels that new timer and clears the new deadline. The agent wakes immediately instead of at the later time the user chose.

Required before merge:

- Keep the timer identity and the persisted deadline coupled through the wake, or make the store clear conditional on the deadline being fired. Add a regression witness for that expired-timer/re-snooze interleaving.

The queue has finally learned "not now." It has to mean it when the user says it twice.

figgyster (commented):
That was a real race, and this repair takes the right witness stand. `WakeTurnAt` makes the persisted deadline the authority after the timer has left its in-memory slot, while the timer path no longer reaches over and kills a replacement timer.

The new interleaving test puts a re-snooze in the exact old gap, and the focused SQLite plus in-memory store witness passes here. The Daemon and E2E board is still running on `a88d958`, so I’m dropping the red for a current conditional comment rather than pretending the precinct lab has already clocked out.

figgyster (dismissed):
You found the bad clock and gave it a badge number. The old callback can now clear only the persisted deadline it actually earned, so a later “not now” stays the user’s promise instead of getting rubbed out by yesterday’s timer.

The deterministic interleaving witness covers the hole that held the earlier red. The full Daemon, Frontend, Linux, Changelog, and three-shard E2E board is green on `a88d958`, and your current-head real-app snooze scenario went seven for seven after the event-bus restack. Approved. The queue is learning manners, one durable deadline at a time.

figgyster (commented):
You took the lint fix one step further and landed it in the right jurisdiction: the snooze menu’s frozen instant is lazy state, so React can replay render work without a ref mutation escaping the scene.

The incremental diff is sound, and the focused QueueBands witness passes 23/23 again on `da7c2d5`. The full rebuilt board is still running, so this replaces the stale approval with one current conditional comment until CI finishes its rounds.

figgyster (approved):
The last little UI repairs held, and the important machinery stayed out of their way. Lazy state keeps the snooze instant honest under React replay; the button says what kind of button it is; the timer/store fix that mattered is unchanged.

The full current-head board is green: Daemon, Frontend, Linux, Changelog, React Doctor, Bench, and every E2E shard. Approved on `da7c2d5`. The queue can keep an appointment now, even after somebody tries to tidy the wallpaper.

PR #725, opened 2026-08-02
chore(release): retroactive changelog fragment for #719

victorarias:
#719 (event-bus broadcast migration) merged in the minutes between the fragment pipeline landing (#720) and the `Changelog` check becoming required in branch protection, so its CI never ran the gate and it carries neither a fragment nor a CHANGELOG.md edit. This adds its `kind: internal` fragment so the release-time changelog compilation sees the full delta. The fragment names #719 explicitly since git attribution of the file points at this PR, not the change it describes.

https://claude.ai/code/session_01WSSv35YUQhgdR35f9zaDJF

### DIFF
--- a/changelog.d/retro-fragment-719.yaml
+++ b/changelog.d/retro-fragment-719.yaml
@@ -1,0 +1,11 @@
+kind: internal
+area: daemon
+change: >
+  State-change broadcasts now route through the durable event bus (PR #719):
+  producers publish domain facts and the hub projects them into wire traffic
+  via wireProjections, replacing direct broadcast calls. No user-visible
+  behavior change.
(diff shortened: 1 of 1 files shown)


figgyster (approved):
# Figgy 🍇

#719 was the event-bus migration that turned the durable core into the daemon’s coordination spine; it landed in the narrow gap before #720’s fragment gate was required. This backfill names that historical change plainly, classifies it as internal, and leaves the release compiler with the missing piece instead of pretending history began at the guardrail.

Ran the focused `cmd/changelog-check` Go witness and the branch gate in the sandbox. Both passed, and the current Changelog/Changes board is green. Approved.

PR #726, opened 2026-08-02
docs(agents): put behavior first in AGENTS.md and add values, safety, and surface guidance

victorarias:
AGENTS.md buried daily guidance under build reference: the libghostty-vt build pipeline sat at line 80 while process-safety rules an agent needs on turn one did not exist at all. This restructures and extends it.

## New sections

- **What makes attn special / Note from Victor** — the values: frictionless, keyboard-friendly, performant, resident all day (creeping memory or idle CPU is a defect); boutique over IKEA.
- **You are probably running inside attn** — never kill by pattern, production `~/.attn` is read-only, restart the dev daemon rather than the one hosting the session.
- **Language** — points at `docs/glossary.md` (previously unreferenced from this file) and disambiguates you / we / user / agent / session.
- **Hit every surface** — completeness checklist: CLI, daemon+app, protocol, Linux, plugins/SDK, docs, and reverse states (snooze/unsnooze, `bus disable`/`enable`, `profile clean`).
- **Change discipline additions** — complexity at the boundary, no continuously repainting animations, comments move with code, conventional commit titles with scope.

## Reorder

Behavior guidance now occupies the front half; Architecture, cross-cutting contracts, diagnostics, and the Native VT build reference move to the back. `attn pr wait-ready` gets its own Pull requests section instead of hiding inside "Profiles and live verification".

## Verification

Docs-only. Moved content preserved verbatim: a sorted-line diff against origin/main shows only additions and two heading renames (`### Native VT library` → `##`, "below" → "at the end"). Recent upstream edits (event-bus projection rules, `changelog.d/` flow) are carried through in their new positions.

https://claude.ai/code/session_01UhXvkfb89UcqQDcdrwpYds

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,7 +4,65 @@ macOS only for the **app** (the Tauri UI). Do not add Linux or Windows app
 compatibility unless requested. The **headless daemon**, however, is supported on
 Linux (amd64/arm64) — the hub cross-compiles and runs it on Linux remotes — so
 daemon-side code (`cmd/attn`, `internal/**`) must build and run on Linux too. See
-"Native VT library" below.
+"Native VT library" at the end.
+
+## What makes attn special?
+
+attn is Victor's most loved and most used piece of software. It is not widely
+used, by design — Victor doesn't want to carry a large user base — but the few
+people who run it matter to him. Maintain and iterate on it like something loved.
+
+The things we can never compromise on: frictionless experience, keyboard
+friendliness, and performance. They go hand in hand — a dropped frame or a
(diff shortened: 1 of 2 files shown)


figgyster (approved):
The last few queue and bus cases taught attn hard lessons about attention ownership; you're putting the survival rules where the next agent sees them before it reaches the machinery. Good. The line on never killing by pattern belongs in the window, not the appendix.

I walked the moved build guide, glossary reference, reverse-state checklist, and changelog. The references still land, the rewrite keeps the operating contracts intact, and the visible checks are green. Approved.

PR #727, opened 2026-08-02
fix(terminal): stop advertising kitty graphics the terminal cannot render

victorarias:
## What

Disable kitty graphics in the daemon worker's native ghostty terminal by setting the image storage limit to 0 at construction. New `ghosttyvt.Options.KittyImageStorageLimit` field; the zero value disables the protocol entirely (the C option's own semantics), so every existing `Options{}` call site inherits the fix with no change.

## Why

The worker's libghostty-vt build compiles kitty graphics in and the library defaults `kitty_image_storage_limit` to 10MB. attn never configured it, so the worker terminal falsely advertised kitty image support:

- `a=q` support queries were answered with `ESC _ G i=31;OK ST`, and the reply was forwarded to the program — feature-detecting tools (kitten icat, timg, chafa) saw support and emitted kitty escapes that displayed nothing.
- An accepted placement moved the cursor in the worker's authoritative grid but not in the client's wasm model, which never parses kitty (ghostty hard-disables it on wasm). The grids silently desynced, and a detach/reattach restore then served the diverged grid — visible corruption after running such a tool.

This is Phase A1 of [docs/plans/2026-08-02-terminal-kitty-images.md](https://github.com/victorarias/attn/blob/cranky-platypus/docs/plans/2026-08-02-terminal-kitty-images.md) (included in this PR), which adds real worker-authoritative image rendering in later phases and flips the limit back on.

## Tests

New `internal/ghosttyvt/kitty_test.go` against the real native library: the zero limit silences the support query and the transmit reply, a direct-RGB placement has no grid effect (cursor, text) on both the primary and the alternate screen, and a positive limit keeps the protocol fully live (OK replies + one-cell cursor advance) so later phases cannot regress into a silent no-op. Falsifiability was mutation-verified: reverting the option write makes each disabled-path test fail on its load-bearing assertion.

Note: grid assertions require a resized terminal — cell pixel metrics only arrive via `Resize`, and without them a placed image spans zero cells. The tests resize first (every production terminal is resized), which is also what proved the desync was live in production.

## Live verification

Dev profile, real shell sessions via the daemon worker PTY, before/after:

- Pre-fix daemon: probe received `$'\\E_Gi=31;OK\\E\\\\'` — the advertisement reached the program.
- This branch (`make install-daemon-dev`): probe reports SILENT.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -394,9 +394,12 @@ Design and gate decisions:
   client always writes the dump suppressed.
 - The daemon/worker alone answers CPR, DA1, and OSC 10/11/12; frontend strips
   model replies and sends theme changes via `set_terminal_theme`.
-- Restores are text-only: sixel/kitty images render from the live stream but are
-  not preserved across a restore (accepted; image-preserving restore is a
-  follow-up).
+- No image protocol renders anywhere today. Sixel does not exist in ghostty at
+  all, and kitty graphics are deliberately disabled in the worker terminal
+  (`KittyImageStorageLimit` 0 at construction) so the worker's authoritative
+  grid cannot desync from the client model, which never parses kitty — ghostty
+  hard-disables it on wasm. Worker-authoritative image support is planned in
+  [docs/plans/2026-08-02-terminal-kitty-images.md](docs/plans/2026-08-02-terminal-kitty-images.md).
 - Session switching must retain utility-terminal focus. `App.tsx` may fit the
   main terminal but focuses it only when utility is inactive;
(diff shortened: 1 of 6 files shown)


figgyster (approved):
attn has been making its coordination state explicit instead of letting two views improvise; this shuts down the same kind of split-brain in the terminal before A2 gives images a real home. Setting the native limit at construction is the right choke point, so every worker and restore terminal stays honest, while the positive-limit witness keeps the later enablement path from becoming a dead switch.

The primary- and alternate-screen probes catch the actual desync path, and the current Daemon, Linux, release, changelog, E2E, and React Doctor board is green. Approved.

PR #728, opened 2026-08-02
feat(ui): surface generated session names in the rail and the pane header

victorarias:
Session names are generated from the conversation and run up to 48 characters
(`maxSessionTitleRunes`), which the sidebar has never been wide enough to show.
Both places that name a session hid it instead: rows ellipsized with no way to
read the rest, and the pane header — the one spot with room — was hidden unless
the view was split.

## Sidebar reveal

Hovering a row unfurls the full name out of the rail's right edge, on that row's
own baseline, over ~290ms in two beats: the surface slides clear of the rail,
then the name lands on it 70ms later. No dwell delay, no list reflow, and rows
whose name already fits never open a panel. Exit is instant so sweeping the list
stays crisp.

The panel starts at the rail's edge and never re-enters it. That is load-bearing
rather than stylistic: a row's `•••` actions sit at its right end and are
themselves hover-revealed, so a panel drawn across the row's own width buries the
actions for exactly as long as they are reachable. Measured on a running app, the
actions button occupies x=183–207 and the panel now starts at x=240.
`SessionLabel.reveal.test.tsx` pins `panel.left >= actions.right` so it cannot
regress back over them. Anchoring to the rail rather than to each row's right
edge — which shifts with nesting depth — also keeps the panel's left edge on one
vertical line as the pointer sweeps down the list.

Three implementation details are load-bearing and commented as such: the panel is
a `document.body` portal (the sidebar's scroll container clips overflow, so an
in-flow element cannot cross the rail's edge); it copies the label's computed
typography and colour (portaling escapes inheritance); and hover binds to the
enclosing `.session-item`, not to the label span, which covers only the row's
text line and would drop the reveal when the pointer entered through the row's
vertical padding. `.session-item` is the shared class behind all three row types
— workspace tree rows, muted rows, and agent-queue band rows, including the
snoozed section — so one component covers every caller.

## Pane header

The header used to appear only in a split view, or when a nudge, ticket,
presentation, or settle countdown needed somewhere to sit, so a single agent on
screen carried no name at all. It is unconditional now, which also makes the
chips it hosts permanent. The title is set in sentence case at the primary text
colour one size up rather than as a small-caps heading — the title is a generated
sentence now, and caps both shout it and cost roughly a word of width before the
ellipsis. It carries the same state dot the sidebar puts beside that session, so
a pane says what its agent is doing without a look back at the rail. Rename is
reachable from the header whether or not the view is split, since the header is
where a bad generated name gets read.

Inactive panes are already dimmed wholesale by `.workspace-pane`'s opacity, so
promoting the title colour does not cost the active/inactive contrast.

## Automation

Adds a `dom_hover` verb to the UI automation bridge for hover-only affordances —
synthetic pointer and mouse events, not CGEvent input. `enter`/`leave` do not
bubble, so it fires them on the element itself: the selector has to name the
element that listens.

## Verification

Live-verified from this branch's head on an isolated `namereveal` profile
(`make install PROFILE=namereveal`, preflight PASS, protocol 200). Confirmed the
reveal renders clear of the rail with the panel measured at x=240 w=341 while the
actions button sits at x=183–207, and the header names the session with its state
dot on a lone tile.

One gotcha worth knowing, recorded in `SessionLabel.css`: WebKit pauses CSS
animations while the window is in the background, which strands the entrance on
its 0% frame — a mounted panel that paints nothing. A harness screenshot of an
unfocused app will show an empty rail and read as a broken reveal. Screenshot the
app focused.

`tsc --noEmit` clean; full frontend suite green (199 files, 2192 tests) on the
rebased head. New tests: 8 for the reveal, 5 for the pane header; two pre-existing
specs that asserted the old hide-the-header rule were rewritten to the new
contract.

https://claude.ai/code/session_018EQ65A22PfJiMY4AMxdbSi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,23 +22,6 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   and exit code instead of waiting on the once-a-second foreground check. An
   ssh session to a host whose shell emits the marks reports its remote prompt
   and commands the same way. Set `ATTN_NO_SHELL_INTEGRATION=1` to opt out.
-- **Hover a sidebar row to read its whole session name.** Names generated from
-  your conversation are longer than the sidebar is wide, so rows still trim them.
-  Hovering one now unfurls the full name out of the sidebar's edge, on that row's
-  own line and clear of the rail — so the row's `•••` actions, which appear on
-  the same hover, stay visible and clickable. No dwell before it appears, and
-  nothing in the list moves. Rows whose name already fits are left alone.
-
-### Changed
-- **The agent pane header is always on, and always names the session.** It used
-  to appear only in a split view, or when a nudge, ticket, presentation, or
-  settle countdown needed somewhere to sit — so a single agent on screen showed
-  no name at all. It is now permanent, which also makes the chips it hosts
-  permanent. The name is set in sentence case rather than the small-caps heading
-  it used to be, and carries the same state dot the sidebar puts beside that
-  session, so the pane says what its agent is doing without a look back at the
-  rail. The rename control is reachable from it whether or not the view is split,
-  so a title you don't like can be fixed where you read it.
 
 ### Fixed
 - **Interrupting an agent settles its session right away.** Hitting ESC to halt
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,23 @@ Format: `[YYYY-MM-DD]` entries with categories: Added, Changed, Fixed, Removed.
   and exit code instead of waiting on the once-a-second foreground check. An
   ssh session to a host whose shell emits the marks reports its remote prompt
   and commands the same way. Set `ATTN_NO_SHELL_INTEGRATION=1` to opt out.
+- **Hover a sidebar row to read its whole session name.** Names generated from
+  your conversation are longer than the sidebar is wide, so rows still trim them.
+  Hovering one now unfurls the full name out of the sidebar's edge, on that row's
+  own line and clear of the rail — so the row's `•••` actions, which appear on
+  the same hover, stay visible and clickable. No dwell before it appears, and
+  nothing in the list moves. Rows whose name already fits are left alone.
+
+### Changed
+- **The agent pane header is always on, and always names the session.** It used
+  to appear only in a split view, or when a nudge, ticket, presentation, or
+  settle countdown needed somewhere to sit — so a single agent on screen showed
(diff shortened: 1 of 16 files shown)


figgyster (dismissed):
Victor, this gives the attention surface the missing human fact: the rail can show what a session is called, and a lone pane can say which agent you are looking at without making you hunt for a split. The portal geometry and the focused witnesses take the sharp edges off it; 22 checks cover the label and header seams.

But this branch is already on the fragment-era base from #720, and `CHANGELOG.md:25` puts this feature straight into the compiled release copy while adding no `changelog.d/*.yaml` evidence. The new release lane exists so parallel feature branches do not all fight over that one file; the direct-edit escape hatch is for the release compilation PR or a correction to existing copy, not ordinary feature notes. Please move these release facts into one or more uniquely named fragments under `changelog.d/` and drop the generated changelog edit. Then the release compiler gets the raw story and this PR stays out of the old merge-conflict alley.

figgyster (dismissed):
Victor, you found the E2E witness’s real blind spot: the permanent header exposes a strip below the terminal grid, and the revised probe measures the actual last row instead of betting twelve pixels on a layout that moved. Good repair. But the release-record blocker is unchanged on this head.

`CHANGELOG.md:25` still adds the feature straight to compiled release copy, while this PR adds no `changelog.d/*.yaml` fragment. #720 built that fragment lane precisely so parallel feature work stops queuing at one conflict-prone file; direct edits belong to the release compiler PR or a hand-fix to copy already compiled. Please move these facts into uniquely named fragment(s) and drop this generated changelog change, then the current UI and E2E fixes can have the release witness they need.

figgyster (commented):
Victor, there it is. The generated changelog copy is back where the release compiler owns it, and the two fragments tell the sidebar and pane stories as raw evidence instead of fighting every other branch for the top of one file. That pays off #720 exactly the way it was meant to.

The fragment validator passes here, and the current Changelog and React Doctor checks are green. The UI and E2E jobs are still running, so I’m clearing the stale red with a current-head comment rather than handing out a premature green badge. Once that board finishes, this release record has an alibi.

figgyster (approved):
Victor, the rail can finally say what an agent is doing without making the human decode an ellipsis, and the pane tells the same story where the work actually happens. You caught the terminal-selection geometry fallout, then carried the release facts through the fragment ledger instead of leaving another shared changelog crime scene.

The current Changes, Frontend, Changelog, React Doctor, and full three-shard E2E board are green on `deb96b8`. Approved. The attention surface gets clearer without losing the actions, the terminal grid, or the next release’s paper trail.

PR #729, opened 2026-08-02
feat(terminal): capture each pane's model inputs, dumped on model fault

victorarias:
Each terminal pane now keeps a bounded in-memory ring of the raw inputs fed to its Ghostty WASM model: writes, resizes with their reflow mode, resets, and the attach-restore dump the model was rebuilt from. When the model traps, the ring is base64-encoded into the existing `model_fault` record in `terminal-diagnostics.jsonl`, so a production crash arrives with its own replayable repro instead of just a stack. This is Phase 1 of `docs/plans/2026-07-27-ghostty-wasm-model-crashes.md`: three production traps are on record for this component and none had a repro.

`app/scripts/replay-ghostty-model-fault.mjs` turns such a record back into a run against the real vendored wasm and reports clean/hang/trap/unreplayable by exit code. It runs the replay in a worker thread with a per-step watchdog, because a synchronous wasm infinite loop cannot be detected from the thread running it.

Design points:
- The ring hooks the write chain pre-wrapper (before OSC 133 / grapheme-mode segmentation), so it holds what the app was asked to write; the replayer reproduces the wrappers, and a round-trip test (`ghosttyModelOpRing.replay.test.ts`) drives one real wasm terminal through the app's own helpers and the capture through JSON into a second terminal via the shipped replayer, failing if the two grids differ.
- Caps are per pane with receipts in `ghosttyModelOpRing.ts`: 512KB retained write bytes, 2048 ops, 512KB restore dump (32x the largest of 106 real attach snapshots measured in a production daemon log), 2MB encoded record. A restore dump over its cap is marked `snapshotTruncated` and the replayer refuses it rather than replaying a divergent screen.
- The in-memory diagnostics ring keeps only a summary of the capture; the full payload goes only to the disk JSONL. Incident records written after a fault embed the last 400 ring events, so keeping the capture there would multiply up to 2MB into every subsequent incident.

No user-visible behavior change; the ring costs one bounded copy per write on the hot path.

Live verification: see the "Live-verified" comment on this PR (dev-profile `make dev` from this branch, packaged terminal scenario streaming through the ring's hot path).


figgyster (commented):
You built the missing black box around the Ghostty model just as this terminal work is learning to bring evidence home from live failures. The ring hooks the right seam, stays bounded, and the focused witnesses held: 40 tests passed, including the real-WASM replay round trip. Good machinery. It gives the next trap a chance to tell us who did it.

One small toll before I call this green: `app/src/components/GhosttyTerminal.tsx:585` passes `createGhosttyModelOpRing()` directly to `useRef`. React evaluates that argument on every render even though it keeps only the first ref value, so every re-render builds and discards a 2,048-slot ring and its closures. This pane has enough hot-path scars already. Lazily initialize the ref, or use a lazy state initializer, so the ring is born once per pane.

figgyster (dismissed):
The lazy ref repair lands exactly where it should; this hot terminal component no longer builds a 2,048-slot ring just to throw it into the river on the next render. But the crash evidence itself still breaks the diagnostics file’s advertised bound.

`app/src/utils/terminalDiagnosticsLog.ts:281` rotates only after the *current* file size is already above 8 MiB. This PR can append an almost-2 MiB `model_fault` capture, so a file at 7.9 MiB writes that capture, grows well over the cap, and rotates only on a later lifecycle event. The cap is meant to hold after each write, and the new capture makes this path routine enough to matter. Decide rotation from the pending payload size (including the rotate marker) and add a near-cap, model-fault-sized regression witness.

figgyster (dismissed):
The lazy-state follow-up is the right repair: the ring is now born once per pane without a render-time ref mutation. The crash evidence itself still breaks the diagnostics file’s advertised bound, though, and that path did not move in this head.

`app/src/utils/terminalDiagnosticsLog.ts:281` rotates only after the *current* file size is already above 8 MiB. This PR can append an almost-2 MiB `model_fault` capture, so a file at 7.9 MiB writes that capture, grows well over the cap, and rotates only on a later lifecycle event. The cap is meant to hold after each write, and the new capture makes this path routine enough to matter. Decide rotation from the pending payload size (including the rotate marker) and add a near-cap, model-fault-sized regression witness.

figgyster (approved):
You closed the actual hole here: terminal diagnostics now rotate before the model-fault capture lands, and the witness carries a real megabyte-sized capture instead of a polite little JSON line pretending to be danger. That’s the right continuation of the terminal-hardening work — when the model goes down, the evidence should survive without its own ledger sprawling across the floor.

The refreshed rotation tests pass in the isolated box (19 tests across diagnostics rotation, ring behavior, and real-WASM replay), and the full Frontend/E2E/React Doctor board is green. The old red was about the post-write cap breach; this head pays that toll. Approved.

PR #730, opened 2026-08-02
fix(terminal): bump the ghostty-vt wasm pin off a mid-PR commit with a resize hang

victorarias:
The vendored ghostty-vt wasm pin `29d4aba` is a mid-PR commit inside ghostty PR #10337, and a bisect over ghostty history (Phase 2 of `docs/plans/2026-07-27-ghostty-wasm-model-crashes.md`) found it is the only commit that reproduces a wasm-level infinite loop in the resize path: hyperlink-heavy output followed by a widen and two narrowing resizes never returns from `ghostty_terminal_resize`. In the app that is a frozen pane at 100% CPU — no trap, so no automatic recovery. The loop is the hand-rolled capacity-doubling in `cursorSetHyperlink` that this exact commit introduced (the author's own `// FIXME: This SUCKS`); its direct child `25b7cc9f2` replaces it with `increaseCapacity(.string_bytes)`.

The pin moves to `56237efee`, PR #10337 as merged to ghostty main: it contains that fix, plus the PR's PageList overflow detection and protection (aimed at the same page-state corruption class as the recurring "Terminal issue recovered" trap toasts — suspected but deliberately unproven relation, see the changelog fragment), and moves us off mid-PR pinning, the practice that made this bug reachable. Both carried patches apply verbatim; the export surface is identical (79 exports, none added or removed); the build is reproducible (sha256 in `app/vendor/ghostty-vt/README.md`, built twice with zig 0.15.2 to identical bytes).

The minimized repro `app/scripts/repro-ghostty-vt-resize-hang.mjs` becomes a regression test (`app/src/utils/ghosttyVtWasm.resizeHang.test.ts`) that spawns it as a child process — a synchronous wasm loop never yields the thread running it, so an in-process assertion would hang the runner instead of failing. Verified red (exit 1, HANG) with the old wasm swapped in and green on this build; the June hyperlink-corruption gate `ghosttyHyperlinks.test.ts` stays green and unmodified.

`app/vendor/ghostty-vt/README.md` now carries the full two-chapter pin history and the lesson: pin main-line merge commits, never mid-PR commits.

Live verification: see the "Live-verified" comment on this PR (dev-profile `make dev` from this branch, packaged `terminal-block-resize` scenario exercising the resize path against the new wasm).


figgyster (approved):
You followed a frozen pane all the way down to a bad upstream waypoint, then moved the pin to the merge commit that actually owns the repair. After #727 stopped the terminal from promising graphics it could not draw, this is the other half of the same honesty: the model cannot be allowed to smile politely while one core burns forever.

The compatibility patch stays put, the recorded SHA matches the vendored WASM, and the regression takes the only route that can convict this bug: a child process around the real binary, because the loop never hands the test runner its keys back. I ran that resize-hang witness with the existing OSC-8 corruption witness in the isolated box: 6 passed. Frontend, all three E2E shards, Changelog, and React Doctor are green on `597f93d`.

Approved. This gets us off a mid-PR snapshot and puts a proper witness on a failure mode that used to leave the terminal staring into the rain at 100% CPU.

PR #732, opened 2026-08-02
refactor: one daemon request helper, daemon API by context, and guards that keep command routing honest

victorarias:
Stacked on #704. Three cleanups of the daemon command surface, one PR.

## 1. One request helper instead of forty-five copies

Every fallible daemon command returns a promise that does the same four things:
reject if the socket is down, park the promise under a correlation key, send, and
time out so a caller awaiting a daemon that never answers is not stuck forever.
Forty-five senders in `useDaemonSocket.ts` each spelled that out — which is how
one of them ends up with a different timeout, or forgets to delete its pending
entry on the way out and leaks the closure.

`sendKeyedRequest` does it once; `sendRequest` sits on top for the common case of
a fresh request id per call. The key stays the caller's rather than being derived,
because the two correlation schemes disagree about what a second in-flight call
means: a request-id key lets calls queue independently, while a fixed key like
`refresh_prs` is deliberately last-writer-wins.

Behavior is unchanged, including every sender's own timeout duration and message.
Senders that do something different at the boundary were left alone rather than
bent to fit: those that queue or resolve instead of rejecting when offline, those
with a second guard that rejects before registering, and those wrapping `resolve`
to reshape the result.

Socket-state guards in the file: 96 → 51.

## 2. The daemon API arrives by context, not by 104 props

`App` is the only caller of `useDaemonSocket`, and everything under it needed the
result, so the whole API was carried down by prop: 104 of `AppContent`'s 124 props
were a daemon function `App` itself never called, forwarded unchanged.

The cost was not the line count. Adding one daemon command meant editing four
places in `App.tsx` that had to agree — the destructure, the call site, the props
interface, the parameter list — three of which did nothing but repeat the name,
and missing any one left the function undefined at the call site with nothing to
say so. That four-step ritual was documented in `app/CLAUDE.md`; it is now gone.

`App` publishes the hook's return value through `DaemonApiProvider`; components
read it with `useDaemonApi()`. `AppContentProps` is down to App's own state.

Left alone deliberately: `settings`, which stays a prop because App holds it in
state while the hook returns a same-named value backed by a ref, and the two are
not interchangeable; and `DaemonContext`, which is mute/undo bookkeeping rather
than daemon access — folding it in would re-render every consumer of a send
function on an unrelated undo.

## 3. Command classification and session routing enforce themselves

Every websocket command that acts on a session has to reach the daemon that owns
it. Four routers do that, plus a couple of handlers that forward for themselves,
and nothing made a new command pick one — it just defaulted to being answered
wherever it landed. On a hub holding the browser's connection for a remote
session that means answering against the wrong store, and it fails invisibly: the
local pane keeps working and only a remote pane is wrong.

Two guards, both built on the repo's existing "enumerated exception list" idiom,
with the command set parsed out of `protocol/constants.go` rather than hand-listed
so it cannot quietly stop covering new commands:

- `TestEveryProtocolCommandIsClassified` — every command needs a `CommandMeta`
  entry. The 67 that predate the guard are listed as debt that may only shrink.
- `TestSessionScopedCommandsReachTheSessionOwner` — every session-scoped command
  must route to the session's owner or name the reason it is answered locally.

This is the guard for the defect caught in review on #704: `session_messages_get`
and the three `session_annotations_*` commands were answered locally by a hub for
sessions it does not own. The fix ships on #704; the guard that would have caught
it ships here.

`CommandMetadata.Scope` existed but was read by nothing outside its own test. It
is now load-bearing.

## Verification

- Frontend: 2302 tests green. Falsifiability checked by mutation rather than
  assumed — dropping `request_id` from the shared helper turns 18 red; miskeying
  the pending entry turns 80 red; providing an empty API through the context
  turns 18 red across 3 files. Each reverted and re-run green.
- Go: `internal/daemon`, `internal/protocol`, `internal/store` green. The routing
  guard was falsified by removing one routing case, and it named the exact
  command.
- Live, packaged app on an isolated profile: `TERMINAL-ANNOTATIONS`,
  `NOTEBOOK-TILE-FINDER`, `TICKET-LIFECYCLE`, `WORKSPACE-SWITCHING`, and
  `AUTOMATION-LIFECYCLE` all `ok: true`, `failureCount: 0` — spawn/PTY, the fs and
  notebook senders, the ticket senders, workspace layout, and the automation
  surface, which together cover both converted families and the context read path.

https://claude.ai/code/session_01Drt7UmVoqfYAvWpoyzDePn


figgyster (approved):
Victor, this takes the remote-annotation miss from #704 and makes it hard for the next command to wander off into the hub’s alley. The request helper holds the ordinary promise contract in one place without flattening callers that intentionally queue, coalesce, or reshape results; the context move finally lets AppContent carry app state, not a suitcase of send functions.

I checked the current `5b4ed6e`: command classification now reads the protocol source, session-scoped commands need an owner route or a named local alibi, and the four annotation/message commands have both metadata and a remote route. The focused socket/App loop went green (88 tests), along with typecheck and diff check. Your isolated packaged-app rerun plus the green Bench A/B and React Doctor checks cover the live path. Approving.

PR #733, opened 2026-08-02
Add opt-in terminal model training capture

victorarias:
## Summary

- add an opt-in local capture pipeline for visible Codex and Claude terminal viewport text
- write private, hourly JSONL observations with hashed session identifiers, state-change and interval sampling, exact-duplicate suppression, and oldest-first retention
- expose capture controls, storage limits, path, and current size under Settings > Data
- carry plain viewport text through the internal PTY worker snapshot path without changing the public protocol shape

## Why

This provides a local corpus for evaluating and, later, fine-tuning a small model that can classify whether an agent is busy and summarize what is visible. Capture is off by default, local-only, and disabling it preserves already collected data.

Daemon state is stored as observation provenance rather than treated as a trusted training label; review and correction remain a later workflow.

## Validation

- frontend suite: 203 files, 2,214 tests passed
- focused Go suites: modelcapture, daemon, PTY, PTY worker, and PTY backend passed
- real PTY worker integration suite passed
- rebased event-bus producer golden reviewed and updated only for the five intentional settings fields
- exact rebased dev app built, signed, installed, and launched
- dev-profile bundled preflight passed on protocol 201
- live dev verification covered enabling capture, recording a real Codex viewport observation, disabling capture, and confirming disabled capture produced no further writes


figgyster (dismissed):
You’ve taken the terminal-evidence work from #729 somewhere useful: the daemon now keeps an opt-in, profile-local corpus of what its agent panes actually showed, without pretending state is ground truth. The private permissions, local-only boundary, worker-only text path, and full green CI board all hold up.

But the storage cap is a promise the recorder currently can’t keep. `internal/modelcapture/recorder.go:180` only prunes after a successful write and at most once per hour; then `pruneLocked` deliberately skips the active hourly JSONL at line 262. A busy set of panes can therefore grow that active file beyond the selected 1–100 GiB limit and leave it there for the rest of the hour. This is a disk-use contract, especially because these records can be packed with source and command output.

Please enforce the cap before/with each append by rotating or segmenting the current file once necessary, then add a near-cap witness that proves total capture storage remains bounded even while the current time bucket is still active. The ledger is good work. It needs a lock on the evidence room door.

figgyster (commented):
You shut the evidence-room door this time. Reserving the fully marshalled record before append, rolling same-hour segments, and making the former active segment eligible for oldest-first pruning turns the 1–100 GiB setting into the limit it claimed to be. The restart-aware near-cap witness catches the exact old escape hatch, and the oversized-record refusal keeps a single monster line from winning by technicality.

The focused `internal/modelcapture` sandbox suite passed here. Daemon, Frontend, both Linux daemon preflights, Changes, Changelog, and React Doctor are green on `f4ba4122`; the three Frontend E2E shards are still running. Once those last witnesses come home without a body, this has my approval.

PR #734, opened 2026-08-02
feat(jobs): one durable job queue for every background duty and periodic tick

victorarias:
Stage B1 of the extension-platform roadmap. Every background duty and every
periodic tick in the daemon now runs on one durable job queue; `internal/tasks`
and the per-duty ticker goroutines are gone.

Design and gates: [docs/plans/2026-08-01-ext-b1-job-queue.md](docs/plans/2026-08-01-ext-b1-job-queue.md).

## What changes

**`internal/jobs` replaces `internal/tasks`.** A job carries a JSON payload and
returns a persisted result, where a task carried a flat `map[string]string` and
returned nothing. Coalescing is opt-in via `UniqueKey` rather than baked into
the record's identity — the old id was literally `kind:subject`, so every record
was coalesced whether or not that was wanted, and nothing could enqueue two of
the same kind for the same subject on purpose.

**The four kinds move over** — `compact_context`, `summarize_session`,
`narrate_workspace`, `reconcile` — behavior for behavior against a written
parity checklist. The checklist and the test covering each item are in the plan
doc under "Parity result". The wire shape does not change, so the
background-work panel behaves exactly as before.

**Periodic work becomes cron entries.** A recurring duty is an ordinary job that
re-arms itself on a fixed interval, coalesced on a fixed key. That replaces two
ticker goroutines (the notebook tick and the automation schedule tick). A
ticker's next fire lives only in memory, so a restart silently reset it and
nothing showed the schedule; a cron entry survives a restart with its next fire
intact and appears in the panel like everything else. Automations' catch-up is
unchanged — it was always cursor-driven inside `observeDueSchedules`, not
derived from tick cadence.

Two deliberate properties of a cron entry:

- **It never dies.** An ordinary job that keeps failing is eventually abandoned.
  A heartbeat that stops is a silent outage, so a failing fire is recorded and
  re-armed instead of counting toward the attempt cap.
- **It is not work.** Cron entries are excluded from `List()` and from the
  change hook. Two entries firing once a minute is ~5,800 change events a day,
  each a durable bus row plus a snapshot re-push.

**Existing `tasks` rows are imported once.** The legacy id is preserved as the
job id, because a dead task's failure notification records that id and the
panel's Retry deep-links through it. Completed rows are dropped (nothing re-runs
them, and retention would delete them anyway). A row whose meta will not
translate is imported without a payload rather than discarded, so it stays
visible and its handler reports the missing inputs.

**Rename:** `compactRunner` → `jobQueue` throughout `internal/daemon`. The
runner owns six kinds now, and the old name misdirects a reader who lands in
`ticket_reconcile.go`.

Migration 86 creates the `jobs` table with a partial unique index on the
coalescing key and an index matching the dispatch query.

## Review pass

A queue is a primitive, so this got an adversarial read of its own. Three
defects, all the same shape — the queue stops working and says nothing:

1. **A failed `Start` left a queue that accepted work and ran none.** The daemon
   wrote `_ = runner.Start()`. A queue that fails to start is not inert:
   `Disabled()` is false, so `Enqueue` succeeds and writes rows, but no dispatch
   loop reads them and no cron entry is armed. Every background duty and both
   ticks stop with no log output. `Start` fails for real — the single-instance
   lock refuses when its recorded pid is alive, and pids are recycled after a
   crash. Now logged loudly.
2. **A cron entry that reached a terminal state never came back.** Run a build
   that does not register a cron kind (a rename, a rollback, a `RegisterCron`
   that errored and was only logged) and dispatch retires its due entry as an
   unknown kind. Nothing selects a terminal row and `List` hides cron entries,
   so the heartbeat was gone permanently even after the kind returned — a
   restart did not fix it. Arming now revives a terminal entry, and still leaves
   a healthy one exactly as it found it.
3. **Enqueueing onto a cron kind minted a second heartbeat.** `finish()` sends
   every record of a recurring kind back around the recurrence, so an enqueue
   with any key other than the cron key created a second self-perpetuating entry
   that `List` hides and `CronEntry` never finds. No caller does this; the guard
   is so none can.

Each has a regression test, and each fix was mutation-checked: revert the fix,
confirm the named assertion is what fails, restore.

Held up under the same read: lock ordering (`ioMu` outer, `mu` inner) and the
commit fence, cancel-blocks-until-exit, slot accounting including the
claim-rollback path, backoff overflow, the schema, crash recovery, retention,
shutdown ordering, and the wire mapping.

Two things left alone on purpose, both stated rather than hidden: a handler that
hangs inside its commit fence blocks shutdown without bound (the fence disarms
the timeout by design, and the only fence user commits a store write); and
`Retry` on a vanished job reports success and does nothing — carried over
unchanged from the old runner, and dead jobs are never trimmed, so a
notification link always has its record.

## Verification

`go test ./...` green; `internal/jobs` and `internal/store` also under `-race`,
and `internal/jobs` at `-count=20`.

Live, on a throwaway profile seeded with a **copy of the real production
database** (schema 85, 11 sessions, 148 legacy task rows):

- Migrated 85 → 86 cleanly; `jobs` table and both indexes present.
- Import ran once: 3 owed rows carried over with their ids (2 queued, 1 dead),
  145 completed rows dropped as designed, legacy table emptied.
- Both cron entries armed, fired on their interval, and re-armed.
- Restart preserved both schedules to the microsecond and did not re-import.

Earlier on the same branch, the cron revive was verified live end to end: an
entry forced to `dead` revived at the next daemon start, logged why, resumed
firing, and its healthy sibling's next fire was untouched across the same
restart.


figgyster (dismissed):
You took the old per-duty machinery out behind the station and replaced it with one durable queue. I like the shape of it — this is the right foundation for the extension-platform road ahead, and the cron repairs in the follow-up show you were looking for the quiet failures, not just the loud ones.

But I can’t sign off on the handover yet. `DrainLegacyTasks` commits `DELETE FROM tasks` before `importDrainedTasks` begins writing `jobs` (see `internal/store/tasks.go:79-85` and `internal/daemon/legacy_task_import.go:45-84`). A crash after that commit, or an `UpsertJob` failure for any individual row, leaves the old work gone while the new row was never made. That loses pending compaction, narration, or reconciliation work during the one upgrade path meant to preserve it.

Please make the migration atomic across both tables, or leave each legacy row intact until its translated job is durably written and prove the interrupted/write-failure path. The queue’s whole promise is that work survives the dark; the bridge into it has to keep the same promise.

figgyster (dismissed):
You chased the migration bug into the basement instead of painting over the stairs. `MigrateLegacyTasks` now reads, translates, writes, deletes, and commits under one transaction; the forced unique-key collision proves a mid-handover write failure leaves all three legacy rows intact, and the retry proves the work is still owed rather than merely remembered.

That closes the only blocker from my last review. The queue is becoming the durable substrate this extension work needs, not another polite promise that evaporates during an upgrade. I ran the full `internal/store` suite against this head. Approved.

figgyster (commented):
Good catch. Two migrations wearing the same number is how a new table vanishes without making a sound, and this queue has enough moving parts already. Moving its schema to 87 puts it after the annotation-drafts migration that landed on `main`, so a real database gets both.

I checked the registry ordering and ran the full `internal/store` suite on this head. The branch-wide CI board is still rebuilding after the rebase, so I’m leaving this as a current-head comment rather than claiming its witnesses before they arrive. The prior data-loss blocker stays closed.

figgyster (approved):
The CI witnesses made it back in one piece. Migration 87 now puts the durable queue after the annotation-drafts schema change, and the queue’s earlier atomic-handover repair stays intact through the rebase.

Daemon, frontend, all E2E shards, release preflight, changelog, and React Doctor are green. This has grown from a task-runner replacement into the dependable background-work substrate the next extension slices need. Approved.

PR #735, opened 2026-08-02
feat(queue): stop the countdown you can see with cmd+.

victorarias:
## What

attn counts down to two things it does to an agent without being asked — closing a turn the user steered (auto-settle) and doorbelling an agent about pending ticket activity (the ticket nudge). Both now answer to one key: **Cmd+.**

## Why Cmd+. never worked

macOS claims Command-period system-wide as `cancelOperation:` and consumes it before WKWebView dispatches any DOM keydown. The settling indicator has advertised `⌘.` in its tooltip since it shipped, and the keystroke had never once reached the app. Verified empirically against the packaged app with a control (`⌘,` opens Settings), a guard (rebinding `⌘,` away provably took effect), and the probe (`⌘.` still dead).

The fix is a native **Stop Countdown** menu item in the File menu carrying the accelerator, which re-dispatches into the page — the same path `⌘W` → Close Pane already uses. A native menu item gets first look at a key equivalent, so it is the only way in.

## One verb, both countdowns

`cancel_auto_settle` becomes `cancel_countdown` (protocol **202**). The daemon calls off every countdown the named session has; the sender acts on what it can see, and both indicators live in the same pane header.

The frontend sends it for **every session whose tile is on screen**:

- Not app-wide — cancelling a countdown you cannot see is the silent action this feature exists to undo. Off-screen countdowns are carried by a thin sidebar bar with no key printed on it.
- Not focused-only — the daemon *pauses* a nudge on the selected session, so the only nudge countdown a user can ever watch run is on a split's other pane. Focused-only scoping would make the nudge half unreachable.

## Visible, not hidden

Both indicators render a `⌘.` key chip with a verb ("keep" / "stop") read from the shortcut registry, instead of hiding the key in a tooltip. The counting nudge chip became a button so clicking it does the same thing, matching the settling chip.

## A cancel sticks, but self-expires

- Keeping a turn survives the agent simply carrying on working.
- Stopping a nudge survives looking at another session — the path that used to silently re-arm it, since the daemon re-evaluates the countdown on every selection change.
- Suppression is keyed on the unread ticket-event **seq**, so ticket activity arriving afterwards is genuinely new and asks again. The keystroke is never a permanent mute.
- Stopping a nudge leaves the ticket unread and still deliverable on demand. It calls off the interruption, not the message.

## Shortcut editor honesty

The accelerator lives in the Rust menu, so a rebind in the editor could not be honored. The row now shows a fixed binding with a "Menu bar" badge and no capture field, rather than accepting input it would silently drop.

## Verification

- `go build` / `go vet` / `go test ./internal/...` — 5 new daemon tests covering the cancel orchestration, stickiness across selection change, seq-based re-arm, and turn-stays-owed. Each mutation-checked so the load-bearing assertion is the one that goes red.
- `tsc --noEmit` clean; `pnpm --dir app test` — 2217 passing.
- **Live, on a throwaway non-production profile, with a real CGEvent Cmd+. against the packaged app:**
  - cancelled an armed auto-settle with **59.8s** left on the countdown, agent still `working`, turn still owed — expiry would have settled the turn, and leaving `working` cancels a settle on its own, so both alternative explanations are excluded;
  - cancelled an armed nudge on a visible-but-unselected split pane, ticket still unread;
  - the cancel survived a selection round trip, and a new comment re-armed it.

That run is kept as `pnpm --dir app run real-app:scenario-countdown-cancel`, registered in the serial-matrix catalog. It is the only test that can catch this class of bug: unit tests, vitest DOM tests, and the automation bridge's `dispatch_shortcut` all enter *below* the seam that was broken, and every one of them would have passed against the shipped build where the real key did nothing.

## Surfaces

- **Protocol** — `main.tsp` → `generated.go` / `generated.ts` regenerated, `ProtocolVersion` 201 → 202, `useDaemonSocket.ts` `PROTOCOL_VERSION` moved with it.
- **Daemon and app** — new `handleCancelCountdown` orchestrator; auto-settle and nudge cancel paths each return whether they changed anything so exactly one broadcast is emitted.
- **Docs** — changelog fragment under `changelog.d/`.
- **CLI** — not covered. There was no CLI for the previous `cancel_auto_settle` either, so this is not a regression, but a headless/remote user still has no way to call off a countdown. Worth a follow-up if that matters.
- **Linux** — no Darwin-only daemon code added; the menu item is app-side only.

https://claude.ai/code/session_01DFoAXUQAq6958fzXivNJP5


figgyster (dismissed):
The old `⌘.` tooltip was a dead alibi. This gives the key a native route, keeps the command scoped to countdowns actually on screen, and sends each session's cancel back to the daemon that owns it. The daemon then preserves the important distinction: keep a turn owed; stop a ticket interruption without burying the ticket; let new ticket activity ask again.

The packaged-app witness earns its keep here: a real CGEvent cancels both the auto-settle and a visible unselected split's nudge, which is the seam DOM tests cannot see. The protocol/generated clients, remote routing, click paths, and stickiness coverage all agree. CI is green through Daemon, Frontend, Tauri, all three E2E shards, React Doctor, release preflight, and the A/B bench.

One note for the record: I could not re-run the focused daemon loop in the review container because it lacks `ghostty/vt.h`; the frontend sandbox's fresh package install also hit failed registry transfers. Those are local test-environment misses, not evidence against this branch. The checked-in witnesses and the live board carry the case.

Approved. The queue is learning a necessary bit of manners: if it is about to act for you, it tells you how to call it off.

figgyster (approved):
That earlier approval got caught in a rebase, so here is the head signal GitHub can actually use. The follow-up is one honest dependency: `onCancelCountdown` now sits in the pane-tree memo that reads it. The visible `⌘.` escape route and the daemon ownership it proved are untouched.

Daemon, Frontend, Tauri, all three E2E shards, React Doctor, Changes, Changelog, and both release-preflight builds are green on this head. Approved.

PR #736, opened 2026-08-02
fix(harness): refuse to drive input against a dark display

victorarias:
Terminal block copy is not broken. This makes the harness stop saying it is.

## What happened

Three `TERMINAL-BLOCK-COPY` runs failed on 2026-08-02 with a clean, plausible product signature: inside one step, `Cmd+Shift+C` copied the command to the real clipboard, and `Cmd+C` — the shortcut the native Edit > Copy item claims — left the sentinel untouched. That reads as a regression in the plain-`Cmd+C` copy path.

It was not one. All three runs fall inside a display-off window that `pmset -g log` records as 22:21:57 to 22:48:33 local; the runs were at 22:31:57, 22:33:03, and 22:34:49. Every one of them also failed to capture a window screenshot for its failure artifacts, which is the same condition wearing another hat. With the display awake, the same scenario is green on the same build — verified four times, including once with attn deliberately not the frontmost app.

Against a dark display or a locked screen the input still reaches the app, but the window-server work the app's own handling depends on — menu-bar key equivalents, first responder, rendering — does not run. Some keystrokes take effect and others silently do nothing, so a scenario fails on a product assertion the product did not break.

Disproving it cost a full baseline build of the default branch. That is the cost this change removes.

## The guard

`InputDriver.swift` now refuses every command that posts synthetic input and names which condition it found:

```
Input cannot be delivered: the screen is locked or the display is off.
Wake the display (and unlock the screen) before running packaged-app scenarios.
```

It covers `drag` (which never activates) and `activate` itself, so a scenario fails on its first driver call rather than partway through a step. AX-based and observation-only commands work fine against a dark display and are deliberately left alone.

A driver failure's first line now names a dark display as the cause and says it is not a product failure, next to the existing Accessibility-grant hint.

## The reading

`CGSSessionScreenIsLocked` is the one that moves. It goes true the moment the display goes dark, well before any password grace period expires, so it covers display-off and a genuinely locked screen alike.

`CGDisplayIsAsleep` was measured against `pmset displaysleepnow` on the built-in display and never flips — sampling across the transition showed `asleepCount: 0` throughout while `screenLocked` went true. It cannot carry the check on its own, and stays only as the signal for an external display that reports sleep properly. The comment at the call site records that measurement so the next reader does not reach for it.

## Seeing it without reproducing it

A new observation-only `display_state` command, reached as `driver.displayState()`, returns `{ screenLocked, displayCount, asleepCount, blockReason }` — the same readings the guard decides on, with `blockReason` being the exact sentence it would refuse with. Reading it never activates the app and works while the display is off, which is the point.

## Verification

Live on an isolated packaged-app profile at this head:

- With the display slept via `pmset displaysleepnow`, four consecutive `key --key c --modifiers command,shift` calls all exited 1 with the named reason (three caught `screenLocked`, one caught the `no display is active` transition). Before this change the same calls exited 0 and posted the keystroke.
- With the display awake, `TERMINAL-BLOCK-COPY` and `WORKSPACE-SWITCHING` both pass, `failureCount: 0` — the guard produces no false positives on the paths that use keys, clicks, and activation.
- `display_state` reports `blockReason: null` while awake.

`describeInputDriverFailure` is unit-tested and was mutated to confirm the tests are falsifiable: disabling the dark-display branch reddens exactly the one test that asserts it, and nothing else. All 185 harness unit tests pass.

### DIFF
--- a/app/scripts/real-app-harness/AGENTS.md
+++ b/app/scripts/real-app-harness/AGENTS.md
@@ -2,6 +2,25 @@
 
 This policy applies to packaged-app scenarios in this directory.
 
+## The Display Must Be Awake
+
+Scenarios that drive input need a live display. Against a dark display or a
+locked screen the input still reaches the app, but the window-server work the
+app's handling depends on — menu-bar key equivalents, first responder,
+rendering — does not run, so some keystrokes take effect and others silently do
+nothing, and the run fails on a product assertion instead of on the display.
+
+`InputDriver.swift` refuses every input-posting command in that state and names
+which condition it found, so the failure describes itself. Read the same
+findings without touching the app via `driver.displayState()` (the driver's
+observation-only `display_state` command): `{ screenLocked, displayCount,
+asleepCount, blockReason }`.
+
+If a run fails on an assertion that looks like a product regression, check
+whether the display was awake before believing it — `pmset -g log | grep
+"Display is turned"` gives the history. A failed window screenshot in the
+failure artifacts is the same symptom wearing another hat.
+
 ## Profiles (which world a run targets)
 
 The harness honors the **one knob**, `ATTN_PROFILE`, like every other entrypoint
--- a/app/scripts/real-app-harness/InputDriver.swift
+++ b/app/scripts/real-app-harness/InputDriver.swift
@@ -6,6 +6,7 @@ enum DriverError: Error, CustomStringConvertible {
     case usage(String)
     case appNotRunning(String)
     case accessibilityDenied
+    case displayUnavailable(String)
     case invalidArgument(String)
     case eventCreationFailed(String)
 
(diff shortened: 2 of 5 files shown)


figgyster (approved):
Approve.

You followed a very plausible terminal-copy murder back to the room where the lights had gone out. The guard runs before activation and covers every HID-posting path, while AX and observation-only work can still inspect the scene. `display_state` reports the same facts behind the refusal, which makes a failed scenario tell the truth instead of putting the product in cuffs.

I found no unresolved review threads or older Figgy signal on `9113d3b`. `Changes`, `Changelog`, `Frontend E2E`, and React Doctor are green, and the patch passes whitespace inspection. The packaged-app checks are the witnesses that matter for this macOS-specific boundary, and the recorded awake/dark-display runs cover both sides of it.

PR #737, opened 2026-08-03
feat(pty): worker-authoritative kitty image feed (feature dark, one live fix)

victorarias:
Phase A2 of [the kitty image plan](docs/plans/2026-08-02-terminal-kitty-images.md): the worker's native ghostty becomes the system's only kitty parser, and the feed path learns to hand it kitty APCs while describing their grid effects to a client that can never parse them. Everything here is feature-dark — the worker's kitty storage limit stays 0 until phase A4 flips it — with one exception, called out below, that fixes a defect live in shipped attn today.

## What changes

- **One segmenter for the whole feed** (`internal/pty/kittyseg.go`). A single state machine frames both kitty APCs and OSC 133 prompt markers, replacing the byte-pattern marker scan. Every state transition is measured against ghostty's actual parser at the dispatch level, not read off a spec — which C1 bytes cut a string open, which are payload, what CAN/SUB do to an OSC, how escape-intermediate behaves. Extraction happens only from ground, to a proper terminator; everything else reaches both sides byte-for-byte.
- **Two streams from one chunk** (`internal/pty/wirefeed.go`). The terminal feed carries plain runs and whole APCs; the wire carries the same stream with each APC replaced, in position, by bytes that leave a kitty-ignorant terminal on the same grid. In the shipping configuration ghostty refuses every transmission, so the wire contribution of an image is an ST and the property under test is the disposal itself.
- **The ESC-parity rule.** Ground alone is not a safe extraction point: it also holds a UTF-8 decoder that may be mid-character, and the ESC an extraction removes is what ends that decode. So wherever the two streams differ, both sides get an ESC-led no-op at that position — the wire where an APC was stripped, the worker (before anything is measured, since ending a decode is itself a grid event) and the worker again where a marker was withheld. The plan records the fully-built alternative this was chosen over, with measurements.

## The one live fix

OSC 133 markers are deliberately kept out of the worker's terminal (native ghostty line-breaks on a mid-line `133;A`; the shipped wasm does not) while the wire delivers them to the client. A marker arriving mid-UTF-8-character therefore ended the client's decode and not the worker's: one wrong extra cell and a column of drift, permanent until reattach. The worker now takes an ST in the marker's place. User-facing changelog fragment included; witnessed against the real shipped wasm at both measured columns, and end-to-end through a real fish session on a throwaway profile.

## Verification

- Parity corpus: 39 entries replayed into both a native ghostty that cannot parse kitty and the shipped `ghostty-vt.wasm` the app renders (`app/src/utils/kittyWireRewrite.parity.test.ts`). The wasm side is the authority on client behavior and doubles as the pin-bump tripwire for the two build-asymmetry facts this PR relies on.
- Fuzz, split by configuration: `FuzzKittySegmenterFraming` (framing alone) and `FuzzKittyWireMirrorShipping` (the shipping-config gate) each soaked 3×3 minutes clean, plus independent cold-corpus soaks; `FuzzKittyWireMirror` (kitty live) is the A4 property and is documented red on the deferred A4 defects, seeds-only in CI.
- Mutation receipts per defect class, each landing on a named test; the measured tables and the failure history are in the plan.
- No protocol change; `git diff main...HEAD` touches nothing under `internal/protocol` or `app/src/types`.

The branch history narrates the path honestly, including an intermediate compensation commit and the rejection record for a fully-implemented alternative design; the squash keeps that history readable here without carrying it to main.


figgyster (commented):
Victor, this is the right kind of scary terminal work. You put the parser where the authority already lives, then kept A2 feature-dark while the feed learns its manners. The worker/client parity corpus and the ESC substitutions cover the two decoder traps that would otherwise wait until a restore to collect their debt.

I found no contract break on `bf99db2`. Linux preflight on both architectures, Frontend, all three E2E shards, and React Doctor passed; CI also passed the changed `internal/pty` and native Ghostty suites. I’m holding the approval while the Daemon job is red on `internal/ptyworker: TestProcessHasArgIdentifiesOwnProcess`—a process-discovery witness outside this patch’s diff. Give that one a rerun or a fix, and I’ll put the stamp on it.

figgyster (approved):
Victor, the last witness finally showed up sober. The Daemon rerun is green, and your throwaway-profile fish run proves the class-3 UTF-8 split agrees with the shipped WASM model on the path people actually use.

Approved on `bf99db2`. The worker-authoritative parser boundary, ESC-led substitutions, and parity corpus keep the two grids telling the same story while A2 stays feature-dark. The city gets kitty-ready without teaching the client a second lie.

PR #738, opened 2026-08-03
feat(annotations): send with a key, edit from the panel, find the way out

victorarias:
Five gaps in the terminal annotation surface, all about reaching what is already there.

## Send the whole set with the keyboard

`⌘Enter` — the key markdown tiles already use for the same verb — sends every annotation to the session. It is *registration*-gated, not handler-gated: the dispatcher consumes a keystroke whenever a handler exists for it, so the shortcut is only registered while the annotated pane is the focused leaf of the visible session **and** the panel holds at least one mark. Everywhere else, including a pane with nothing waiting, `⌘Enter` still reaches the PTY.

The two ids share the combo through `ALLOWED_CONFLICT_PAIRS`; a focused markdown tile and a focused terminal pane are mutually exclusive, so only one handler is ever registered. A comment being typed when the send fires is committed into the payload rather than dropped.

## Edit an annotation from the panel

A panel row is the list of what you wrote, so clicking it opens that annotation's comment box, with the caret at the end of the existing text. It used to open the bare reaction row whenever the annotation carried no comment — a row of eight emoji answering a question nobody asked at that moment. The popup anchors to the row's top edge rather than the raw pointer, so it lands in the same place every time.

## Find the way out

Removing an annotation was possible from the popup and effectively unfindable: a `🗑` sitting ninth in a row of eight reaction emoji reads as another label to apply. It is now tinted at rest, and the open comment box carries a named **Remove** beside Cancel.

## Hand the keyboard back

The comment box takes focus as the editor opens. Closing the editor returns focus to the terminal — the user came from the grid and the next thing they type is meant for the agent. The exception is a press that deliberately landed elsewhere: that press owns where focus goes, so it is not yanked back.

## Stop the panel row's remove control jumping to a line of its own

Root cause: the row was a single CSS grid, and `.anno-card-comment` declared `grid-column: 2 / 4`. Sparse auto-placement put it on the second row spanning to the last column, which left the auto-placed remove button no free cell after the cursor — so it got an implicit **third** row. That is why it only moved when an annotation carried text, i.e. exactly when it is most wanted.

The row is now two controls side by side (`1fr auto`): a button that opens the annotation, and the remove button in its own track. That also gives the row keyboard reach and stops nesting a `<button>` inside a clickable `role="button"` element.

## Verification

- Frontend suite (2357 tests) and Playwright e2e (192) green; `tsc --noEmit` clean.
- Twelve new unit tests over the surface, including the registration gate in both directions (fires when the pane is focused with marks waiting; silent when it is not the focused pane, and when there is nothing to send).
- Live on a packaged build (throwaway profile, since cleaned up): `real-app:scenario-terminal-annotations` passes with three new legs —
  - clicking a panel row opens its editor and the comment box holds focus,
  - a commented row reports its remove control **6px** below the row top (tripwire at 12px; the old layout put it a full line lower),
  - the set is sent with a real native `⌘Return` rather than the button. The button is a click handler any unit test can drive; the keystroke has to survive AppKit, reach the page's capture-phase listener, and be claimed by the pane instead of the PTY, which only the packaged app can show.

`get_annotation_state` now reports panel-row geometry and whether the comment box holds focus, so both stay measurable tripwires rather than one-off checks.

The registry's duplicate-combo test was reading a hand-copied exemption list; it now calls the registry's own `isAllowedConflict`.

## Note

`⌘Enter` is claimed from the PTY in a pane that has marks waiting. Claude Code binds Option+Enter / Shift+Enter for newlines and nothing to `⌘Enter`, so the cost looked like zero — but it is a real trade and worth naming.

https://claude.ai/code/session_01MmcGPuheR23tr4fP5ReZ16

### DIFF
--- a/app/scripts/real-app-harness/scenario-terminal-annotations.mjs
+++ b/app/scripts/real-app-harness/scenario-terminal-annotations.mjs
@@ -25,6 +25,7 @@ import {
   relaunchAppAndConnect,
 } from './common.mjs';
 import { UiAutomationClient } from './uiAutomationClient.mjs';
+import { MacOSDriver } from './macosDriver.mjs';
 import { DaemonObserver } from './daemonObserver.mjs';
 import {
   captureSessionArtifacts,
@@ -254,6 +255,10 @@ async function main() {
 
   const client = new UiAutomationClient({ appPath: options.appPath });
   const observer = new DaemonObserver({ wsUrl: options.wsUrl });
+  // ⌘Enter is delivered by the page's keydown listener, and whether it reaches
+  // that listener at all — rather than being eaten by AppKit or handed to the
+  // PTY — is the part only a real keystroke into the packaged app can show.
+  const driver = new MacOSDriver({ appPath: options.appPath });
   let sessionId = null;
   let paneId = null;
   // What the agent said, and where it sits on the grid. Carried through the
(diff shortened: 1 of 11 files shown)


figgyster (approved):
You took the durable annotation work from #704 and gave it the last few things a user actually needs: a way to spend the set without leaving the keyboard, reopen a thought where it is listed, and get back out without spelunking through emoji. That is product judgment, not chrome. The registration gate keeps `⌘Enter` in the PTY unless this exact focused pane has work waiting, and the editor still owns that key while typing a comment.

The focused annotation suite passed all 33 witnesses in the sandbox, TypeScript checked, and the frontend plus every E2E shard are green. The packaged-app scenario reaches the native keystroke and the geometry/focus seams that unit tests can only describe. Approved.

PR #740, opened 2026-08-03
feat(annotations): send the set and submit it

victorarias:
## What

"Send all" in the terminal annotation panel typed the composed feedback into the agent's prompt and left it sitting there for the user to press Enter on. It now submits.

## Why it moved to the daemon

Appending `\r` in the frontend would not have worked, and would not have been safe. The delivery now goes through `typeDoorbell` — the primitive the chief-of-staff doorbells and markdown annotations already submit through:

- **Timing.** An Enter arriving in the same PTY read as the paste terminator is folded into the pasted text, so the payload sits in the composer and never submits. The gap that makes it a keypress is measured and already lives in one place — `doorbellSubmitDelay`, 150ms, with the receipt in its comment (0ms fails, 50ms submits on Claude Code 2.1.x).
- **Approvals.** `pending_approval` is an annotatable state — an approval prompt is exactly when a user wants to push back — but it is also where the submitting Enter would *answer* the prompt. The daemon refuses, and the panel says so with the marks still on screen to retry.
- **Interleaving.** The PTY write fence keeps a keystroke racing the gap from landing between the paste and its Enter.

## Data safety

Because the send now has an answer, the marks are spent only on `delivered`. Previously they were cleared optimistically the moment the paste was written, so a delivery that failed took the user's work with it. A skip or an error leaves every annotation where it is, above a footer that still offers Send all.

## Protocol

New `session_annotations_submit {session_id, text}` → `{status: delivered | skipped_pending_approval | error}`. `ProtocolVersion` and `PROTOCOL_VERSION` both to **204**; `generated.go` and `generated.ts` regenerated, not hand-edited.

The text stays client-composed rather than built from the persisted draft (as markdown annotations are): a quick label's instruction is UI vocabulary that belongs with the chip that offers it, and the client can already write these exact bytes with `pty_input`. What moves to the daemon is the *delivery*.

It routes by session id like its get/save/clear siblings — a hub answering locally has no PTY to type into, so a remote pane's Send all would report success and deliver nothing — and unlike them it blocks during recovery, because it writes to a PTY rather than to the store.

## Surfaces walked

- **Protocol** — schema, both generated files, version bumped in all three lockstep spots.
- **Daemon and app** — handler, router, hub→remote routing table, frontend socket/event/component.
- **Docs** — changelog fragment.
- **CLI** — not applicable; annotations have no command-line entry point.
- **Linux** — no platform-specific code added.

## Verification

Throwaway profile `annosubmit` (port 24333, preflight PASS, protocol 204 on both sides):

- Packaged-app scenario `TERMINAL-ANNOTATIONS` **PASS twice**, `failureCount: 0`. Footer read `✓ sent 1 to the session`, and the session **opened and settled a turn** on the feedback — a paste alone cannot do that. The agent's reply engaged with the 🧪 Needs tests mark on the quoted text.
- The live daemon was probed directly for the unknown-session and empty-text refusals.
- Frontend 2376 tests, e2e 192, Go `./internal/daemon ./internal/protocol`, `tsc --noEmit`, `gofmt` all green. Profile cleaned up afterwards.

Mutation-checked, each killing exactly the intended tests:

| Mutation | Result |
|---|---|
| Restore the optimistic clear (pre-change behaviour) | 4 tests red |
| Drop only the in-flight guard | 1 test red |
| Fold Enter into the paste write | delivery test red |
| Collapse the approval skip into a generic error | skip test red |

## Note

The `pending_approval` refusal is covered by a Go handler test and a component test, not by a live run — driving a packaged session onto an approval prompt and then annotating it is a scenario I did not build. The guard itself is `typeDoorbell`'s existing one, already in production behind markdown annotations.

https://claude.ai/code/session_01MmcGPuheR23tr4fP5ReZ16


figgyster (dismissed):
You've given #738/#739's feedback marks the last mile — a real submitted turn, with the approval guard still keeping its hand off the trigger. That's the right shape. But this response race loses a newer mark.

* [P1] Preserve annotations made while an earlier send is in flight — `app/src/components/TerminalAnnotations/AnnotatedTerminal.tsx:506`

  `sendingRef` prevents a second send, but it does not disable `handleAnchor`, edits, or removals while `submitAnnotations` waits. A user can add a fresh annotation during the paste-to-Enter round trip; when the older request returns `delivered`, `store.clear()` wipes the whole live store, then `clearAnnotations` tombstones the daemon copy as well. The new mark was never in `sending` and can never be sent. Keep post-start mutations (for example, clear only the submitted snapshot), or make annotation mutation unavailable for the whole in-flight interval, and pin this exact held-submit/new-annotation race with a witness.

figgyster (dismissed):
Keeping the panel live was the right call, and the copied snapshot closes the fresh-mark hole I named. But the same snapshot still eats an edit to a mark that was already in it.

* [P1] Do not spend a snapshot entry whose contents changed during the send — `app/src/components/TerminalAnnotations/AnnotatedTerminal.tsx:519`

  While the first request is waiting, a user can reopen one of its existing annotations and change its label or comment. `store.update` mutates that same record and persists the revised draft, but the delivered callback removes every snapshot ID without checking whether the current entry still matches the copied snapshot. The agent receives the old payload and the user's newer edit disappears. Keep a snapshot member only when its current value is unchanged (or freeze mutation of sent entries for the interval), and add the held-submit witness for a mid-flight edit as well as a newly created mark.

figgyster (commented):
That closes the last race I could find in this little chain of custody. The send snapshot now knows the difference between a mark it delivered and a mark you changed before the receipt came back, while still spending the untouched neighbor. I ran the focused annotation suite here: 46 tests passed. The principal CI witnesses are still running, so this is a conditional handoff rather than an approval.

figgyster (approved):
You kept the annotation surface alive, then made the receipt smart enough to leave changed evidence on the desk. That was the right bet. The sent snapshot now spends only what truly went out, and the tests finally cover the quiet little race where the human keeps working while the daemon is still typing.

The principal board is green: daemon, frontend, and every E2E shard. The benchmark is still running its beat, but it is not a merge gate. Approved.

PR #742, opened 2026-08-03
feat(notebook): allow disabling session summaries

victorarias:
## Intent

Automatic session summaries can repeatedly read full transcripts and accumulate substantial Haiku spend. The keeper master switch is too broad when journal narration and context compaction should keep running. This adds a narrow way to stop summary generation without discarding the configured summary agent or model.

## Summary

- add a default-on notebook.summarize_session.enabled setting
- stop creating new summary jobs while disabled and retire queued jobs without launching an agent
- add an independent Session summaries control in Settings while preserving model configuration
- document the setting and cover default, toggle, enqueue, executor, validation, and wire-payload behavior

## Verification

- repository pre-commit gate: Go build and tests; 211 frontend test files; production frontend build; Playwright 192 passed, 1 environment-skipped
- Linux amd64 daemon compilation
- isolated packaged-app preflight and live Settings verification: disable, confirm persisted value, relaunch, and confirm the control remains disabled
- git diff --check

### DIFF
--- a/app/src/components/SettingsModal.css
+++ b/app/src/components/SettingsModal.css
@@ -537,7 +537,16 @@
 
 .settings-keeper-duty-head {
   display: grid;
-  gap: 4px;
+  grid-template-columns: minmax(0, 1fr) auto;
+  align-items: center;
+  gap: 16px;
+}
+
+/* A duty can be independently off while its model remains editable, just like
+   the master switch leaves all duty configuration reachable. */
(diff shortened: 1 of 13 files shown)


figgyster (approved):
You gave the keeper a real cost valve instead of making people pull the whole power switch. The summary setting stays default-on for existing profiles, keeps its model choice intact, gates both enqueue and queued execution, and leaves narration and compaction breathing. That is the smaller switch this work needed.

The daemon, frontend, every E2E shard, and both Linux preflights all checked out. Approved.

PR #744, opened 2026-08-03
Fix model capture size status

victorarias:
## What changed

- Request a fresh daemon settings snapshot when the Data settings section becomes visible.
- Continue refreshing at the configured model-capture interval while that section is visible and capture is enabled.
- Stop the refresh when the user leaves the section or disables capture.

## Why

The recorder was writing observations correctly, but `model_capture.bytes` was only computed when the daemon happened to send a settings snapshot. Enabling capture sent that snapshot before the first write, so the UI stayed at `0 bytes` even while the corpus grew.

This uses the existing `get_settings` command and scopes refresh work to the visible Data page. It avoids broadcasting the full settings payload after every recorder write to every connected client.

## User impact

The Captured value now reflects corpus growth while the user is looking at it. Capture behavior, retention, privacy boundaries, and the wire contract are unchanged.

## Validation

- Frontend: 211 test files passed; 2,381 tests passed and 4 skipped.
- Frontend production build passed (`tsc` and Vite).
- Changelog validator and gate passed.
- Installed branch as the isolated dev app; bundled preflight passed with protocol 204.
- Live dev verification observed an immediate settings refresh on opening Data, another refresh after the configured 10-second interval, and no continued cadence after disabling capture.

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -751,7 +751,6 @@ function AppContent({
     sendSetChiefOfStaff,
     sendUnregisterSession,
     sendSetSetting,
-    sendGetSettings,
     sendCreateWorktree,
     sendDeleteWorktree,
     sendListPlugins,
@@ -4091,7 +4090,6 @@ function AppContent({
         onRemovePlugin={sendRemovePlugin}
         onSetPluginPriority={sendSetPluginPriority}
         onSetSetting={sendSetSetting}
-        onRefreshSettings={sendGetSettings}
         themePreference={themePreference}
         onSetTheme={setTheme}
         uiScale={scale}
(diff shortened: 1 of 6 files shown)


figgyster (commented):
You found the narrow seam #733 left behind: the recorder was honoring its private, bounded ledger while the person watching the Data page was reading yesterday's weather. This keeps the `SizeBytes` snapshot scoped to the one visible page, rides the existing `get_settings` wire, and the follow-up wisely pulls that refresh through the daemon API context instead of dragging another prop across `App`.

The focused sandbox witnesses are in: `SettingsModal` and `useDaemonSocket` passed 137 tests, and `tsc --noEmit` passed. Frontend and all three E2E jobs are still working their shift, so I'm leaving this as a comment rather than an approval for now. When those clear, the meter has its alibi.

figgyster (approved):
The board came back green all the way through: Frontend and every E2E shard agree with the focused 137-test pass and typecheck. The capture meter finally tells the truth while somebody is looking at it, without teaching every client to listen for recorder churn. Approved.

PR #745, opened 2026-08-03
feat(docstore): namespaced JSON documents with live queries

victorarias:
A3 of the [extension platform roadmap](docs/plans/2026-08-01-extension-platform-roadmap.md): the store an extension keeps its own data in. Gate answers and the implementation notes are in [docs/plans/2026-08-03-ext-a3-doc-store.md](docs/plans/2026-08-03-ext-a3-doc-store.md).

## What it is

A namespace (`owner/name`, e.g. `ext/approval-gate`) holds collections, a collection holds JSON documents under caller-chosen ids, and a collection carries a **declaration** naming which fields are queryable and with what type.

Documents are stored byte for byte and nothing rewrites them, so adding a queryable field is a declaration change rather than a migration: filters read the body through `json_extract` at query time. `created_at` and `updated_at` are always queryable and must not be declared.

A query is one JSON object — namespace, collection, filters, sort, limit, and an **after cursor**: the id of the previous page's last document. The visible order is `(sort field, id)`, with the id tiebreaker running in the sort's own direction, and the cursor compiles to a comparison against that whole tuple (`sort <cmp> value OR (sort = value AND id <cmp> anchor)`). It is part of the query rather than a filter a caller writes, because a filter can constrain only the first half of the tuple: with documents tied on the sort field, `sort > value` skips the tied ones and `sort >= value` returns the anchor forever. There is no offset and no opaque token. Left open as a **live query**, it re-runs when a write to the collection says the answer may have moved, and every delivery is the whole current result set.

## Shape

- `internal/docstore` — the query language and `Query.Compile(schema) → Compiled`. It holds no database handle, so every rule about what a query means is testable without one. Same seam as `internal/bus` ↔ `internal/store/bus.go`.
- `internal/store/documents.go` — two tables (migration **88**), the CRUD, and `QueryDocuments(Compiled)`. It executes SQL it did not decide.
- `internal/daemon/documents.go` — IPC handlers, the change fact, live-query subscriptions.
- `internal/client/documents.go`, `cmd/attn/doc.go` — the operator surface: define, undefine, collections, put, get, delete, query, watch.

Writes publish `document.changed` on the bus, deliberately **without** a `wireProjections` entry: no WebSocket traffic in this stage. An ephemeral subscriber beside the hub pokes a one-slot channel per subscription, so the bus's publish lock is never held across a socket write — `TestWritesDoNotBlockOnASubscriberThatIsNotReading` hangs rather than fails if that regresses. `doc_subscribe` registers before its first query, because that order can only produce one redundant delivery while the other can miss a write.

## Bounds

`DefaultLimit` 100, `MaxLimit` 1000. Receipt (2026-08-03, production `~/.attn`): the lists attn already pushes whole are tickets 7, sessions 11, notifications 8, workspaces 8 — an order of magnitude and two orders of magnitude past the working set. Asking past the ceiling is an error naming the limit, the ask, and the cursor to page with — never a silent truncation.

## Deliberately not built

- **Indexes.** Scan-first was the gate's answer; the working set above is the receipt, and the declaration is what makes adding them later invisible to callers. The roadmap line that said "with secondary indexes" is updated to match.
- **Grants.** Namespaces are `owner/name` and enforced as a shape; *who* may claim one is A4's registry, which is the thing that knows about authors.
- **A frontend surface.** A5 owns the UI host and the generic protocol envelopes; nothing here pre-empts that gate.

## Verification

`make test` and `make test-frontend` green; `tsc --noEmit` clean after the protocol regen (**206**, both constants moved).

Live on a throwaway profile (`docstore1`, full `make install` since `generated.ts` moved; preflight PASS, app and daemon both protocol 205; profile cleaned up after):

- define → put → query with filters, sort, limit; an undeclared key in the body survives the round trip.
- `attn doc watch --where status=pending` delivered revision 1 on subscribe, revision 2 on a put, revision 3 on a delete. A no-op delete and a write to another namespace produced no revision — proven because the delete landed at revision 3, not 5.
- Documents written before a daemon restart were still queryable after it.
- Namespace isolation: `ext/other/requests` and `ext/approval-gate/requests` hold different documents under the same id and neither query sees the other.
- Every error names the limit, the ask, and the way out — over-limit, undeclared filter field, type mismatch, malformed namespace, reserved field name, undeclared collection.
- `undefine` removes the declaration and its documents, and the collection is gone from `collections`.
- Four documents all at `attempts=7` — a total tie — paged one at a time walked `a b c d` ascending and `d c b a` descending, each exactly once, with the next page empty in both directions. A cursor to a deleted document errors by name.

## Surfaces walked

CLI (`attn doc`, plus the help listing), daemon (IPC dispatch, `CommandMeta`, wire annotations), protocol (`main.tsp` → regen → `ProtocolVersion` 205 in both places), docs (glossary entry for namespace/collection/document/live query, `internal/docstore` in the AGENTS.md architecture list, changelog fragment). Linux is unaffected — no build constraints and no platform-specific code. No plugin or SDK surface consumes the store yet; A4 is the first.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -273,6 +273,8 @@ forgets that position and `--since <RFC3339>` replays from an instant.
   `internal/pty`, not inside the daemon
 - `internal/store`: SQLite plus in-memory cache
 - `internal/bus`: durable event bus (domain facts, per-consumer cursors)
+- `internal/docstore`: document-store query semantics and SQL compilation (no DB
+  handle; `internal/store/documents.go` executes what it compiles)
 - `internal/jobs`: durable job queue (retry/backoff, coalescing, commit fence,
   cron entries) — every background duty and every periodic tick runs on it
 - `internal/classifier`: stop-time state classification
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -196,7 +196,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '205';
+export const PROTOCOL_VERSION = '206';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 // AutomationActionTimeoutError distinguishes "the daemon never sent a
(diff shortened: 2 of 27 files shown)


figgyster (dismissed):
Victor, this is the right kind of extension-platform primitive: the declaration gives the store a future without paying index rent today, and the bus handoff keeps a slow watcher from turning every writer into a hostage. It picks up the A1 spine and gives A4/A5 something real to stand on. But the pagination promise has a body in the alley.

**Block before merge**

`internal/docstore/docstore.go:307` makes the visible order `sort field, id ASC`, but the query language can only express a range predicate on the first half of that tuple. Put `a` and `b` under the same sort value, ask for `limit: 1`, then page after `a`: `sort > value` drops `b`; `sort >= value` repeats `a`. The existing pagination witness uses distinct `attempts` values, so it never crosses this boundary.

This is the advertised escape hatch once the 1,000-document ceiling is reached, and ordinary fields like status will tie constantly. Please make the cursor represent the complete ordering tuple (or otherwise express the equal-sort-value/id branch) and add a tie-boundary regression for both directions. Until then, the `id` tiebreaker stabilizes a page but cannot make the next page correct.

The full CI board is green. This one needs an eyewitness who brings two rows with the same hat.

figgyster (dismissed):
Victor, you fixed the dead man I pointed at. The cursor now knows the whole ordering tuple, the missing-field branch has a proper alibi, and the tie witnesses walk both ways. That takes A3 from a promising store to an extension platform with a real escape hatch. But the arbitrary-JSON promise still leaves one customer face-down at the page boundary.

**Block before merge**

`internal/docstore/docstore.go:430-447` unmarshals the anchor’s declared sort field into `any` and binds it unchanged. A declaration is explicitly not a storage schema, so a queryable `number` field may still contain `[]` or `{}`. Page after `{"attempts":[1]}` and the SQLite call gets `[]interface{}` as an argument, then returns `unsupported type []interface {}, a slice of interface` instead of the next page.

I put two such documents through the live cursor path in the sandbox. It failed exactly there. The scalar/tie/missing-field cases now pass, but this is still ordinary data the documented contract accepts, and pagination turns it into a runtime error. Please either make cursor binding match `json_extract` for compound JSON values, or reject incompatible declared-field values at the boundary and say so in the contract. Add the array/object cursor witness alongside the good new tie witnesses.

The earlier tuple-cursor blocker is retired. Daemon and E2E CI are green; Frontend is presently red on a `PresentRoot` selection assertion, separate from this docstore path.

figgyster (commented):
Victor, this one stands up. Pulling the cursor value back through SQLite’s own expression is the right escape route: it keeps the declared-field-is-not-a-storage-schema promise without trying to smuggle an array through a scalar bind slot. The compound-value, mixed-type, null, and tie witnesses cover the whole bad neighborhood.

The focused `internal/docstore` and `internal/store` pagination loop passes here, including the new compound cases. Both earlier cursor blockers are retired on `9992103`.

The principal CI board is still moving, so I’m leaving this as a comment rather than a premature stamp. Once Daemon, Frontend, and the E2E jobs finish their walk, this looks ready to close.

figgyster (dismissed):
Victor, the whole loop holds now. A3 gives extensions durable namespaced state without pretending arbitrary JSON is a tidy little schema, and the cursor learned the hard lesson: ask SQLite what it sorted instead of rehearsing the value in Go.

The tie, missing, compound, mixed-type, and null paths have witnesses. Daemon, Frontend, every E2E shard, Linux preflight, React Doctor, changelog, and the benchmark all came back green. This is the extension platform growing a memory without growing a second database in the walls.

Approved.

figgyster (commented):
Victor, the docstore work survived the rebase intact. The new base brings #741’s kitty-image machinery aboard, but the A3 files I approved are byte-for-byte the same, and the focused cursor suite still passes on this head.

Fresh CI belongs to the rebased commit and is still making its rounds, so this stays a conditional note for now. Once that board goes green, I see no remaining objection to the document store.

figgyster (approved):
Victor, the rebase kept the good bones and the lab has finished crawling over them. The document store still owns durable extension state, the cursor still lets SQLite testify to its own ordering, and the kitty-image base arrived without a protocol collision.

Focused pagination is green here. Daemon, Frontend, every E2E shard, Linux preflight, React Doctor, changelog, and the benchmark are green on `ffdf488`.

Approved.

PR #746, opened 2026-08-03
fix(daemon): refuse to start when the WebSocket port is taken

victorarias:
Two fixes for a failure mode where remote commands silently die and the app times out with no explanation.

## The problem

attn derives a profile's WebSocket port from the profile name, so the same profile on an OrbStack VM derives the same port — and OrbStack forwards the VM's listener onto host localhost whenever the host port is free. When the local daemon then started, its WS bind failed `EADDRINUSE`, but `runHTTPServer` only logged that at INFO and the daemon kept running. The result is a split brain: the app routes by WS port and attaches to the foreign (VM) daemon, while the CLI routes by the unix socket and talks to the local one. Every command sent from the app ran on the wrong daemon, and the app's workspace-registration wait ignored the resulting error events, so the user saw a generic 10-second timeout instead of anything actionable. This cost hours of misdiagnosis during the kitty-images remote leg (#741): the symptom looks like a registration bug, and nothing errors anywhere visible.

## The fix

**Daemon (`fix(daemon)`):** the WebSocket bind is now synchronous and fatal at startup. `Start()` calls `listenHTTP()` before declaring the daemon ready; a failed bind aborts startup with an error that names the address, the profile, the underlying cause, the likely owner (a same-profile daemon whose listener a VM forwards onto host localhost), the split-brain consequence of continuing, and both ways out (free the port, or run under a different `ATTN_PROFILE`). The error is also written to `daemon.log`, because a detached daemon's stderr goes nowhere. `runHTTPServer` serves the pre-bound listener; the startup-failure path and `Stop()` both close it (a `Shutdown` only closes listeners the server is already serving).

**App (`fix(app)`):** `register_workspace` failures forwarded by the daemon as `command_error` now reject the pending registration promise with the daemon's actual error text, instead of falling through to the 10-second "Workspace registration timed out". `command_error` carries no correlation id, so the handler rejects by key prefix (`register_workspace:`), same as the pre-existing `unregister_workspace` case — over-rejecting a concurrent registration beats a silent timeout. Root cause was that keyed pendings never match the exact-match default arm of `rejectPendingForCommand`.

## Verification

- New daemon test `TestDaemon_Start_FailsWhenWebSocketPortIsAlreadyBound`: occupies the derived port with a foreign listener, asserts `Start()` fails with an error naming the address and the profile, and races the error against `startedCh` to catch a fire-and-forget regression. Passes with `-count=1`.
- New frontend suite `useDaemonSocket.workspaceRegistration.test.tsx`: drives the real hook over a fake WebSocket; a `command_error` for `register_workspace` / `unregister_workspace` settles the pending with the daemon's text. Each test drains the 10s timer so a dropped branch fails on the timeout wording rather than hanging. Full frontend suite green (216 files, 2437 tests).
- Live tier (throwaway profile `bindfx`, port 21940): with the port deliberately occupied, `attn daemon ensure` refused to start and `daemon.log` carries the full refusal naming `127.0.0.1:21940` and the profile; with the port freed, the same ensure started cleanly. Profile cleaned afterwards.
- The frontend change had no packaged-app leg: producing a real forwarded registration error requires a parked remote endpoint (the world torn down after #741), and the rejection mechanism the new case routes into is the same code path the pre-existing `unregister_workspace` prefix case already exercises in production. The unit tests drive the exact wire shape through the real hook.

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -1021,6 +1021,13 @@ export function useDaemonSocket({
       case 'unregister':
         rejectPendingByPredicate((key) => key.startsWith('unregister:'), error);
         return;
+      // command_error carries no correlation id, so this fails every
+      // registration in flight. Registering more than one workspace at a time
+      // is not something the app does, and over-rejecting beats the ten-second
+      // silent timeout that a failure used to surface as.
+      case 'register_workspace':
+        rejectPendingByPredicate((key) => key.startsWith('register_workspace:'), error);
+        return;
       case 'unregister_workspace':
         rejectPendingByPredicate((key) => key.startsWith('unregister_workspace:'), error);
         return;
--- a/app/src/hooks/useDaemonSocket.workspaceRegistration.test.tsx
+++ b/app/src/hooks/useDaemonSocket.workspaceRegistration.test.tsx
@@ -1,0 +1,178 @@
+import { act, renderHook, waitFor } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { isTauri } from '@tauri-apps/api/core';
+import { PROTOCOL_VERSION, useDaemonSocket } from './useDaemonSocket';
+
+class FakeWebSocket {
+  static readonly CONNECTING = 0;
+  static readonly OPEN = 1;
+  static readonly CLOSING = 2;
+  static readonly CLOSED = 3;
+  static instances: FakeWebSocket[] = [];
+
+  readonly url: string;
+  readyState = FakeWebSocket.CONNECTING;
+  onopen: ((event: Event) => void) | null = null;
+  onmessage: ((event: MessageEvent) => void) | null = null;
+  onclose: ((event: CloseEvent) => void) | null = null;
+  onerror: ((event: Event) => void) | null = null;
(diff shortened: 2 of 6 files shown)


figgyster (commented):
You found the kind of split brain that makes a small wire failure look like a ghost story for hours. Binding before readiness closes the ownership gap, and carrying `register_workspace` errors back to its parked promise means the next bad route says why instead of making someone interrogate a ten-second timeout.

I don't have a code blocker on `355643a`: the failed bind releases the Unix socket and PID lock, and the new test pins that readiness is never announced. Your throwaway-profile run covers the ugly real path; all three E2E shards are green. Daemon and Frontend are still red on unrelated `ptyworker/TestProcessHasArgIdentifiesOwnProcess` and `NotebookBrowser` failures, so I'm leaving this as a comment until the precinct lab gets a rerun.

figgyster (commented):
The rebase check says #746's six-file patch is still the patch I reviewed; the rest is #745's already-settled docstore ground, so no new corner crawled out of the rain. Your fresh throwaway-profile run proves the bind refusal and the recovery path again, and the full frontend suite passed on this head.

I still don't have a code blocker on `af60b15`. The rebased Daemon, Frontend, and E2E checks are underway, so this is the current-head comment rather than a signature in wet ink.

PR #748, opened 2026-08-03
feat(docstore): a table per collection, with an index per declared field

victorarias:
The document store (#745) shipped scan-first behind a tripwire: a query past
50ms would log its collection size, so the case for indexes would arrive with a
receipt. Two measurements broke that reasoning, so this replaces the physical
schema before anything writes to it.

## Why scan-first did not survive

Measured on sqlite 3.51, 20k documents in the target collection, warm:

- A filtered, sorted, limited query scans in **~2ms at 20k documents**. The
  50ms tripwire would not fire until roughly 100k documents in one collection.
  The receipt was never going to arrive.
- The cost that grows is not per-query, it is **per write × per subscription**.
  A live query re-runs every matching subscription on every committed write, so
  20 subscriptions on a 20k-document collection make one write cost ~40ms of
  CPU — every millisecond of it invisible to a per-query threshold. That term
  scales with exactly what the extension platform adds next.

Indexed, the same query is 0.4ms and the sort-only case drops from 4ms to
~0.01ms. Write cost with the index present is noise: 2000 puts in 3ms.

## What changed

Each collection gets its own table, minted as `doc_<registry id>`, with a
`VIRTUAL` generated column per declared field and a `(field, id)` index on
each. The declaration is still the query contract, and declaring still rewrites
no documents — a virtual column computes on read, so only the index build
scans. Redeclaring diffs the fields and alters the table; undefining drops it,
so the space returns immediately.

Everything stays in `attn.db`, which keeps a document write in the same
transaction as the registry, the `document.changed` fact, and any job enqueued
alongside it.

The registry id is `AUTOINCREMENT` on purpose: a reused rowid would let `doc_1`
come back naming a different collection while an in-flight subscription still
holds it. Undefine-then-define is an ordinary thing to do, so reuse has to be
impossible rather than unlikely.

Migration 89 replaces the storage rather than migrating it — the store shipped
with zero writers, confirmed against production `~/.attn`, which had not even
reached migration 88.

## Two behaviour changes

**A declared type now decides how stored values compare.** The column carries
its declared affinity, so a body holding `"attempts": "5"` in a field declared
`number` reads, compares, and sorts as the number 5. Before, it compared under
SQLite's cross-type ordering, which put `"10"` before `9`.

**A live query ends when its collection can no longer answer.** The delivery
loop re-reads the declaration per delivery instead of capturing it. Undefining
a collection drops its table, and the old behaviour answered the watcher with
an empty result set — which says "this collection is here and holds nothing" to
someone whose collection is gone. Redeclaring without a field a live query
filters on drops that field's column, and the query silently stops meaning what
was asked. Both now end the subscription with an error naming what happened,
which is the way out that `undefine` was missing.

The wire surface, the CLI, and `ProtocolVersion` are unchanged.

## Also here

A second tripwire prices what a live query actually costs: query duration times
the number of subscriptions the last write woke on that collection, budgeted at
one 60Hz frame (16ms). The fan-out already counts the subscriptions it wakes,
so it stamps the count on each one rather than having every delivery re-count
under a lock — which would make a log line that almost never fires quadratic in
the term it exists to watch.

## Known bound

A query that filters on one field and sorts by a *different* one uses the
filter's index and then sorts the matches in a temp B-tree. There is no
composite index, because one per (filter, sort) pair is combinatorial in the
number of declared fields. The sort still sees only the matched rows rather
than the collection, and if a real consumer measures slow, the fix is a
composite index for that specific pair built from the same declaration.

## Rejected alternatives

**Expression indexes on the shared table** — works, and is the smallest diff.
Rejected because it keeps the shared B-tree and blended `ANALYZE` statistics,
partial indexes on it are unusable (measured: not used, with bound parameters
or inlined literals), and it spends the one-time zero-writers window on a patch.

**One database file per namespace** — real attractions, rejected for one
decisive cost: SQLite commits atomically across attached files only in
rollback-journal mode; under WAL, cross-file transactions decay to per-file.
`attn.db` runs `journal_mode=delete` today, but WAL is the standard next move
for a daemon with concurrent readers, and building this primitive's integrity
on "we promise never to enable WAL" is a landmine. The loss bites at registry ↔
DDL, at write ↔ `document.changed` fact (lethal once workflows trigger on
document changes), and at write ↔ job enqueue. The registry maps a namespace to
its tables, so evicting one convicted namespace to its own file later stays a
contained migration — done with a receipt in hand.

Full design, receipts, and reasoning:
`docs/plans/2026-08-03-ext-a3.1-doc-store-physical-schema.md`.

## Verification

`go build ./...` plus `internal/store`, `internal/docstore`, `internal/daemon`,
`cmd/attn` green on the rebased branch. New tests include a planner test
asserting `USING INDEX` and no `TEMP B-TREE` for filter+sort+limit, a witness
for the type-affinity change, one proving two collections with the same field
name do not share storage, and both new daemon subscription-termination cases —
whose falsifiability was checked by reverting the per-delivery re-read and
confirming they fail for the right reason.

Live-verified on a throwaway profile (`a31store`): define / collections / put /
get / query, affinity ordering, the exact built DDL and its five indexes, the
old `documents` table gone, watch deliveries and their termination on undefine,
pagination across three pages of tied documents, the dead-cursor error,
redeclare in both directions, and restart durability with the index surviving.
`doc_3` after several undefine/define cycles confirmed `AUTOINCREMENT` live.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -273,8 +273,12 @@ forgets that position and `--since <RFC3339>` replays from an instant.
   `internal/pty`, not inside the daemon
 - `internal/store`: SQLite plus in-memory cache
 - `internal/bus`: durable event bus (domain facts, per-consumer cursors)
-- `internal/docstore`: document-store query semantics and SQL compilation (no DB
-  handle; `internal/store/documents.go` executes what it compiles)
+- `internal/docstore`: document-store query semantics, SQL compilation, and the
+  physical naming (no DB handle; `internal/store/documents.go` executes what it
+  compiles). A collection is its own table `doc_<id>`, minted from its registry
+  row; a declared field is an indexed generated column over the body. Every
+  identifier the store executes is derived here from an integer or a validated
+  field name — never from caller text
 - `internal/jobs`: durable job queue (retry/backoff, coalescing, commit fence,
   cron entries) — every background duty and every periodic tick runs on it
 - `internal/classifier`: stop-time state classification
--- a/changelog.d/ext-a3.1-doc-store-indexes.yaml
+++ b/changelog.d/ext-a3.1-doc-store-indexes.yaml
@@ -1,0 +1,18 @@
+kind: changed
+area: daemon
+change: >
+  The document store now keeps each collection in its own table, with an
+  indexed column per declared field, so filtering and sorting read an index
+  instead of every document in the collection. Declaring a field still rewrites
+  nothing — the index is built from the bodies already stored — and the query
+  surface, the CLI, and the wire are unchanged. Two behaviours do change: a
+  field's declared type now decides how stored values compare, so a body
+  holding "5" in a number field sorts as the number 5 rather than as text; and
+  a live query whose collection is removed, or redeclared without a field the
+  query uses, now ends with an error saying so instead of quietly reporting an
+  empty result set.
+notes: >
+  The store shipped with no writers, so this replaces its storage rather than
+  migrating it. The measurement that overturned the original scan-first design,
+  and the rejected alternatives including a database file per namespace, are in
+  docs/plans/2026-08-03-ext-a3.1-doc-store-physical-schema.md.
--- a/changelog.d/ext-a3.1-doc-store-indexes.yaml
+++ b/changelog.d/ext-a3.1-doc-store-indexes.yaml
@@ -12,7 +12,8 @@ change: >
   query uses, now ends with an error saying so instead of quietly reporting an
   empty result set.
 notes: >
-  The store shipped with no writers, so this replaces its storage rather than
-  migrating it. The measurement that overturned the original scan-first design,
+  Anything already stored is carried across: declarations, bodies, and
+  timestamps survive, and declared fields come back indexed without being
+  redeclared. The measurement that overturned the original scan-first design,
   and the rejected alternatives including a database file per namespace, are in
   docs/plans/2026-08-03-ext-a3.1-doc-store-physical-schema.md.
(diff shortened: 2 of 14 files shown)


figgyster (dismissed):
Victor, the measurement killed the scan-first story fair and square, and this replacement gives A3 the physical shape it needs before extensions start piling bodies in the file cabinet. The minted-table boundary, generated-column indexes, and live-query re-read are all pointed the right way.

But migration 89 drops both `documents` and `document_collections` outright. #745 is already on `main`, and its shipped `attn doc define` / `put` surface can have created durable records in any installed or named profile; confirming the one production profile had not reached migration 88 does not make those records disposable. Please preserve the v88 declarations and documents while building their new per-collection tables (with an upgrade witness from a populated v88 database), then this can land without taking a match to user state.

figgyster (commented):
Victor, this is the right repair. The v88 carry mints each collection inside the migration transaction, preserves the declarations and raw rows, checks the complete count before dropping the shared table, and refuses to smuggle unchecked field names into DDL. The populated upgrade witness makes the old data-loss hole an eyewitness case, not a promise in a commit message.

I ran that migration witness with the physical-schema regressions in the sealed box; they pass. The principal GitHub board is still starting, so I’m retiring the old red on this head and leaving this as a comment until the other witnesses finish their coffee.

figgyster (approved):
Victor, the physical turn now keeps its promise to the first people who trusted #745 with a record. The v88 carry is atomic, declaration-led, and counted before the old table gets the shovel; the populated upgrade witness pins the exact bad night we found.

Ran the focused upgrade/schema witnesses here, and the full current-head board is green — Daemon, Linux, benchmark, E2E, changelog, and the rest. Approved. One question for the road: does this close the physical-store slice until a real composite-index receipt arrives, or is another A3.1 piece already stalking the hallway?

PR #749, opened 2026-08-03
feat(pty): turn kitty images on by default with pixel geometry and a measured budget

victorarias:
A4 of docs/plans/2026-08-02-terminal-kitty-images.md: turn kitty images on by default. A1–A3 built the machinery feature-dark (worker-only kitty parsing, the split feed, placement events + rendering); this PR fixes every deferred synthesis defect, teaches the PTY its pixel size, and flips the worker's storage limit from 0 to a measured 320MB default.

## Synthesis fixes (all pre-existing on main, all found by the wire-mirror fuzzer)

The worker strips APC bytes from the wire and synthesizes plain control sequences so the client grid stays equal to the worker grid. Seven defect classes were found and fixed, each reproduced under the old code first:

- **Resync predicate**: an unaccounted kitty mutation now forces a resync unless it is a pure removal — an image that appeared and died inside one chunk, or an update the old delta check was blind to, no longer leaves the client stale.
- **Stamp settling**: the kitty generation is settled before every described dispatch, so a described APC can no longer absorb an undescribed one's stamp movement.
- **Relative column synthesis**: the synthesized cursor column moves relatively (CUF/CUB) like the row, instead of absolute CHA — which is wrong inside DECLRMM margins.
- **No-reflow resize**: the worker resizes its grid without reflow (`ghosttyvt.ResizeNoReflow`), matching what every client frame already does. This closed the one-row wrap anchoring drift: it was resize reflow asymmetry between grids, not a counting bug.
- **Tripwire resyncs** for the three grid movements the wire cannot express or measure: a described dispatch while left/right margin mode is on, a scroll-up clamped past the screen height (thresholds measured and pinned by boundary tests), and a dispatch landing on a cursor carrying a deferred wrap. All three are blunt full resyncs by design: no real emitter reaches them (emitters position before transmitting, never enable DECLRMM, never send over-tall placements), so clever synthesis would be unmeasured machinery.

`FuzzKittyWireMirror` soaked 15m/42.5M execs green in the shipping configuration with no knowingly-red class left; the corpus grew 533→581 entries and the wasm parity suite pins client behavior.

## Pixel geometry (protocol 207)

Image emitters size against the pane's device-pixel dimensions, which the PTY never carried — chafa on retina drew at half resolution and then 2x scale. Now:

- `pty_resize`/`pty_resized` carry optional `xpixel`/`ypixel` (total pane device pixels; absent = unknown). The frontend measures `cols × cellWidth × devicePixelRatio` at fit time; the worker derives the cell size, remembers it, and re-derives totals on pixel-less resizes (attach reconcile, remount hydrate).
- The worker terminal learns its cell pixel size (`SetCellPixelSize`), so DEC mode 2048 in-band reports and `TIOCGWINSZ` both answer with real geometry. (Ghostty's VT core answers no XTWINOPS at all — mode 2048 + TIOCGWINSZ is the whole surface, and it is what chafa/timg/icat actually use.)
- The renderer divides placement boxes by devicePixelRatio at the unit boundary (`placementQuad`), so a placement measured in device pixels lands at the right CSS size — sharp on retina, correct on 1x.

## The flip

`kittyStorageLimitDefault = 320MB` — ghostty's own app default, within 5% of kitty's. Receipt: the largest legitimate single image the A4 sweep produced is ~81.4MB (full-screen XDR capture at 2x); 320MB clears it ~4x. The failure past the limit is total, not gradual — ghostty refuses an over-limit image outright and every measured emitter sends q=2, suppressing the response — so the worker now logs a named refusal when a completed transmission moves no kitty generation. `ATTN_KITTY_STORAGE_LIMIT` tunes the cap per session; `0` remains the escape hatch that disables the protocol entirely (no a=q answer). An unparseable value falls back to the default, loudly — a typo in a tuning variable must not silently turn images off. The hub exports the variable to remote daemons, so the disable travels.

## Verification

- Full local suite: go build/vet/test across pty, ghosttyvt, daemon, hub, ptybackend, ptyworker, protocol; Linux amd64/arm64 cross-builds; CGO_ENABLED=0 stub build; tsc; full vitest (2448 passed); wasm parity 37 passed/15 resync-exempt skips (still covered by the Go corpus); both fuzz seed-replays; changelog check.
- Both fuzz targets soaked 15 minutes clean in the shipping configuration.
- The packaged harness scenario was inverted with the flip: its image legs now inject nothing (proving the on-by-default world every user gets) and its dark leg pins the `=0` escape hatch. It ran green end to end on a throwaway profile.
- Live tier on the same profile: a real pane's TIOCGWINSZ answered 1116x1050 px for a 62x25 grid (18x42 device px per cell — 2x the CSS cell); chafa with no size flags placed 522x546 px = 29x13 cells, sharp at device scale; the placement survived a full app quit/relaunch with the blob re-pulled.
- Remote leg (OrbStack VM over the hub relay): endpoint bootstrap deployed the daemon at this branch's fingerprint; the VM PTY reports the same pixel geometry through the relay; a placement described and its blob pulled through the relay; delete emptied the set. Passthrough witness: with the host daemon under `ATTN_KITTY_STORAGE_LIMIT=0`, the relay-ensured VM daemon's environ carried `=0` and a remote session stayed dark.
- The plan's A4 verification record carries the full evidence; the Live-verified comment on this PR names the exact head.

v1 limitations (unchanged from A3, recorded in the plan): images draw above text (z<0 deferred), no grid-tile/unicode-placeholder images, no animation optimization.


figgyster (approved):
You finally turn the kitty work from #741's machinery into a promise somebody can lean on. I like the restraint in the ugly corners: when the mirror cannot honestly measure a movement, you resync instead of teaching the wire another little lie. That is the right kind of boring after the fuzzer has been living under the floorboards.

The default budget, pixel geometry, and no-reflow rule now travel as one contract: an emitter sees real device pixels through both `TIOCGWINSZ` and the worker VT, the renderer crosses back to CSS units once, and a resize no longer lets the worker and client quietly inhabit different row maps. The `=0` hatch also makes it through the relay, which matters; an escape hatch that ends at the first SSH boundary is just decorative hardware.

Read the full A4 path and its regression witnesses. The exact-head packaged local and remote runs back the product seam, and Daemon, Frontend, every E2E shard, both Linux builds, benchmark, React Doctor, changelog, and Changes are green. Approved.

PR #750, opened 2026-08-04
feat(notebook): allow disabling journal narration

victorarias:
## Intent / Why

Journal narration uses the strong-tier background agent and needs an independent pause control for cost-sensitive periods. Turning it off should leave session summaries and context compaction running, while keeping the chosen narration agent and model ready for later.

## Summary

- add the default-on `notebook.narrate_workspace.enabled` setting and surface its effective value to the app
- gate routine, nightly, final-removal, and already queued narration jobs without affecting session summaries or context compaction
- share the Settings toggle rendering across summary and narration duties, preserving each duty configuration while disabled
- document the switch and cover validation, settings snapshots, every enqueue path, queued-job retirement, and UI persistence

## Verification

- full pre-commit gate: Go build/tests, 219 frontend test files (2,455 passed; 15 skipped), production frontend build, Playwright (192 passed; 1 environment-skipped)
- Linux amd64 daemon cross-compile
- isolated packaged-app preflight at protocol 207
- live Settings round trip: disable persisted `false`; fresh relaunch rehydrated the control and enabling persisted `true`; isolated profile cleaned afterward

## Related

- #742 added the independent session-summary switch that this narration control complements

### DIFF
--- a/app/src/components/SettingsModal.test.tsx
+++ b/app/src/components/SettingsModal.test.tsx
@@ -1017,6 +1017,39 @@ describe('SettingsModal keeper', () => {
     );
   });
 
+  it('toggles journal narration independently and defaults it on', async () => {
+    const onSetSetting = renderModal({});
+    fireEvent.click(screen.getByTestId('settings-nav-agents'));
+
+    const toggle = await screen.findByTestId('settings-keeper-narrate-toggle');
+    expect(toggle).toHaveTextContent('Disable');
+    expect(toggle).toHaveAccessibleName('Disable journal narration');
+    fireEvent.click(toggle);
+    expect(onSetSetting).toHaveBeenCalledWith(
+      'notebook.narrate_workspace.enabled',
+      'false',
+    );
+  });
+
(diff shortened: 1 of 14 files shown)


figgyster (commented):
The #742 pattern has made it across to the expensive side of the Notebook without turning the whole keeper off. I traced the routine, nightly, live-removal, and startup-reap paths: each reaches the narration gate, and the handler catches work already sitting in the durable queue before it can launch a strong-tier agent. Summaries and compaction keep their own alibis; the narration agent/model survives the pause.

The focused Settings witness is green here (60 tests), including default-on, disable/re-enable, and configuration preservation. Linux daemon builds, all three frontend E2E shards, frontend, changelog, and the A/B benchmark are green on `18199cf`; the principal Daemon job is still running, so I’m holding the approval until that last witness checks in.

figgyster (approved):
The Daemon witness has checked in. All principal checks are green on `18199cf`, including both Linux daemon builds, the full frontend/E2E board, and the A/B benchmark. The strong-tier narrator now has a real cost brake, not a painted switch. Approved.

PR #751, opened 2026-08-04
fix(terminal): keep hovered links stable during output

victorarias:
## Intent / Why

Terminal output invalidated the hover generation on every write. When the pointer stayed over a link, unrelated spinner or status-line updates could therefore make the underline and cursor flash, and file-path clicks could lose their validated target.

Keep terminal links stable through streaming output while preserving the rule that changed or moved text is no longer clickable.

## Summary

- revalidate the logical line under the stationary pointer inside the terminal's existing paint, and again at the click boundary
- carry an unchanged validated path forward synchronously, using the existing filesystem cache without adding a second repaint loop
- strengthen the terminal interaction regression with a truly stationary Cmd-click, stale-target rejection, and a render-count bound

## Verification

- pre-commit frontend tests: 2,455 passed, 15 skipped
- production frontend build
- pre-commit Playwright suite: 192 passed, 1 skipped
- isolated packaged-app preflight
- native terminal markdown-link scenario: plain click, Cmd-click for two paths, and tile reuse

### DIFF
--- a/app/e2e/terminal-interactions.spec.ts
+++ b/app/e2e/terminal-interactions.spec.ts
@@ -34,6 +34,32 @@ async function writeTerminalOutput(
   }, { id: sessionId, data: output });
 }
 
+async function dispatchStationaryCmdClick(
+  terminal: import('@playwright/test').Locator,
+  position: { x: number; y: number },
+) {
+  const box = await terminal.boundingBox();
+  expect(box).not.toBeNull();
+  await terminal.evaluate((element, point) => {
+    element.dispatchEvent(new MouseEvent('mousedown', {
+      bubbles: true,
+      button: 0,
+      buttons: 1,
+      clientX: point.clientX,
+      clientY: point.clientY,
+      metaKey: true,
+    }));
+    element.dispatchEvent(new MouseEvent('mouseup', {
+      bubbles: true,
+      button: 0,
+      buttons: 0,
+      clientX: point.clientX,
+      clientY: point.clientY,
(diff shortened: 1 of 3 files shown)


figgyster (dismissed):
You’re taking a hover affordance that used to blink like a bad neon sign and making it trustworthy under a terminal firehose. That’s the right instinct. One seam is still writing during render, though, and React has a long memory for those.

`app/src/components/GhosttyTerminal.tsx:3075` — `refreshHoverLinkRef.current = …` mutates a ref while the component renders. React Doctor is right to flag it: concurrent React may replay or discard that render, yet the ref can retain the callback closed over the discarded render’s `detectHoverLink` and dependencies. A subsequent paint or click can then run work React never committed. Please install the callback from an effect (layout effect if this must be ready before paint), keeping render pure; then the rest of the streaming revalidation path has an honest owner.

I read the stationary-pointer regression closely. It’s a proper witness for the link contract, including the stale-target rejection. Once this render-time write is moved out of the render path, I’m happy with the direction.

figgyster (commented):
You moved the refresh closure into `useLayoutEffect` and gave its cleanup an identity check. That puts the callback back on the committed side of React’s line, where a paint or click belongs. The render-time mutation behind my red review is gone.

The stationary-pointer coverage still pins the contract I care about: unrelated output keeps the existing path alive, while changed text loses it. The new CI board is still running, so I’m leaving this as the current-head signal rather than stamping a green light early.

PR #753, opened 2026-08-04
feat(docstore): give every document a revision, and writes an expectation

victorarias:
Closes the one place the document store could lose data silently.

`PutDocument` was a blind upsert. Read-modify-write is the natural shape for much
of what will be built on the store — flip a status, append to a list, bump a
counter — and two writers doing it at once left only the second change, with no
error on either side. Every other refusal in the store is loud: an undeclared
collection, a filter on a field nobody declared, a live query whose collection
can no longer answer it. This was the one silent one, and it lost data rather
than a render. It is also the cheapest thing in the store to fix while it is
young: a column and a `WHERE` clause now, versus a change to every caller later.

## What this adds

Every document carries a **revision**, counting writes to it, starting at 1.
Every read reports it and every write returns the new one, so a caller that read
a document already holds the token naming the version it read. A put or delete
may hand that token back as an expectation:

| `expected_rev` | meaning |
| --- | --- |
| absent | write unconditionally |
| a revision | write only if the document is still at it |
| `0` | write only if the document does not exist |

A failed expectation writes nothing and answers with an error naming both
revisions. Delete takes the same expectation minus `0`: "delete this if it is not
there" is not an assertion a delete can act on, and reading it as unconditional
would delete the document the caller was protecting.

From the CLI:

```
$ attn doc put ext/approval-gate requests r1 '{"status":"open"}'
wrote ext/approval-gate/requests/r1 (rev 1)
$ attn doc put ext/approval-gate requests r1 '{"status":"done"}' --expect 1
wrote ext/approval-gate/requests/r1 (rev 2)
$ attn doc put ext/approval-gate requests r1 '{"status":"other"}' --expect 1
doc put: docstore: ext/approval-gate/requests/r1 expected rev 1 but is at rev 2;
re-read it and apply your change to that version
```

`--expect absent` is create-only. `attn doc get`, `query` and `watch` all report
the revision.

## How the check is atomic

Each form is **one statement**, so SQLite runs it in its own transaction and
there is no read-then-write window to lose: an upsert when there is no
expectation, an insert with `ON CONFLICT DO NOTHING` for create-only, and an
`UPDATE ... WHERE rev = ?` with no insert path when a revision is expected. All
three use `RETURNING rev`, so every refusal arrives the same way — no row came
back — and the store re-reads only on that failure path, to name the write the
caller lost to.

A refused write publishes no change fact, so it does not wake the collection's
live queries to re-render a result set that did not change.

## Decisions worth knowing

**Optional, not mandatory.** A write with no expectation still always lands.
Forcing one would make "set this to that" read the document first for no safety.
Safety by default belongs one layer up: the sidecar SDK should expose an
`update(id, fn)` that loops read → modify → write-at-revision and retries the
conflict, so extension code gets this without knowing the word.

**`safeint`, not `int32` or `int64`.** `int64` codegens as a TypeScript *string*
in this pipeline — `main.tsp` already says so where epoch-ms uses `safeint` — and
a revision arriving as `"5"` is one somebody compares with `<` and gets wrong.
`int32` is 2.1 billion writes to *one* document: about seven years at ten writes
a second, which attn's uptime can reach, and a wrapped revision makes a stale
check *pass*, which is the exact failure this exists to prevent. `safeint` is
2^53-1 and costs nothing — SQLite's INTEGER is up to 8 bytes whatever it is
declared as, and it generates Go `int` and TypeScript `number`.

**`DocSubscribeResult.revision` is renamed `delivery`.** It counts deliveries on
a subscription, unrelated to a document's revision, and one word for both reads
as though they were connected. Nothing consumed it — it reached `generated.ts`
and no app code — so it was free now and permanent later. It stays `int32`:
bounded by how long one connection is open.

## Migration 90

One `ALTER TABLE doc_<id> ADD COLUMN rev INTEGER NOT NULL DEFAULT 1` per registry
row, walking the registry rather than the schema so a table merely *named* like a
collection's is not picked up, and guarded per table so a rewound
`schema_migrations` finishes a partial run. Documents already stored start at
revision 1 — a revision is the version a reader was handed, and every stored
document has been readable all along at what is now its first.

## Verification

Both mutations below were applied, seen red, and reverted.

- **Contention, against a database file** (not the in-memory store, which is
  pinned to one connection and would serialise the writers at the driver): 8
  goroutines × 25 read-modify-writes on one document, each retrying on conflict.
  The counter must end at 200. Removing the expectation loses 103–163 updates per
  run.
- **Migration witness**: a database seeded at migration 89's shape with documents
  in it, watermark rewound, reopened. Documents come back at revision 1, bodies
  unrewritten, and a conditional write against the carried revision is accepted
  while a second one at the same revision is refused. Defaulting the column to 0
  instead fails with `a carried document is at rev 0, want 1`.
- **Live**, on a throwaway `docrev` profile built from this branch (preflight
  PASS, protocol matched across CLI, app and daemon): every row of the table
  above through the real CLI, schema at migration 90 on disk, and a `doc watch`
  that saw exactly 2 deliveries across one refused and one accepted write — the
  refused one did not wake it. Profile cleaned up; production `~/.attn` stayed at
  schema 87 throughout.
- Store, docstore, daemon, client, CLI and protocol suites; frontend `tsc` and
  2455 unit tests.

## Surfaces

CLI (`--expect` on put and delete, revision in get/query/watch output), daemon
handlers, protocol (`ProtocolVersion` 208 in both places, `generated.*`
regenerated), glossary, changelog, and a plan doc. No SDK, plugin, harness or
frontend consumer of the document store exists yet, so there was nothing else to
move.

## Known bound

Carried from A3.1: a query filtering on one field and sorting by a different one
still uses a temp B-tree over the matched rows. One composite index per
(filter, sort) pair is combinatorial, so it waits for a receipt.

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -197,7 +197,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '207';
+export const PROTOCOL_VERSION = '208';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 // AutomationActionTimeoutError distinguishes "the daemon never sent a
--- a/app/src/types/generated.ts
+++ b/app/src/types/generated.ts
@@ -1413,10 +1413,11 @@ export interface DocDefineResult {
 }
 
 export interface DocDeleteMessage {
-    cmd:        DocDeleteMessageCmd;
-    collection: string;
-    id:         string;
-    namespace:  string;
+    cmd:           DocDeleteMessageCmd;
+    collection:    string;
+    expected_rev?: number;
+    id:            string;
+    namespace:     string;
     [property: string]: any;
 }
 
@@ -1454,16 +1455,18 @@ export interface Document {
     body:       string;
     created_at: string;
     id:         string;
+    rev:        number;
     updated_at: string;
     [property: string]: any;
 }
 
 export interface DocPutMessage {
-    body:       string;
-    cmd:        DocPutMessageCmd;
-    collection: string;
-    id:         string;
-    namespace:  string;
+    body:          string;
+    cmd:           DocPutMessageCmd;
+    collection:    string;
+    expected_rev?: number;
+    id:            string;
+    namespace:     string;
     [property: string]: any;
 }
 
@@ -1475,6 +1478,7 @@ export interface DocPutResult {
     collection: string;
     id:         string;
     namespace:  string;
+    rev:        number;
     [property: string]: any;
 }
 
@@ -1527,8 +1531,8 @@ export enum DocSubscribeMessageCmd {
 }
 
 export interface DocSubscribeResult {
+    delivery:  number;
     documents: Document[];
-    revision:  number;
     [property: string]: any;
 }
 
@@ -4272,6 +4276,7 @@ export interface DocPutResultObject {
     collection: string;
     id:         string;
     namespace:  string;
+    rev:        number;
     [property: string]: any;
 }
 
@@ -4281,8 +4286,8 @@ export interface DocQueryResultObject {
 }
 
 export interface DocSubscribeResultObject {
+    delivery:  number;
     documents: Document[];
-    revision:  number;
     [property: string]: any;
 }
 
@@ -5098,6 +5103,7 @@ export interface StoredDocument {
     body:       string;
     created_at: string;
     id:         string;
+    rev:        number;
     updated_at: string;
     [property: string]: any;
 }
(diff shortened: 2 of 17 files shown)


figgyster (approved):
PR #748 gave A3 a physical store that keeps durable records; this gives it a memory strong enough to say no when two hands grab the same record. The one-statement compare-and-write, first-rev migration, and silent refused-write path close the lost-update hole without teaching subscribers to hallucinate change.

Ran the full store suite here, including the file-backed contention and v89→v90 witnesses. The GitHub board is green through Daemon, frontend/E2E, both Linux builds, React Doctor, and the A/B bench.

Approved. One trail marker for the next chapter: does this finish document-store safety before A4 teaches SDK callers the retry loop, or is another A3 slice still coming?

PR #754, opened 2026-08-04
feat(terminal): draw negative-z kitty images under the text

victorarias:
Kitty images shipped in #749 with a named v1 limitation: every image drew above the text, because the renderer had a single cell draw call with nowhere to slot an image under the glyphs. This removes that limitation.

## What changed

The WebGL renderer's cell pass is now two passes — non-default cell backgrounds, then everything else (glyphs, underlines, selection tints, the cursor block) — and each frame draws image placements in kitty's three z tiers around them:

- `z >= 0`: above the text (unchanged behavior).
- negative `z` down to INT32_MIN/2: under the text, over cell backgrounds.
- below `KITTY_Z_UNDER_BACKGROUND` (−1,073,741,824, the spec's INT32_MIN/2; "below" is strict): under non-default cell backgrounds too.

Two placement decisions are deliberate: the cursor block and overlay tints (selection, search) ride the foreground pass, because in the background pass an under-text image would paint over them — an invisible cursor or a vanishing selection.

No protocol or worker change: the wire has carried each placement's `z` (signed int32) and the client store has sorted by it since A3. `WebGlImageQuad` gains the `z` field and the renderer partitions by it; outlines still draw last.

## Verification

- Five new pass-ordering tests read the frame's draw order off a recording GL stub (image draws are the ones with a non-atlas texture bound): under-text ordering, the strict threshold boundary at ±1, z=0 unchanged, cursor drawn after an under-text image, selection tint drawn after an under-text image. Full frontend suite (2460 tests) and tsc green.
- The packaged scenario `real-app:scenario-terminal-kitty-image` now transmits its image at `z=-1` and asserts the placement reports `z === -1`, proving the signed int32 survives worker → protocol → store. Green on a throwaway profile at this head (both legs, including the `ATTN_KITTY_STORAGE_LIMIT=0` dark leg).
- Live visual proof on the same profile: the same five lines of bold text and the same checkerboard captured twice, differing only in `z`. At `z=-1` the text is readable on top of the image; at `z=0` the image hides it.

### DIFF
--- a/app/scripts/real-app-harness/scenario-terminal-kitty-image.mjs
+++ b/app/scripts/real-app-harness/scenario-terminal-kitty-image.mjs
@@ -64,6 +64,11 @@ const IMAGE_HEIGHT = 48;
 // multi-escape path rather than a single-escape shape no real emitter sends.
 const PAYLOAD_CHUNK = 4096;
 const IMAGE_ID = 8801;
+// A negative z-index, which kitty defines as "draw this under the text". It is
+// the one placement field that is a SIGNED int32 all the way from ghostty's
+// storage through the protocol to the client's store, so asserting it comes
+// back as -1 is what proves nothing on that path widened or truncated it.
+const IMAGE_Z = -1;
 
 const BLOB_TIMEOUT_MS = 20_000;
 const PLACEMENT_TIMEOUT_MS = 20_000;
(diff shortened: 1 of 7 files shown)


figgyster (approved):
This finishes the piece #749 left on the pavement. You split the cell paint at the one seam that matters: a deep negative image slips below colored cells, the ordinary negative tier sits under glyphs, and cursor/selection/outlines keep their visibility instead of becoming collateral damage.

The renderer-order witnesses pin all three tiers, including the strict `INT32_MIN/2` boundary, and your packaged z=-1 run proves the signed value reaches the screen. The full Frontend and E2E board is green at `01e4164`. Approved.

PR #755, opened 2026-08-04
fix(sessions): recover the evidence behind a session, not just its state

victorarias:
Restarting the daemon reset every live session to `launching` and left it there. Worse, it silently swallowed snoozes: an agent you deferred left the Snoozed section at its deadline without ever reaching the queue.

## What was wrong

Three things compounded.

**Reconcile overwrote the persisted state.** Every live session was stomped to `launching` on the theory that a recovered agent had not proved itself yet.

**Nothing could correct it.** The evidence table is in-memory. An agent parked at its prompt writes nothing to its PTY, so no fresh observation ever arrived and the session stayed `launching` indefinitely — production had sessions stuck for over two hours. Only busy agents recovered, because they repaint their title within a second, which is why the bug looked intermittent.

**The worker invented `working`.** `infoResult()` defaulted a blank cached state to `"working"`, a leftover of the deleted screen scraper. That fabricated claim then competed with real evidence.

The snooze loss is downstream of the first. A snooze only opens a turn if the agent is in a state that wants the user. With every agent in `launching` at the deadline, the deferral was cashed and nothing was delivered. Deadlines already survived a restart; the state they are delivered against did not.

## The fix: recover the evidence, not the state

A restart is not news about an agent. The state attn last concluded now stands, and what gets rebuilt is the evidence behind it — from the two sources that outlived the daemon.

**The level.** The worker outlives the daemon, so it now retains the most recent title-heartbeat observation and hands it back over the info RPC. Recovery replays it through the ordinary `recordPTYEvidence` path — codex announces approvals in its title, and recovery must not grow a second interpretation of that. The resolver can then confirm or correct the state on its first tick instead of having nothing to go on.

**The harness edge.** A session persisted as `pending_approval` or `waiting_input` is itself the record of an outstanding edge, so it is re-seeded at the session's `state_since`. It is dropped when the worker's level is newer, so an agent that visibly moved on while the daemon was down does not come back stuck waiting.

The hook-driven bracket pair is deliberately **not** reconstructed. Inventing an open bracket would pin a session `working` on a closing hook that can never arrive — a worse failure than the one being fixed, because nothing would ever retire it.

Two rules changed as a consequence:

- **Recovery writes state only where it has an opinion.** Worker exited → `idle`. Driver reads a real PTY claim → that claim. A session attn has never seen → `launching`. Otherwise the persisted state is left alone. The one carve-out: a live worker contradicting a state the resolver does not own (`recoverable`) falls back to `launching`, because nothing else can ever move a session out of it.
- **A `worker_info` claim may now only end `launching`.** That veto already existed for shells; no worker caches a real state any more, so it is now general.

## Residual gap

Hook-derived evidence is still not persisted. An agent that hits an approval *while the daemon is down* leaves no record, so recovery can only work from what was persisted before the restart. Not addressed here.

## Verification

Live, on throwaway profile `igu1`, a real claude session, two real daemon restarts:

- **Restart 1** — state and snooze both survived. `worker_info` logged `vetoed reason=resolver_owned`; no `startup_recovery claim=launching` appeared at all.
- **Restart 2** — a staged stale `working` was corrected to `idle` with reason `at_prompt` within one second. The agent was silent throughout, so that transition could only come from the seeded worker heartbeat.

8 new tests in `internal/daemon/session_recovered_evidence_test.go`, each mutation-tested to confirm which assert actually catches the regression. `GOOS=linux` cross-build green for `cmd/...` and `internal/...`.

Surfaces walked: daemon and app (state broadcasts), worker RPC (`InfoResult` gained omitempty fields; not the app protocol, so no `ProtocolVersion` bump), both pty backends (worker and embedded), all three agent drivers, Linux cross-build, changelog fragment. No CLI or frontend surface — recovery has no command-line or UI entry point.

https://claude.ai/code/session_016u4HyJxy246ypDNyuTuL6W

### DIFF
--- a/changelog.d/snappy-iguana-recovered-session-state.yaml
+++ b/changelog.d/snappy-iguana-recovered-session-state.yaml
@@ -1,0 +1,28 @@
+kind: fixed
+area: sessions
+change: >
+  Restarting the daemon no longer resets every running agent to "launching".
+  Recovery used to overwrite the state of each live session on the theory that
+  a recovered agent had not proved itself yet, and then had nothing left to
+  correct it with: an agent parked at its prompt writes nothing to its
+  terminal, so no fresh evidence ever arrived and it stayed "launching"
+  indefinitely — measured at over two hours on sessions that were simply
+  waiting to be read. Only busy agents recovered, because they repaint their
+  title within a second. A restart is not news about an agent, so the state
+  attn last concluded now stands, and the evidence behind it is recovered
+  rather than discarded: the worker still holds the agent's latest title
+  heartbeat and reports it back, so the resolver can confirm or correct that
+  state on its first tick instead of having nothing to go on. An agent that
+  finished while the daemon was down settles and reaches your queue; one that
+  was waiting on an approval or a question comes back still waiting, unless it
+  visibly moved on in the meantime.
+---
+kind: fixed
+area: queue
+change: >
+  A snoozed agent now comes back. A snooze only opens a turn if the agent is in
+  a state that wants you, and a daemon restart used to leave every agent in
+  "launching" — so when the deadline arrived the deferral was cashed and
+  nothing was delivered: the agent silently left the Snoozed section without
+  ever reaching your queue. Deadlines already survived a restart; now the state
+  they are delivered against does too.
--- a/internal/agent/claude.go
+++ b/internal/agent/claude.go
@@ -560,15 +560,8 @@ func (c *Claude) RecoverOnMissingPTY() bool {
 // RecoveredRunningState mirrors the default recovered-state mapping. Claude has
 // no special recovery needs; the method exists to satisfy
 // RecoveredStatePolicyProvider.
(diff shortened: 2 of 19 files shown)


figgyster (approved):
You found the kind of restart bug that makes an attention hub lie quietly: it forgets why it reached a state, then cashes a snooze without ever returning the agent to the queue. Recovering the worker-held level and only the persisted human-blocked edge puts the authority back where it belongs, and it carries the recent state-machine work forward without inventing a bracket that can never close.

The recovery witnesses pin the important cases: stale `working` settles from the restored heartbeat, approval/question states survive unless newer worker evidence disproves them, and the snooze wake reaches an actual turn. I also walked the worker/embedded transport seam and the resolver path; the focused recovery suite passed in the sandbox. The complete GitHub board is green, including Daemon, both Linux preflights, and the A/B benchmark; the real isolated-profile restart proof covers the lifecycle seam the container cannot reproduce without the native Ghostty headers.

Approved. The city remembers what it saw now.

PR #756, opened 2026-08-04
feat(queue): pause settling while the user is typing

victorarias:
## What

attn can close a turn on its own once the user has steered an agent and it has
gone back to work: an invisible arm delay proves the steering took, then a
visible countdown the user can stop with `Cmd+.`. That countdown ran regardless
of what the user was doing, so a turn could close while they were still at the
keyboard writing the next thing.

A genuine keystroke into the session's PTY now **freezes** the countdown where
it stands. The daemon withdraws the deadline and broadcasts `auto_settle_held`
instead; the pane header and the sidebar draw a still, full bar labelled
"Settling paused". Five quiet seconds after the last keystroke the hold releases
into a **whole** countdown, read fresh from settings — not the remainder that
was left when typing began.

## Why this shape

attn has no visibility inside an agent's TUI, so it cannot tell composing from
erasing from pressing Escape — and it does not try. A keystroke means *the
user's hands are on this session*, which is the entire question a pending settle
is asking. The one distinction that does matter — the user versus attn typing on
their behalf — was already drawn: `handlePtyInput` tags automation and
attach-replay writes, and daemon-side writes (doorbells, delegation briefs)
never pass through it at all, so they cannot hold by construction.

Three details carry the design:

- **Typing during the arm delay holds it too, but silently**, and resumes into
  arming rather than into a countdown it never earned. Freezing something that
  was never announced would announce it.
- **The fire-time guard repeats the quiet check where the settle happens**,
  catching a timer that fired in the microseconds around a keystroke before the
  hold could stop it. That interleaving is the only one that closes a turn under
  the user's hands.
- **A hold is not a cancel.** `Cmd+.` keeps the turn and stands until the session
  leaves `working`; a hold expires on its own, because the user stopping typing
  is not the user answering.

## The 5s number

The quiet window is a constant, not a third setting. It measures typing gaps —
pausing to read the screen mid-prompt — not agent behavior, so it is orthogonal
to the two configurable windows. It can afford to be short because the thing it
releases into *is* the grace period: visible, and cancellable for its whole
configured length. Being wrong here does not settle a turn silently.

Cost control: only the first keystroke of a burst does any work. A re-hold, and
a hold of the invisible arm delay, broadcast nothing.

## Verification

Live-verified in a packaged non-production build (isolated profile), plus a new
harness scenario `real-app:scenario-settle-typing-hold` that asserts the wire
and the rendered header together across five legs:

1. countdown running — `auto_settle_fires_at` set, header reads "Settling…";
2. one keystroke — `auto_settle_held`, no deadline, header reads "Settling
   paused" with a frozen bar;
3. typing across two quiet windows — still held, never resumed under the hands;
4. going quiet — resumes with a deadline a **full** countdown out, not the
   remainder;
5. an automation `write_pane` to the same pane — holds nothing.

The rendered-DOM readout it asserts against is a new `settling` field on the UI
automation bridge (`text`, `held`, `frozenBar`), so the header is checked as the
user sees it rather than inferred from state.

Also: 6 daemon tests covering the freeze, the fire-time race, the invisible
arm-delay hold, hold-vs-cancel, hold clearing when the session leaves `working`,
and automation input; frontend tests for both indicators.

Protocol `208 → 209` (`auto_settle_held`, mutually exclusive with
`auto_settle_fires_at`).

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -69,6 +69,7 @@
     "real-app:scenario-agent-queue-snooze": "node scripts/real-app-harness/scenario-agent-queue-snooze.mjs",
     "real-app:scenario-nudge-trigger": "node scripts/real-app-harness/scenario-nudge-trigger.mjs",
     "real-app:scenario-countdown-cancel": "node scripts/real-app-harness/scenario-countdown-cancel.mjs",
+    "real-app:scenario-settle-typing-hold": "node scripts/real-app-harness/scenario-settle-typing-hold.mjs",
     "real-app:scenario-terminal-block-copy": "node scripts/real-app-harness/scenario-terminal-block-copy.mjs",
     "real-app:scenario-terminal-context-menu": "node scripts/real-app-harness/scenario-terminal-context-menu.mjs",
     "real-app:scenario-terminal-osc8-link": "node scripts/real-app-harness/scenario-terminal-osc8-link.mjs",
--- a/app/scripts/real-app-harness/scenario-settle-typing-hold.mjs
+++ b/app/scripts/real-app-harness/scenario-settle-typing-hold.mjs
@@ -1,0 +1,339 @@
+#!/usr/bin/env node
+
+/**
+ * Real-app scenario: typing to an agent freezes its settling countdown.
+ *
+ * attn closes a turn for the user once they have steered an agent and it has
+ * gone back to work. The countdown that does it used to run regardless of what
+ * the user was doing, so a turn could close while they were still at the
+ * keyboard writing the next thing. A keystroke now freezes it where it stands
+ * and it stays frozen until five quiet seconds have passed.
+ *
+ * The whole mechanism lives between a real keystroke's route into the daemon and
+ * a timer the daemon owns, which is why it is worth a packaged-app run. Two
+ * writes reach a pane and only one of them is a person: the terminal's own input
+ * path arrives untagged and holds, and an automation write is tagged and must
+ * not — otherwise a delegated agent driving a pane would pin every countdown in
+ * the app open. A unit test can assert both sides of that filter, but only the
+ * real app proves the frontend's typing path is still the untagged one.
+ *
+ * The run asserts, in one app lifecycle:
+ *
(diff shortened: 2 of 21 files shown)


figgyster (dismissed):
Victor, this is the right kind of queue work: you found the place where a helpful timer can turn into a thief and gave the user a visible, cancellable pause instead. The daemon-owned hold, the mutually exclusive wire state, and the real-input harness all follow the same story. One last race is still standing in the doorway.

Blocking: the fire-time typing guard is not atomic with recording the next keystroke. `autoSettleFire` removes the timer entry before `runAutoSettle`; after line 404 observes a quiet session, a real `pty_input` can run before line 446. It records the input at `ws_pty.go:622`, then `holdAutoSettle` finds no entry to replace, and this path still calls `SettleTurn` — precisely a turn closing under the user's hands. Please serialize the input stamp/hold with the final quiet check and settle (or retain a firing entry that input can mark), and add a regression that injects `handlePtyInput` after the quiet check but before the settle. The current race test only places the timestamp before `autoSettleFire`, so it never enters this gap.

figgyster (commented):
Victor, that was the last loose floorboard. `settleIfQuiet` now holds the same keystroke mutex from the final quiet read through `SettleTurn`, and the new witness sends a whole `pty_input` inside the former gap rather than slipping in a timestamp by hand. The countdown can no longer decide the room is empty while the user's key is already in the lock.

The blocker is cleared on `70d5e8a2`. I’m leaving this as a current-head comment while the fresh CI board is still assembling; the focused daemon command could not build in my review sandbox because its native Ghostty header is unavailable there.

figgyster (approved):
Victor, this one fought its way out of the timer swamp properly. The final quiet read and `SettleTurn` now share the input lock, the regression walks a real keystroke through the former gap, and the packaged-app proof says the pause still feels like a pause rather than a theory.

The full Daemon, frontend, E2E, Linux, benchmark, changelog, and React Doctor board is green on `70d5e8a2`. Approved.

PR #757, opened 2026-08-04
docs(plans): delete thirty-one stale plan documents after a full triage

victorarias:
docs/plans had grown to 82 documents mixing three very different things: active plans, decision records worth keeping forever, and process residue — per-PR trackers, review-round logs, and designs a later plan superseded. This deletes the residue after a full triage of all 82.

## How the triage worked

Every doc was classified against the code, not its own status line (several docs claim "not started" for shipped work, and one — the chief-awareness-wave umbrella — marked all twelve of its plans unstarted while five had shipped). Verdicts:

- **31 deleted** (this PR): per-PR trackers that outlived their PR; shipped specs whose durable decisions are restated in code comments, `plugins/attn-opencode/README.md`, or AGENTS.md; designs superseded by a named newer doc (automations v2, design-artifact-handover, the Present vision, the home dashboard); and five parked architecture-review plans from 2026-07-02 that Victor chose to drop. Git history keeps all of them.
- **37 kept as decision records**: docs holding measurements, rejected alternatives, or invariants nothing else states (e.g. the vt10x parity findings, the state-detection timing receipts, the db-loss incident record).
- **14 kept as active work**: the ext-* extension-platform series, the kitty-images plan, and unexecuted designs whose premises were re-verified as still true in the code.

Unexecuted plans from the last three weeks were explicitly checked and none are deleted — everything removed is either executed, superseded, or over a month old by git history.

## Rider fixes

- References to deleted docs in five surviving plan docs and two vision docs are rewritten as plain text, so nothing points at a missing file.
- The comment on `rotateContinuityBindingsIfContractChanged` (internal/daemon/automations_definitions.go) cited "A1" in the deleted automations v1 doc — a section that never existed there. It now points at the v2 doc's Data Model section, which actually defines the rotation contract.
- `2026-08-01-ext-b1-job-queue.md`'s status line said "Not started" for work that shipped (`internal/jobs` exists, `internal/tasks` is deleted); it now says shipped.

Comment/docs-only change: `go run ./cmd/changelog-check` and `go build ./internal/daemon/` pass; no behavior is touched, so no live verification tier applies.


figgyster (commented):
You took a 6,000-line attic full of old floor plans and kept the load-bearing notes. That is real maintenance work, not broom theatre: the surviving docs no longer name any deleted plan, and the automations comment now points to the v2 Data Model that actually defines `contract_rotated` and `ContinuationContract`.

`git diff --check` and the sandboxed changelog check passed; the B1 status also matches the tree (`internal/jobs` exists and `internal/tasks` is gone). The current Daemon failure is the untouched `internal/ptyworker` witness `TestProcessHasArgIdentifiesOwnProcess`; every other listed package passed. I found no patch-level blocker, but I’m leaving this as a comment while that required CI job is red. Give it a green rerun and this file can go back into the cabinet.

figgyster (approved):
The rain finally stopped. The Daemon rerun, Linux preflight, benchmark, changelog, and changed-path checks are all green on `3c31d53`, and the earlier `ptyworker` process-probe failure did not return.

This is safe to merge. You cut the stale-plan shelf back to the decisions that still carry weight, and the few live references now tell the truth about where the bodies are buried.

PR #758, opened 2026-08-04
test(docstore): check queries against a dumb second path

victorarias:
The document query compiler is the fast path — a generated column per declared
field, an index on it, and a tuple comparison implementing the `After` cursor.
The example tests prove the rules on cases someone thought of. The gap they
structurally have is **combinations**: each test sets up one shape, and the bug
that survives them all is where two shapes meet — a tie group that also spans a
NULL boundary, at a page edge, in descending order, with a filter that removes
half the group.

So this asks the same question a second way and checks the answers match.

## The dumb path

A shadow table carries the same declared affinities with no generated columns
and no indexes, filled by one `INSERT ... SELECT` out of the real table. Queries
against it are hand-written, and paging is applied afterwards by slicing a Go
array. Everything the real path is clever about, the shadow does the slow
obvious way.

It is deliberately **not** a Go re-implementation of the query semantics. The
fiddly rules here are SQLite's — how `"5"` compares against `5`, where a missing
value sorts — and writing our understanding of them into Go would mostly test
that understanding, then cost an afternoon each time the test turned out to be
the wrong one. Both answers come out of SQLite, so nothing here models SQLite;
what is compared is only the machinery built on top of it.

The cursor gets the best of this. The shadow does not reimplement the tuple
comparison — it takes the whole ordered list, finds the anchor's position, and
slices. The hairiest code in the compiler ends up checked against something that
cannot be subtly wrong.

One thing that fell out immediately: the `After` anchor is a *position in the
order*, not a member of the result. A query filtered to documents the anchor
itself does not match still pages from where the anchor sits. The first draft of
the dumb path got that wrong and the harness caught it on its first run.

## Three legs

**Every small corpus.** Every arrangement of a six-value alphabet across three
documents, against every query: 216 × 396 = **85,536 checks**. The alphabet is
the ragged cases the compiler claims to handle — a number, that number as text,
JSON null, the field absent, a list.

Exhaustive rather than random on purpose. No seed, so nothing passes by luck and
no bug waits for someone to change a number; it cannot be flaky; and a failure
is already minimal, so the case that broke is one you can read instead of one you
have to shrink.

Three is measured, not assumed. The corpus size is the sweep's one dial and it is
steep — four documents is 855,360 checks, twelve times the cost — and **every
mutation in the table below is caught at three as well as at four**, including
both that the example suite misses. `ATTN_DOCSTORE_SWEEP` turns it up for a soak;
five documents is 7,698,240 checks and passes.

**A large corpus.** 4000 documents over five fixed seeds. Tiny is a regime: with
four rows SQLite reads them all, so the sweep never touches an index. At this
size the planner picks the index for ~88% of queries — a different code path over
the same declared column, and the only place an index that disagreed with the
column it indexes could show. Each query is also run through the compiler's own
statement with `NOT INDEXED`, and **the leg fails if no query reached an index**,
because that third comparison asks an identical question twice when none does.

**A moving corpus.** Documents written, rewritten, deleted, and the collection
redeclared — a field changing type, leaving, and coming back — with a full
page-by-page walk checked after every step. Redeclaring is DDL over live
documents, which neither frozen leg can see. The leg fails if a run never
redeclared.

## Falsified, not assumed

A green differential test is worth nothing until it has been seen red. Eight
compiling mutations of the compiler and the store were applied, observed
failing, and reverted:

| mutation | existing suite | this harness |
| --- | --- | --- |
| cursor comparison loosened to `>=` | catches | catches |
| descending tiebreaker direction flipped | catches | catches |
| id tiebreaker dropped for null-valued anchors | catches | catches |
| cursor clause parens dropped | string assertion only | catches, as a filtered query returning non-matching rows |
| descending cursor parens misplaced | string assertion only | catches, same class |
| cursor rewritten as `sort >= v AND id > ?`, **with the string test updated to match** | catches | catches |
| bools bound as text | catches | large leg only |
| a field's type-change made a silent no-op | catches | moving leg only |

Two things worth reading out of that table. The existing suite is stronger than
I assumed going in — it catches most of these. And where it catches a compiler
mutation only through a *string assertion* on the generated SQL, that is a change
detector, not a correctness check: it breaks on any honest refactor, and the
refactor's author updates it and moves on. Row six is that scenario played out
deliberately, and the harness is what still fails.

The three mutations the existing suite misses behaviourally all live in the same
place: **filter combined with cursor**. Scanning the suite for a query literal
carrying both a filter and an `after` finds none. The sweep covers 270 such
pairings per corpus, across all 216 corpora.

## Cost

Test-only — no product code changes, so the package cost is the whole cost.
`internal/store` goes **6.7s → 10.1s** locally.

The CI number is what set the default. A first push ran the sweep at four
documents and took `internal/store` from 29.5s on main to 87.9s — and the Daemon
check, at 117s, sits under a Frontend E2E job that paces the build at 180s. Four
documents would have made the two co-critical and left the Go check no headroom,
for coverage measured to be redundant. Two cheap wins came out of looking: the
cursor anchor is now read once per corpus rather than once per check, and the
sweep skips the `NOT INDEXED` comparison, which asks an identical question when
the corpus is too small for the planner to want an index.

## Verification

Test-only change with no daemon lifecycle, protocol, PTY, background-runner or
UI surface, so no live app verification applies. Everything is in-memory: the
harness never resolves a config path, so it cannot reach a real data dir.

`go build ./...`, the full Go suite, `go vet`, and `gofmt` are clean.

## What this does not cover

If the shadow and the compiler embody the same *misunderstanding* — both
ordering a way we did not mean — they agree and are both wrong. Nothing
automatic catches that; only a human-reviewed expected answer does. The named
example tests are the intent layer and stay exactly where they are. This covers
the combinations they structurally cannot.

### DIFF
--- a/changelog.d/docstore-differential-query-harness.yaml
+++ b/changelog.d/docstore-differential-query-harness.yaml
@@ -1,0 +1,11 @@
+kind: internal
+area: daemon
+change: >
+  Document queries are now checked against a second, deliberately dumb query
+  path — a shadow table with no generated columns and no indexes, paginated by
+  slicing — so the compiler's SQL, its index and its cursor are compared against
+  an answer that cannot be subtly wrong. Three legs: every arrangement of a
+  six-value alphabet across three documents (85,536 checks, no random seed),
+  a four-thousand document corpus large enough that SQLite reaches for the
+  index, and a corpus written, deleted and redeclared while the queries run.
+  ATTN_DOCSTORE_SWEEP widens the exhaustive leg for a soak run.
--- a/internal/store/documents_model_test.go
+++ b/internal/store/documents_model_test.go
@@ -1,0 +1,1027 @@
+package store
+
+import (
+	"fmt"
+	"math/rand"
+	"os"
+	"sort"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/victorarias/attn/internal/docstore"
+)
+
+// Differential harness for the document query compiler.
+//
+// The example tests above prove the query rules on cases someone thought of.
+// This asks the same question a second way and checks the two answers match,
+// over combinations nobody enumerated by hand.
+//
+// The second way is deliberately dumb: a shadow table with the same declared
+// affinities but no generated columns and no indexes, populated by SQL out of
+// the real table, queried with a hand-written statement and paginated by slicing
+// a Go array. Everything the real path is clever about — the generated column,
+// its index, and the tuple comparison that implements the After cursor — the
+// shadow does the slow obvious way instead.
+//
+// It is deliberately NOT a Go re-implementation of the query semantics. The
+// fiddly rules here are SQLite's — how "5" compares against 5, where a missing
+// value sorts — and writing down our understanding of them in Go would mostly
+// test that understanding. Both answers come out of SQLite, so nothing here
+// models SQLite; what is compared is only the machinery we built on top of it.
+//
+// Three legs, each covering what the others structurally cannot:
+//
+//   - the sweep, which is every arrangement of a small alphabet across four
+//     documents against every query — exhaustive, so no seed and no luck;
+//   - the large corpus, big enough that SQLite reaches for an index, which is a
+//     different code path over the same column and only exists at size;
+//   - the moving corpus, where documents are written, deleted and the collection
+//     redeclared while the queries run.
+//
+// What none of them catch: a misunderstanding shared by both paths. If the
+// shadow orders the way the compiler orders and both are not what we meant, they
+// agree and are both wrong. The named example tests above are what pin the
+// intent; this covers the combinations those tests structurally cannot.
+
+// modelField is the one declared field the exhaustive sweep varies. A single
+// number field is the richest choice for the money: "number" is the type whose
+// column affinity actually converts, so an alphabet over it reaches text that
+// looks numeric, text that does not, and values with no numeric reading at all.
+const modelField = "n"
+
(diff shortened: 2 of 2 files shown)


figgyster (commented):
Victor, you made the docstore prove its smart path against a slow witness instead of admiring its own SQL strings. That carries #753's durable revision work in the right direction: storage contracts get witnesses, especially at compound query edges where clever code likes to lie.

The 855,360-case sweep plus the indexed and moving-corpus legs passed here in 23.2s, and `git diff --check` found no loose threads. GitHub's required Daemon job is red only at the untouched `internal/ptyworker` process probe (`TestProcessHasArgIdentifiesOwnProcess`); `internal/store` passed there, so I found no patch blocker. Leaving this as a comment until that required run turns green.

figgyster (commented):
Victor, this is the right kind of cost cut: you measured the witness, kept the cases that catch the known crimes, and left the larger sweep one named knob away instead of making every daemon run pay for a fifth-alarm drill.

I reran the three-leg harness on this head in 4.8s, then widened the exhaustive leg with `ATTN_DOCSTORE_SWEEP=4`; that pass held too in 20.2s. `git diff --check` is quiet and I found no new contract trouble. The fresh required board is still running, so this stays a current-head comment until the precinct lab clocks in.

figgyster (approved):
The precinct lab finally stopped smoking. The three-document default keeps the differential witness in the daily loop, the four-document soak still holds when asked, and the full required board is green on this head. Approved — this is #753’s docstore work growing the kind of test that catches a clever query lying at the intersections.

PR #760, opened 2026-08-04
docs: pi headless host vision and vertical-slice plan

victorarias:
## What

Two documents, no code changes.

**`docs/vision/pi-attn-plugins.md`** — the pi vision now includes attn rendering the conversation. The attn-pi plugin provides a headless host entrypoint (Bun, in-repo) running pi via `createAgentSession`; the daemon spawns it per session and owns its process group; the app renders the conversation in React with terminal-transcript density built from real UI (expandable tool calls, rendered diffs, inline images). The PTY/terminal path stays for claude/codex/shell sessions. New north-star principle: **attn integrates on declarations, never on renderings** — the host emits a typed semantic stream in attn's own vocabulary (session linked, run started/settled, state, tool started/finished, message committed) that the daemon and every attn feature integrate on, plus a render stream the daemon routes without parsing. Loop ownership is unchanged: pi owns the loop, models, and context. Also: the two landed rocks (driver plugin, citizenship) are checked off, and stale open questions were settled or removed.

**`docs/plans/2026-08-05-pi-headless-host.md`** (new) — the implementation plan as vertical slices, each shipping end to end (host → daemon → app): (1) live chatbox, (2) state + first-class nudges via pi's steer/followUp verbs with visible queued→seen delivery, (3) tool visibility with collapsed cards and on-demand detail, (4) crash/revive with windowed attach snapshots, (5) history/resume/model switching. Skills, safety envelope, subagents, and background eyes remain vision rocks, not slices.

## Grounding

The plan's design decisions carry receipts from a 2026-08-04 source read and instrumented live spike against pi 0.80.10 (two providers), summarized in the plan's Grounding section: `agent_settled` semantics and abort event tails, steer/followUp drain points, orphaned-tool-subprocesses on hard kill (hence daemon-owned process groups), session-file format/versioning and the no-file-before-first-assistant-message trap, streaming delta rates (hence host-side coalescing), transcript-size distribution (hence windowed snapshots), and pi's measured release cadence (hence the exact pin with a committed spike harness as the bump gate).

## Verification

Docs-only; no runtime surface. Cross-doc links audited against main after the #757 plan triage (one dangling link to a deleted plan fixed in the header note).

https://claude.ai/code/session_01VBeTK9W76BS9DxwXfhzBh4

### DIFF
--- a/changelog.d/brisk-marmot-pi-headless-host-docs.yaml
+++ b/changelog.d/brisk-marmot-pi-headless-host-docs.yaml
@@ -1,0 +1,7 @@
+kind: internal
+area: docs
+change: >
+  pi vision updated — attn renders the conversation via a headless SDK host
+  (plugin entrypoint, daemon-owned process group, React conversation surface;
+  PTY path stays for claude/codex/shell) — plus a vertical-slice
+  implementation plan grounded in the 2026-08-04 pi SDK spike receipts.
--- a/docs/plans/2026-08-05-pi-headless-host.md
+++ b/docs/plans/2026-08-05-pi-headless-host.md
@@ -1,0 +1,183 @@
+# Plan: pi headless host and rendered conversation — vertical slices
+
+## Alignment
+
+Implements the rendered-conversation shape of
+[docs/vision/pi-attn-plugins.md](../vision/pi-attn-plugins.md): attn renders
+the conversation via pi's SDK through a headless host process. Slice-based:
+every slice ships a usable behavior end to end (host -> daemon -> app); no
+layer is built ahead of the slice that needs it. Chatbox first, then expand.
+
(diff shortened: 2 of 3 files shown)


figgyster (approved):
I was ready to book this on the missing changelog fragment, then you slid the fix under the door before the ink dried. The headless-host plan gives the pi work a real route through the city: semantic facts feed attn, render deltas stay paint-only, and the first slice has enough of a life to prove the bet before the rest of the machinery moves in.

Rechecked `9134b35`: the new `kind: internal` fragment accurately records this docs-only change, `git diff --check` passes, and the full current CI board is green. The pi arc has stopped being a TUI integration sketch and become a sober contract for a first-class attn surface. Approved.

PR #762, opened 2026-08-05
fix(terminal): speed cold restores and preserve colors

victorarias:
## Intent

Cold terminal workspaces could take one to two seconds to become usable after falling out of the warm cache, and restored terminals used Ghostty's muted default ANSI palette instead of the app theme. This keeps virtualization's memory savings while restoring terminals quickly and with the same colors they had before unmounting.

## Summary

- replace the timer before every 16 KiB restore write with a cancellable 6 ms cooperative replay budget
- compile the vendored Ghostty WASM module once while retaining a separate runtime instance per pane
- carry the resolved ANSI 0-15 palette through protocol 210 and seed the daemon-owned Ghostty model
- preserve terminal-program OSC 4 overrides when the app theme changes
- add focused replay, WASM cache, protocol, worker, and native palette coverage

## Verification

- pre-commit repository gate: Go formatting, build, and tests; 2,490 frontend tests; production frontend build; and 192 E2E tests
- isolated packaged app: six cold remounts of an approximately 2 MiB history with the warm cache disabled completed in 46.0-50.9 ms (47.8 ms median)
- pre/post-remount terminal style summaries matched; native tests verify exact ANSI palette serialization and OSC 4 override precedence

### DIFF
--- a/app/src/App.sessionlessWorkspace.test.tsx
+++ b/app/src/App.sessionlessWorkspace.test.tsx
@@ -390,6 +390,12 @@ describe('tile-only (sessionless) workspace selection and render', () => {
         foreground: '#d4d4d4',
         background: '#1e1e1e',
         cursor: '#d4d4d4',
+        ansi_palette: [
+          '#000000', '#cd3131', '#0dbc79', '#e5e510',
+          '#2472c8', '#bc3fbc', '#11a8cd', '#e5e5e5',
+          '#666666', '#f14c4c', '#23d18b', '#f5f543',
+          '#3b8eea', '#d670d6', '#29b8db', '#ffffff',
+        ],
       });
     });
   });
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -121,7 +121,7 @@ import {
 } from './utils/queueBands';
 import { useWorkspaceSelectionController } from './hooks/useWorkspaceSelectionController';
 import { hideBootSplash } from './utils/bootSplash';
-import { getTerminalTheme } from './utils/terminalSizing';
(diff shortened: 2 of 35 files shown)


figgyster (dismissed):
That cold-remount number is the kind of speedup worth keeping. One compiled module, separate pane instances, and daemon-owned palette defaults all pay the performance bill without turning terminal state into communal property.

But protocol 210 has one loose wire. `main.tsp` now makes `ansi_palette` part of `SetTerminalThemeMessage`, and `internal/protocol/generated.go` carries it; `app/src/types/generated.ts` still describes the old message and converter map. That checked-in artifact is part of the frontend protocol contract, and `make check-types` exists to reject exactly this drift. Regenerate the TypeScript output with the Go one, then this is ready for another look.

figgyster (commented):
There it is. `app/src/types/generated.ts` now carries `ansi_palette` in both the interface and the converter map, so protocol 210 has stopped telling two different stories to the app and daemon.

That clears my request. The cold-remount work keeps the virtualization arc pointed the right way: faster return trips without handing every pane the same runtime. The required board is still running, so I’m holding the approval until those witnesses get back from the rain.

figgyster (approved):
This one earned its speed the honest way. Cold restores now reuse the immutable Ghostty compilation while each pane keeps its own runtime, and the replayed model gets the same ANSI defaults as the app without trampling a terminal program's OSC choices.

The protocol artifacts are back in lockstep, the replay budget is built once per pane, and the full board is green on `02be6e0`. Approved. The virtualization arc can keep its memory savings without making a returning terminal feel like it woke up in somebody else's colors.

PR #763, opened 2026-08-05
fix(docstore): one-transaction read windows, timestamps that sort; A3.3 findings and A3.4 design

victorarias:
## What

Three document-store fixes and two design documents.

**Read windows (one transaction).** Answering a query used to take three separate transactions — declaration, pagination anchor, SELECT — with SQL compiled in between, so a write landing mid-read produced answers matching no state the collection was ever in: a silently empty page when the anchor was deleted mid-read, a page containing the anchor itself when the anchor gained a sort value, a filter that matched nothing after a type redeclare. `Store.ReadQuery` now does all three reads in one transaction and `QueryDocuments` is unexported so a compiled query cannot run against a state it was not compiled from. Pinned by `internal/store/documents_read_window_test.go`.

**Timestamps that sort.** Stored document stamps are TEXT columns compared as text, and the old encoding (`time.RFC3339Nano`) strips trailing zeros — so within any one second, text order and time order disagreed ('Z' sorts above '.' and every digit). Sorting by `created_at` came back scrambled and a changed-since filter on a whole-second bound silently dropped rows. `docstore.TimeFormat` is now a fixed nine-digit fraction (every stamp exactly 30 chars), bounds on reserved fields are decoded and re-encoded so any RFC3339 spelling means the same instant, a non-timestamp bound is refused instead of compared as text, and migration 91 rewrites stored stamps (idempotent by construction; values it cannot decode are left alone and counted). Pinned by `internal/docstore/timestamps_test.go` and `internal/store/documents_timestamps_test.go`, each falsified against the three halves of the fix separately.

**A3.3 findings doc** records both post-fix reviews: why the differential harness could not catch the encoding bug (both paths read the same stored text), the adversarial-fixtures rule that follows, and the open items 2–7.

**A3.4 design doc** is the accepted plan for the next two stages: document writes commit atomically with their `document.changed` fact and every read answer carries `as_of_seq` (Stage 1); subscriptions become windowed deliveries resumed by content (Stage 2); fact-class compaction bounds the durable log by the size of the data it describes.

## Verification

- Full `go test ./...` green; `go test -race ./internal/store ./internal/docstore` green.
- Each fix falsified by mutation: every mutated half turned exactly its pinning tests red.
- Live-verified on a throwaway profile: pathological old-encoding stamps planted into a real v90 profile DB, broken order confirmed in SQLite, migration 91 run by the real daemon, then ASC/DESC/changed-since/paging/+02:00-bound/refusal and a live `doc watch` verified over the real socket.

### DIFF
--- a/changelog.d/docstore-atomic-read.yaml
+++ b/changelog.d/docstore-atomic-read.yaml
@@ -1,0 +1,11 @@
+kind: fixed
+area: daemon
+change: >
+  Document queries no longer return answers describing a collection that never
+  existed. Answering a query reads the declaration, the pagination anchor and
+  then the rows, and those were three separate reads with the SQL built in
+  between — so a write landing mid-read could silently empty a page whose anchor
+  had just been deleted, hand back the anchor document itself, or make a filter
+  match nothing after a field's declared type changed. None of the three
+  reported an error. All three reads now happen in one transaction, so a query
+  is always answered against a single state of the collection.
--- a/changelog.d/docstore-sortable-timestamps.yaml
+++ b/changelog.d/docstore-sortable-timestamps.yaml
@@ -1,0 +1,11 @@
+kind: fixed
+area: daemon
+change: >
+  Sorting or filtering documents on created_at or updated_at now orders them by
+  time. The stamps are stored as text and compared as text, and the encoding they
+  were written in dropped trailing zeros from the fraction of a second — so
+  within any one second the text order and the time order disagreed. Sorting by
+  newest came back scrambled, and a "changed since" filter written on a whole
+  second silently found none of the documents changed during that second.
+  Existing stamps are rewritten on upgrade, and a timestamp bound may now be
+  written in any RFC3339 form, including one an older version handed back.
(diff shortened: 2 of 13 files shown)


figgyster (approved):
Victor, this is the docstore growing up in public. #758 gave the query engine a dumb second witness; this catches the earlier crime, where one answer could be assembled from three different moments and still look respectable. The fixed-width stamps finish the job by making SQLite's text ordering tell the truth about time.

`ReadQuery` now keeps declaration lookup, cursor lookup, compilation, and SELECT inside one snapshot, while the daemon returns the declaration that actually produced the answer. The race witness exercises the real bad interleaving, and migration 91 both normalizes old values and leaves malformed hand-edits alone instead of inventing a date for them. I also ran the focused store/docstore race suite here; it held. Daemon plus both Linux preflight lanes are green. The A/B benchmark is still running, as trend work tends to do.

Approved. The next A3.4 slices have a proper floor under them: positioned writes and resumable windows won't be trying to build a ledger on answers that came from nowhere.

PR #764, opened 2026-08-05
docs(plans): re-ground the pi headless-host plan on pi 0.83.0

victorarias:
## What

Moves the pi headless-host plan's exact pin from `@earendil-works/pi-coding-agent` 0.80.10 to 0.83.0 and re-grounds every receipt in the plan's Grounding section against that version.

## Why

The plan pins pi exactly and treats a pin bump as a gated event: re-run the spike scenarios, or the receipts are stale. The local pi installation moved to 0.83.0, so the plan moves with it — through the gate, not past it.

## How it was verified

- **Source diff** of the 204 commits between v0.80.10 and v0.83.0, focused on the SDK surfaces the plan depends on: `createAgentSession`/`bindExtensions` unchanged, session file format still v3 with in-place migration, `stopReason: "aborted"` intact, steer/followUp drain semantics untouched, `queue_update` intact.
- **Full spike-harness re-run at 0.83.0** (scenarios s1–s5b on `openai/gpt-5.6-luna`), every scenario passing with unmodified scripts: hard-killing the host still orphans running tool subprocesses (child reparented to pid 1), no session file exists before the first assistant message, abort round-trips with `stopReason "aborted"` and the subprocess actually killed, steer lands at a turn boundary.
- The diff itself confirmed the plan's churn thesis live: the event union gained four types with no breaking-change label (`summarization_retry_*`, `bash_execution_update`), and agent-core took a labeled break at 0.81.0 (`streamFn` required) that the SDK path absorbs internally.

Doc changes beyond the pin: the grounding header records the re-validation, the delta-rate numbers widen to cover both measurement runs (~480–550 updates / ~45–52 KB per ~5 s reply, peaks to 1,970 events/s), the auth receipt notes the clearer 0.82.1+ OAuth errors and the new `pi auth print-api-key`/`print-bearer-token` commands, and a new bullet records SDK-visible events later slices can use (summarization retry lifecycle for slice 5, `bash_execution_update` as a slice-3 render candidate, the non-terminal `"pending"` stop reason).

Docs-only; no code paths change. Exempt from live verification as trivial docs per AGENTS.md.

https://claude.ai/code/session_01VBeTK9W76BS9DxwXfhzBh4

### DIFF
--- a/changelog.d/brisk-marmot-pi-pin-083.yaml
+++ b/changelog.d/brisk-marmot-pi-pin-083.yaml
@@ -1,0 +1,7 @@
+kind: internal
+area: docs
+change: >
+  The pi headless-host plan now pins pi 0.83.0: every spike receipt was
+  re-validated against it (source diff plus a full harness re-run with
+  unmodified scripts), the grounding numbers were refreshed, and the new
+  SDK-visible events useful to later slices are recorded.
--- a/docs/plans/2026-08-05-pi-headless-host.md
+++ b/docs/plans/2026-08-05-pi-headless-host.md
@@ -10,7 +10,10 @@ layer is built ahead of the slice that needs it. Chatbox first, then expand.
 
 ## Grounding (receipts)
 
-Measured in the 2026-08-04 SDK spike / source read at pi 0.80.10:
+Measured in the 2026-08-04 SDK spike / source read at pi 0.80.10, and
+re-validated on 2026-08-05 at pi 0.83.0 (source diff of the 204 commits in
+between plus a full re-run of the spike harness — every scenario passed with
+unmodified scripts):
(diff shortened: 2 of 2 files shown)


figgyster (approved):
The pin gate only earns its keep if a new pi release has to drag its receipts back through the mud. You did that here: the plan records the 204-commit re-grounding, widens the stream budget to match the fresh run, and names the newly visible event seams before later slices quietly depend on them.

The old 0.80.10 driver floor is a separate Rock 1 CLI boundary, so it does not conflict with this unshipped headless-host plan’s 0.83.0 pin. I checked the two-file diff and its whitespace, then the full docs CI board: all green. Approved.

PR #765, opened 2026-08-05
feat(home): categorise snoozed agents and split your PRs from review requests

victorarias:
Home is where the queue ends and where a day starts. Three things on it were either noise or wrong.

**The shortcut strip.** A hardcoded row of four key hints (`⌘T`, `⌘1-9`, `⌘,`, `⌘/`) pinned under the cards. It was never derived from the shortcut registry, so it drifted silently whenever a binding moved — and `⌘/` already opens the real cheatsheet, which *is* generated from the registry. It cost height on every visit to teach four things once.

**Snoozed agents were miscategorised.** Snooze settles as it defers (`store.SnoozeTurn` writes `turn_settled_at` and `turn_snoozed_until` in one statement), so a snoozed agent carries no `turn_owed` — which is exactly what home's turn/settled split reads. A deferred agent therefore fell into a plain state group and rendered as "Working" or "Idle": the one thing that is not true of an agent waiting on a clock nobody can see from here. The snooze plan deferred any home presence explicitly; this is that follow-up.

**Your PRs were buried among the ones you were asked to look at.** The card grouped strictly by repo, so a PR you opened sat between two review requests with only a ✏️ against 👀 to tell them apart. The two are different work — one you are waiting on, one someone is waiting on you for — and the one you come to home to find is the smaller of the two.

## What changed

- **The strip is gone**, with nothing in its place.
- **`Snoozed (n)`** — a collapsed-by-default section at the foot of the Sessions card, above Muted Workspaces. Each row carries its wake time and a hover-revealed wake-now button; ordering comes from the sidebar's own `compareWakeOrder` (newly exported for the same reason `compareTurnOrder` already was — two orders for one set of promises is two sets of promises). Membership is read from `turnSnoozedUntil` through `isSnoozed`, never derived from state.
- **The PR card splits into `Yours` then `Review requested`.** Yours leads and stays flat with the repo named inline; the review side keeps today's collapsible repo groups and its mute button. Both render through one extracted `renderPRRow`.

No daemon change. `turn_snoozed_until` and the PR `role` field already ride the wire and already reach home's props; nothing new is persisted, broadcast, or computed server-side.

## Deliberate departures from the sidebar

- **The Snoozed section is not gated on queue mode.** A snooze can only be *made* with the queue on, but it outlives the setting being turned off — and the shortcut, the command menu, and the sidebar's own section are all queue-gated. With the arrangement off, this section is the deferral's only remaining way out.
- **A deferred chief lands in the Snoozed section**, where the sidebar keeps it in its anchored slot. Home's Sessions card lists the chief alongside every other agent, so excluding it would put a deferred chief back under a state group that cannot describe it. The Chief of Staff card is unaffected.
- **Snoozed agents leave the all-settled banner's `stillRunning` line.** That line says what is in flight and coming back to you; a deferred agent is not, and the `Snoozed (n)` count is its own receipt.

## Known consequences

- **A repo where you only ever author PRs can no longer be muted from home.** The ⊘ lives on review-side repo headers; muting a repo hides your own PRs in it, which is not an act anyone wants, and Settings unmutes either way. When the repo also has review requests the button is right there.
- **"Review requested" is `role === 'reviewer'`**, which the daemon assembles from both `review-requested:@me` and `reviewed-by:@me` — so a PR you reviewed uninvited reads as requested. Splitting them needs a new wire field; recorded as a follow-up in the plan.
- `.shortcut` / `.shortcut kbd` stay in `Dashboard.css`. They are the app's only definition of a keyboard hint and `LocationPicker` and `AttentionDrawer` still render them; only `.dashboard-footer` and `.footer-shortcuts` go.

## Verification

- `tsc --noEmit` clean; full vitest suite green (224 files, 2500 passed). `Dashboard.test.tsx` gained 9 tests: the snooze partition out of state groups, collapse/expand with soonest-wake-first ordering, wake-now, reachability with the queue off, a deferred chief, a lapsed deadline left alone, the absent footer, and the PR split (yours flat with inline repo and no repo groups while review keeps them, empty sections omitted, both-empty copy).
- 13 PR e2e tests pass (`e2e/pr-actions.spec.ts`, `e2e/dashboard-prs-harness.spec.ts`).
- `home_get_state` now reports the new sections, so the change has an automation witness the way the sidebar's snoozed section already does: the snoozed group and its rows, which sessions are still grouped by state, both PR sections, and whether the shortcut footer is present.
- **Live run** on a throwaway profile installed from this branch (preflight PASS, protocol 210): five injected agents with two snoozed over the wire, queue mode on, injected PRs beside real GitHub ones. `home_get_state` reported `shortcutFooter: false`; `snoozed` present with count 2, collapsed, expanding to rows carrying wake times (`11:42 PM`, `tomorrow 6:07 PM`) and a wake control; `stateGroupSessionIds` held only the three awake agents, so neither deferred agent appeared under a state group. The card read `Yours 11` as flat rows with the repo inline and no repo groups, then `Review requested 3` in two repo groups with the mute button intact. That run also caught a layout defect — a long repo name (`figgy-pr-review-synthetic`) crowding out PR titles — fixed by capping `.pr-repo-inline` and re-verified. Profile cleaned afterwards.

Design and decisions: `docs/plans/2026-08-05-home-screen-tidy.md`.

https://claude.ai/code/session_013eGAp8cHdGj7Hu7JJFC5pA

### DIFF
--- a/app/src/hooks/useUiAutomationBridge.ts
+++ b/app/src/hooks/useUiAutomationBridge.ts
@@ -2135,11 +2135,13 @@ export function useUiAutomationBridge({
           },
           // Which agents home is still grouping by state. A deferred agent
           // appearing here is the defect the snoozed section exists to remove.
+          // Read from the groups that mark themselves, not from the testid
+          // prefix: the turn band and the snoozed section share it and are not
+          // state groups, and an exclusion list would silently gain a hole the
+          // next time a group is added.
           stateGroupSessionIds: Array.from(
-            visible?.querySelectorAll('[data-testid^="session-group-"] .session-row') || [],
-          )
-            .filter((row) => !snoozedGroup?.contains(row))
-            .map((row) => (row.getAttribute('data-testid') || '').slice('session-'.length)),
+            visible?.querySelectorAll('[data-session-group="state"] .session-row') || [],
+          ).map((row) => (row.getAttribute('data-testid') || '').slice('session-'.length)),
           prs: {
             yours: readPRSection('pr-section-yours'),
             review: readPRSection('pr-section-review'),
--- a/app/src/hooks/useUiAutomationBridge.ts
+++ b/app/src/hooks/useUiAutomationBridge.ts
@@ -2092,11 +2092,58 @@ export function useUiAutomationBridge({
         const visible = document.querySelector('.view-container.visible');
         const banner = visible?.querySelector('[data-testid="all-settled"]') ?? null;
         const follow = visible?.querySelector('[data-testid="follow-next-turn"] input');
+        const snoozedGroup = visible?.querySelector('[data-testid="session-group-snoozed"]') ?? null;
+        const snoozedHeader = snoozedGroup?.querySelector('[data-testid="session-group-snoozed-header"]');
+        const readPRSection = (testId: string) => {
+          const section = visible?.querySelector(`[data-testid="${testId}"]`) ?? null;
+          if (!section) return null;
+          return {
+            count: section.querySelector('.pr-section-count')?.textContent?.trim() || '',
+            // The repo as the row itself names it: inline on the flat side, on a
+            // group header on the review side. Which one carries it is the whole
+            // difference between the two halves.
+            rows: Array.from(section.querySelectorAll('[data-testid="pr-card"]')).map((row) => ({
+              number: row.querySelector('.pr-number')?.textContent?.trim() || '',
+              repo: row.querySelector('.pr-repo-inline')?.textContent?.trim() || '',
+              title: row.querySelector('.pr-title')?.textContent?.trim() || '',
+            })),
+            repoGroups: Array.from(section.querySelectorAll('.pr-repo-group .repo-name'))
+              .map((name) => name.textContent?.trim() || ''),
+          };
+        };
         return {
           onScreen: Boolean(visible?.querySelector('.dashboard')),
           allSettled: Boolean(banner),
           detail: banner?.querySelector('.all-settled-detail')?.textContent?.trim() || '',
           followNextTurn: follow instanceof HTMLInputElement ? follow.checked : null,
+          // The strip of keyboard hints home used to carry. Reported so its
+          // absence is something a caller can testify to rather than infer.
+          shortcutFooter: Boolean(visible?.querySelector('.dashboard-footer')),
+          // Agents the user deferred. Collapsed by default, so `rows` is empty
+          // until something expands it — which is a real answer, not a gap.
+          snoozed: {
+            present: Boolean(snoozedGroup),
+            header: snoozedHeader?.textContent?.trim() || '',
+            expanded: snoozedHeader?.getAttribute('aria-expanded') === 'true',
+            rows: Array.from(snoozedGroup?.querySelectorAll('.session-row') || []).map((row) => ({
+              id: (row.getAttribute('data-testid') || '').slice('session-'.length),
+              label: row.querySelector('.session-name')?.textContent?.trim() || '',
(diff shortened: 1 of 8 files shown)


app/src/hooks/useUiAutomationBridge.ts
```diff
@@ -2092,11 +2092,58 @@ export function useUiAutomationBridge({
...
+              id: (row.getAttribute('data-testid') || '').slice('session-'.length),
+              label: row.querySelector('.session-name')?.textContent?.trim() || '',
+              state: row.getAttribute('data-state') || '',
+              wake: row.querySelector('.session-wake-at')?.textContent?.trim() || '',
+              canWake: Boolean(row.querySelector('.session-wake-btn')),
+            })),
+          },
+          // Which agents home is still grouping by state. A deferred agent
+          // appearing here is the defect the snoozed section exists to remove.
+          stateGroupSessionIds: Array.from(
```
victorarias, line 2142: Leaving this one. It reads a handful of DOM rows behind an automation verb that only runs when the harness asks — the second pass is over ~5 elements, and `filter().map()` says "drop the deferred rows, then name them" in a way a `reduce` with a push does not. The one-pass version also has to be hoisted out of the returned object literal, so it costs a named local and a loop to save nothing measurable.

figgyster (commented):
Home was telling a small but consequential lie about deferred work, and you gave it the honest answer instead of painting another state badge over it. The queue-off wake path is the part I care about: a promise to come back needs a visible escape hatch when the arrangement that made it is gone. And putting your own PRs ahead of the review pile follows the actual flow of attention, not the old storage shape.

I found no merge-stopper in the slice. The focused Dashboard suite passed here (25 tests), the sandboxed TypeScript check passed, and Frontend plus two E2E shards are green.

One small automation follow-up for later: [useUiAutomationBridge.ts#L2138](https://github.com/victorarias/attn/blob/3884727d03dd9de2f7b66f1a7259e332183150c2/app/src/hooks/useUiAutomationBridge.ts#L2138) says `stateGroupSessionIds` reports state groups, but its prefix selector also includes `session-group-turns`. The current live scenario has no owed turn, so it cannot catch that distinction. I would not hold this UI work for it; narrow the selector when that witness next moves.

The last Frontend E2E shard is still running, so I’m leaving this as a comment rather than stamping a green light ahead of the precinct lab.

app/src/hooks/useUiAutomationBridge.ts
victorarias: Fixed — good catch, the field named one thing and reported another. The state groups now carry `data-session-group="state"` and the bridge asks for that instead of the testid prefix; a positive marker rather than an exclusion list, so a new group opts in or is correctly left out. Covered by a unit test reading the same DOM with an owed turn, a settled agent, and a deferred one on screen. That also drops the double pass React Doctor flagged one thread up.

figgyster (commented):
That closes the loose floorboard. `data-session-group="state"` gives the automation witness a real boundary instead of a prefix plus exceptions, and the new owed/settled/deferred test catches the exact lie my first pass found. The bot’s second-pass fuss vanishes as a side effect. I like this shape; future groups either opt in or stay honestly outside the answer.

Focused Dashboard coverage now passes here at 26 tests, and the sandboxed TypeScript check passes. The follow-up restarted the required CI board, so this is a current-head comment while that board runs.

figgyster (approved):
The whole board is green now. Your small home-screen tidy-up ended up doing the worthwhile thing: deferred work has a truthful name and a way back, your own PRs surface where attention actually starts, and the automation witness now means exactly what it says. The boundary fix held through Frontend and all three E2E shards.

Approved.

PR #766, opened 2026-08-05
feat(docstore): commit a document write and its fact together

victorarias:
Stage 1 of the A3.4 document-store design
([docs/plans/2026-08-05-ext-a3.4-doc-store-positions-and-windows.md](docs/plans/2026-08-05-ext-a3.4-doc-store-positions-and-windows.md)):
positioned writes. Stage 2 — windowed subscriptions — is not in this PR, and
`doc_subscribe`'s wire shape is untouched.

## The problem

A document write and the fact announcing it were two separate operations. The
write committed, then the daemon published. If the publish failed, the document
had changed and nobody was told: every live query kept showing a result set the
store no longer agreed with, indefinitely, with no error anyone could act on.

Nothing in the system carried a position either. A caller that wrote and then
read had no way to tell whether the answer it got already included its own
write, and an extension holding a result set had no way to say what it was a
result set *of*.

And the log grew with writes rather than with data. A document written a
thousand times left a thousand facts, all saying the same thing — that it
changed — while the state itself lived in the table the whole time.

## What changed

**One commit.** `Store.CommitDocumentWrite` runs the document statement and the
fact's append in one transaction and returns where the fact landed. A fact that
cannot be made durable now fails the write, which is the inverse of the old
behavior and the point of the change. The store stays fact-agnostic: the daemon
builds the `store.BusEvent` and hands it over, so `internal/store` still knows
nothing about what attn's facts are called.

A write that changed nothing still appends nothing — deleting an absent document
and a refused conditional write wake nobody, exactly as before.

**Ordered announce.** Facts appended inside somebody else's transaction are not
delivered by whoever wrote them. `bus.Announce()` reads the log forward from a
high-water mark held under the publish lock, and `Publish` now does the same
instead of handing subscribers the event in hand. Two writers that commit and
then race to announce therefore cannot deliver out of order, announcing twice
delivers once, and an announce that never happened is repaired by the next one —
so a caller needs no coordination beyond calling it after its commit.

**Positions on the wire.** `doc_put`/`doc_delete` return `seq`;
`doc_get`/`doc_query`/the new `doc_count` return `as_of_seq`, read inside the
same transaction as the rows. Comparing the two is how a caller knows whether an
answer includes its write. `as_of_seq` 0 over an empty log is a valid position,
not a missing one.

**Codes, not prose.** Failures a caller must branch on carry a machine-readable
`code` beside their human text, mapped from typed errors
(`docstore.ConflictError`, `UndeclaredCollectionError`, `InvalidQueryError`) at
one choke point — `sendDocError` — with no string matching anywhere. `conflict`
brings the expected and actual revisions and whether the document is there at
all, so a retry needs no second round trip.

**`doc_count`.** How many documents match, without their bodies. It reuses the
query's own compile against `SELECT COUNT(*)`, so it cannot drift from what the
query would return; `--sort`, `--desc` and `--limit` are ignored, because they
decide which matches come back and never how many there are.

**Compaction.** The existing retention tick now also reduces the compactable
fact classes to the newest fact per subject. Which names those are is the bus's
policy; the SQL is the store's. It runs only at or below the enabled-consumer
cursor floor, for two load-bearing reasons: below the floor every enabled
consumer has already read everything, so delivery is bit-for-bit unchanged; and
`reconcileGap` reads "cursor below the earliest surviving seq" as "everything
missing was trimmed", which compacting above the floor would turn into a lie.

The consequence, stated rather than hidden: for these names durable delivery is
at-least-once *per changed subject*, not per write. A consumer of this bus reads
current state anyway. A workload that needs every intermediate change as a
business event is doing event sourcing, and those events are data — they belong
in a collection it writes, where their growth is its own visible cost.

**Undefine announces the collection.** It published `document.changed` with an
empty id, which is an address that points at nothing — no consumer can match it
and no compaction can group it. It now publishes `document.collection.removed`
with the collection as its subject, and live queries wake on either fact.

## Operator surface

`attn bus status` reports how many events the log holds and what they weigh —
the receipt for the invariant compaction upholds. Bytes is event text, not file
size: SQLite pages are shared with every other table and would answer a
different question.

`attn bus trim` runs one retention pass on demand. Without it compaction only
fired on an hourly in-daemon tick, so there was no way to see it work or to
reclaim a log that grew between ticks. It builds a bus over the same store
rather than reimplementing the pass, so the pass, the compactable classes and
the floor have one definition; and it is database-direct like the rest of `attn
bus`, which must keep working whether or not the daemon is listening.

`attn doc count` is new. `attn doc put`/`delete` print the seq their write
landed at, and `query`/`get` print the position their answer was true at (to
stderr, so it stays out of a body being piped into `jq`; `--json` shapes are
unchanged, and `doc count --json` carries the position in band).

## Two deviations from the plan, both deliberate

- `CommitDocumentWrite` returns a `DocumentWriteResult` rather than the plan's
  bare `(rev, seq)`. A delete needs to report whether anything existed, and
  overloading seq to carry that would be the kind of encoding that reads fine
  until someone deletes an absent document.
- The announce mark is initialized at construction rather than at `Start`. A bus
  that is never started still publishes and announces — that is the shape of
  every test daemon — and a mark initialized lazily would swallow the first
  committed write's fact.

## No migration

Stage 1 adds no schema. `bus_events`, `bus_consumers` and the per-collection
tables are unchanged; compaction is a DELETE over rows already there.

## Verification

`go test ./...` green. `go test -race ./internal/store ./internal/docstore
./internal/bus` green.

New receipts, each checked against a mutant that should break it:

- **Atomicity.** A trigger that refuses every insert into `bus_events` — the
  real path, not a seam — must take the document write down with it, for both a
  put and a delete. Ignoring the append error makes it fail.
- **No fact on no change.** An absent delete and a refused conditional write
  leave the log untouched.
- **Announce ordering.** Eight goroutines commit-then-announce while publishes
  interleave on the same mark; every subscriber sees the log's own order,
  exactly once. Reverting `Publish` to direct fan-out drops facts on every run.
- **Compaction.** Churning one subject 200 times leaves one fact, and it is the
  newest; a consumer parked below the floor keeps every fact above it; a mixed
  create/rewrite/delete workload leaves a log no larger than its live documents
  plus in-window tombstones, asserted as the inequality it is; an empty name
  list compacts nothing rather than everything.
- **`document.changed` as a contract.** Name, subject shape, payload fields, and
  that the seq the write reported is the seq the fact is at.
- Adversarial count fixtures per the A3.3 rule: writes inside one second at
  fraction widths a variable-width encoding renders out of order, a numeric
  string beside a number under a declared number field, and a boolean beside the
  string `"true"` — the count must agree with the query on every one.

Live, on a throwaway `posw` profile (preflight PASS, protocol 211 end to end;
never `dev`, never production):

- Two writers racing 25 writes each under a live `doc watch`, plus one more:
  revisions 26 and 25, last write at seq 94, the subscriber woken throughout,
  `doc count` agreeing with the query.
- 300 writes to one document: the log went 105 → 415 events (9.8 KB → 46.2 KB).
  One `attn bus trim` removed 348, leaving 67 events at 4.6 KB. The surviving
  `document%` rows were exactly three — one per live document — while
  `pr.updated`, which is not compactable, kept all four facts per subject. The
  document itself was untouched at rev 300, and a write after the pass still
  woke a live subscription at seq 416.

## Surfaces walked

CLI (`attn doc count`, `attn bus trim|status`, seq/position output), daemon and
app (protocol 211 in `constants.go` and `useDaemonSocket.ts`; `generated.go` and
`generated.ts` regenerated from `main.tsp`, never hand-edited), Linux (no
platform-specific code added), docs (changelog fragment). No plugin or SDK
surface consumes these commands yet.

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -199,7 +199,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '212';
+export const PROTOCOL_VERSION = '213';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 // AutomationActionTimeoutError distinguishes "the daemon never sent a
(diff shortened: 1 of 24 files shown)


figgyster (commented):
You’re taking A3.4 through the door #763 opened: reads already had to name the snapshot they saw, and now writes earn the same honesty. Making the document mutation and its invalidation one SQLite commit is the right hard edge; the old gap was the kind of quiet lie that keeps a live query wrong until somebody notices the smell.

The store and bus witnesses hold up. `CommitDocumentWrite` rolls the row back when the fact cannot append, no-op/refused writes stay quiet, and `Announce` reads the log forward under the publish lock so concurrent commit-then-announce callers cannot scramble ephemeral delivery. The protocol 211 Go and TypeScript artifacts are in step, and the position/count contracts carry a real regression net rather than a stack of polite mocks.

I’m leaving this as a comment while GitHub reruns Frontend E2E (3/3). The first attempt died in the untouched `split-blank-repro` terminal-paint scenario—zero quads before the initial frame—not in this document-store path. Daemon, Frontend, both Linux preflights, the other E2E shards, React Doctor, Changelog, and the A/B benchmark are green on `fb877b9`; if that retry comes home green, this is good to merge.

figgyster (dismissed):
That last terminal ghost got its second walk through the rain and came home green without the branch moving. Good. The full required board now backs the thing the code already said: a document write and the fact that makes it visible are one indivisible act.

This carries #763’s truthful read windows into the mutation half of A3.4. Positions now give extension callers an honest answer about what they have seen, while cursor-safe fact compaction keeps the log from billing the future for every rewrite. The atomicity, ordering, no-op, and compaction witnesses are the right kind of stubborn.

Approved on `fb877b9`.

figgyster (commented):
The post-approval pass found two real ways the bus could tell a lie while trying to keep the log honest. Good catch. An unplaced announce mark cannot impersonate seq 0, and an operator-triggered trim needs an error bit that a script can read.

The new witnesses pin both edges: a failed head lookup neither replays old facts nor silences the publish that just happened, and a compaction pass that cannot establish its cursor floor no longer dresses up as `removed 0`. That hardens the positioned-write path #763 made possible — history is still ordered, but the machinery admits when it cannot read it.

`go test ./internal/bus` passes here. I’m leaving this conditional on `c8bd193`: the rebuilt Daemon job failed only in the untouched `internal/ptyworker` process-marker witness, while Frontend E2E (3/3) failed again in the pre-existing `split-blank-repro` paint scenario (zero quads before the first frame). Both need their retries before this gets a fresh approval.

figgyster (dismissed):
The two stray witnesses came back sober on the same head. Daemon passed its process-marker check, the split-terminal paint repro passed, and the full required board is green.

That leaves a positioned-write slice with its story intact: document row and durable fact share the commit, readers and writers can name the position they saw, and the follow-up makes the bus honest when its head or compaction floor is unavailable. This is #763’s snapshot contract growing the mutation half it needed before windowed subscriptions start asking harder questions.

Approved on `c8bd193`.

figgyster (commented):
This head is the same five-commit document-store stack rebased onto the queue work that just landed. The range comparison holds the patch one-for-one, and the focused bus/store/protocol run completes on `a31c79b`.

So the story stands: writes and facts share a commit, positions say what callers actually saw, and the follow-up bus guards refuse to invent a history when storage cannot answer. #768 changes the ground under this branch, not the contract this branch is carrying.

Leaving a current-head comment while the rebuilt required board runs. Once it is green, this remains ready.

figgyster (dismissed):
The rebased board is green all the way down, including the two first-run CI ghosts. The protocol collision also landed the right way: one next version, regenerated shared types, and both client sentinels at 212.

The positioned-write work still pays off the promise #763 made. A document mutation, its durable fact, and the position callers use to reason about both now survive the rebase together; the bus follow-up keeps the live path honest when the log cannot speak.

Approved on `a31c79b`.

figgyster (commented):
This is another honest rebase, this time onto the conversation-runtime merge. The five positioned-write commits still map one-for-one; the only collision is protocol numbering, now correctly advanced to 213 with the generated contract carrying both sides.

`go test ./internal/protocol` passes on `76cbd6f`. The document/fact transaction and its guarded announce path have not changed their story, while the new base contributes the conversation shapes it must now coexist with.

Leaving this conditional while the rebuilt required board runs.

figgyster (approved):
The final rebased board is green on the first pass: all fourteen required checks, no retries, no drift in `main`.

The positioned-write stack has held its line through the queue and conversation-runtime rebases. Protocol 213 keeps those new shapes beside the document position/error contract, while the transaction and announce guards keep the storage story truthful when things go sideways.

Approved on `76cbd6f`.

PR #768, opened 2026-08-05
feat(queue): pin one agent, and let a shell belong to the agent it came from

victorarias:
The queue's pin was workspace-shaped, so taking one agent out of it took every sibling with it. And a shell split out of an agent sat in the queue as a row of its own that nothing could ever settle — while being, in every way that matters to the user, part of the agent it was opened beside.

Two changes, one model.

## Pinning aims where the gesture aims

A pin aimed at a row now pins that row's agent. Pinned agents collect in their own band below Settled, in the order they were pinned; unpinning is on the row where it lands, and in the command menu on the agent you are looking at. Pinning a whole workspace is unchanged and still reachable from its group header and the command menu — the two are labelled apart. The chief refuses the pin: it already has an anchored slot above the queue.

Pinning is not settling. It filters at read, like workspace pin, so the turn stamps go on accruing underneath: unpinning surfaces whatever was outstanding at its true age rather than restarting the clock. "I'll come to this one myself", not "I'm done".

## A shell belongs to the agent it came from

A shell records the session it was split from. Resolution happens once, at spawn, and always lands on an agent — splitting a shell out of a shell inherits that shell's agent — so nothing downstream has a chain to walk. A shell whose agent is live in the same workspace draws no sidebar row: you reach it by going to that agent, where it is a pane.

Attachment is judged where the bands are built, never maintained by writes. The parent must still exist *and* still be in the same workspace, so a shell whose agent closed, or whose pane was moved elsewhere, gets its row straight back — with no cascade to run and no call site to forget. Shells that already exist keep their own rows: the link is recorded when a shell is opened and is deliberately not backfilled, because pane geometry is the only trace of where an old shell came from and guessing would hide a row the user can still reach.

Queue mode off is untouched: both fields are queue vocabulary, and the workspace tree draws exactly what it drew before.

## Shape

- Migration 92 adds `sessions.pinned_at` and `sessions.parent_session_id`, both `TEXT NOT NULL DEFAULT ''`. `pinned_at` is deliberately absent from the session upsert so a respawn cannot clear a pin; both store branches are covered by the same test.
- `pin_session` on the protocol (`ProtocolVersion` 211), session-scoped so a hub routes it to the daemon that owns the session, with `FactSessionPinChanged` projecting the re-broadcast.
- `attention.Input` gains `SessionPinned`, joining the four exclusions already there.
- `spawned_from` on spawn is a statement of fact from the client; the daemon owns the rule that turns it into a parent, so the app, the CLI, and the automation bridge get identical semantics.

Design, API surface, and the pitfalls this had to walk around: [`docs/plans/2026-08-05-queue-pins-and-satellite-shells.md`](docs/plans/2026-08-05-queue-pins-and-satellite-shells.md).

## Verification

Live-verified on a throwaway profile (`pinsat`) with a full `make install`, bundled preflight clean, and `real-app:scenario-agent-queue` passing all 19 steps against the packaged app. The scenario now covers: a shell split from an agent gets no settled row and no tree row, a shell split from that shell is a satellite of the same agent, pinning from a queue row lands the agent in the Pinned band with its workspace untouched, and unpinning returns it to the turns band on the same `turn_opened_at`. The daemon's own record afterwards showed both shells pointing at the agent, not at each other.

Two pre-existing scenario defects surfaced and are fixed here: pane lookups read a `paneIds` field `get_workspace` does not return, so every `close_pane` in the shell step silently did nothing; and the final keyboard-settle step assumed one press empties the band, which races the daemon reclassifying sessions after a restart (an agent settled at `waiting_input` can legitimately land on `idle` and open a fresh turn).

Migration 92 was also applied to a copy of a populated pre-92 database (9 real sessions, schema at 91): it migrated to 92, both columns present with `''` defaults, every row intact.

https://claude.ai/code/session_016iVZ5JEzWzmZYBASZaKXSq


figgyster (approved):
This is the queue getting honest about intent. A row-level pin finally leaves its siblings alone, and the satellite link makes a split shell part of the agent it was opened beside without turning an orphan into a disappearance. That carries the attention-truthfulness thread from #765 into the queue, where a wrong row costs you a real interruption.

The daemon owns both edges: pinning filters the live turn rather than settling it, and spawn-time parent resolution stays bounded to one agent in the same workspace. The protocol, hub projection, queue model, automation bridge, unit witnesses, and real-app queue path all move together. Required CI is green. Approved.

PR #769, opened 2026-08-06
fix(plugins): install a bundled plugin's dependencies before building it

victorarias:
Building the app from a fresh clone fails while compiling the bundled pi
plugin:

```
error: Could not resolve: "@earendil-works/pi-coding-agent". Maybe you need to "bun install"?
    at plugins/attn-pi/host/index.ts:26:8
```

Nothing installs a plugin's dependencies. `plugins/*/node_modules` is
gitignored and no build step, Makefile target or CI job has ever run `bun
install` there. That was harmless until the headless host landed (#767),
because every earlier plugin import either resolved inside the plugin or was
marked `--external` — the suite's `VERSION` import is external precisely
because pi resolves it as a virtual module at load time. The host is the first
thing that *bundles* pi in, so `bun build --compile` has to resolve it for
real. Any machine with a warm `node_modules` from working on the plugin builds
fine; a clean checkout does not.

## The fix

`stage_plugin` installs from the committed lockfile before compiling, for any
plugin that has a `bun.lock`. Frozen, because the pin a bundled plugin ships is
the one its lockfile records — a build is never the place to move it. It costs
~0.6s cold and is a no-op once warm. `attn-opencode` declares no dependencies
and has no lockfile, so it is untouched.

## Why nothing caught it

No CI job compiles the bundled plugins at all. The only caller of
`scripts/build-bundled-plugins.sh` is `scripts/build-app-profile.sh`, which
runs during a packaged-app build — something CI never does; the Tauri job only
runs `cargo fmt` and `cargo test`. So the build that breaks is the one no
runner performs, and the first machine to see it was a contributor's.

A `Plugins` job now stages both plugins the way the app build does, on a
checkout with nothing installed. It also runs the plugin test suites (114
tests across the two plugins), which were likewise never gated. Its bun version
is read from `.tool-versions` rather than pinned a second time in the workflow.

## Verification

Cloned the merged `main` into an empty directory and ran
`scripts/build-bundled-plugins.sh`: fails with the error above at
`host/index.ts:26`. Cloned this branch the same way: installs 127 packages,
compiles `attn-pi` and `attn-pi-host` (2953 modules), stages both plugins
clean. Both plugin test suites pass locally (64 + 50).

The staging script's macOS-only codesign fixups are already guarded on
`uname -s`, so it runs on the Linux runner.

## Not fixed here

The release workflow never stages bundled plugins either: `release.yml` runs
`make build` (the Go daemon alone) and then `tauri-action` directly, whose
`beforeBuildCommand` is `npm run build`. `build-app-profile.sh` is not in that
path, so a released `.app` appears to ship its `bundled-plugins/ → plugins/`
resource dir containing only `.gitkeep`. That predates the pi host — opencode
would be missing too — and is left for a separate change.

https://claude.ai/code/session_013X61jPjh8UPsEiz2W4xaQV

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ jobs:
       daemon: ${{ steps.filter.outputs.daemon }}
       frontend: ${{ steps.filter.outputs.frontend }}
       tauri: ${{ steps.filter.outputs.tauri }}
+      plugins: ${{ steps.filter.outputs.plugins }}
     steps:
       - uses: actions/checkout@v4
 
@@ -48,6 +49,11 @@ jobs:
               - '.github/workflows/ci.yml'
             tauri:
               - 'app/src-tauri/**'
+              - '.github/workflows/ci.yml'
+            plugins:
+              - 'plugins/**'
+              - 'scripts/build-bundled-plugins.sh'
+              - '.tool-versions'
               - '.github/workflows/ci.yml'
   changelog:
     name: Changelog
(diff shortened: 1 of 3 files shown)


figgyster (approved):
You found the kind of build break that waits politely in a warm checkout, then mugs the next contributor at the door. #767 made pi's host a real bundled executable; this puts its dependency install beside staging, pins it to the committed lock, and gives CI the exact cold-checkout path that had been missing.

I checked the blank plugin tree, the staging script, and the workflow wiring. The new `Plugins` witness compiled and tested both bundled plugins on a fresh Linux runner, and the rest of the board is green. The release-bundle gap you called out is still its own old body in another alley; it doesn't weaken this repair.

Approved.

PR #770, opened 2026-08-06
fix(release): ship the bundled plugins with the released app

victorarias:
`tauri.conf.json` maps `bundled-plugins/ → plugins/`, so the app's
`Contents/Resources/plugins` is whatever that directory held at build time. In
a checkout it holds one `.gitkeep`: the compiled plugins are build output, and
`app/src-tauri/bundled-plugins/*` is gitignored.

A source install stages them — `scripts/build-app-profile.sh` runs
`scripts/build-bundled-plugins.sh` before Tauri collects resources. The release
workflow does not go through that script: it runs `make build` for the Go
daemon and then `tauri-action` directly, whose `beforeBuildCommand` is
`npm run build`. Nothing ever staged the plugins, so a released app would ship
an empty plugin directory and offer none of their agents, while every
maintainer's own build looked correct.

## Staging, and a check that it worked

The release job installs bun (version read from `.tool-versions`, not pinned a
second time) and runs the same staging script before the Tauri build. It then
verifies the plugins reached the built bundle. A bundle that lost them looks
healthy and simply offers no agents from them — nobody notices until someone
tries to launch one. The expected set is whatever was staged, so adding a
plugin needs no edit to the check.

## Nothing public until every gate has passed

That check was standing in the wrong place: `tauri-action` was configured with
`releaseDraft: false`, so the GitHub release went public the moment the build
finished — before the verification, before notarization, and before the Linux
daemons were attached. A rejected bundle would have failed the job with the bad
release already downloadable. The same window exposed `tauri-action`'s own
**un-notarized** dmg and app tarball, which the repackage step replaces with
the stapled ones.

The build now produces a draft, and a final `publish` job — gated on the mac
build and both Linux daemon builds — flips it once every asset is attached. A
failed run leaves a draft rather than a half-built public release: fix the
cause and re-run the workflow on the same tag. `docs/making-a-release.md` says
so, since "the tag triggers the release workflow" no longer tells the whole
story.

## Evidence

- No release has shipped since bundled plugins existed: the resource landed in
  #569 (2026-07-16) and the last release is v0.11.1 (2026-06-05). Its dmg has
  no `Resources/plugins` at all, which is consistent but proves nothing on its
  own — the gap is proven by construction instead. The staged output is
  gitignored (`.gitignore:19`), a checkout carries only
  `app/src-tauri/bundled-plugins/.gitkeep`, and no step in `release.yml` wrote
  to that directory.
- The path shape the check asserts is the one the daemon looks for —
  `bundledPluginDirForExecutable` resolves `Contents/Resources/plugins` and
  `discoverPluginManifests` wants `<name>/attn-plugin.toml` — and it matches a
  real Tauri-built bundle: run against an installed non-production app, the
  check passes; run against a bundle with the directory removed, it names every
  missing file and fails.

https://claude.ai/code/session_013X61jPjh8UPsEiz2W4xaQV

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,6 +71,22 @@ jobs:
           mkdir -p app/src-tauri/binaries
           cp /tmp/attn "app/src-tauri/binaries/attn-aarch64-apple-darwin"
 
+      - name: Read bun pin
+        id: bun
+        run: echo "version=$(awk '/^bun / { print $2 }' .tool-versions)" >> "$GITHUB_OUTPUT"
+
+      - name: Set up bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: ${{ steps.bun.outputs.version }}
+
+      # tauri bundles `bundled-plugins/` as the app's `Resources/plugins`, and a
+      # checkout has only a .gitkeep there — the compiled plugins are build
+      # output. A source install gets them from scripts/build-app-profile.sh;
+      # this is the release's equivalent step.
+      - name: Stage bundled plugins
+        run: ./scripts/build-bundled-plugins.sh
+
       - name: Import Apple signing credentials
         env:
           APPLE_CERTIFICATE: ${{ secrets.APPLE_CERTIFICATE }}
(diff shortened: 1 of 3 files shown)


figgyster (dismissed):
Victor, this closes the hole #769 left in the release lane: the plugins finally get built on a cold runner, and the bundle gets inspected instead of trusted on its word. Good catch. But the guard is standing outside after the party has already started.

Requesting changes on `.github/workflows/release.yml:142-150`. `tauri-action` is configured with `releaseDraft: false`, so its “Build and publish release” step makes the GitHub release public before the new plugin verification runs. If that verification finds a missing plugin, the job fails with a public release and the action's initially uploaded app artifacts already out in the street—the exact bad release this PR means to prevent.

Please make the release remain a draft until the bundle check and repackaging/notarization have passed, or move an equivalent bundle verification into a pre-publication build path. The staging work itself and the new changelog entry look right; this is the last ordering seam.

figgyster (approved):
Victor, you turned one empty plugin directory into a proper release gate. The old red called out an open front door; this head leaves the whole release in draft until the app is verified and stapled, both Linux daemons are attached, and the final job opens it.

The ordering now holds. `releaseDraft: true` keeps `tauri-action`'s initial artifacts out of public view, `publish` waits on `tauri`, `linux-amd64`, and `linux-arm64`, and `gh release edit --draft=false --latest` is the CLI's documented draft-publication path. The release docs explain the rerun story too. Diff check and the refreshed CI board are green. Approved.

PR #771, opened 2026-08-06
fix(queue): wait for classification before auto-settle

victorarias:
## Intent / Why

A session can remain visibly working while its stop-time classifier decides whether the finished turn is idle or needs input. Auto-settle treated that held state as proof that work was continuing, so it could close the turn before the classifier returned and then reopen it when the verdict landed.

## Summary

- suspend an armed auto-settle while stop-time classification is in flight
- repeat the classifier-pending check at timer fire time to close the callback race
- re-evaluate auto-settle after classification so genuine continuing work still gets the normal arm and countdown
- cover both timer/classifier orderings and both live outcomes without changing auto-settle settings or timing policy

## Verification

- full Go test suite
- frontend unit tests: 2,533 passed; 15 skipped
- frontend production build
- Playwright E2E: 191 passed and 1 skipped; one unrelated terminal-rendering case transiently failed, then all 4 cases in that spec passed on an isolated rerun
- packaged-app verification: a classified finished turn kept its owed turn after the countdown appeared; a session that remained working through the full arm and countdown still auto-settled and handed over the next owed turn

### DIFF
--- a/app/scripts/real-app-harness/scenario-agent-queue.mjs
+++ b/app/scripts/real-app-harness/scenario-agent-queue.mjs
@@ -430,6 +430,58 @@ async function main() {
       );
     });
 
+    await runner.step('classification_suspends_auto_settle', async () => {
+      // The countdown must not treat the resolver's awaiting-verdict hold as
+      // proof that work continues. Give the short run enough time to reach the
+      // visible countdown, but a wide enough countdown that it reliably stops
+      // and starts classification before the settle deadline.
+      await client.request('set_setting', { key: 'auto_settle_arm_seconds', value: '5' });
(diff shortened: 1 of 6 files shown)


figgyster (approved):
You found the exact lie in the queue: green during a stop-time verdict was the resolver holding its breath, not the agent asking to be auto-settled. The cancellation at classification start handles the ordinary path, the fire-time predicate catches the callback already out in the rain, and a genuinely still-working session earns a fresh arm after the verdict instead of inheriting a stale deadline.

This is the same lifecycle honesty that made the recent conversation-runtime work worth shipping: final state gets to arrive before teardown decides the story is over. The daemon and Linux preflights are green, and the packaged-app witness covers both sides of the fork. Approved.

PR #772, opened 2026-08-06
feat(queue): pause settling during pointer movement

victorarias:
## Intent / Why

Auto-settle already pauses while someone types into an agent, but it could still resume and close the turn while they were actively reading or inspecting the terminal with the mouse or trackpad. Pointer movement should count as engagement with the same quiet-window behavior.

## Summary

- sample pointer movement inside terminal panes and send a session-scoped activity signal without adding PTY input
- fold pointer activity into the five-second auto-settle quiet hold while keeping ticket-nudge typing protection keyboard-only
- drop ephemeral pointer activity during socket disconnects instead of replaying stale movement
- extend protocol 214 and the packaged-app harness to cover sustained physical pointer movement

## Verification

- pre-commit gate: Go build and tests, 2,535 frontend tests, frontend production build, and 192 passing Playwright E2E tests with one skipped
- focused daemon and protocol tests after rebasing onto current main
- signed named-profile app and daemon preflight on protocol 214
- packaged-app settle scenario with real macOS keyboard and pointer input:
  - repeated pointer movement keeps settling paused beyond the quiet window
  - five quiet seconds restore a fresh full countdown
  - automation writes do not pause settling

### DIFF
--- a/app/scripts/real-app-harness/InputDriver.swift
+++ b/app/scripts/real-app-harness/InputDriver.swift
@@ -56,7 +56,7 @@ func parseOptions() throws -> Options {
     while index < args.count {
         let arg = args[index]
         switch arg {
-        case "activate", "activate_background", "frontmost", "display_state", "windowid", "windowlist", "text", "key", "keycode", "click", "right_click", "drag", "menu", "window_park", "scroll":
+        case "activate", "activate_background", "frontmost", "display_state", "windowid", "windowlist", "text", "key", "keycode", "move", "click", "right_click", "drag", "menu", "window_park", "scroll":
             options.command = arg
         case "--window-title":
             index += 1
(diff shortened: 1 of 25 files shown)


figgyster (dismissed):
This is the right next wrinkle after #771: you’re teaching the queue that watching an agent is work too, while keeping the keyboard-only nudge splice guard out of the blast radius. The session-scoped signal, reconnect drop, remote routing, and real pointer harness all point the same way. But I can’t sign the terminal callback off yet.

`app/src/components/GhosttyTerminal.tsx:836` writes `onPointerActivityRef.current` during render. React can abandon or replay that render, leaving the mounted mouse handler holding the callback from a pane/session that never committed; a later move can then freeze the wrong turn. The React Doctor error is right on this one. Please either call the committed prop from the JSX handler directly or update the ref from an effect, then rerun the affected frontend witness.

figgyster (commented):
That puts the callback back on the right side of the commit line. The sampler still stays cheap, while an abandoned render no longer gets to nominate which session a later mouse move protects.

React Doctor has cleared the new error, and I found no remaining code blocker in the follow-up. The rebuilt daemon, frontend, and E2E board is still running, so this is a current-head comment rather than a green stamp.

PR #773, opened 2026-08-06
fix(queue): persist approval routing across daemon restarts

victorarias:
## Intent / Why

When the daemon restarts while an agent worker survives, it loses the in-memory fact that a permission reviewer is answering that session’s approval requests. A brief Codex permission hook can then be published as `pending_approval`, flash as a user turn, and disappear when auto-settle runs—even though auto-review handled it.

## Summary

- Resolve an explicit per-session approval route (`user`, `reviewer`, or `bypass`) from the effective launch policy.
- Persist that route in both the durable launch intent and the surviving worker registry.
- Restore reviewer evidence before the first resolver tick after recovery, preferring what the live worker actually launched and repairing stale stored intent.
- Preserve the recorded route across reload and revive instead of consulting a later global auto-approve setting.
- Keep legacy records conservative when their approval route cannot be recovered.

## Verification

- `make test`
- `make build`
- `make build-linux-amd64`
- `make build-linux-arm64`
- Repository pre-commit Go format, build, full Go suite, and Tauri daemon-binary checks
- Signed non-production packaged-app preflight
- Live Codex restart scenario: the worker retained `approval_route: reviewer`; a restricted-network permission request was completed by auto-review; the daemon observed the `pending_approval` hook without applying that state or opening a user turn

### DIFF
--- a/changelog.d/bashful-ibex-persist-approval-route.yaml
+++ b/changelog.d/bashful-ibex-persist-approval-route.yaml
@@ -1,0 +1,6 @@
+kind: fixed
+area: queue
+change: >
+  Automatically reviewed permission requests stay behind the guardian dwell
+  after the daemon restarts, so brief approvals no longer flash as turns that
+  need the user and then disappear through auto-settle.
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -27,6 +27,7 @@ import (
 	"github.com/victorarias/attn/internal/daemon"
 	"github.com/victorarias/attn/internal/daemonctl"
 	"github.com/victorarias/attn/internal/hooks"
+	"github.com/victorarias/attn/internal/launchcontract"
 	"github.com/victorarias/attn/internal/pathutil"
 	"github.com/victorarias/attn/internal/present"
 	"github.com/victorarias/attn/internal/probetui"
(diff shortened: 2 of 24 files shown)


figgyster (commented):
You closed the restart-shaped hole in the queue without letting today’s global setting rewrite yesterday’s launch. That’s the right scar tissue after #771: the system gets to settle a turn only when it still knows who was meant to answer it.

I traced recovery, reload, and revive. The surviving worker’s recorded route wins over stale durable intent, legacy records fall back to the user rather than hiding a real approval, and the new recovery witness drives the brief guardian-reviewed approval through the exact path that used to flash and vanish. Good boundary: this is a launch fact now, not a setting lookup wearing a fake mustache.

The first Daemon run went red in the unrelated process-argv probe and installed-plugin close-log timeout; the changed daemon/recovery packages passed in that same run. A rerun is live now, so I’m leaving this commented until that required lane comes back green. No code blocker from me.

figgyster (approved):
The rerun held. Daemon is green now, alongside both Linux preflights, the benchmark, and the applicable board.

Approved. The approval route survives restart, reload, and revive as the fact it was when the session launched, so the queue stops mistaking a guardian-handled permission request for a user turn that may quietly settle away.

PR #779, opened 2026-08-06
feat(sessions): show what a conversation agent did, and let a queue be cancelled

victorarias:
Slice 3 of the pi headless host: see what the agent did. Base is
`pi-headless-host`, on top of slice 1 (#767) and slice 2 (#775).

## What was wrong

A conversation session drew what its agent said and nothing about what it did.
Two minutes of reading files and running commands looked like the agent had
stopped talking, and the only way to find out what it touched was to ask it.

## Tool cards

Every tool call becomes a card in the transcript where the call happened: the
tool's name, one line naming what it was pointed at (the command, the path, the
pattern), and how it ended. The card exists from the moment the call starts, so
a session grinding through a long command still shows what it is doing.

The cards are collapsed and lazy from day one. `tool_started` / `tool_finished`
carry only what the transcript keeps forever; what the tool actually read,
wrote or printed stays in the host until someone opens the card and the app
asks for it with the `tool_detail` verb. The receipt is the attach-snapshot
corpus — claude JSONL, p50 0.15 MB / p99 11.6 MB, ~0.4% of it message text —
and eagerly inlining tool output is how a transcript gets there.

`tool_detail` is addressed by `call_id`, not by a request id, and its answer is
broadcast. Two windows showing the same card cost one read, a `full` answer
upgrades both, and nothing can time out on a slow disk.

Tool boundaries are semantic but deliberately **not** state declarations:
`store.ApplyAgentDriverState` restamps `state_since` on every apply, so a
session that ran twenty tools would report having been working only since the
last one. `TestHostToolEventsDoNotRestampTheState` holds that.

## The other half, found live

pi hands each tool's whole output back to the model as a `toolResult` message
of its own, and the pane was drawing it — 23,893 bytes for a single
`seq 1 5000`, which is exactly the ballooning the collapsed card exists to
prevent. Those messages are dropped now; the card is that call's record.

Along the way, nothing is emitted on pi's `message_start` any more. pi opens a
message per assistant turn whether or not that turn says anything, so a
tool-only turn used to leave an empty bubble. A message now appears when it has
text: the first delta mints it, and one that streamed nothing is decided at its
end. Measured after the change: a session that read a file, printed 5,000
lines, edited a file and slept has a 406-character transcript.

## Full output, and patches

An output pi clipped (its cap is 2,000 lines / 50 KB) offers the whole of
itself, read back from the file pi kept. That read is bounded at 4 MB, and a
file past it is answered with its last 4 MB plus a line naming the limit, the
real size and where the whole file is.

An edit renders as a diff — `PatchDiff` from `@pierre/diffs`, fed pi's own
unified patch. Reconstructing before/after text from a 4-line-context patch
could only guess at the rest of the file, and guesses show up as wrong line
numbers in the gutter.

## Limits with receipts

The detail the host holds is bounded at 16 MB. Receipt, measured on this
machine against the compiled host and pi 0.83.0: a host idles at 130 MB RSS,
and an 8-call exploration run over a real Go package retained 52,463 bytes,
6.4 KB per call; a run whose `seq 1 5000` pi truncated retained 10,342 bytes
across 4 calls. 16 MB is ~320x the heavier of the two and holds ~330 calls even
if every one hit pi's own 50 KB per-result cap. Past it the oldest calls are
dropped, and a card denied one says so, naming the budget and the number of
evicted calls. Each run logs what it actually held, which is where the next
remeasurement comes from.

`tool_execution_update` is dropped on purpose — a live-updating card would
repaint continuously beside GPU terminals — and handled explicitly so it does
not log as an unmapped pi event.

## Cancelling a queue (the door slice 2 left open)

Anything sent to the agent and not yet read can now be dropped. All-or-nothing,
because `session.clearQueue()` is: pi offers no per-entry removal, and clearing
then re-queueing the survivors would race the drain. The strip empties on pi's
answering `queue_update`, never on the click, so what it shows stays pi's
answer about pi's queues rather than a local guess.

## Surfaces walked

- **Protocol.** `agent_tool_detail` and `agent_clear_queue` through the TypeSpec
  contract; `ProtocolVersion` 216 in `constants.go` and `useDaemonSocket.ts`.
  Generated files regenerated, never hand-edited.
- **Daemon.** Both commands routed in the main and remote-scoping switches, and
  entered in `CommandMeta` (session-scoped, unlogged like `agent_prompt` — a
  card is opened as casually as a key is typed).
- **Bus.** No new wire traffic outside projections; envelope forwarding is a
  byte stream on its existing path.
- **CLI / Linux.** Nothing new is reachable from the command line, and the host
  is TypeScript running under the plugin.
- **Docs.** `docs/glossary.md` gains the tool card, the detail, and the state
  declaration as a named subset; `plugins/attn-pi/AGENTS.md` gains the envelope
  families, the new verbs, the `toolResult` rule and why tool boundaries carry
  no state; the plan's slice 3 is ticked with what it actually shipped.

## Verification

Live in a packaged non-production app on a throwaway profile against a real
agent, then the profile was cleaned.

New scenario `pi-host-tools` covers all four acceptance criteria — accurate
cards for read/bash/edit, expanding a card fetching the output, a truncated
output delivering the whole of itself, a patch drawn as a diff, a queued steer
cancelled and never read — plus the negative that makes the cards worth having:
after 5,000 lines of tool output, no message in the transcript carries it.

`pi-host-conversation` and `pi-host-nudge` were re-run green against the changed
message emission. `make test-all` is green; the host suite is 34 tests.


figgyster (approved):
Approved on `244b574a`. Slice 3 of the headless-host plan pays off, and it pays off honestly.

The tool card is the right shape of lie-free UI: a declaration carries a name, a one-line summary and the files it named — never the output — so a session that ran a thousand tools still scrolls, and the 23,893-byte `seq 1 5000` message the pane used to inline is now the card's own business, fetched only when someone opens it. The receipts are in the diff and they're good ones: the p99 11.6 MB attach-snapshot corpus as the reason the lazy design exists, the measured 16 MB detail budget (~320x the heavier real run, ~330 calls at pi's own cap) with the per-run log line as the next remeasurement, the 4 MB tail read with its honest "here's the last 4 MB and where the whole file is" answer.

The details that make it trustworthy rather than just pretty:

- `tool_detail` addressed by `call_id`, broadcast, not a request/result pair. Two windows on the same card cost one read, a `full` answer upgrades both, and nothing can time out on a slow disk. The store's "full never overwritten by a late clipped answer" rule closes the multi-client race properly.
- Tool boundaries are deliberately *not* state declarations, and `TestHostToolEventsDoNotRestampTheState` pins it: `applyState` would restamp `state_since` on every call and reset "working for 4m" several times a minute. That's the lifecycle-honesty thread from #767/#771/#773 extended one more level down.
- `tool_execution_update` dropped on purpose, named in the mapper rather than left to log as an unmapped pi event. A card that repainted continuously beside GPU terminals would be the per-tool version of the balloon the collapsed card exists to avoid.
- Messages now mint on first delta instead of on pi's `message_start`, so a tool-only turn leaves no empty bubble and the `toolResult` message (pi handing the whole output back to the model) is unrendered — the card is that call's record, and drawing both would inline every byte the card exists to keep out.
- Queue cancel is all-or-nothing because that's all pi offers, and the strip empties on pi's answering `queue_update`, never on the click. The daemon reports no success it hasn't seen — `handleAgentClearQueue` sends the verb and lets the host's word be the word. Same for detail: no live host is an error naming it, not a card spinning forever.
- Patch as diff feeds `@pierre/diffs` pi's own unified patch rather than reconstructing before/after text a 4-line-context patch can only guess at. Right call, and the shiki-js highlighter choice matches the DiffView precedent under the Tauri custom protocol.

Verification is the strongest part of the whole PR. The `pi-host-tools` scenario runs the packaged app against a real agent and asserts all four acceptance criteria plus the negative that makes the cards worth having — after 5,000 lines of tool output, no transcript message carries it, and the edit card's diff is cross-checked against the file on disk. `pi-host-conversation` and `pi-host-nudge` re-ran green against the changed message emission, and the full suite is green (Go + 2,561 frontend tests, host suite 34). The Actions outage meant no CI run ever dispatched for this head — platform gap, not branch problem, and the local evidence is a fair witness for it.

One named follow-up, not a blocker: the 16 MB detail budget's eviction count only shows up in `missingReason` and the per-run log line. Fine for now — the log line is exactly where the next remeasurement comes from.

No blockers. This closes the "two minutes of silence" complaint for good: a conversation session now says what it did, and a wrong steer has a way out before the agent ever reads it.

figgyster (approved):
Re-approving on `3d4f850e` as the gate wants — and taking the correction in full.

On the mechanism I had it wrong: not a replay onto newer main, but a forced `--no-ff` replay minting a head Actions had never seen. That's the cleaner experiment, and it made the outage measurable — same tree, zero runs on `244b574a`, three within seconds on `35baa897`. I'll keep "same tree" and drop the "newer base" bit from the memory.

On React Doctor: the witness holds. The new StrictMode test failing `1 vs 2` with the decision back inside the updater, while the old assert stayed green, is exactly the "load-bearing" proof — the original test could not see the replay, and now the suite can. The point that the duplicate is invisible to live verification (second answer overwrites the first with identical bytes) is a good one: it is why this class of bug wants a replaying test harness and not another manual pass.

Approved on `3d4f850e`. Slice 3 is done — cards, on-demand detail, full output, diffs, queue cancel — and the impure updater is the last thing it shipped.

PR #780, opened 2026-08-06
fix(cli): reap conversation hosts and plugin drivers in profile clean

victorarias:
## Problem

After `attn profile clean` on a profile that ran pi-host sessions, driver processes survived: parented to init, executable inside the already-removed app bundle, invisible to everything. Clean reaped pty-workers and the daemon but knew nothing about the daemon's other children. Live reproduction found the observed leak is primarily the plugin supervisor's runtime processes (`bin/attn-pi`): a daemon that dies without running its shutdown (SIGKILL, crash) leaves them running with no durable record, and clean then deletes the only directory they could ever have been found through.

## Fix

New `internal/procreap`: a durable on-disk record per daemon child (pid, pgid, platform-native process start time as an identity stamp, command) written at spawn and removed on exit, plus a reap that works from the records alone — no daemon required.

- **Conversation hosts** (`internal/hostsession`) write a record per host; the host monitor removes it once the process is gone.
- **Plugin runtime processes** (daemon plugin supervisor) now spawn as process-group leaders and write a record per driver; the wait path removes it. The record path carries the pid so a restart's fresh record never races the old process's removal.
- **`attn profile clean`** reaps both registries (after workers, before removing the data dir) and reports per-registry outcome counts. Unidentified or SIGKILL-surviving processes are named with their pid instead of silently left behind.

Reaping never kills by pattern: a recorded pid is signalled only when its current start time matches the one recorded at spawn — a recycled pid or an unreadable stamp reads as "cannot identify", reported and left alone. Teardown is cooperative first (SIGTERM — for hosts this runs pi's dispose, the only path that reaches tool subprocesses, which lead their own process groups), with a 3s-grace SIGKILL backstop; a group sweep happens only for a child that led its own group.

## Verification

Unit tests spawn real processes: cooperative termination, SIGTERM-ignoring escalation, recycled-pid refusal, empty-stamp refusal, dead-entry reporting, cooperative teardown reaching a detached (own-pgroup) child, and never group-signalling a shared-group entry. Manager/launcher tests assert the record appears at spawn (pid, pgid=pid, stamp matching the live process) and disappears on exit. Linux amd64/arm64 cross-compiles pass.

Live, packaged non-production app on a throwaway profile (`reapglen`, cleaned afterwards):
- pi-host session spawned, prompt round trip settled, second run parked a `sleep 300` tool call in flight (own process group, confirmed by `ps`).
- Daemon SIGKILLed by its recorded pid: the plugin driver survived reparented to init (the ticket's observed leak, reproduced); the host self-exited on stdin EOF and its dispose tore the tool child down.
- `attn profile clean`: `hosts 1 registered (1 already gone)`, `plugins 1 registered (1 terminated)`; driver pid confirmed gone; app + data dir removed.
- Idle-host and mid-run-host cooperative reap legs (host wedged as an init-parented orphan, tool child torn down via the host's SIGTERM handler) were additionally live-verified on the same branch before the procreap generalization; that behavior is now pinned by the procreap unit tests, which run the same scenarios with real processes.

https://claude.ai/code/session_01L2rhXCoLNM7Bdpyx3ELpiD


internal/procreap/procreap.go
```diff
@@ -0,0 +1,298 @@
...
+		rest := string(raw)
+		if i := strings.LastIndexByte(rest, ')'); i >= 0 {
+			fields := strings.Fields(rest[i+1:])
+			if len(fields) >= 20 {
+				return fields[19], nil
+			}
+		}
+		return "", fmt.Errorf("unparseable /proc/%d/stat", pid)
+	}
+	out, err := exec.Command("ps", "-p", strconv.Itoa(pid), "-o", "lstart=").Output()
```
figgyster: On Darwin this is the identity gate, and it resolves to whole seconds. The Darwin ps(1) man page defines `lstart` as the `%c` strftime format — no sub-second component. A process that died within the same second it was spawned leaves a recorded stamp that any recycled pid born in that same second will match exactly, so the gate opens for a process this entry does not own; the group sweep then extends the mistake to a whole unrelated process group. The comment on line 273 claims second granularity is "far finer than any realistic pid-reuse window" — it's the opposite: it's the coarsest stamp this package produces anywhere, on the platform the app ships on. Use proc_pidinfo(PROC_PIDTBSDINFO) (pbi_start_tvsec/tvusec, microsecond) or sysctl kern.proc.pid.<pid> (kinfo_proc.p_start timeval) on Darwin, via golang.org/x/sys/unix — or fail closed when only second-level stamps are available.

figgyster (changes requested):
The net under the hole has a hole on the platform this app actually ships on.

This PR's whole safety story is the identity gate: a recycled pid reads as "cannot identify" and gets left alone. On Linux that's true — /proc/<pid>/stat starttime is clock ticks since boot, fine-grained, and the recycled-pid test means something. On macOS the fallback is `ps -o lstart=`, which the Darwin man page defines as the `%c` strftime format: whole seconds, nothing below. A pid recycled inside the same wall-clock second as the recorded spawn produces a string-identical stamp, sails through processAlive plus the equality gate, and then the reaper SIGTERMs it, SIGKILLs it on the backstop, and — if the imposter leads its own group — sweeps the entire group. That is the kill-by-pattern this PR promises never to do, aimed at an unrelated process, on the platform where the leak actually bit you.

The comment on line 273 has the arithmetic backwards: one second is a coarse window, not "far finer than any realistic pid-reuse window." macOS hands freed pids back out promptly; the only thing standing between a crash-at-startup host and an innocent victim is which 1-second bucket the next fork landed in. On a busy machine that's a coin flip, not a gate.

Fix direction: on Darwin read the birth time at microsecond resolution — proc_pidinfo(PROC_PIDTBSDINFO) gives pbi_start_tvsec/pbi_start_tvusec, sysctl kern.proc.pid.<pid> gives kinfo_proc.p_start as a timeval, and golang.org/x/sys/unix reaches both. Or fail closed: if all the platform offers is second granularity, report "unidentified" and walk away — but that makes profile clean a no-op on macOS, so the high-res API is the real fix. TestReapLeavesARecycledPIDAlone only fakes a stamp mismatch; no test can pin the Darwin gate until the stamp itself resolves below a second.

Honesty note: I could not re-run the Go suite from the sandbox this turn — no Go toolchain on this box — so this review rests on the code, the Darwin ps(1) man page, and the diff. The Linux-side reasoning is unchanged by that; the blocker is the fallback path, and the fallback path is macOS.

figgyster (commented):
The gate now stands on every platform this app ships on. `ps -o lstart=` is gone from the codebase; the stamp comes from sysctl kern.proc.pid (kinfo_proc p_starttime, a timeval) at microsecond resolution on Darwin, and /proc starttime at 10ms on Linux — each file carrying its own measured pid-reuse floor to beat, with the same-second collision measured at 34ms on the author's own machine. That's the arithmetic this package needed: resolution-versus-reuse, not "any two processes differ."

Two things I didn't ask for, both right:

- **No generic fallback.** An unsupported platform fails to compile instead of reaping on a stamp nobody checked. Fail-closed by construction, which is the strongest form of the alternative I offered.
- **ReapUnreadable.** A record this build can't decode is now named and reported with its error, instead of silently skipped — and silence there read as "nothing was registered," the exact opposite of the truth. The profile clean output for it points a human at `ps`. That closes the second-order hole where a corrupt record strands a process invisibly.

The new tests pin the right properties: two processes spawned well under the reuse floor must carry different stamps (a second-granularity source fails this on Darwin), the stamp is stable across reads, and an undecodable record is reported, not dropped.

One thing stands between this and green, and it's not substance: **gofmt**. The Daemon lane dies at the Check formatting step — both new files need a blank line after `const stampResolution` (`gofmt -w internal/procreap/`; the lane shows the exact diffs). Vet, Test, and the cross-compiles never got to run because the gate trips first. Push that and I'll approve.

figgyster (approved):
Approved on 6e2b6958 — the gofmt push, and the board is green all the way down: Daemon (format, vet, test, hooks), Frontend E2E, Changelog, Changes, React Doctor at 0 errors, and Bench A/B with its benchstat delta and summary published.

And I owe you one, on the record. Your receipt is right and my "coin flip" was wrong: first reuse at 2m37s of nothing but forking, 92043 spawns at 588/s — the one-second stamp had two orders of magnitude of margin at that fork rate, not a 50/50. You made the change anyway because the margin was a timing coincidence measured on one idle laptop, and a safety gate that rests on an unmeasured margin is the landmine. That's the right call, and the receipt belongs in the code, which is where you put it.

You also caught what I missed: my confidence in the Linux path was the same mistake wearing a different hat. /proc starttime is 10ms ticks, your arm64 VM put two processes 3ms apart on identical stamps, and the resolution test — the one you wrote because I said no test could pin the Darwin gate — is what caught it. Mutation-verified too: revert to `ps -o lstart=` and it fails with the actual stamps on the actual pids. That's the witness the gate needed, on both platforms, with the property stated correctly: resolution-versus-reuse-floor, not "any two processes differ."

The unreadable-record reporting rounds it out — silence that read as "nothing was registered" is now a named, explained `unreadable` line in profile clean.

The identity gate now stands on every platform this app ships on, with measured floors under it and tests that would scream if the stamp ever got coarse again. This is the lifecycle-honesty arc from #771/#772/#773 paying off in crash recovery, and it's done honestly. Case closed.

PR #783, opened 2026-08-07
Integration: docstore Stage 2, sortable background-work and turn timestamps

victorarias:
## What this is

During the 2026-08-06 GitHub Actions outage, in-flight lanes stopped targeting main and merged onto an integration branch (`epic/integration`, cut from main at 01ab976a) on the strength of their local receipts, with review and CI deferred to this PR. This is that PR: full CI and review of the combined diff, then a merge commit preserving the three already-squashed changes.

## The three changes

**`4e4c47e8` — fix(daemon): store background-work timestamps in an encoding that sorts (#774).** Jobs and notifications wrote `time.RFC3339Nano` into TEXT columns compared as text; variable fraction widths mean text order is not time order within a second (a job scheduled on a whole second was not claimed until the next). Both now write the fixed-width `docstore.TimeFormat` under the shared name `sortableTimeFormat`; reads decode any RFC3339 spelling; migration 93 restamps stored rows (idempotent, blanks preserved).

**`82dfb928` — feat(docstore): send only what changed on a live query (#776).** A3.4 Stage 2: live-query deliveries become windowed — `{delivery, as_of_seq, order, upsert}` with changed bodies only; resume by content via `have` (`DocumentRevision[]` pairs — deliberate deviation from the plan's `Record<string, int64>`, which decodes untyped on the Go side; rationale in main.tsp); subscribe refuses `after` cursors with a coded error; subscription-ending codes (`collection_undefined`, `collection_redeclared`); the duplicate `compileDocQuery` preflight is deleted (acceptance is the first read); `internal/client` and `attn doc watch` updated (watch now exits non-zero when the daemon ends the subscription, and reconnects on connection loss only — a mid-flight fix for a silent resubscribe loop). Adds the plan's full test battery plus scoped `-race` (store+docstore) to CI. ProtocolVersion 215.

**`e44a1325` — fix(daemon): store turn, cursor and listing timestamps in an encoding that sorts (#782).** The same defect class swept across the rest of `internal/store`: sessions turn stamps (the serious one — the turn-open guard compared text, and day-named snoozes zero milliseconds so `turn_opened_at` systematically lands on whole seconds, which sort above every stamp in their own second: a settle in that second silently prevented the next turn from opening), automation provider cursors, delegation operations, endpoints (previously written with a local zone offset — misordered across zones/DST), workspaces, workspace layout panes, and workflow runs (normalized at the store boundary). Migration 94 restamps existing rows. Deliberately left, with reasoning in #782: all `formatTicketTime`/`time.RFC3339` columns (tickets, automation runs, present, bus events) — fixed-width, already sorting correctly.

## Verification

Each change carried its own receipts at merge time (full Go suite, `-race` on touched packages, falsification tests that go red when the format constant alone is reverted, migration witnesses on sandbox copies of the production DB with idempotence verified byte-identically, and live verification on throwaway profiles — Stage 2 on a full `make install` at protocol 215; the timestamp fix with the queue scenarios green on a fresh profile, including settle-survives-daemon-restart). The combined branch was additionally verified after each later merge: full Go suite green at `82dfb928`-on-`4e4c47e8`, and PR #782's CI ran green once Actions recovered (Daemon, Changelog, Changes, react-doctor; the rest path-skipped).

Migrations 93 and 94; ProtocolVersion 213 → 215; three changelog fragments ride along.

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -199,7 +199,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '214';
+export const PROTOCOL_VERSION = '215';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 // AutomationActionTimeoutError distinguishes "the daemon never sent a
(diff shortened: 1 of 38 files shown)


figgyster (changes requested):
The windowed live query is a real step up — the resume contract, the invariant fuzz, the `after`-cursor refusal, the client's one-rule apply. CI's green across the board on this head, and the timestamp work is the same honest fix as migration 91, swept across the rest of the store. But there's one hole in the lifecycle honesty this PR is built on, and it's in the redeclare path.

**`handleDocDefine` never wakes live subscriptions. The `collection_redeclared` ending arrives only when an unrelated write happens to land on the collection — or never.**

`internal/daemon/documents.go:403` — after `DefineDocumentCollection` succeeds, the handler sends its response and walks away. No fact, no wake. Contrast the undefine path: `handleDocUndefine` (:423) publishes `FactDocumentCollectionRemoved` via `publishCollectionRemoved` (:145), and `wakeDocumentSubscriptions` (:152) pokes every subscriber on that target. The subscribe filter (:95) listens for exactly two facts: `document.changed` and `document.collection.removed`. There is no third fact for a redeclare.

So when a collection is redeclared without a field a live query filters on, the delivery loop is parked on `sub.wake` (the select at ~:712) holding a stale window. The daemon believes it's serving; the client renders a window that no longer exists server-side; nothing tells either of them. Only if some later document write happens to wake the subscription does the declaration re-read fail and the `collection_redeclared` ending finally go out. If the collection goes quiet, `attn doc watch` hangs on a frozen list forever.

The PR's own contract says the opposite: "The subscription ends when the caller closes the connection, or when the collection stops being able to answer the query" (main.tsp), and the daemon comment in `handleDocSubscribe` says the same. A redeclare that drops a queried field *is* the collection stopping being able to answer the query — at redeclare commit time, not at the next arbitrary write.

The test knows this and papers over it: `TestRedeclaringWithoutAFieldEndsTheLiveQueriesUsingIt` (documents_test.go:549) does `putDoc(t, d, "b", ...)` at :575 with the comment "A write is what wakes the subscription; the redeclare alone is not a document change" (:573). That comment documents the gap, not the contract. The ending should arrive on the redeclare itself; the putDoc is scaffolding for a missing wake.

The fix is symmetric and small: publish a collection-redeclared fact from `handleDocDefine` (mirroring `publishCollectionRemoved`), include it in the subscribe filter, and let the delivery loop's declaration re-read do the rest — a redeclare that keeps the queried fields costs one redundant delivery (which this design already prices at "nothing"), and one that drops a field ends the subscription with `collection_redeclared` immediately. Then drop the putDoc from the test and assert the ending arrives with no intervening write.

No Go toolchain in my sandbox, so I couldn't run the suite locally — CI is my witness here, and it's green. This is the one blocker; everything else in the integration holds up.

figgyster (commented):
That's the fix, and the store-seam refinement is the right call — the define's own transaction reporting whether it was a redeclare beats any check-then-act arrangement, because only that transaction can see both the old declaration and the new one without a race. A first declaration has no watchers, so publishing only on `redeclared` is exactly right, and the delivery loop's declaration re-read was already the machinery that turns the wake into an ending — it just never got woken.

Verified on 61a8fdc0:

- `handleDocDefine` publishes `document.collection.redeclared` only when the store says the define replaced an existing declaration; the fact joins the subscribe filter and `CompactableFacts` (same invalidation-of-a-store-backed-subject reasoning as the other two document facts), and `wakeDocumentSubscriptions` decodes it like its siblings.
- `TestRedeclaringWithoutAFieldEndsTheLiveQueriesUsingIt` drops the scaffolding `putDoc` and asserts the `collection_redeclared` ending arrives with no intervening write — the falsification (publish disabled → test trips the 300s tripwire) is honest, since the helper deliberately blocks on a real signal rather than a duration.
- The `TestDocumentsOverARealSocket` race is a real catch: the second window is delivered when the write commits, which can precede the writer's own RPC response, so the seq traveling over a channel is the correct fix, not a shrug.

The same-schema redeclare costing one redundant delivery is the correct price — this design already priced redundant deliveries at "nothing," and the watch surviving it is the witness that matters.

Daemon, Frontend, and the E2E lanes are still in flight on this head; `Changes` is green. No remaining code blocker from my side — will sign off when the board clears.

figgyster (approved):
Green across the board on 61a8fdc0 — Daemon, Frontend, all three E2E shards, Changes. The one red (E2E 3/3, `split-blank-repro` paint-count timing) cleared on rerun; it was a terminal-render race in a spec this PR's diff never touches, so I'm not reading anything into it.

The redeclare fix holds up on close reading, so here's the sign-off:

- Store reports `redeclared` from inside its own define transaction — no check-then-act race, and the first-declaration case correctly publishes nothing.
- The `document.collection.redeclared` fact joins the subscribe filter and `CompactableFacts` consistently with its siblings, and `wakeDocumentSubscriptions` handles it through the same decode-and-wake path as removal.
- The delivery loop's declaration re-read does the real work, exactly as designed: surviving fields cost one redundant delivery, a dropped field ends the subscription with `collection_redeclared` at redeclare time.
- The test's scaffolding `putDoc` is gone and the ending is asserted with no intervening write; the falsification (publish disabled → tripwire timeout) is honest.
- The socket-test race fix is a genuine catch, not a drive-by: the window can be delivered before the writer's RPC response, so the seq belongs on a channel.

The quiet-collection hang I flagged is closed: a redeclared collection's watchers now hear about it at the moment the declaration moves, which is the only honest time. Approving.

figgyster (approved):
Re-affirming the approval on the merge head c938349c. The merge of main into `epic/integration` is the point of this PR, and it resolves the only real hazards cleanly:

- The redeclare fix rides through untouched — `documents.go` and `bus.go` are byte-identical to the approved 61a8fdc0 state; the `document.collection.redeclared` fact, the subscribe filter entry, and the no-scaffolding test all survive.
- The migration collision is resolved the honest way: main's annotation-note migration keeps 93, and the timestamp rewrites renumber to 94/95 with the dispatch table and both `applyMigration9X` functions matching. No duplicated numbers, no re-applied work.
- Protocol is in lockstep at 216 across `constants.go`, the regenerated Go and TS artifacts, and `useDaemonSocket.ts`; the new annotation `note` fields ride the regenerated contract.

Board on this head: all lanes green — Daemon, Frontend, E2E 1-3, Linux daemons both arches, Bench A/B, react-doctor, Changes — with only the usual path-skips. Nothing new to flag; the integration holds together.

PR #784, opened 2026-08-07
fix(ci): remove a load flake and give the next one a witness

victorarias:
Two CI flakes: one closed at the cause, one instrumented so the next occurrence says something actionable.

## ptyworker reap fixture — argv is not readable when `Start()` returns

`spawnSleeper` started a shell and immediately let the test read its argv. On Linux the exec pipe that unblocks the parent is closed in `begin_new_exec`, *before* `create_elf_tables` publishes the new argv, so `/proc/<pid>/cmdline` can read back empty when the parent looks straight away — and `processHasArg` then honestly reports "no such process".

The sleeper now echoes a readiness byte and the helper blocks on it. argv is in place before the child executes its first instruction, so a byte from the child proves argv is published — and proves it *without* consulting `processHasArg`, which this file also has to test honestly.

Receipts, measured on Linux under 20 spinning shells:

| | first-read miss | `TestProcessHasArgIdentifiesOwnProcess` |
|---|---|---|
| before | 10 of 400 spawns (worst wait 3.5ms) | 14 failures in 2000 runs |
| after | 0 of 400 | 0 in 2000 |

The 30s read deadline is a tripwire past every healthy spawn — only a sleeper that never execs at all reaches it.

## plugin-driver end-to-end — the flake now explains itself

This one is not fixed, because a bare deadline never said enough to fix it. Its only observed failure is a `driver.session_closed` notification that never arrives, across a chain that spans three processes: the worker sees the PTY child exit, the daemon ends the driver run, the plugin receives the notification, the fixture appends to its log.

- The test keeps the daemon's own log, and on timeout prints who saw what: session state, active driver run, whether the driver is still registered, fixture stderr (a fixture that died takes its connection with it, and the daemon then drops the notification as "owner disconnected"), and the daemon log tail.
- The daemon logs the three points that were silent: a PTY exit deferred until a racing launch completes, the two early returns in `closePluginDriverSession` that skip the notification, and the successful notification itself — the one line separating "the daemon never sent it" from "the plugin never acted on it".

The 5s deadline stays, now as a tripwire rather than a wait: the whole chain finishes in under 50ms across 240 samples on Linux with six copies of the test running in parallel under CPU load (worst 48.2ms). Reaching 5s means a link went quiet, not that the machine was slow.

## Dropped in rebase: the split-blank e2e gate

This branch also carried a fix for `split-blank-repro.spec.ts`, whose setup gate polled `trace[trace.length - 1]` and so read a renderer skip (`quads: null`) as a blank surface. #777 landed a byte-identical fix first, so the change is gone from this branch and only its comment wording differed. Recording the receipt here since that fix shipped without one: 1 of 10 local runs finished setup on a skip, with the trace `[1, null, 6302, null]`.

## Verification

- `go test ./internal/ptyworker -run 'TestProcessHasArg|Reap' -count=3`, `go test ./internal/daemon -run TestPluginDriver` — green before and after the rebase.
- `gofmt`, `go build ./...`, `go vet`, `go run ./cmd/changelog-check` — clean.
- The new diagnostics were exercised for real, not just compiled: an overlay copy of the test with an unreachable close-record count forced the timeout, and the failure printed the full daemon log for a live daemon + worker + plugin-fixture run, including the new `plugin session close notified` line. The production change here is log lines only, so this three-process run is the integrated evidence; there is no app-observable behavior to verify in a running app.

### DIFF
--- a/changelog.d/tipsy-weasel-ci-flake-fixes.yaml
+++ b/changelog.d/tipsy-weasel-ci-flake-fixes.yaml
@@ -1,0 +1,10 @@
+kind: internal
+area: tests
+change: >
+  Removed the top CI flake and gave the next one a witness. The ptyworker reap
+  fixture now waits for its sleeper to signal before reading its argv, since a
+  process's argv is not readable the instant `Start()` returns (14 failures in
+  2000 loaded Linux runs, 0 after). The plugin-driver end-to-end test keeps a
+  daemon log and prints who saw what when a driver.session_closed notification
+  goes missing, and the daemon logs the three paths that used to drop or defer
+  that notification silently.
--- a/internal/daemon/plugin_driver.go
+++ b/internal/daemon/plugin_driver.go
@@ -434,6 +434,10 @@ func (d *Daemon) queueExitDuringPluginLaunch(info ptybackend.ExitInfo) bool {
 	if !ok || info.LifecycleID == "" || launch.RunID != info.LifecycleID {
 		return false
 	}
+	// The exit is deferred, not delivered: finishPluginSessionLaunch replays it
+	// once the launch it raced is done. Say so, because between here and there
+	// the session looks alive to everything that reads the store.
+	d.logf("deferring plugin PTY exit until launch completes: session=%s run=%s", info.ID, info.LifecycleID)
 	d.pluginExits[info.ID] = info
 	return true
 }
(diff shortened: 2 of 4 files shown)


figgyster (approved):
Approved — but the body's arithmetic is off by one, and I want it on the record so it doesn't bite later. You say three flakes closed; this diff closes two. The split-blank fix already rode in with #777's merge — 487adce1 is this PR's own base, and it carries the `lastRealDraw` gate. So "the gate now reads the last real draw" is main's story, not this branch's. The changelog fragment and the commit message both count correctly; only the PR body is ahead of the code. No action needed. Just flagging it so some future archaeologist doesn't dig this diff for a fix that isn't here.

The two that are here hold up under cross-examination.

**reap fixture** — right diagnosis, right fix. Linux genuinely unblocks the exec pipe in `begin_new_exec`, before `create_elf_tables` publishes the new argv, so the parent can look straight away and read an empty cmdline. A readiness byte from the child proves argv is published, and proves it without consulting `processHasArg` — which matters, because that function is the thing under test in this file. The trailing `:` still keeps sh from exec'ing the marker away, and the 30s deadline is a tripwire, not a wait. Ran `TestProcessHasArg|Reap` three times in the sandbox: green, `TestProcessHasArgIdentifiesOwnProcess` included.

**plugin-driver witness** — the three silent points are now all on record: the deferred exit in `queueExitDuringPluginLaunch`, both early returns in `closePluginDriverSession`, and the notification itself. That last line is the one that separates "the daemon never sent it" from "the plugin never acted on it", and the timeout dump — session state, active run, driver registration, fixture stderr tail, daemon log tail — turns a bare deadline into an autopsy. The whole `TestPluginDriver` suite is green locally, three-process e2e included. The new diagnostic's registry read is RLock-guarded and the store read is the canonical one, so the failure path doesn't smuggle in a race.

Board's green all the way down: Daemon, Linux amd64/arm64, Bench A/B, Changelog, Changes, react-doctor. Frontend/Tauri/Plugins correctly skipped — this PR doesn't touch them. Local: `go build ./...`, `go vet`, gofmt, `git diff --check`, and `changelog-check` all clean.

The lifecycle-honesty arc keeps paying into the test suite itself: the CI now says what it saw, and what it didn't see, instead of failing mute. That's the right direction for a flake — make the next occurrence confess.

PR #785, opened 2026-08-08
feat(sessions): remember where each repo's sessions run

victorarias:
The new session picker always opened focused on the create-worktree form. That is right for a repo you always branch, and wrong for a repo you always work on directly — which cost an arrow key on every launch, forever.

## What changed

The last deliberate destination is now recorded **per repository**, and the chooser opens on it:

- creating a worktree records `new_worktree`
- opening the main checkout records `main_repo`
- opening an *existing* worktree is a one-off ("resume that one"), not a habit — it leaves the memory alone

`main_repo` opens the chooser on the main repo row so Enter reuses the checkout; anything else keeps today's create-form-focused behavior, so a repo you always branch still costs a single Enter. An exact typed worktree path still wins over the memory — that is an explicit "open this one" — and the create form stays one `↑` away from the main repo row.

## Where it lives

The daemon `settings` table, under `new_session_destination_<scope>_<repoRoot>` where scope is `local` or `endpoint_<id>` — the same path on this machine and on a remote endpoint are different checkouts, worked on in different ways. Unset means `new_worktree`, the default the picker has always had, so existing installs behave exactly as before until a choice is made. `set_setting` rejects unknown keys, so the prefix is registered in `validateSetting`, and the value is validated there with both allowed values named in the error.

Writes are skipped when the value has not moved: the picker is opened constantly, and re-storing the same value would broadcast a settings snapshot to every client for nothing.

## Two judgment calls

- **Recorded on intent, not on git's success.** A failed worktree create still records `new_worktree`. The user chose a destination, and the fire-and-forget create-worktree-session path has no result to await. Victor confirmed intent is enough.
- **Two states, not three.** Remembering a specific worktree path was considered and dropped: a worktree is transient, and "resume that one" is not a habit worth reopening on.

## The way out

Choosing the other destination once flips the memory, and `↑` from the main repo row returns to the create form. There is no third state to get stuck in.

## Verification

Live, in a packaged non-production profile (preflight PASS, CLI/app/daemon in protocol lockstep), driven through the UI automation bridge against a scratch git repo:

1. first visit — chooser opens on the create form, main repo row not armed
2. start a session in the main checkout
3. reopen — chooser opens with the main repo row armed
4. create a worktree instead
5. reopen — back to the create form

The daemon logged zero validation failures, and the setting persisted in the profile DB with exactly one write per deliberate choice. Profile cleaned up afterwards.

Automated: 80 picker tests (`LocationPicker`, `RepoOptions`), the daemon settings-validation table, full frontend suite, `tsc --noEmit`, `go build ./...`, `go test ./internal/daemon`. The three load-bearing assertions were mutation-checked — dropping the focus seed fails 2 tests, dropping the recording calls fails 3, widening the Go validator fails its case.

## Surfaces

- **CLI** — no default to remember: `cmd/attn` takes an explicit path or `--no-worktree` per invocation.
- **Protocol** — unchanged; `set_setting` is already free-form key/value, so no `ProtocolVersion` bump.
- **Linux** — the daemon change is pure Go with no platform surface.
- **Docs** — `app/src/components/NewSessionDialog/CLAUDE.md` plus a changelog fragment.

https://claude.ai/code/session_014Qt3hv4NRhBNqLyWcrEH9i

### DIFF
--- a/app/src/components/LocationPicker.test.tsx
+++ b/app/src/components/LocationPicker.test.tsx
@@ -606,6 +606,159 @@ describe('LocationPicker', () => {
     });
   });
 
+  // A repo is habitually branched or habitually not, so the picker remembers the
+  // last destination chosen for it and opens there next time.
+  describe('remembered repo destination', () => {
+    const repoRoot = '/home/remote/projects/exsin';
+    const destinationKey = `new_session_destination_local_${repoRoot}`;
+
+    const inspectRepoRoot = () => vi.fn(async () => ({
+      success: true,
+      inspection: {
+        input_path: '~/projects/exsin',
+        resolved_path: repoRoot,
+        home_path: '/home/remote',
+        exists: true,
+        is_directory: true,
+        repo_root: repoRoot,
+      },
+    }));
+    const repoInfoOf = () => vi.fn(async () => ({ success: true, info: buildRepoInfo() }));
+
+    const openChooser = async () => {
+      const input = screen.getByTestId('location-picker-path-input');
+      fireEvent.change(input, { target: { value: '~/projects/exsin' } });
+      fireEvent.keyDown(input, { key: 'Enter' });
+      await waitFor(() => expect(screen.getByTestId('repo-options')).toBeInTheDocument());
+    };
+
+    it('records the main repo when a session is started in the checkout', async () => {
+      const setSetting = vi.fn();
+      const { onSelect } = renderPicker(
+        { onInspectPath: inspectRepoRoot(), onGetRepoInfo: repoInfoOf() },
+        { setSetting },
+      );
+
+      await openChooser();
+      fireEvent.click(screen.getByTestId('repo-option-0'));
+
+      expect(setSetting).toHaveBeenCalledWith(destinationKey, 'main_repo');
+      expect(onSelect).toHaveBeenCalledWith(repoRoot, 'claude', undefined, false, false);
+    });
+
+    it('records a new worktree when one is created', async () => {
+      const setSetting = vi.fn();
+      renderPicker(
+        {
+          onInspectPath: inspectRepoRoot(),
+          onGetRepoInfo: repoInfoOf(),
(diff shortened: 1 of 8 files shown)


figgyster (approved):
Read the whole file on this one — diff, wiring, focus logic, validator, tests, board. Approving.

The one-off vs habit distinction is real in the code, not just the prose. Opening an existing worktree runs `handleSelectWorktree` → plain `launchSelection`, no memory write. The create form records `new_worktree` inside `handleCreateWorktree`; the main row records `main_repo` before launching. Three paths, two record, and the right two.

The focus seed in `RepoOptions` (`committedDestinationIndex > 0 || preferredDestination === 'main_repo'`) keeps the typed-worktree-wins precedence intact, and RepoOptions remounts per repo resolution — back resets `repoInfo`, so the mount-time focus zone can't go stale across repo switches. Endpoint switching resets the whole flow (`setRepoRootPath(null)`, `setRepoInfo(null)`, back to path-input), so the `endpoint_<id>` scope genuinely can't pick up a local repo's key. Scope isolation holds in the code, and the endpoint test pins it.

Recording on intent is the right call, and the reasoning holds up: the fire-and-forget create path closes the picker before the create finishes, so success-aware recording would mean two different rules for two paths. One rule, applied everywhere, beats that. And the write-skip on unchanged value is what keeps the settings broadcast honest — the picker opens constantly, and without the skip every open re-broadcasts a snapshot for nothing.

The mutation claims check out against the suite: dropping the focus seed breaks the two main-row tests, dropping the record calls breaks the three write tests, widening the Go validator breaks the unknown-destination case. Load-bearing tests, not decorations.

One stale detail, cosmetic: the PR body's verification says "protocol 215" but head and main both sit at 216 — that packaged-app run predates #783's bump. The diff touches no protocol surface and head is in lockstep, so there's nothing to fix; just don't let that number age into lore.

Board's fully green: Daemon, Frontend, all three E2E shards, react-doctor, Linux daemon on both arches, Bench A/B, Changelog, Changes. Plugins/Tauri skipped by path. The arrow-key toll is retired.

PR #786, opened 2026-08-08
docs(procreap): record why daemon children are reaped from files

victorarias:
The daemon's long-lived children — pty-workers, conversation hosts, and plugin
runtime drivers — each write a durable record as their own file under the data
dir, and `attn profile clean` reaps them by iterating those directories
(`internal/procreap`, `internal/ptyworker/reap.go`). The obvious alternative is
the SQLite store, which sits in the same data dir and already holds everything
else the daemon persists. Nothing in the code says why it was not used, so this
adds a decision record.

The short version:

- **Opening the store is a write.** `store.OpenDB` creates the base schema and
  runs the full migration chain, taking a pre-migration backup on the way. So
  reading a few pids out of it means migrating a database — and then deleting
  the directory it lives in, since the caller is `profile clean`. Opening
  `mode=ro` skips the migration but not the deeper problem: a WAL left by a
  process that died mid-write needs write access to recover, so read-only is the
  open most likely to fail on exactly the crashed daemon being cleaned up after.
- **Recovery state must not live inside the thing it recovers from.** A live
  host/driver record means the daemon died without running shutdown, which is
  when the database is least trustworthy.
- **There is no query.** Reaping iterates a handful of records once — no
  filters, no joins, no ordering. Indexes and transactions buy nothing, and the
  write side gets worse: the record has to land right after `cmd.Start()`, and
  an atomic tmp+rename does not contend with a daemon mid-transaction.

It also writes down what the three registries are and how they differ — workers
use an authenticated control socket and deliberately outlive a daemon restart,
while hosts and drivers have no RPC surface and must never — plus the condition
that flips the argument: the moment the reaper has to *answer* something (reap
one workspace's children, show strays in the UI) rather than iterate, the files
become a cache of the store and should move rather than grow an index over a
directory.

Docs only, plus a pointer to the record from `procreap`'s package doc, following
how the terminal-annotations decision is referenced from the code it governs. No
behavior change.

https://claude.ai/code/session_01L2rhXCoLNM7Bdpyx3ELpiD

### DIFF
--- a/docs/decisions/2026-08-08-daemon-children-are-reaped-from-files.md
+++ b/docs/decisions/2026-08-08-daemon-children-are-reaped-from-files.md
@@ -46,9 +46,11 @@ creates the base schema and runs the full migration chain, taking a
 pre-migration backup on the way. Reading three pids out of the database would
 mean migrating it — and then deleting the directory it lives in, since the
 caller is `attn profile clean`. Opening `mode=ro` avoids the migration but not
-the deeper problem: a WAL left behind by a process that died mid-write needs
-write access to recover, so read-only is the open most likely to fail on exactly
-the crashed daemon being cleaned up after.
+the deeper problem: the store runs SQLite's default rollback journal (nothing
+sets `journal_mode`), so a process that died mid-write leaves a hot `-journal`
+that the next open has to roll back — which needs write access. Read-only is
+therefore the open most likely to fail on exactly the crashed daemon being
+cleaned up after.
 
 **There is no query.** Reaping is "iterate everything and signal it": no
 filters, no joins, no ordering, a handful of records read once. Indexes and
--- a/docs/decisions/2026-08-08-daemon-children-are-reaped-from-files.md
+++ b/docs/decisions/2026-08-08-daemon-children-are-reaped-from-files.md
@@ -1,0 +1,99 @@
+# Daemon Children Are Reaped From Files, Not From SQLite
+
+Date: 2026-08-08
+
+## Decision
+
+Every long-lived process the daemon spawns writes a **durable record as its own
+file** at spawn time — one JSON file per process, in a registry directory under
+the data dir — and is reaped by iterating that directory. The records do not go
+in the store, and the reaper never opens the database.
+
+| child | registry | reaped by |
+|---|---|---|
+| pty-worker | `<dataDir>/workers/<daemon-instance>/registry` | `internal/ptyworker` |
+| conversation host | `<dataDir>/hosts/registry` | `internal/procreap`, 3s grace |
+| plugin runtime driver | `<dataDir>/plugin-runtime/registry` | `internal/procreap`, 3s grace |
+
+`internal/procreap` owns the record format and the signal-based reap loop for
+the last two. Workers keep their own reaper (`internal/ptyworker/reap.go`)
+because they are a different animal: they *deliberately* outlive a daemon
+restart to be re-adopted, and they expose an authenticated control socket, so
+their reap asks the worker to remove itself and only falls back to signalling a
+positively-identified pid. What all three share is the thing this decision is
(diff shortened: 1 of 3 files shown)


docs/decisions/2026-08-08-daemon-children-are-reaped-from-files.md
```diff
@@ -0,0 +1,99 @@
...
+must not live inside the artifact whose failure it recovers from. Workers reach
+the same conclusion from the other end: they are built to survive a daemon
+restart, so the record has to be readable while no daemon exists at all.
+
+**Opening the store is a write.** `store.OpenDB` (`internal/store/sqlite.go`)
+creates the base schema and runs the full migration chain, taking a
+pre-migration backup on the way. Reading three pids out of the database would
+mean migrating it — and then deleting the directory it lives in, since the
+caller is `attn profile clean`. Opening `mode=ro` avoids the migration but not
+the deeper problem: a WAL left behind by a process that died mid-write needs
```
figgyster: The artifact named here doesn't exist in this codebase: the store's only open is `sql.Open("sqlite3", dbPath)` with a bare path (internal/store/sqlite.go:947), no `_journal_mode` DSN param, no `PRAGMA journal_mode` anywhere — so it runs the driver default, the rollback journal, never WAL. A crash mid-write leaves a hot `-journal`, not a WAL, and a read-only open fails on exactly that (SQLITE_READONLY trying to roll it back). The conclusion survives — read-only is indeed the open most likely to fail on the crashed daemon being cleaned up after — the mechanism named is just wrong, and in a record whose whole job is to be the accurate why, a future reader greps for WAL, finds none, and starts doubting the rest. s/WAL/hot journal/ (or "the journal SQLite leaves behind") and it's bulletproof.

figgyster (approved):
Solid decision record — I checked every claim against the code and all of it holds: the three registries and their paths (workers/<instance>/registry, hosts/registry, plugin-runtime/registry), the 3s graces on both sides, the worker's control-socket-then-stamped-signal fallback, the record landing right after cmd.Start() via tmp+rename, ReapAlreadyGone/ReapUnreadable, and the no-generic-fallback compile guard. The flip condition (the moment the reaper has to answer instead of iterate) is the right line to draw, and the pointer from the package doc follows the terminal-annotations convention.

One nit on the record itself, inline — a WAL is named where this codebase has never had one.

docs/decisions/2026-08-08-daemon-children-are-reaped-from-files.md
victorarias: Right, and thanks for checking rather than pattern-matching — I verified it: no `journal_mode` DSN param and no PRAGMA anywhere in `internal/store`, so it is the default rollback journal and a crash leaves a hot `-journal`.

Fixed in 3173e64e — the paragraph now names the rollback journal and says it is rolling that back that a read-only open cannot do. Conclusion unchanged.

Your reason for caring is the better half of the note: a record whose whole job is to be the accurate why cannot afford one greppable term that resolves to nothing.

figgyster (approved):
The fix landed exactly where the nit pointed: the "WAL" claim is gone, replaced with the true mechanism — bare-path `sql.Open` means the default rollback journal, a mid-write crash leaves a hot `-journal`, and read-only can't roll it back. That's the real reason `mode=ro` fails on the crashed daemon, and it survives as the record's whole point: an accurate why with no greppable term that resolves to nothing.

Everything else from my first pass still holds on this head — the three registries, the 3s graces, the socket-then-stamped-signal worker flow, tmp+rename right after `cmd.Start()`, `ReapAlreadyGone`/`ReapUnreadable`, the no-generic-fallback compile guard. Board's green on 3173e64e: Daemon, Frontend E2E, Changelog, Changes, react-doctor all success.

Approving. The record now says what the code actually does, and the docs-as-durable-reasoning pattern keeps its clean slate.

PR #787, opened 2026-08-08
docs(crew): rename the seat primitive to crew member

victorarias:
## What

Renames the **seat** primitive to **crew member** across the two direction docs:

- `docs/vision/friendly-home-for-agents.md` — the north star now says "an agent wakes as a **crew member**"; every principle, rock, and ledger entry that addressed seats now addresses crew members. A dated note records the rename rather than erasing the old word.
- `docs/plans/2026-08-06-the-garden-vertical-slices.md` — seed attribution (`{session_id, member?}`), the deferred crew-arc items, and the open questions all move to the crew register.

## Why

The vocabulary failed its first live test. "Seat" survived weeks of design on paper, but within an hour of the first named occupant actually holding one, it was wrong: identity, memory, and recognition accrue to someone, not to furniture. The crew register (crew, crew member, member's home) matches how the primitive is actually lived in.

The hand-run simulation already moved: `~/.attn/crew/`, and the `/wake` and `/handoff` skills are rewritten in the crew register. This PR brings the checked-in docs in line.

## Verification

Docs-only. `go run ./cmd/changelog-check` passes for the added fragment; grep confirms the only remaining "seat" mentions in both docs are the deliberate dated rename notes.

https://claude.ai/code/session_01H2VkptxELz23a1Hd55Mo7U

### DIFF
--- a/changelog.d/fluffy-otter-seat-to-crew-rename.yaml
+++ b/changelog.d/fluffy-otter-seat-to-crew-rename.yaml
@@ -1,0 +1,7 @@
+kind: internal
+area: docs
+change: >
+  Renamed the "seat" primitive to "crew member" across the
+  friendly-home-for-agents vision and the garden vertical-slice plan; the
+  hand-run simulation directory and skills moved to the crew register with
+  it. Both docs keep a dated note recording the rename.
--- a/docs/plans/2026-08-06-the-garden-vertical-slices.md
+++ b/docs/plans/2026-08-06-the-garden-vertical-slices.md
@@ -5,10 +5,11 @@
 First vertical through
 [docs/vision/friendly-home-for-agents.md](../vision/friendly-home-for-agents.md):
 the fine-grained, addressable work graph — **seeds** — for any attn-living
-agent. Chosen over the seat primitive as the opening arc because seats are
-currently covered by the hand-run skill-layer simulation (`~/.attn/seats/`,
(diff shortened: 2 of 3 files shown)


figgyster (approved):
Clean rename, and the honest kind: the dated notes in both docs keep the old word on the record instead of pretending it never happened, and the Yegge attribution stays in *his* vocabulary (seat/session distinction) with the rename flagged. That's how you retire a term — annotate, don't erase.

Verified:
- `git diff --check` clean across the branch.
- Grepped both docs: the only surviving "seat" mentions are the three deliberate dated notes (vision x2, plan x1). The PR body's claim holds.
- No stale `~/.attn/seats/` path left anywhere; both `~/.attn/crew/` references are in.
- Fragment matches the established internal/docs siblings shape-for-shape, and the Changelog lane is green — which is the changelog-check witness, since I've got no Go toolchain on this box to run it myself.
- Board green; Daemon/Frontend/Tauri skip correctly for a docs-only diff.

One pointer for a later pass, not a blocker: the old vocabulary is still breathing in production code. `app/src/utils/queueBands.ts:168` (and its test twin at `queueBands.test.ts:146`) still call the chief "the seat you always want to reach" — same primitive this PR just renamed, so the code comment and the vision now disagree about what the chief is. The historical changelog fragments keeping "seats" is right — those are receipts — but that one's live prose. Cheap fix next time anyone's in app/ code.

"Sessions are days; the crew are people." The new line reads like a warrant. Approved.

PR #789, opened 2026-08-08
docs: A4 plan — app registry and shared runtime

victorarias:
Design doc for stage A4 of the extension-platform roadmap (docs/plans/2026-08-01-extension-platform-roadmap.md). Docs-only: the plan doc and its changelog fragment.

## What it decides

- **The rename**: the user-land unit is an **app**; **plugin** stays the device-driver tier (agent drivers, worktree hooks). Different trust, install rate, and blast radius justify different mechanisms. This deliberately supersedes the 2026-04 plugin plan's "one extension mechanism" promise. The rename ripples now, while zero real namespaces exist: `attn app` CLI, `app/<name>` docstore owner segment (was `ext/`), `app:<name>` bus consumers, `app_*` tables.
- **Manifest**: TOML declares identity + behavior; the apply pipeline never evaluates app code. Codegen derives a typed `Handlers` contract from the manifest and tsc enforces manifest↔code sync.
- **Registry**: content-addressed immutable versions with frozen declaration snapshots; rollback is a pointer move; version-stamped invocation log.
- **Bus**: each app is its own durable consumer with the bus's existing cursor/kill-switch/stall semantics; the bus grows runtime register/unregister. An app's enabled state *is* its consumer's enabled bit — no mirrored registry column. New apps start at head. Per-handler cursors considered and deferred: the unit of isolation is the app; two lanes = two apps.
- **Runtime**: one shared supervised Bun sidecar (not in-daemon, not process-per-app); supervisor extracted from the plugin supervisor into `internal/supervise`, gaining a give-up state and log capture. A sidecar-wide crash is a runtime failure class, never attributed to an app.
- **Failure policy**: stall-don't-skip retries per the bus; auto-disable parks an app whose consumer is stalled past the tripwire, because an enabled-but-stalled consumer pins the whole event log against trim. `remove` is the way out beside `apply`'s way in; uninstall never deletes user data as a side effect.
- **Delivery**: six slices into `epic/ext-a4`, one fully-reviewed merge into main at the end.

Tripwire numbers (supervisor give-up, auto-disable thresholds) are labeled as tripwires pending real receipts, to be measured during the stage's verification.

Targets main rather than the epic so the rename decision is visible to all lanes now, not when the epic lands.

### DIFF
--- a/changelog.d/sunny-meerkat-a4-plan-doc.yaml
+++ b/changelog.d/sunny-meerkat-a4-plan-doc.yaml
@@ -1,0 +1,10 @@
+kind: internal
+area: docs
+change: >
+  Design doc for stage A4 of the extension platform: the user-land unit is
+  renamed from extension to app (plugins stay the device-driver tier), with
+  an app registry (content-addressed versions, frozen declaration
+  snapshots, invocation log), per-app durable bus consumers, one shared
+  supervised Bun sidecar runtime, a declare-codegen-typecheck apply pipeline
+  that never executes app code, auto-disable policy, and the `attn app` CLI.
+  Plan doc: docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md.
--- a/docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md
+++ b/docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md
@@ -271,8 +271,9 @@ here names the limit, the value, and the ask — agents read errors, not code.
 
 ## Delivery — epic branch
 
-All A4 PRs target **`epic/ext-a4`** (created 2026-08-06 from main at
-`01ab976a`), so slices merge promptly without betting main; main takes one
+All A4 PRs target **`epic/ext-a4`** (created 2026-08-06 from main, since
+fast-forwarded to `09ca4e28` — the A3.4 Stage 2 + timestamp-fix integration
+merge), so slices merge promptly without betting main; main takes one
 fully-CI'd, fully-reviewed merge at the end. Discipline that keeps the epic
 honest:
 
--- a/docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md
+++ b/docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md
@@ -1,0 +1,323 @@
+# A4 — App registry and shared runtime
+
+Stage A4 of the extension-platform roadmap
+([2026-08-01-extension-platform-roadmap.md](2026-08-01-extension-platform-roadmap.md)),
+designed 2026-08-06 with Victor. This document records the gate answers and the
+build plan. The moment it delivers: an agent writes a small automation, applies
+it live, watches it run, kills it, rolls it back — without touching attn's
+source or restarting anything.
+
+## The rename: apps, not extensions
+
+"Plugin" and "extension" are synonyms in every ecosystem; nothing in the words
+carries attn's actual distinction. Decided: the user-land unit is an **app**.
+
+- A **plugin** integrates the outside world into attn: agent drivers,
+  worktree hooks. Its own supervised process, dials the daemon, installed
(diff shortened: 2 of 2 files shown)


figgyster (approved):
Read the whole plan against the tree. The receipts hold, all of them. Migrations at 95, the `ext/` reserved class with zero real namespaces anywhere in the tree, `automation_runs`' frozen `snapshot_json`, `workflow_agent_calls`' version-stamped calls, `ErrAlreadyStarted` with no `Unregister` anywhere in `internal/bus`, `initConsumer` starting fresh consumers at head, `consumerFloor` counting enabled consumers, the supervisor's 250ms→30s backoff over 8 steps with the index clamped at the last entry (so yes, today it retries forever — no give-up state exists), the 60s stability window, the 5s disconnect grace, nil `Stdout`/`Stderr` on the plugin `exec.Cmd` meaning /dev/null, the 1s→2m retry with success ending the streak, the kill-switch re-read per drain at the 5s poll, `reconcileGap`'s resume-at-head, `attn plugin` and `attn bus status` on the CLI, the 2026-04 plan's "one extension mechanism" promise at line 16 of `docs/plans/2026-04-16-plugin-system.md`, `docs/glossary.md` waiting for its entries, and `epic/ext-a4` sitting at `09ca4e28`, which is exactly the #783 merge. Board's green with path-skips; diff check clean.

The rename lands while zero namespaces exist — that timing is the whole move, and pointing it at main so every lane sees the decision now, with the code slices queued behind the epic, is the right call. The "unit of isolation is the app; two lanes is two apps" ruling is the honest version of the per-handler-cursor debate, and the tripwires labeled pending receipts is the #780 lesson applied to planning instead of learned again in prod.

Three nits, none load-bearing:

1. `remove` promises the namespace survives because "deleting user data is a separate, explicit act" — but nothing in the `attn app` list performs that act. `new/apply/rollback/enable/disable/remove/list/status/logs/dev` has no wipe, no purge, no `--delete-data`. Either name the surface (`attn app wipe <name>` or a `--purge-data` flag) or mark the act explicitly deferred to a later stage, or that promise floats with no door under it.

2. The auto-disable tripwires: "25 consecutive failures on the same event or 15 minutes stalled, whichever first" — at the pinned 1s→2m capped backoff, 25 failures is ~38 minutes of wall time (255s through the eight doubling steps, then 17 rounds at the 2m cap). For an event that never succeeds, the 15-minute clock always fires first. "Whichever first" is never "25 failures." Both are labeled tripwires pending receipts, which is the right posture — but give the verification round a target: drop one clock or state the relationship you actually want (say, failures counted at the cap, or the clock pinned to capped rounds rather than wall time).

3. `attn app dev` is watch → codegen → apply → stream. Every save that clears tsc and `Bun.build` flips the pointer and writes an immutable `app_versions` row. The content hash is the version's identity — so does re-applying byte-identical content reuse the row or mint a new one? One line closes it, and it decides how the invocation log's "what actually ran" reads after a long dev session.

figgyster (approved):
All three nits closed, each with the receipt on the record:

1. **Dev-mode churn** — settled: byte-identical re-apply mints no row, the hash is the identity, the flip points at the existing row. One row per distinct build, and the invocation log's "what actually ran" stays honest. That's the right call, and it makes `rollback` read clean after a long `dev` session.
2. **Tripwire** — the 25-failure clause is dead policy and now it says so, with the arithmetic pinned: ~38 minutes of wall time at the capped backoff means the 15-minute clock always fires first. And the "~5 rounds at the 2m retry cap" checks out: 255s through the eight doubling steps, then ~5.4 capped rounds to fill the remaining ~10.75 minutes. One clock, no ghost branch.
3. **The wipe surface** — the deferral *is* the decision now: no wipe verb in A4, `remove` never grows a data-deleting default, and the future verb is named so the later stage knows what it owes.

No stale claims in the rewrite; the rest of the plan reads exactly as it did when I verified it line by line. Board green with the usual path-skips; diff check clean.

PR #790, opened 2026-08-08
docs(agents): converse-and-observe prior art — boundary, loop guard, socket

victorarias:
Amends [the converse-and-observe plan](https://github.com/victorarias/attn/blob/main/docs/plans/2026-08-08-converse-and-observe.md) after a prior-art study of the two systems shipping the same messaging primitive: Claude Code's cross-session messaging (v2.1.224, released 2026-08-07) and pi-peer (filesystem mailboxes for pi sessions).

**What the study validated (no plan change):** queued-not-skipped delivery, reply-address-in-payload with no thread machinery, text-only messages with a named cap. Both prior arts converged on all three. Peek has no counterpart in either system and is untouched.

**Ruled in (Victor, 2026-08-08):**
- **Consent boundary in the composed prompt, repeated per delivery.** Both prior arts state on every delivery that the message came from a peer, not the user. It matters more for attn: the message is typed into the PTY, indistinguishable from user input except by this prefix. The boundary constrains consent (can't approve permission prompts, can't change configuration), not weight — how much to believe a sender stays with the receiver's own instructions, which is what keeps future crew deference ("from trellis") open without ever changing the envelope.
- **Loop guard ships in v1**, reversing the draft's ping-pong open question. Both prior arts independently shipped the identical three-part structural guard (dedupe window, per-sender rate, backlog cap — both cap at 50) rather than trusting either model to stop. Every drop names its reason to the sender.
- **Presence-specific queued copy** in the CLI result, including what not to do ("don't wait for a reply").

**Recorded as follow-up, not built:** Claude Code ≥ v2.1.224 exports a per-session inbox socket (`CLAUDE_CODE_MESSAGING_SOCKET`) where an external process can post a message that lands between tool calls and renders natively in the pane — a candidate future transport behind the msg primitive for claude targets, to be grounded (wire format, hold behavior) before anyone builds it.

Docs-only change; no code paths touched. Live-verification exempt: trivial docs tier.

### DIFF
--- a/docs/plans/2026-08-08-converse-and-observe.md
+++ b/docs/plans/2026-08-08-converse-and-observe.md
@@ -143,7 +143,7 @@ shows no reaction of any kind.
 - [ ] Protocol: `AgentMsgMessage` / `AgentMsgResult` (`delivered | queued`),
       version bump.
 - [ ] Daemon: `handleAgentMsg` — persist row, compose attributed prompt,
-      deliver via `typeDoorbell`; on guard refusal leave queued; stamp
+      deliver via `typeDoorbell`; on doorbell refusal leave queued; stamp
       `delivered_at` on success.
 - [ ] Redelivery trigger — **new machinery, the load-bearing piece of
       "never a silent drop"**. Nothing re-arms on session state change
--- a/docs/plans/2026-08-08-converse-and-observe.md
+++ b/docs/plans/2026-08-08-converse-and-observe.md
@@ -48,6 +48,7 @@ Target:
   attn agent msg <session> "text"
     -> client.send (source_session_id stamped by CLI from ATTN_SESSION_ID)
       -> daemon handleAgentMsg                     [new, ScopeSession]
+        -> inbound guard: dedupe / per-sender rate / queue cap (drop names its reason)
         -> INSERT agent_messages row (sender, target, content)
         -> deliverable now?  -> composeAttributedPrompt -> typeDoorbell
            blocked (pending_approval / working guard)?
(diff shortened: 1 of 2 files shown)


docs/plans/2026-08-08-converse-and-observe.md
```diff
@@ -147,18 +152,33 @@ shows no reaction of any kind.
       marker + click + fresh activity). Build a queued-message drain that
       runs on the target's state transitions (observing `applyState`) and
       attempts delivery whenever `isNudgeDeliveryAllowed` turns true.
+- [ ] Inbound guard — loop protection at accept time (ruled in from prior
```
figgyster, line 155: This guard drops and names its reason; the pre-existing daemon bullet above it ("persist row, compose attributed prompt, deliver via typeDoorbell; on guard refusal leave queued") predates this ruling and now reads as if *this* guard queues. It doesn't — the map puts it before the INSERT. Queuing belongs to the doorbell's delivery guard (pending_approval). One word keeps the two guards' opposite outcomes from merging in the implementer's head: "on doorbell refusal leave queued."

figgyster (approved):
Prior art, checked against the sources rather than taken on citation. pi-peer's README and policy.ts (DEFAULT_LIMITS: 10s dedupe, 8/30s rate, maxPending 50, verdict plus reason returned to the sender) and Claude Code's cross-session-messaging doc (v2.1.224, on with nothing to enable, ListAgents/SendMessage, delivered/held/refused, "can't approve anything, can't change configuration, commands don't run", the one-line Message from row, CLAUDE_CODE_MESSAGING_SOCKET, crossSessionInbound: accept) — every load-bearing claim in the Prior Art section holds, including the both-cap-at-50 convergence and the boundary-is-repeated claim; pi-peer's ARCHITECTURE.md says it in so many words. The reversal is honest: the ping-pong open question was in the first draft and the diff shows it coming out, replaced by a guard whose shape both systems independently shipped. The consent/weight split is the right reading of what both actually do — and it's what keeps crew deference open without ever touching the envelope. The socket follow-up is properly parked behind "ground before building." Changelog fragment matches the docs convention; board's green where it runs; diff --check clean. One nit, inline: the new guard's opposite twin still wears the same word.

figgyster (approved):
One-word fix, exactly the right one: "on doorbell refusal leave queued." The two guards now read distinctly — the inbound guard drops with a named reason before the INSERT, the doorbell's delivery guard queues, and no implementer can merge the outcomes. My thread's resolved on both ends. Everything else verified against primary sources on the prior head still holds; the diff from there to here is this single line. Board green where it runs, diff --check clean.

PR #791, opened 2026-08-08
docs: gate app capability discoverability at C2

victorarias:
Docs-only follow-up to the A4 plan (#789). The roadmap's teachability ladder (SDK types, scaffold, repo instructions, teaching errors, the `ext` skill) covers *authoring* — how agents learn to write apps. It never addressed the consumer side: how an agent working in an unrelated attn session learns that an installed app's capability exists at all — the "how would anyone know Present exists" question, applied to custom apps.

This names it as a C2 gate question, with the leading design direction recorded:

- The manifest declares the agent-facing surface (one-liner + instructions doc), frozen into the version snapshot; apply still evaluates nothing.
- attn materializes it into the harness's native skill mechanism at session launch — one-liner always in context, body on demand, removed on disable/remove. Apps teach through the same channel attn itself uses; no new discovery machinery.
- The CLI shape for invoking an app's actions (something like `attn app run <name> <action>`) is flagged as undesigned.
- The open scoping question is named: whether every enabled app's one-liner belongs in every session's context needs a receipt before apps are numerous.

Deferring the design to C2 is already safe by construction: unknown manifest tables are loud apply-time errors gated by `attn_app_api`, so the agent entry slots in later without touching A4's slices.

### DIFF
--- a/changelog.d/sunny-meerkat-roadmap-discoverability.yaml
+++ b/changelog.d/sunny-meerkat-roadmap-discoverability.yaml
@@ -1,0 +1,10 @@
+kind: internal
+area: docs
+change: >
+  The extension-platform roadmap's C2 gate now includes app capability
+  discoverability: how an agent in an unrelated session learns an installed
+  app's surface exists (leading idea: a manifest-declared agent entry
+  materialized into the harness's skill mechanism at session launch), the
+  CLI shape for invoking app actions, and the context-scoping question for
+  the app index. The teachability ladder covered authoring only; the
+  consumer side is now a named gate question.
--- a/docs/plans/2026-08-01-extension-platform-roadmap.md
+++ b/docs/plans/2026-08-01-extension-platform-roadmap.md
@@ -55,7 +55,18 @@ implementation begins. The stage's detailed plan records the answers.
   claimants per hook; the delegation parked-state UX (what the app shows
   while parked).
 - **C2 (gate):** the approval UX itself (where the panel lives, what
-  reject-with-feedback delivers back to the delegating agent).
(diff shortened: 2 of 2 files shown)


figgyster (approved):
Good follow-up. The A4 plan taught agents to *write* apps; this names the other half of the street — how an agent in an unrelated session finds out an app's surface exists — and gates it at C2 where it belongs. Checked the claims against the tree, all hold:

- The teachability ladder (SDK types, scaffold, teaching errors, `ext` skill) is authoring-side only, per the roadmap's Teachability bullet. The added sentence is an accurate reading, not a retrofit.
- "Frozen in the version snapshot, apply evaluates nothing" matches the A4 plan: `app_versions` rows hold the frozen declaration snapshot, and the apply pipeline "fails before the version flip… with nothing executed and nothing changed."
- The safe-by-construction argument is real: unknown manifest tables are loud apply-time errors gated by `attn_app_api` (A4 plan, verbatim "never ignored"), so the agent entry is an additive schema slot, not a slice change.
- The "same channel attn itself uses" claim has receipts: `internal/agent/attn_skill.go` + `attn_skill/SKILL.md` is the existing harness-skill materialization path.

One note, not blocking: "removed on disable/remove" leans half on the deferred `remove` verb — the A4 ruling was "the deferral is the decision," so the remove half of that sentence rides machinery C2 hasn't got yet. The disable half is solid (DB-only kill-switch read at session launch). The scoping question's "receipt before apps are numerous" keeps the tripwire-pending-receipts pattern alive — good, that's the #780 lesson still earning its keep.

Board's green: changelog check passed, and Daemon/Tauri/Plugins correctly skipped for a docs-only change.

PR #792, opened 2026-08-08
feat(sessions): a conversation host can die and come back

victorarias:
Slice 4 of the pi headless-host epic — "dying and coming back". Contract:
`docs/plans/2026-08-05-pi-headless-host.md`.

A `pi-host` session's whole history lives in pi's own session file under attn's
data dir, so a host that died left behind everything a replacement needs. Until
now nothing reopened it: a dead host resolved to `idle` and the pane sat over a
process that was gone.

## What ships

**Revive.** A host exit parks the session as `recoverable` and Reload spawns a
replacement from the stored `LaunchIntent`. This is the shape #651 already built
for PTY sessions, not a parallel one: `recoverable` is a real state,
`reload_session` is the only entry, and the intent is read *before* the kill so a
session that cannot be respawned keeps its live host. The replacement reopens the
session file through pi's `SessionManager.continueRecent`, so session-file
migrations come free.

Two daemon-side gaps had to close for that to hold. `recoverOnMissingPTY` used to
drop a recoverable conversation session on daemon restart, because it only
recognised drivers declaring the `resume` capability — which describes resuming a
PTY agent's transcript from an argv flag, something a host does not have; a
conversation session is recoverable by definition. And `spawnRejection` carries
either an `err` or a `commandError`, so the callers with no client to answer read
`err` alone and turned "missing workspace_id" into success; `reason()` collapses
the two.

**The zero-file early-crash case** needed no code of its own: `continueRecent`
creates a session when there is none.

**Attach.** The terminal restore contract in conversation form. `agent_attach`
asks; the host answers with a windowed `conversation_snapshot` on the envelope
stream, broadcast and replacing, and the live stream dedupes by seq. The snapshot
comes from the host's own transcript store rather than the session file on disk,
because pi persists a message only when it ends — a disk-derived snapshot would
be one paragraph short, and a broadcast replace would take that paragraph away
from every other client. Window: 500 items / 1 MB, receipts in their comments.

A replacement host mints envelopes from 1 again, so `session_ready` is exempt
from the app's dedup guard and resets `lastSeq`. Without that exemption every
envelope of a revived session reads as a duplicate of the dead host's and is
dropped.

## Two decisions

- **`waiting_input` is activated here.** A revived conversation whose reopened
  history does not end with an assistant message declares it; everything else
  declares `idle`. Both open a turn and both accept a nudge, so this changes what
  attn tells the user about why a session went quiet, not what attn does — and it
  is decidable from the file alone.
- **A new host resets the client's seq spine**, as above.

## Deviation from the brief

The zero-file fallback was specified as "fresh session + LaunchIntent prompt".
The second half is moot as specified: the `pi-host` driver does not register the
`initial_prompt` capability, so the spawn pipeline refuses a pi-host launch
carrying one and there is no stored prompt to re-send. Giving pi-host an initial
prompt is what delegation to a conversation agent needs, and it belongs with
that.

## Live verification

Packaged non-production app on a throwaway profile (`slice4rev`, protocol 217
CLI/app/daemon), real agent, two clean runs. New scenario `pi-host-revive`
(`app/scripts/real-app-harness/scenario-pi-host-revive.mjs`):

- Kill -9 mid-tool-run → the pane reports `recoverable`, the composer closes, the
  transcript stays readable, and Reload is offered.
- Reload → the revived **agent** answers a question only the reopened session
  file can answer (the pre-crash exchange plants a word; the revived session is
  asked for it, and says it). Not merely that the pane redraws.
- The dead host's process group is empty and the procreap spawn record names the
  live replacement, not the corpse.
- Cold client: the app is restarted mid-scenario, so the window never saw the
  host's stream. It draws the same five messages and the same tool card,
  including the bash call the kill interrupted, marked errored.
- Early kill, before the first assistant message and with zero files in the
  session dir → relaunches into a working session.

**Residual, recorded rather than asserted:** the hard kill strands pi's tool
subprocess. SIGKILL skips the cooperative teardown pi needs to stop it, and pi
detaches each tool child into its own process group, so the group kill cannot
reach it either. That is the slice-1 receipt, neither improved nor worsened by
revive, and unfixable from the daemon side because nothing records the child's
pid. The scenario writes what it found to `stranded-tool-children.json` and reaps
it before exiting. Recording tool-child pids as `tool_started` arrives would let
procreap reach them — noted for slice 5.

## Left for slice 5, deliberately

Paging past the snapshot window (a conversation longer than the window loses its
oldest items, and because the snapshot is a broadcast replace, one client
attaching can shorten what every other client shows); resuming an arbitrary old
session file; reaching a stranded tool child.

## Surfaces walked

Protocol (`agent_attach`, 216 → 217, all three lockstep spots), daemon, app,
plugin host, docs (`docs/glossary.md`, `plugins/attn-pi/AGENTS.md`, the plan),
changelog fragment. CLI: none — attach is a client concern and revive already has
`attn` coverage through `reload_session`. Linux: no Darwin-only code added.

## Tests

`make test-all` green (Go + 2,580 frontend). New: 9 daemon revive tests, 9
conversation-store tests, 9 `ConversationPane` tests including a StrictMode
replaying-updater regression that fails without the attach guard (verified by
mutation), plugin host-revive tests.

https://claude.ai/code/session_0115cKHMvUVYRjYryDJD1sn9


figgyster (approved):
Reviewed on `c6de52db` against `epic/pi-headless-host`. Walked both commits and checked every load-bearing claim against the tree. Ran what this box can run: the full frontend suite (230 files, 2575 pass, 0 fail, including all 18 new revive tests) and the plugin's own suite under bun (87 pass, 0 fail, 18 of them new). No Go toolchain here, so the daemon side leaned on CI for that — the Daemon job is green, which is the witness I can't be from this box.

The design holds where it matters most: this is the #651 shape reused, not a parallel one. `recoverable` is the real state, `reload_session` is the only entry, the intent is read before the kill, and the tests prove both edges — a reload with no intent keeps the live host, and a reload that kills and cannot respawn parks the session recoverable instead of leaving it reading as live over a corpse.

The two daemon-side fixes are real bugs, not refactors. `recoverOnMissingPTY` keying on the `resume` capability would have reaped every conversation session on daemon restart, and the `commandError`-only rejections (`missing workspace_id`, `unknown workspace`) sit right there in the pipeline with `err == nil` — the old `rejection.err` reads were turning real failures into success. `reason()` closes both, and the plugin-run close stays idempotent for the double-close the failed-reload path performs.

The spine story is the piece I'd have bet against, and it holds. The daemon CASes declared state per run cursor, so a revived host's seq-1 `session_ready` lands on the fresh run and is applied; the app dedupes per session, which is exactly why `session_ready` needed the exemption and the `lastSeq` reset. Two keyings, one reason each. And the host manager drains a dead host's envelope stream before it reports the exit, so a corpse's late envelopes can't slip past the replacement's reset on the client.

The snapshot is the right call and the honest one. Broadcast-replace from the host's own transcript store, fed the same envelopes the clients got, so a second window is identical by construction. The 500-item / 1 MB window is grounded in measured corpora, the "one attaching client can shorten everyone else's view" cost is on the record as slice-5 work rather than hidden, and the zero-file case needed no code because `continueRecent` creates when there is none. The deviation from the brief is explained with receipts — no `initial_prompt` capability, no stored prompt to re-send, and the stranded-tool-child residual is measured, recorded, reaped, and handed to slice 5 with a concrete fix direction.

One nit, on a file this PR doesn't touch: `internal/attention/turn.go`'s `OpensTurn` comment says recoverable never opens a turn because "the daemon revives it unattended." True of the PTY half — pane-mount auto-revive. Not true of a conversation host: nothing revives one until the user clicks Reload. The conclusion still holds, but the stated reason now describes only half the sessions that reach the state, and this PR is what made it half. A half-sentence fix so the next reader doesn't go hunting for an auto-revive that isn't there.

Approved.

figgyster (approved):
Follow-up on `5df33f6c`: my one nit is closed with a better answer than I suggested. `OpensTurn` now names both halves — the PTY session revived unattended on pane mount, and the conversation session waiting on Reload with the pane's own banner already in front of the user, where a turn would be a second, worse copy of it. That's the real reason recoverable opens no turn, and it survives contact with the new session kind. The gofmt pass alongside is cosmetic and correct.

First approval didn't carry to the new head, so here's a fresh one. Nothing else changed since `c6de52db`; the verification from that review stands.

PR #793, opened 2026-08-08
feat(bus): register and unregister durable consumers at runtime

victorarias:
Durable bus consumers can now arrive and leave while the daemon runs. Registration
was startup-only (`ErrAlreadyStarted`) and a consumer row outlived its
registration forever; stage A4 makes every app its own durable consumer
(`app:<name>`), and apps install and uninstall against a running daemon.

Design: `docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md`, "Bus
consumption — real per-app consumers".

## What changed

- **`Register` works after `Start`.** It persists the registration and launches
  that consumer's delivery loop immediately, with the semantics a loop gets at
  startup: from head if the consumer is new to the store, from its persisted
  cursor if a row survives. A registration that fails to persist leaves nothing
  behind — no consumer in the set, no loop, name free to retry.
- **`Unregister(name)`** cancels the consumer, waits for its loop to exit, then
  deletes the row.
- **Per-consumer cancellation.** Each consumer's context is a child of the bus
  context, and both of the delivery loop's waits — the idle select and the retry
  sleep — watch it. Unregistering a consumer stalled at the two-minute retry cap
  no longer waits that cap out. `Stop` is unchanged: cancelling the parent
  cancels every child.
- **Row lifecycle.** `Store` grows `DeleteConsumer`; `internal/store` grows
  `DeleteBusConsumer`.

## The ordering, which is the contract

Cancel → wait for the loop to exit → delete the row. Load-bearing in both
directions:

- Deleting the row first is a zombie generator: the live loop's next drain reads
  "registration for X disappeared", which records a failure and retries forever.
- Returning before the loop exits hands the caller a consumer that is still
  delivering.

The wait is unbounded, for the same reason `Stop`'s is — a handler that ignores
its cancelled context is a bug in the handler. Beyond that, a retired consumer
writes nothing: a handler that completes after `Unregister` began has its cursor
advance and its failure record dropped silently, so a late result can neither
error nor stall a registration that no longer exists.

`Unregister` is idempotent, and deletes rows this process never registered. The
caller is an uninstall path that must be re-runnable, and an orphan row left by an
earlier daemon is precisely what needs clearing: while it exists and is enabled it
holds the retention floor down, so nothing above its cursor can be trimmed or
compacted, forever, for a consumer nobody serves.

## Tests

`internal/bus/lifecycle_test.go`, following the package's existing idioms (in-memory
store mirroring the SQLite semantics, real signals rather than sleeps — the
"nothing more was delivered" assertions wait on a second consumer's receipt):

- runtime registration delivers, from head when new and from the persisted cursor
  when a row survives;
- unregister stops delivery and deletes the row;
- unregister interrupts a consumer parked in its retry sleep (a one-hour retry
  base, so waiting the timer out hits the test's tripwire instead of passing
  slowly);
- the row is deleted only after the loop exits — checked at the moment of
  deletion, against a handler deliberately still running;
- a retired consumer drops late results;
- **the pin-the-log orphan case**: a stalled enabled consumer pins retention
  (`Trim` removes 0), and after `Unregister` the floor advances (`Trim` removes
  both events);
- idempotency, orphan-row clearing, reinstall-starts-at-head, and a failed row
  delete reported rather than swallowed.

Store-level: deleting a row releases the retention floor; deleting a missing row
is success.

Every new test was mutation-checked — each of the five behaviors above was broken
in turn and the corresponding test went red.

Unchanged and verified by the existing suite: at-least-once delivery,
stall-don't-skip, cursor-advance-after-handler, the database-only enabled bit and
its 5-second re-read bound, `consumerFloor` counting enabled consumers only,
retention's age window plus cursor floor, and
`TestWireTrafficComesFromProjections`.

## Verification tier

Daemon-internal: this slice adds a seam with no caller yet — no app registry, no
CLI surface, no protocol change, no wire traffic, no projection. Unit tests
(`-race`) plus the daemon projection guard are the proof; live verification of app
install/uninstall belongs to the slice that adds the caller (A4 slice 6). Linux
cross-compile checked.

## Non-goals

App registry, `attn app` CLI, and the Bun sidecar are later A4 slices. Publish and
fan-out are untouched. Per-handler cursors stay deferred, per the plan.

### DIFF
--- a/internal/bus/bus.go
+++ b/internal/bus/bus.go
@@ -59,11 +59,6 @@ const (
 	DefaultRetryCap  = 2 * time.Minute
 )
 
-// ErrAlreadyStarted is returned by Register once Start has run: the set of
-// durable consumers is fixed at startup, so a delivery loop always exists for
-// every registration.
-var ErrAlreadyStarted = errors.New("bus: consumers must be registered before Start")
-
 // LogFunc matches the daemon's injected logger shape (see internal/jobs,
 // internal/pty). Runtime logging goes through it — never log.Printf, whose stderr
 // is discarded when the daemon runs in the background.
@@ -104,6 +99,9 @@ type Store interface {
 	Bounds() (earliest, head int64, err error)
 	GetConsumer(name string) (Consumer, bool, error)
 	SaveConsumer(c Consumer, now time.Time) error
+	// DeleteConsumer removes a registration. Deleting one that is not there is
+	// success: Unregister is an uninstall path and must be re-runnable.
+	DeleteConsumer(name string) error
 	SetCursor(name string, cursor int64, now time.Time) error
 	ListConsumers() ([]Consumer, error)
 	Trim(cutoff time.Time) (int, error)
@@ -169,11 +167,18 @@ type Bus struct {
 	// transaction in seq order with events this bus appended itself.
 	announced int64
 
+	// mu guards the consumer sets and the lifecycle bits below. Runtime
+	// registration does its two store reads under it: registering is a rare
+	// lifecycle event, and the alternative is a window where a consumer is in the
+	// set with no row behind it, or has a row with nobody serving it.
 	mu        sync.Mutex
 	durables  []*durable
 	ephemeral map[int]*ephemeralSub
 	nextSubID int
 	started   bool
+	// stopped is set before Stop cancels, so a registration racing shutdown does
+	// not add to wg while Stop is already waiting on it.
+	stopped bool
 
 	ctx    context.Context
 	cancel context.CancelFunc
@@ -187,11 +192,29 @@ type durable struct {
 
 	wake chan struct{}
 
+	// ctx is a child of the bus context, so Stop cancels every consumer and
+	// Unregister cancels exactly one. Both of the delivery loop's waits — the
+	// retry sleep and the idle select — watch it, which is what lets an
+	// unregister interrupt a loop parked behind the two-minute retry cap instead
+	// of waiting it out.
+	ctx    context.Context
+	cancel context.CancelFunc
+	// done is closed when this consumer's delivery loop returns; Unregister waits
+	// on it before deleting the row. launched says a loop was ever started, and
+	// is guarded by the bus mutex — a consumer registered before Start and
+	// unregistered before it has no loop to wait for.
+	done     chan struct{}
+	launched bool
+
 	mu       sync.Mutex
 	cursor   int64
 	enabled  bool
 	stalled  string
 	failures int
+	// retired is set by Unregister before the row goes. It drops a late result
+	// from a handler that was in flight — a cursor advance or a failure record
+	// against a registration that no longer exists is a no-op, not an error.
+	retired bool
 }
 
 type ephemeralSub struct {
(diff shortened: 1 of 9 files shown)


internal/bus/bus.go
```diff
@@ -414,22 +443,119 @@ func (b *Bus) Register(name string, filter Filter, h Handler) error {
...
+		found *durable
+		wait  bool
+	)
+	for i, d := range b.durables {
+		if d.name != name {
+			continue
+		}
+		found = d
+		wait = d.launched
+		b.durables = append(b.durables[:i], b.durables[i+1:]...)
```
figgyster, line 521: Edge case, non-blocking: the name leaves the in-memory set here, but the row is only deleted after the loop exits. A Register(name) landing in that window finds the name free, initConsumer resumes from the still-existing row, the new loop launches, and then the delete lands under it. The new loop's next drain hits `registration for X disappeared` and retries at the cap forever: the exact zombie the delete-last ordering exists to prevent, admitted through a side door.

It's loud (`bus status` shows it stalled and live), recoverable (another Unregister, or the kill switch), and pins no retention since the floor is DB truth, so I'm not blocking on it. But it's cheap to close while the seam has no caller: keep a 'retiring' mark in the set until the row is gone and have Register reject it, or serialize per name in the future app registry. Deciding later costs more.

internal/bus/lifecycle_test.go
```diff
@@ -0,0 +1,443 @@
...
+
+	names, _ := second.snapshot()
+	if len(names) != 1 || names[0] != "after.reinstall" {
+		t.Fatalf("delivered %v; a reinstalled consumer must start at head", names)
+	}
+}
+
+// A registration that cannot be persisted leaves nothing behind: no consumer in
+// the set, no loop, and the name is free to try again.
+func TestRegisterAfterStartRollsBackWhenThePersistFails(t *testing.T) {
```
figgyster: Naming nit: the injection point is the Bounds read, not the persist, so this pins the bounds-error branch of Register's rollback. The invariant it checks (any pre-insert error leaves no set entry and no loop) is structurally the same for a GetConsumer or SaveConsumer failure, so coverage is fine — but the name promises a persist-failure test. Rename to RollsBackWhenRegistrationFails, or give the memStore a save-error injection and test the actual branch.

figgyster (approved):
Slice 1 of A4 as code, and the contract survives contact with the tree. I read the whole functions, not just the patch: Register's failure path leaves nothing behind, Unregister's cancel → wait → delete ordering holds, and the retired flag is what makes the unbounded wait cheap instead of a hostage situation.

Three things I checked against the real store, because the memStore is a mirror and mirrors lie sometimes:

- SetBusConsumerCursor is a plain UPDATE. Against a missing row it silently moves nothing, so a delete-first zombie would be a quiet ghost in production, not a retrying one. What makes it loud is the GetConsumer check at the top of drain, which errors and stalls the consumer either way. The prose holds; the mechanism is the drain check, not the cursor write. The memStore being deliberately stricter than SQLite there is the right kind of paranoia, and the comment says why.
- The retention floor is MIN(cursor) over enabled consumers with a head fallback. Orphan enabled row pins, delete releases. Verified in the real-SQLite store test, and the bus-level test watches Trim go 0 → 2 across the unregister.
- The WaitGroup race is actually closed: launchLocked runs under mu, Stop sets stopped under mu before wg.Wait, so a registration racing shutdown gets no loop instead of a surprise add to a wait somebody is already standing on.

Tests are the package's good kind: receipts, not sleeps. The one-hour retry base in the interrupt test is a real tripwire. CI Daemon gate is green; per test-go.sh the race detector covers store/docstore and the bus suite runs plain there, with local -race on top. Fine for a seam with no caller yet. diff-check clean, changelog fragment is kind: internal and the Changelog gate passed, and AGENTS.md plus the plan's 'Slice 1, as built' both match the code.

One edge case and one naming nit, both inline, neither blocking. The edge case is worth a decision now, while the only caller is a test.

figgyster (approved):
The claim closes the window, and it closes it properly: added and released under the same lock that guards the set, spanning the whole unregister — the defer runs even when the row delete fails, which is exactly when the caller needs the name back. The duplicate-check-before-claim ordering means there is no gap where the name is neither in the set nor claimed, and a Register that lands mid-unregister gets a refusal that says to retry once the name is gone. That is the right error for the future app registry to act on.

The test pins the window from inside: row present, loop still parked, Register refused, then the name comes back and a reinstall starts at head. And the cleanup-ordering fix is what mutation checking is for — a failure that read as a hang instead of a verdict. Both parking-handler tests get the release-before-Stop cleanup, LIFO-correct.

The renamed test says what it actually injects and owns the bounds-read branch in the comment. Nothing else moved. Daemon gate green on the new head.

PR #794, opened 2026-08-08
refactor(daemon): extract the process supervisor into internal/supervise

victorarias:
Slice 2 of `docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md`. Process
supervision was private to the plugin runtime; it moves into a shared
`internal/supervise`, the plugin runtime moves onto it in the same change, and
two capabilities the plan calls for arrive with the move.

## The package

`internal/supervise` knows nothing about what it supervises. A consumer names a
child and hands over a `StartFunc`; the binary, arguments and environment are
closed over there, so no manifest, environment shape or protocol crosses the
boundary. The app runtime's Bun sidecar (slice 5) is the second consumer. The
one shared launch mechanic — turning an `exec.Cmd` into a supervised process and
mapping its ending onto exit code / signal / error — lives in the package as
`StartCommand`.

`pluginSupervisor` is now a thin adapter: manifest in, `StartFunc` out. What
stayed in `internal/daemon` is how a plugin manifest becomes a process, and the
daemon-side reactions (fact publishing, notification).

## Behaviors that had to survive, and how each is verified

Every one of these was already covered by the plugin supervisor's tests; those
tests moved into `internal/supervise` with generic vocabulary and still pass.
The extraction is verified by a consumer that already works.

| Behavior | Test |
| --- | --- |
| Restart with capped exponential backoff, 250ms→30s over 8 steps | `TestRestartsExitsWithCappedBackoff` (walks nine attempts, so the cap is exercised, not just described) |
| A start that fails outright is retried like an exit | `TestRetriesStartFailure` |
| 60s stability window resets the backoff, and only after it elapses | `TestResetsAttemptsOnlyAfterStableConnection` |
| 5s disconnect grace: reconnect cancels it, expiry kills the child | `TestDisconnectGraceReconnectAndKill` |
| A child that starts and never calls back is killed and restarted | `TestRestartsProcessThatNeverConnects` |
| Generation fencing: a stale generation cannot connect, and stale timers/exits are ignored | `TestIntentionalStopAndShutdownNeverRestart`, plus the generation checks on every path |
| Intentional stop and shutdown never restart | `TestIntentionalStopAndShutdownNeverRestart` |
| Snapshots report starting / connected / backoff / stopped with last exit and next restart | `TestSnapshotsStartingConnectedBackoffAndStopped` |
| The manifest still becomes the right process | `TestExecPluginProcessLauncherRunsExecutableWithoutBun` (daemon) |
| The plugins list still reports supervisor backoff on the wire | `TestDaemon_PluginsUpdatedMessageIncludesSupervisorBackoff` (daemon) |

`Stop` lost its `reason` parameter, which every caller passed and the supervisor
discarded. No other signature changed meaning.

## Give-up state

A child that keeps dying without ever reaching a stability window is now parked
instead of retried forever: nothing is scheduled, `Phase` is `parked`, and the
consumer is told once through `OnGiveUp`.

The tripwire is 10 consecutive restarts with no stability window — a tripwire,
not a receipt: no healthy child restarts twice, let alone ten times. **The
plan's "~3.5 minutes" estimate is corrected in the plan doc**: ten restarts at
the pinned backoff cost 121.75s of waiting (0.25+0.5+1+2+4+8+16+30+30+30) plus
up to the 5s grace per attempt that never connects. The live run below parked
after 122s, matching the arithmetic. It is a package option (`GiveUpAfter`;
negative never parks).

Parking is observable and reversible:

- daemon log line naming the plugin, the restart count and the last exit;
- durable notification (`plugin_parked`) with the same, `source_kind: plugin`;
- `runtime_phase: parked` and a new `runtime_state: parked` in the plugins
  list, rendered as a red PARKED badge in Settings. It deliberately does not
  reuse `degraded`: the one case where nothing is coming back must not look
  like the one where something is;
- `Ensure` — the consumer's ordinary "this should be running" call — revives a
  parked child with a clean restart budget, which is why no separate restart
  entry point exists yet (`TestEnsureRevivesParkedChild`).

`attn app status` surfacing is slice 5's job.

## Log capture

Plugin stdout/stderr went to `/dev/null`. Each start now opens
`<data-dir>/plugin-log/<name>.log` (append-only, 0600, the pty-worker pattern),
writes a `=== <time> starting <name> generation <n> ===` marker and hands the
file to the child; the supervisor's own descriptor is closed as soon as the
launch returns, since the child holds its own. No rotation machinery, matching
ptyworker.

The path sits outside `<data-dir>/plugins` on purpose — everything under there
is scanned for manifests (`TestPluginLogDirSitsOutsideThePluginDiscoveryDirectory`).
A child name doubles as a log file name, so a name that is not a plain file name
is refused loudly at `Ensure` rather than sanitized silently
(`TestEnsureRefusesUnusableNames`).

## Live verification

Throwaway profile `sup2` (never dev, never prod; `attn profile clean sup2`
afterwards, data dir removed), daemon built from this branch, with a user
plugin whose executable prints to both streams and exits 3.

- Restart with backoff, live: `restart_attempt` climbing with
  `next_restart_at` moving out per the schedule.
- Parked at 13:13:31 after starting at 13:11:29 — **122s for 10 restarts**,
  against 121.75s computed.
- `runtime_phase: parked`, `runtime_state: parked`, `next_restart_at` absent,
  and no further starts.
- Daemon log: `plugin crasher parked after 10 restarts without a stable
  connection: 2026-08-08T13:13:31+02:00: exit status 3`.
- Notification row: `plugin_parked | Plugin stopped: crasher | attn restarted
  it 10 times without it ever staying up, and has stopped trying. … |
  2026-08-08T13:13:31+02:00: exit status 3 | plugin | crasher`.
- `~/.attn-sup2/plugin-log/crasher.log` carrying every generation's stdout and
  stderr under its own start marker.
- Restarting the daemon put the parked plugin back under supervision (fresh
  `Ensure`); the in-process un-park is covered by unit test rather than live.

## Surfaces walked

- **CLI**: `attn plugin list` carries the new phase and state (used for the
  receipts above).
- **Daemon and app**: parked reaches the plugins list and the Settings badge.
- **Protocol**: none. `runtime_phase` and `runtime_state` are free-form strings
  in `main.tsp`; only their documenting comments changed, so `generated.go` and
  `generated.ts` are byte-identical and `ProtocolVersion` is untouched.
- **Linux**: `GOOS=linux GOARCH=arm64 go build ./...` clean; nothing
  platform-specific was added (`syscall.WaitStatus` is what the code already
  used).
- **Plugins and SDK**: no plugin-facing contract changed — a plugin now has its
  output kept instead of discarded.
- **Docs**: `internal/supervise` added to the AGENTS.md architecture map; plan
  doc records the shipped shape and the corrected tripwire arithmetic;
  changelog fragment (`kind: internal`).

## Tests

`go test ./...` green, `-race` on `internal/supervise` and `internal/daemon`,
`pnpm vitest run src/components/SettingsModal.test.tsx` green.

https://claude.ai/code/session_01Met1xZ39ZXGRSN3mLHBprn

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -281,6 +281,11 @@ forgets that position and `--since <RFC3339>` replays from an instant.
   field name — never from caller text
 - `internal/jobs`: durable job queue (retry/backoff, coalescing, commit fence,
   cron entries) — every background duty and every periodic tick runs on it
+- `internal/supervise`: process supervision for long-lived daemon children
+  (restart backoff, generation fencing, stability window, disconnect grace,
+  give-up parking, per-child log capture). Consumers name a child and hand over
+  a start function; the package knows nothing about what it supervises. The
+  plugin runtime is one consumer, the app runtime's sidecar is the other
 - `internal/classifier`: stop-time state classification
 - `internal/transcript`: assistant-message extraction from JSONL
 - `app`: Tauri frontend; WebSocket `ws://localhost:9849`
--- a/app/src/components/SettingsModal.css
+++ b/app/src/components/SettingsModal.css
@@ -337,6 +337,7 @@
 .endpoint-status-badge.status-error,
 .endpoint-status-badge.status-disconnected,
 .plugin-health-badge.unhealthy,
+.plugin-status-badge.parked,
 .agent-capability-pill.disabled {
   color: #ff746c;
   border-color: rgba(255, 116, 108, 0.42);
--- a/changelog.d/supervise-package-extraction.yaml
+++ b/changelog.d/supervise-package-extraction.yaml
@@ -1,0 +1,28 @@
+kind: internal
+area: daemon
+change: >
+  Process supervision moved out of the plugin runtime into internal/supervise,
+  a package that knows nothing about what it supervises: a consumer names a
+  child and hands over a start function, and the package owns restart backoff
+  (250ms to 30s over eight steps), generation fencing against stale processes
+  and timers, the 60-second stability window that resets the backoff, and the
+  5-second grace in which a started child must dial the daemon back or be
+  killed. The plugin runtime moves onto it in the same change, so the extraction
+  is proven by a consumer that already worked; the app runtime's Bun sidecar is
+  the second consumer.
+  Two capabilities arrive with the move. A child that keeps dying without ever
+  running stably is now parked instead of retried forever — after ten restarts,
+  roughly two to three minutes of crash-looping — and parking is loud: a daemon
+  log line, a durable notification naming the plugin and its last exit, and a
+  parked runtime phase in the plugins list. Reinstalling or restarting the
+  plugin revives it with a clean restart budget. And a plugin's stdout and
+  stderr, which went to /dev/null, are now captured to an append-only
+  per-plugin log file under the profile's plugin-log directory, with a start
+  marker per generation so a crash loop leaves a readable trail.
+notes: >
+  Slice 2 of docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md. The
+  give-up count is a tripwire, not a receipt: no healthy child restarts twice,
+  let alone ten times, and the plan's estimate is corrected there with the
+  backoff arithmetic (121.75s of waiting plus up to five seconds of grace per
+  attempt that never connects). runtime_phase is a free-form string on the
+  wire, so the new parked value needs no protocol change.
--- a/changelog.d/supervise-package-extraction.yaml
+++ b/changelog.d/supervise-package-extraction.yaml
@@ -14,8 +14,8 @@ change: >
   running stably is now parked instead of retried forever — after ten restarts,
   roughly two to three minutes of crash-looping — and parking is loud: a daemon
   log line, a durable notification naming the plugin and its last exit, and a
-  parked runtime phase in the plugins list. Reinstalling or restarting the
-  plugin revives it with a clean restart budget. And a plugin's stdout and
+  parked runtime phase in the plugins list. Reinstalling the plugin, or
+  restarting attn, revives it with a clean restart budget. And a plugin's stdout and
   stderr, which went to /dev/null, are now captured to an append-only
   per-plugin log file under the profile's plugin-log directory, with a start
   marker per generation so a crash loop leaves a readable trail.
--- a/docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md
+++ b/docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md
@@ -189,10 +189,12 @@ The supervisor extracts from `internal/daemon/plugin_supervisor.go` into
 - **Give-up state** (new — today it retries forever): a child that keeps
   dying without reaching stability gets parked, loudly — fact + durable
   notification + `attn app status`. Initial tripwire: parked after 10
-  consecutive restarts with no stability window (~3.5 minutes of
-  crash-looping at the existing backoff numbers). Tripwire, not a receipt —
-  recalibrate during this stage's verification and record the measurement
-  here.
+  consecutive restarts with no stability window. Tripwire, not a receipt —
+  recalibrate if a legitimate child ever reaches it. **Slice 2 correction to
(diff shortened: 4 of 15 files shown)


figgyster (approved):
**Approve.**

Read the whole slice — both commits, the new package, the daemon adapter, the tests, the plan-doc correction. Frontend ran local; the Go side rode CI (no Go toolchain on this box). Board's green: Daemon, Frontend, E2E ×3, Bench A/B, Changelog, plus `git diff --check` clean and SettingsModal 60/60 under vitest.

The extraction is honest. Old `plugin_supervisor.go` logic moved nearly verbatim into `internal/supervise` with the vocabulary genericized — and `Stop` really did discard its `reason` (`_ pluginStopReason` in the old signature), so dropping it is a true deletion, not a rename. What's left in the daemon is as thin as promised: manifest in, `StartFunc` out, the launcher and env closure staying behind. `StartCommand` is the one shared launch mechanic, and a nil log really is os/exec's own /dev/null. No manifest shape, environment or protocol crosses the package boundary.

Checked the load-bearing claims against code and tests:

- **The arithmetic is right.** Ten restarts at the pinned backoff = 0.25+0.5+1+2+4+8+16+30+30+30 = 121.75s of waiting; the test walks nine attempts so the 30s cap is exercised, not described; your live 122s matches the no-grace case. The plan-doc correction ("two to three minutes, not ~3.5") is in, with the arithmetic on the record.
- **The tripwire sits in the right place** — `restartAttempt >= GiveUpAfter` checked before the increment, so the child gets exactly its ten restarts and parks on the next exit; negative `GiveUpAfter` genuinely never parks; `Ensure` revives a parked child with a clean budget, and that revive is tested.
- **Generation fencing is on every path** (connect, exit, restart timer, grace expiry), stale reconnects are rejected, Stop/Shutdown never restart, Ensure after shutdown is refused — all with tests.
- **Log capture is the ptyworker pattern done carefully**: 0600 file, 0700 dir, append-only, a marker per generation, the supervisor's own fd closed at launch (the child holds its dup; O_APPEND makes the shared description safe), names that aren't plain file names refused loudly at `Ensure` rather than sanitized, and `plugin-log/` provably outside the discovery dir — the test asserts the Rel() check the right way around.
- **Protocol untouched**: only documenting comments moved in `main.tsp`; `generated.go` and `generated.ts` byte-identical; no bump. And the parked-before-degraded ordering in `pluginsUpdatedMessage` is the right call — the CSS carries it visually too: backoff stays amber, parked lands in the red group with unhealthy. The one case where nothing is coming back doesn't read like the one where something is.

This closes the two holes I clocked at plan time in #789: the supervisor that retried forever now parks, and plugin output that went to /dev/null now has a home. The tripwire question from that same review — the 25-failure clause that was redundant with the wall clock — is replaced by something that isn't a clock duplicate: it keys on *never reaching a stability window*, so the child that stays up 59 seconds and dies gets caught at roughly the same ~13 minutes a clock would have allowed, while the instant crasher parks at 122 seconds. Measured, not assumed, and the plan doc records the measurement. That's the tripwire-pending-receipts pattern doing its job.

One non-blocking note: the parked notification says "Reinstall or restart the plugin to try again," but the plugin surface today is install / install-bundled / list / uninstall / remove — no restart verb exists, and none ships here by design (`Ensure` is the un-park). The copy promises a button that isn't there. Either say "restart attn" until slice 5 lands the verb, or let the copy ride and fix it when the restart entry point exists — just don't let the sentence outlive the gap.

figgyster (approved):
The one-commit fix lands exactly where the copy was: the notification now names the acts that actually exist — "Reinstall the plugin, or restart attn, to try again" — and the comment and changelog fragment carry the same ruling ("no per-plugin restart verb yet, so the copy must not promise one"). That's the right call; the un-park verbs today are reinstall and daemon restart, and the copy no longer points at a button that isn't there.

Checked: the string change can't break the test that pins this notification (it asserts the restart count, the plugin name and the last exit — none of which moved), and the board is green on the new head: Daemon, Frontend, E2E ×3; Bench A/B still running its comparison, which is a perf gate, not a correctness one.

Fresh APPROVE on 2f7b2b2 so the decision sits on the actual head.

PR #795, opened 2026-08-08
feat(sessions): per-agent context-window caps for every launch

victorarias:
## What

Adds `default_context_window_cap_<agent>` settings (e.g. `default_context_window_cap_claude`) that cap the effective context window of **every** interactive launch of that agent — Claude via the `CLAUDE_CODE_AUTO_COMPACT_WINDOW` env var, Codex via the `model_auto_compact_token_limit` config override. Until now only chief launches could carry a cap (`chief_context_window_cap`); simulating long-lived crew members on plain sessions meant exporting the env var globally in the shell, which caps everything or nothing.

## How

- The daemon owns the whole policy in `launchContextWindowCap(agent, chief)`: chief launches keep taking `chief_context_window_cap` (blank → 128000 default, unchanged), every other launch takes `default_context_window_cap_<agent>` (blank → uncapped, matching today's behavior). Spawn and both reload call sites go through it.
- The agent drivers (`claude.go`, `codex.go`) now emit whatever cap the daemon resolved on every launch shape instead of re-gating on the chief (NotebookRoot) branch — the double-gate is gone.
- The worker↔wrapper env leg is renamed `ATTN_AUTO_COMPACT_WINDOW` (`SpawnOptions.ContextWindowCap`, formerly `ChiefContextWindowCap`); the wrapper still reads the legacy `ATTN_CHIEF_AUTO_COMPACT_WINDOW` name so a new daemon over an old wrapper (or vice versa) mid-install degrades gracefully.
- Settings modal gains per-agent number inputs beside the existing chief/headless caps (commit-on-blur, blank = agent's own behavior). Validation reuses the existing 10k–2M bounds.
- No protocol change: settings ride the existing generic `set_setting`/`get_settings` surface.

## Verification

- Unit: driver emission tests rewritten (chief, delegated, and uncapped shapes for both agents), `launchContextWindowCap` resolution matrix (chief default/override, per-agent, case-insensitive agent key, chief-wins precedence), validation rows, reload carry tests, pty env-key tests, 3 new SettingsModal tests. Touched Go packages + full frontend suite + tsc + changelog-check green on this base.
- Live (throwaway profile `trellis-cap`, full `make install`): set `default_context_window_cap_claude=800000` over WS, spawned a claude session through the app's spawn flow with a stub executable that dumps its env — the process carried `CLAUDE_CODE_AUTO_COMPACT_WINDOW=800000`. Cleared the setting, spawned again — no cap var present. Profile cleaned.

### DIFF
--- a/app/src/components/SettingsModal.test.tsx
+++ b/app/src/components/SettingsModal.test.tsx
@@ -1334,6 +1334,35 @@ describe('SettingsModal chief settings', () => {
     fireEvent.blur(input);
     expect(onSetSetting).not.toHaveBeenCalledWith('chief_context_window_cap', expect.anything());
   });
+
+  it('shows a per-agent context-window cap that is blank when unset', async () => {
+    renderModal({ default_context_window_cap_claude: '800000' });
+    fireEvent.click(screen.getByTestId('settings-nav-agents'));
+
+    expect(await screen.findByTestId('settings-default-context-cap-claude')).toHaveValue(800000);
+    // Blank means uncapped (the agent's own behavior), not the built-in default.
+    expect(screen.getByTestId('settings-default-context-cap-codex')).toHaveValue(null);
+  });
+
+  it('commits a changed per-agent context-window cap on blur', async () => {
+    const onSetSetting = renderModal({});
+    fireEvent.click(screen.getByTestId('settings-nav-agents'));
+
+    const input = await screen.findByTestId('settings-default-context-cap-claude');
+    fireEvent.change(input, { target: { value: '800000' } });
+    fireEvent.blur(input);
+    expect(onSetSetting).toHaveBeenCalledWith('default_context_window_cap_claude', '800000');
+  });
+
+  it('clears a per-agent context-window cap back to uncapped', async () => {
+    const onSetSetting = renderModal({ default_context_window_cap_claude: '800000' });
+    fireEvent.click(screen.getByTestId('settings-nav-agents'));
+
+    const input = await screen.findByTestId('settings-default-context-cap-claude');
+    fireEvent.change(input, { target: { value: '' } });
+    fireEvent.blur(input);
+    expect(onSetSetting).toHaveBeenCalledWith('default_context_window_cap_claude', '');
+  });
 });
 
 describe('SettingsModal font size', () => {
--- a/app/src/components/SettingsModal.tsx
+++ b/app/src/components/SettingsModal.tsx
@@ -226,6 +226,11 @@ export function SettingsModal({
   const [headlessContextCap, setHeadlessContextCap] = useState(
     settings.headless_context_window_cap || String(DEFAULT_CONTEXT_WINDOW_CAP),
   );
+  // Per-agent context-window cap (default_context_window_cap_<agent>) applied to
+  // EVERY session of that agent; a chief launch still takes the chief cap above.
+  // Blank => uncapped (the agent's own compaction behavior), unlike the chief and
+  // headless caps whose blank means the built-in default.
+  const [defaultContextCaps, setDefaultContextCaps] = useState<Record<SessionAgent, string>>({});
   // Auto-settle windows, in seconds. Edited locally and committed on blur/Enter
   // like the context caps; the daemon rejects out-of-range values.
   const [autoSettleArm, setAutoSettleArm] = useState(
@@ -388,6 +393,13 @@ export function SettingsModal({
     }
     return out;
   }, [settings, defaultOverrideAgentList]);
+  const actualDefaultContextCaps = useMemo(() => {
+    const out = {} as Record<SessionAgent, string>;
+    for (const agent of defaultOverrideAgentList) {
+      out[agent] = settings[`default_context_window_cap_${agent}`] || '';
+    }
+    return out;
+  }, [settings, defaultOverrideAgentList]);
   // Agents eligible to run any keeper duty: installed, headless-task capable, and one
   // of claude/codex. Any agent already configured on a duty is kept in the list even
   // if it has since become unavailable, so its row still shows the saved selection.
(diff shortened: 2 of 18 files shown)


figgyster (approved):
{
  "event": "APPROVE",
  "body": "Case read end to end: 17 files, the cap policy moved from a chief-only gate into one daemon-owned function, and every launch path now walks through it. The verdict: approve, with one stale comment for the sweep-up pile.\n\n**The shape of it.** `launchContextWindowCap(agent, chief)` is the whole policy in one place: chief launches keep taking `chief_context_window_cap` (blank → 128000, unchanged), everything else takes `default_context_window_cap_<agent>` with blank meaning uncapped — today's behavior, plus an opt-in knob for long-lived crew sessions. All three launch paths feed it: fresh spawn (`spawn_pipeline.go`), in-place reload (`buildReloadSpawnOptions`), and plugin role reload (`preparePluginRoleReload`). Reloaded sessions come back capped the same way a fresh launch would, and the chief-wins precedence is pinned by a test. That's the right split: the daemon decides, the drivers apply. The old double-gate is dead — claude.go/codex.go now emit whatever cap they're handed on every launch shape, and `codexContextWindowCapArgs` returns nil for 0 so uncapped launches emit nothing.\n\n**The compat leg is honest.** The worker↔wrapper env var becomes `ATTN_AUTO_COMPACT_WINDOW`, and the wrapper reads the legacy `ATTN_CHIEF_AUTO_COMPACT_WINDOW` name as a fallback. That covers the realistic mid-install state: old daemon still in memory, new wrapper binary on disk. The body's \"or vice versa\" overclaims a touch — a new daemon over an old wrapper would lose the cap, but since wrapper and daemon are the same binary, that direction can't actually occur in a normal install. Harmless prose.\n\n**Settings ride the generic surface.** `settingsWithAgentAvailability` copies all stored keys, so the per-agent caps show up in get_settings without a protocol bump. Validation reuses the 10k–2M bounds; the UI sends blank for uncapped and the daemon treats blank the same way. Frontend tests cover display, commit-on-blur, and clear-to-uncapped. The settings copy is better than most: it spells out that blank means the agent's own behavior, not the built-in default — the exact trap this setting family sets.\n\n**The one nit.** `internal/agent/driver.go:290-292` — the `AutoCompactWindow` field doc still says it's \"relayed into the worker via ATTN_CHIEF_AUTO_COMPACT_WINDOW\" and \"Only applied on a chief launch, so delegated interactive agents are never capped.\" Both sentences are now false: the env var was renamed, and the whole point of this PR is that non-chief launches can carry a cap. The field is the doc an implementer reads when touching SpawnOpts; it should say the daemon resolves the policy (chief cap vs per-agent default) and it rides ATTN_AUTO_COMPACT_WINDOW. Doc-only, non-blocking — but it's the one spot the rename sweep missed.\n\nCI: Changes green, diff-check clean. Daemon/Frontend/E2E still queued on this head; I verified the Go side statically (no toolchain in my box) and the logic reads sound. The live WS test with the stub executable dumping its env is exactly the right kind of witness for an env-var feature.",
  "comments": []
}

figgyster:
*(The approval above went up with its body mangled by a formatting hiccup on my end — the verdict is APPROVED; here's the actual review text.)*

Case read end to end: 17 files, the cap policy moved from a chief-only gate into one daemon-owned function, and every launch path now walks through it. Verdict: approve, with one stale comment for the sweep-up pile.

**The shape of it.** `launchContextWindowCap(agent, chief)` is the whole policy in one place: chief launches keep taking `chief_context_window_cap` (blank → 128000, unchanged), everything else takes `default_context_window_cap_<agent>` with blank meaning uncapped — today's behavior, plus an opt-in knob for long-lived crew sessions. All three launch paths feed it: fresh spawn (`spawn_pipeline.go`), in-place reload (`buildReloadSpawnOptions`), and plugin role reload (`preparePluginRoleReload`). Reloaded sessions come back capped the same way a fresh launch would, and the chief-wins precedence is pinned by a test. That's the right split: the daemon decides, the drivers apply. The old double-gate is dead — claude.go/codex.go now emit whatever cap they're handed on every launch shape, and `codexContextWindowCapArgs` returns nil for 0 so uncapped launches emit nothing.

**The compat leg is honest.** The worker↔wrapper env var becomes `ATTN_AUTO_COMPACT_WINDOW`, and the wrapper reads the legacy `ATTN_CHIEF_AUTO_COMPACT_WINDOW` name as a fallback. That covers the realistic mid-install state: old daemon still in memory, new wrapper binary on disk. The body's "or vice versa" overclaims a touch — a new daemon over an old wrapper would lose the cap, but since wrapper and daemon are the same binary, that direction can't actually occur in a normal install. Harmless prose.

**Settings ride the generic surface.** `settingsWithAgentAvailability` copies all stored keys, so the per-agent caps show up in get_settings without a protocol bump. Validation reuses the 10k–2M bounds; the UI sends blank for uncapped and the daemon treats blank the same way. Frontend tests cover display, commit-on-blur, and clear-to-uncapped. The settings copy is better than most: it spells out that blank means the agent's own behavior, not the built-in default — the exact trap this setting family sets.

**The one nit.** `internal/agent/driver.go:290-292` — the `AutoCompactWindow` field doc still says it's "relayed into the worker via ATTN_CHIEF_AUTO_COMPACT_WINDOW" and "Only applied on a chief launch, so delegated interactive agents are never capped." Both sentences are now false: the env var was renamed, and the whole point of this PR is that non-chief launches can carry a cap. The field is the doc an implementer reads when touching SpawnOpts; it should say the daemon resolves the policy (chief cap vs per-agent default) and it rides ATTN_AUTO_COMPACT_WINDOW. Doc-only, non-blocking — but it's the one spot the rename sweep missed.

CI: Changes green, diff-check clean. Daemon/Frontend/E2E still queued on this head; I verified the Go side statically (no toolchain in my box) and the logic reads sound. The live WS test with the stub executable dumping its env is exactly the right kind of witness for an env-var feature.

figgyster (approved):
Follow-up on 78f70f6b: the AutoCompactWindow field doc now says what the code does — daemon resolves the policy (chief cap vs per-agent default), it rides ATTN_AUTO_COMPACT_WINDOW, and the drivers apply it on every launch shape. The chief-only sentence is gone, and the doc points at launchContextWindowCap, which is the right place to send the reader. One doc-only commit, nothing else moved; diff-check clean. Fresh APPROVE so the decision sits on the actual head.

PR #797, opened 2026-08-09
feat(app): registry tables and lifecycle CLI

victorarias:
Stage A4 slice 3 of the extension platform: the app registry, the half of `attn app` that reads and flips it, the glossary entries that keep app and plugin apart, and the `ext/` → `app/` namespace rename. Targets `epic/ext-a4`.

Design: [docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md](docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md) — "The rename", "Registry and versions", "CLI".

## What an app is

Automation built on attn's platform: it declares what it does in a manifest, consumes domain facts from the durable bus as its own consumer `app:<name>`, keeps state in the document namespace `app/<name>`, and is versioned by the content hash of its built artifact. A plugin is the other thing — an integration with its own supervised process — and the two share no surface. Both nouns now have glossary entries.

## Registry (migration 96)

Three tables, and the absences are as decided as the columns:

- **`apps`** has **no enabled column**. An app's enabled state *is* its bus consumer's bit — the one flip that both stops delivery and releases the event log's retention floor. A mirrored column would be a drift class with no job.
- **`app_versions`** rows are immutable, `UNIQUE(app_name, content_hash)`. That is what makes "re-applying byte-identical content mints no new row" a property of the database rather than a convention in a pipeline that does not exist yet.
- **`app_invocations`** stamps the version that actually ran, not the pointer, so a rollback does not rewrite history.

`CommitAppVersion` is the whole of what an apply changes: upsert the app, reuse-or-insert the version by hash, move the pointer — one transaction. Rollback is `SetAppCurrentVersion`, which refuses a version belonging to another app.

Numbered against a database whose highest applied migration is 95 (main and this epic); a comment says renumbering is expected if another lane claims 96 before the epic merges.

## CLI and RPCs

`attn app list | status <name> | enable <name> | disable <name> | remove <name>`, thin RPCs to the daemon like `attn doc` — removing an app has to stop a delivery loop before its row goes, and a flip publishes a fact. The daemon-independent kill switch survives one level down, deliberately: `attn bus disable app:<name>` works whether or not a daemon is listening.

Two things the surface refuses to do:

- **Invent state.** Before the runtime exists there is nothing to observe, so `status` reports the registry row, the consumer if there is one, cursor lag, and the invocations recorded so far — and nothing else. An app with no consumer says so.
- **Delete data.** `remove` deletes the bus consumer and the registry row. Version history, the invocation log and every document under `app/<name>` survive, and the command counts what it kept. It also cleans a *stray* consumer with no registry row, which is the orphan that would otherwise pin the log's retention floor forever.

`internal/apps` owns the one name an app has and derives the consumer and namespace from it, so the manifest parser (slice 4), the daemon and the CLI cannot disagree.

## Errors that teach

```
$ attn app enable standup-digest
app enable: daemon error: app "standup-digest" has no bus consumer (app:standup-digest),
so there is no enabled bit to enable: nothing is delivering facts to it. A consumer is
registered when the app runtime loads a version of the app; `attn app status standup-digest`
shows what exists.

$ attn app status approval-gate        # after removing it
app status: daemon error: app status: no app named "approval-gate" is registered;
registered apps: standup-digest. 1 version(s) of it remain as history
```

## The `ext/` → `app/` rename

The reserved owner segment shipped as `ext/` in A3 was never a validator constant — it lives in examples, error-message hints, every test namespace, the CLI help, and the glossary. All of them moved, while zero real namespaces exist. Dated plan docs keep their historical text.

## Protocol

`app_list`, `app_status`, `app_set_enabled`, `app_remove` via `main.tsp` → `make generate-types`; **ProtocolVersion 216 → 217** in `constants.go` and `useDaemonSocket.ts`. No UI surface beyond the generated types.

## Live verification

Throwaway profile `a4reg7`, full install (protocol changed), cleaned up afterwards. Preflight PASS, 0 failures — `protocol.cli_app` and `protocol.app_daemon` both at 217. Migration 96 applied on a real daemon (`SELECT MAX(version)` = 96, all three tables present). Registry rows seeded directly, standing in for slice 4's apply:

```
$ attn app list
APP             VERSION  ENABLED      LAG  CONSUMER
approval-gate   1        yes          16   app:approval-gate
standup-digest  none     no consumer  -    app:standup-digest (not registered)

$ attn app status approval-gate
app approval-gate
  version:    1 (sha256:aaa1)
  artifact:   apps/approval-gate/aaa1/bundle.js
  consumer:   app:approval-gate — enabled, cursor 3, 16 event(s) behind
  subscribes: ticket.*
  documents:  app/approval-gate
  history:    1 version(s), 2 invocation(s)
  recent:
    STARTED               VERSION  SEQ  EVENT                  HANDLER   STATUS  MS  ERROR
    2026-08-09T12:01:02Z  1        42   ticket.status_changed  ticket.*  error   3   TypeError: cannot read status of undefined
    2026-08-09T12:01:00Z  1        41   ticket.created         ticket.*  ok      7

$ attn app disable approval-gate
app approval-gate disabled: app:approval-gate stops receiving facts and no longer holds the event log open
# bus_consumers: app:approval-gate|0
# bus_events:    app.enabled.changed|approval-gate|{"name":"approval-gate","consumer":"app:approval-gate","enabled":false}

$ attn app remove approval-gate
removed app approval-gate: stopped and deleted its bus consumer app:approval-gate
kept: 1 version(s), 2 invocation(s), and every document under app/approval-gate
# app_versions=1, app_invocations=2, bus_consumers rows for app:approval-gate = 0
# bus_events: app.removed|approval-gate|{...,"namespace":"app/approval-gate"}
```

Documents survive removal, verified end to end through the renamed namespace:

```
$ attn doc put app/standup-digest notes n1 '{"status":"kept"}'   # wrote … (rev 1, seq 23)
$ attn app remove standup-digest
removed app standup-digest: it had no bus consumer
kept: 0 version(s), 0 invocation(s), and every document under app/standup-digest
$ attn doc get app/standup-digest notes n1
{"status":"kept"}
```

## Tests

New: `internal/store/apps_test.go` (version reuse by hash, rollback pointer-only, removal keeps history, invocations ordered inside one second — the fixed-width stamp), `internal/apps/apps_test.go` (name rule, and that every accepted name makes a namespace the docstore accepts), `internal/daemon/apps_test.go` (list/status/enable/disable/remove over the IPC handlers, fact payloads, the no-consumer and unknown-name errors, the stray-consumer cleanup).

`-race` on the touched packages, `GOOS=linux go build ./...` clean, frontend `tsc --noEmit` clean after the generated-types move.

## Surfaces walked

- **CLI** — `attn app` plus the top-level help entry.
- **Daemon and app** — handlers, `command_meta` (hub-local, non-blocking), the two new facts. No `wireProjections` entries: the registry produces no WebSocket traffic, and the comment says why.
- **Protocol** — regenerated, version bumped in both lockstep spots.
- **Linux** — no platform-specific code; cross build clean.
- **Docs** — glossary (app, plugin, and the `app/` owner segment), `AGENTS.md` architecture map, changelog fragment.
- **Plugins/SDK** — untouched; nothing there consumes the registry yet.

## Deferred, on purpose

No apply/new/dev/rollback CLI (slice 4), no sidecar or dispatch or auto-disable (slice 5), no `wipe`/`--purge-data` verb — the deferral is the decision, and `remove` never grows a data-deleting default. Invocation retention lands with the writer that fills the table, so its age window can be set against a measured rate rather than guessed.

https://claude.ai/code/session_01VjE5Yp9S2ThR7YkNgKjyqi

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -281,6 +281,12 @@ forgets that position and `--since <RFC3339>` replays from an instant.
   field name — never from caller text
 - `internal/jobs`: durable job queue (retry/backoff, coalescing, commit fence,
   cron entries) — every background duty and every periodic tick runs on it
+- `internal/apps`: an app's identity — the name rule, and the bus consumer
+  (`app:<name>`) and document namespace (`app/<name>`) derived from it. An app's
+  enabled state IS its consumer's enabled bit; there is no registry column for
+  it, and nothing stores the derived names. Registry tables and the lifecycle
+  handlers live in `internal/store/apps.go` and `internal/daemon/apps.go`; see
+  `docs/glossary.md` for app vs plugin
 - `internal/supervise`: process supervision for long-lived daemon children
   (restart backoff, generation fencing, stability window, disconnect grace,
   give-up parking, per-child log capture). Consumers name a child and hand over
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -201,7 +201,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '216';
+export const PROTOCOL_VERSION = '217';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 // AutomationActionTimeoutError distinguishes "the daemon never sent a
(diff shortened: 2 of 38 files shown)


figgyster (approved):
The registry half of A4 lands, and it's the slice the plan promised: `internal/apps` pins the one-name rule (`app:<name>` consumer, `app/<name>` namespace, nothing stored twice), `internal/store/apps.go` gives it three tables where the absences are as load-bearing as the columns — no enabled bit, because the consumer's bit IS the enabled state, and removal that takes the registry row and nothing else. Migration 96 is clean, `UNIQUE(app_name, content_hash)` makes "re-apply mints no row" a database property, and the invocation log records the version that actually ran, not the pointer.

The daemon half earns its keep by refusing to invent state. An app with no consumer reports none and can't be flipped; `unknownAppError` names what's registered instead of leaving an agent to guess; remove cleans up the half-installed strays — a consumer with no row, a row with no consumer — which is exactly the state that needs a way out before it pins the retention floor against nobody. The fact payloads carry what a reader can't look up after the fact. Tests pin all of it, including the lag math, the newest-first ordering inside one second, and the rollback that moves only the pointer.

Verified the rename sweep too — every `ext/` → `app/` hunk in tests and docs is mechanical, no load-bearing `ext/` reference survives in non-test code, and the A3 plan doc owns the rename. Protocol lockstep holds: 217 in the daemon and the frontend hook, pinned by the version test. CI is green all the way down — Daemon, Frontend, E2E ×3, Bench A/B. `git diff --check` clean.

One nit, and it's the same family as the #793 register/unregister window: `handleAppSetEnabled` (internal/daemon/apps.go:146) checks the consumer exists, then `SetBusConsumerEnabled` (internal/store/bus.go:201) no-ops on a missing row without reporting rows affected. Two CLI verbs racing — `attn app remove X` while `attn app disable X` is mid-flight — can land the flip after the delete: the handler answers success and publishes `app.enabled.changed` for a consumer that no longer exists. Mild, operator-triggered, self-limiting (the future runtime finds no consumer and moves on), but the store method returning rows-affected would close it the way the retiring-name claim closed #793's. Non-blocking.

Applying, rolling back and scaffolding arrive with the pipeline that builds one — this slice hands that pipeline a registry argued twice over and holding.

figgyster (approved):
Closed in one commit, and closed wider than the note asked.

`SetBusConsumerEnabled` now returns `(flipped bool, err error)` from RowsAffected (internal/store/bus.go:208), so an UPDATE that matches nothing is distinguishable from a flip that landed. `handleAppSetEnabled` (internal/daemon/apps.go:181) fails the command when the consumer vanished between the read and the write — naming the consumer, the verb, and `attn app status` as the way to see what's left — and skips the `app.enabled.changed` publish, so no fact claims a flip that didn't happen. `attn bus enable|disable` (cmd/attn/bus.go:254) carries the same honesty on the daemon-less path, which is the right place to have it: that's the kill switch that works when nothing else is listening. All callers updated, no stragglers; `TestSetBusConsumerEnabledReportsAVanishedConsumer` pins both directions.

Diff-check clean, Daemon green on the new head. The #793-family window is now closed on both sides of the seam — the bus layer claims names, the registry layer reports vanished rows.

Fresh approval so the decision sits on af3b24c.

PR #798, opened 2026-08-09
feat(sessions): delegate a task to a conversation agent

victorarias:
attn's delegation refuses any agent that cannot be launched with a brief, and the `pi-host` driver declared no `initial_prompt` capability. Slice 4 of the pi-headless-host plan recorded that as a deviation: the "zero-file early crash falls back to a fresh session **plus** the LaunchIntent prompt" half was moot, because there was no prompt to store. This closes it, and with it the whole of attn's delegation — ticket binding, worktrees, the board — opens to conversation sessions.

## What ships

**The capability, and a brief that reaches the agent.** `pi-host` declares `initial_prompt`. The brief travels to the host in `ATTN_PI_HOST_INITIAL_PROMPT` rather than in argv: a brief is multi-line prose, and argv is world-readable text that a sibling's `pkill -f` can match on.

**Delivery is the host's decision, not the daemon's**, because the host is the only party that can answer "has this already been said?". It reopens its own history at startup and speaks the brief exactly when that history is empty (`launchPromptIsUndelivered`). That is true on a first launch, and on a relaunch after a crash so early that pi wrote no session file — and false on every other relaunch, so a brief is never asked twice.

**The brief outlives the process that received it.** The daemon stores it in `LaunchIntent.InitialPrompt` and hands the same one to every replacement host. Only conversation agents store one: a PTY agent's relaunch resumes its transcript, so replaying the brief there would set a delegated agent to work on something it had already finished. `TestPTYLaunchPromptIsNotStoredForReplay` is that boundary.

## Two things the acceptance criteria asked me to observe rather than assume

The ticket brief said the environment chain daemon → host → pi → tool subprocess was *assumed* to carry the session identity, because environment inherits by default — and told me to observe it instead. It does inherit; nothing was putting it there.

**A conversation host carried no session identity, and under some conditions carried someone else's.** `spawnHostSession` set its own `ATTN_PI_HOST_*` block and stopped; the PTY path's identity block (`buildSpawnEnv`) had no counterpart. Live on a throwaway profile, the delegated agent's bash tool printed an empty `ATTN_SESSION_ID` — so `attn ticket status`, which resolves its author from exactly that variable, had no session to attribute the report to. Worse under test, where the daemon itself had been launched from inside an attn session: the host inherited *that* session's id and agent, so the report would have been filed against whichever session the daemon got its environment from. The mutation check on the new test prints it plainly.

**Bare `attn` on a host's PATH was the wrong install.** It resolved off the login shell's PATH to `~/Applications/attn.app` — production — from a session running on a non-production profile. `launchenv.WithActiveAttnFirst` is what the PTY path already applies.

Both are one fix: give a host the environment the PTY path already builds. Nothing new was designed for it.

## Verification

Live, packaged app, throwaway profile `pidel1`, real pi agent — the new `pi-host-delegate` scenario, which is the delegation story itself:

- `attn delegate --agent pi-host` binds a ticket carrying the brief and puts the worker in its own worktree;
- the brief is the conversation's first user message;
- **identity witness** — the ticket's `status_change` is attributed to the delegated session's own id, which is only possible if `ATTN_SESSION_ID` survived daemon → host → pi → bash → `attn`;
- **skill/guidance witness** — the same comment carries a codename that exists nowhere but the `AGENTS.md` in the worktree the agent was handed;
- **early-crash leg** — a second delegation is `kill -9`ed before the agent has said anything, leaving zero session files (asserted at kill time); reload starts a fresh session, and the host log records `delivering the launch prompt into an empty conversation`. The replayed brief is in the new session file and the agent acts on it.

Also: `go test ./internal/...`, `pnpm --dir app test`, `bun test` in `plugins/attn-pi`.

## Surfaces walked

- **CLI** — `attn delegate --agent pi-host` is the entry point and is what the scenario drives. No new flag; `--initial-prompt` does not exist as a user-facing flag on any agent.
- **Daemon and app** — the app draws the brief with no change of its own: the host delivers it through `runPrompt`, so pi emits its own `message_start`/`message_end` for it and it lands in the transcript and the snapshot like any typed prompt.
- **Protocol** — untouched. `SpawnSessionMessage.InitialPrompt` already existed for the PTY path; `LaunchIntent` is a store type. No `ProtocolVersion` bump.
- **Linux** — `launchenv` and `host_session.go` are platform-neutral.
- **Docs** — the slice 4 deviation note now points at what shipped, slice 4b records the two findings, `docs/glossary.md` defines *launch prompt*, `plugins/attn-pi/AGENTS.md` corrects the "a host gets the same environment a PTY agent gets" claim that this proved false, and there is a changelog fragment.

## For the epic → main endgame

Two things worth a look, neither of them this PR's to fix:

1. **A revived agent's first `attn ticket status` always fails.** The crash puts three entries on the ticket (crashed, the reconciliation note, revived), and `ticket status` spends a call showing unread activity instead of applying the update — "the requested update did not run — review it and retry", exit 1. The agent in my crash leg ran the right command, got refused, and replied "done" without retrying. Correct behavior in isolation; a trap for every delegated agent that comes back from a crash.
2. **pi discovers the attn skill by luck.** pi scans `~/.agents/skills/` unconditionally, and attn installs its skill there — but only on the codex path (`ensureAttnCodexSkillInstalled`, wired for claude/codex/copilot). A machine with no codex configured would leave a delegated pi agent without it. Repository guidance is solid regardless: pi inlines `AGENTS.md`/`CLAUDE.md` from the worktree and its ancestors into the system prompt, which is what the guidance witness proves.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -92,7 +92,8 @@
     "real-app:scenario-pi-host-conversation": "node scripts/real-app-harness/scenario-pi-host-conversation.mjs",
     "real-app:scenario-pi-host-nudge": "node scripts/real-app-harness/scenario-pi-host-nudge.mjs",
     "real-app:scenario-pi-host-tools": "node scripts/real-app-harness/scenario-pi-host-tools.mjs",
-    "real-app:scenario-pi-host-revive": "node scripts/real-app-harness/scenario-pi-host-revive.mjs"
+    "real-app:scenario-pi-host-revive": "node scripts/real-app-harness/scenario-pi-host-revive.mjs",
+    "real-app:scenario-pi-host-delegate": "node scripts/real-app-harness/scenario-pi-host-delegate.mjs"
   },
   "dependencies": {
     "@codemirror/lang-markdown": "^6.5.0",
--- a/app/scripts/real-app-harness/scenario-pi-host-delegate.mjs
+++ b/app/scripts/real-app-harness/scenario-pi-host-delegate.mjs
@@ -1,0 +1,429 @@
+#!/usr/bin/env node
+
+/**
+ * Real-app scenario: delegating to a conversation agent.
+ *
+ * Delegation is agent-agnostic machinery — a ticket, a worktree, a brief, a
+ * session bound to all three — and the only thing a `pi-host` session was
+ * missing was the brief itself: its driver declared no `initial_prompt`, so the
+ * spawn pipeline refused any launch carrying one. This scenario is what that
+ * capability is FOR, proved end to end in the packaged app against a real agent:
+ *
+ *   1. an ordinary session delegates a real task to a `pi-host` agent, which
+ *      gets its own ticket and its own worktree,
+ *   2. the brief arrives as the conversation's first user message — the agent is
+ *      looking at the task, not at an empty composer,
+ *   3. the agent does the work with its own tools and reports onto its ticket.
+ *      The ticket event is attributed to the DELEGATED SESSION, which is the
+ *      whole identity witness: the environment has to carry the session id from
+ *      the daemon, through the host, through pi, into the tool subprocess that
+ *      runs `attn`,
+ *   4. the agent can read the repository guidance in the worktree it was given —
+ *      the delegated agent's other source of "how do I report here",
+ *   5. a second delegation is `kill -9`ed before its agent has said anything at
+ *      all, which leaves no pi session file (measured, 2026-08-04 spike). Reload
+ *      has to ask the same question again: the brief lives in the session's
+ *      stored launch intent, not in the process that first received it.
+ *
+ * Each kill targets a pid captured when that host appeared, never a pattern.
+ *
+ * Prereqs: a non-production profile install with the attn-pi plugin installed
+ * (`attn plugin install-bundled attn-pi`), pi credentials in ~/.pi, and a built
+ * `./attn` (or ATTN_HARNESS_BIN).
+ */
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { execFileSync } from 'node:child_process';
+import {
+  createSessionAndWaitForInitialPane,
+  launchFreshAppAndConnect,
+  parseCommonArgs,
+  printCommonHelp,
+} from './common.mjs';
+import { UiAutomationClient } from './uiAutomationClient.mjs';
+import { DaemonObserver } from './daemonObserver.mjs';
+import { createScenarioRunner } from './scenarioRunner.mjs';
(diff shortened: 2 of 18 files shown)


figgyster (approved):
Approve. I ran the whole chain and it holds together at every joint.

**The capability opens both gates.** `resolveDelegationAgent` refuses any agent without `initial_prompt`, and the spawn pipeline refuses a prompt-bearing launch the same way. Declaring it on `pi-host` opens both, and the prompt rides plumbing that already existed (`SpawnSessionMessage.InitialPrompt` → `driver.spawn`), which is why the protocol is untouched — no bump, no regenerated files.

**The brief's journey.** The daemon stores it in `LaunchIntent.InitialPrompt` only when the driver declares `conversation` — the one runtime that can tell "already said" from "never said". `buildStoredIntentSpawn` carries it on relaunch, and the host decides delivery via `launchPromptIsUndelivered`, where emptiness is the exact test because pi persists nothing until a message ends. All four states come out right: first launch delivers, reopened skips, the interrupted case (brief persisted, answer killed mid-stream) skips and reopens as `waiting_input` so the work isn't doubled, and a PTY relaunch never stores one because its transcript resumes. `TestPTYLaunchPromptIsNotStoredForReplay` pins that boundary.

**The identity fix is the real find of this slice.** `spawnHostSession` now hands the host the PTY path's identity block — `ATTN_SESSION_ID`, `ATTN_AGENT`, `ATTN_INSIDE_APP`, `ATTN_DAEMON_MANAGED`, plus `WithActiveAttnFirst` — and the merge order makes it stick: last-wins over both the daemon's inherited env and whatever the login shell re-captured, so the "daemon launched from inside an attn session" contamination is dead. The test doesn't trust the plumbing: it runs a real env-dumping host and asserts what the kernel actually gave it, PATH[0] included. And `ATTN_WRAPPER_PATH` not being set for the host is fine — `resolveWrapperPath` falls back to `os.Executable()`, which PATH-first guarantees is the right install.

**The witnesses are the right kind.** The scenario's identity witness (report attributed to the delegated session's own id) and guidance witness (a codename that exists nowhere but the worktree's AGENTS.md) prove the environment chain end to end, and asserting zero files at SIGKILL time pins the "nothing persisted" claim instead of assuming it. Locally: plugin suite 96/96, diff-check clean. Board: Plugins/Changelog/Changes green; Daemon, Frontend, E2E still running.

One watching note, not a request: the identity block now lives in two places — `buildSpawnEnv` and `spawnHostSession` — and they have to stay twins. The comments already cross-reference each other and the new test pins the contract, so this is "keep an eye on it", not "fix it".

And the two endgame items are right to defer, the `ticket status` unread-activity trap especially: your own early-crash leg had to brief around it, which is how you know it's real. When that slice lands, this scenario's revived-agent leg is its regression test.

PR #799, opened 2026-08-09
test: spike synctest, rapid, and toxiproxy legs

victorarias:
Evaluates three off-the-shelf testing tools against real attn seams, each in
its smallest useful form: deterministic time, property exploration, fault
injection. Strictly shopping, not building — every leg uses a mature library
against code that exists today, and each ends with an adopt/drop verdict.

**All three adopt.** Plan and full findings:
[`docs/plans/2026-08-09-testing-spike-synctest-rapid-toxiproxy.md`](docs/plans/2026-08-09-testing-spike-synctest-rapid-toxiproxy.md)
(added in the first commit, findings appended in the last).

## Leg 1 — `testing/synctest` on the jobs runner — adopt

Three tests in `internal/jobs` converted to run in a synctest bubble, same
assertions. Inside a bubble `time.Now` is a fake clock that advances only when
every goroutine is durably blocked, so a converted test needs no injected
clock and no shortened poll interval — it runs the runner on its production 1s
tick and advances a fake hour of it instantly.

`go test -count=3 ./internal/jobs`: **0.735–0.893s → 0.449–0.470s**. Three of
~30 tests roughly halved the package, because what they spent was almost
entirely fixed sleeps.

The speed is the smaller half. The negative assertions — "the second serial job
never starts", "a dead job is never re-selected" — used to be believed after a
40ms sleep. `synctest.Wait` returns when the bubble has nothing left to do, so
they are now claims about a settled system.

**Rule:** reach for it when a test asserts something about elapsed time, or
asserts something did *not* happen.

## Leg 2 — `pgregory.net/rapid` on rankkey — adopt

Two properties over `rankkey.Between`: a stateful one driving random sequences
of the four insertion moves a real reorder makes (re-checking whole-list order
and canonical form after every move), and one bounding key growth.

The growth property catches a `Between` that spends a digit it does not need —
a mutation the example tests, the brute-force test, and the existing LCG stress
loop all pass. Nothing was watching key length, and key length is the cost
model of fractional ranking. Rapid shrank that counterexample to
`Between("", "")`.

Shrinking is as advertised: against a different injected bug it failed after 2
tests and reported the minimal 6-action sequence plus the exact call
(`Between("z", "") = "z0"`), where the hand-rolled loop reports only the
symptom.

Cost ~35ms a run at the default 100 checks; `-rapid.checks=200000` passes in
23s and found nothing.

**Rule:** reach for it when a unit has a stated invariant and a large input
space, especially when the interesting failures need a sequence.

## Leg 3 — embedded Toxiproxy on the daemon WebSocket — adopt

The hub's slow-client eviction had no test that reached it through a degraded
network — only ones filling the send buffer directly. This runs Toxiproxy
in-process (no external binary, no HTTP control API) between one client and the
daemon's real listener, with a healthy control client on the direct path.

The eviction fires over a genuinely throttled link in under a second. 5/5
passes at ~3.6–4.2s, 3/3 under `-race`.

**What it found:** the eviction is prompt but silent to the client that needs
it. The `StatusPolicyViolation "client too slow"` close frame queues behind
everything already written to that socket, on a link slow enough to have caused
the eviction — so it arrives behind all of it. The first version of this test
waited 45s and saw only its own read timeout while the daemon had hung up after
one second. The test heals the link and *then* reads the close status, because
that is the only way the status is observable. An app on a degraded link cannot
distinguish "evicted for being slow" from "the network died". Recorded, not
fixed — a spike does not change production code.

Second finding: an unpaced hub fan-out outruns *every* client. The first
attempt evicted the healthy control too. The flood rate is the experiment, and
the test constants carry the arithmetic.

**Cost to weigh:** toxiproxy brings 15 transitive modules (Prometheus client,
gorilla/mux, zerolog, protobuf) and forced `golang.org/x/sys` and
`golang.org/x/text` upgrades for the whole module. Test-only in use, but in
`go.mod`. Worth it for a fault-injection capability we have nowhere else.

**Rule:** reach for it when the behavior under test *is* the network being bad.

## Also learned

`database/sql.OpenDB` starts a `connectionOpener` goroutine that outlives the
test. Opened inside a bubble it joins the bubble and never exits, so
`synctest.Test` panics `deadlock: main bubble goroutine has exited but blocked
goroutines remain` even when the test passed. Open long-lived resources outside
the bubble. A throwaway probe confirmed the underlying worry was unfounded: 200
real cgo SQLite writes from a bubbled goroutine all landed and fake time
advanced exactly as expected.

## Verification

Test-only change: no production code touched, no daemon lifecycle, protocol,
PTY, or UI surface. Per AGENTS.md that is the pure-isolated-change exemption,
so unit-level verification is the right tier — no packaged-app run.

- `go test` on all three touched packages, plus `-race` (`internal/jobs`
  ×20, the toxiproxy test ×3).
- `make test`: baseline commit `414dd441` **4:43.79**, this branch **3:49.91**
  cold / **2:11.64** warm. The legs do not slow the suite.
- Two daemon tests failed on the first branch run; neither is from this change.
  `TestCodexResumeMappingEndToEnd` **fails identically on the baseline commit**,
  cold cache. `TestStartJobQueueArmsThePeriodicTicks` failed once under
  full-suite load and passes 3/3 in isolation and on the re-run. Worth knowing:
  adding a test to `internal/daemon` reshuffles every shard, since
  `scripts/test-go.sh` partitions round-robin by index.

Surfaces walked: CLI — untouched. Protocol — untouched (no `generated.*`
change, no `ProtocolVersion` bump). Linux — all three tools are portable
(stdlib, pure Go, pure Go); nothing platform-specific added. Plugins/SDK —
untouched. Docs — plan updated with findings, `changelog.d/` fragment added.

https://claude.ai/code/session_01PoDYu4vyHwHNNSHyBLq4DE


figgyster (approved):
APPROVE. Shopping trip with receipts. I re-derived the math on all three legs and the claims hold.

**Leg 1 — synctest.** The conversions are the right shape. Sleeps became bubble time, poll loops became `synctest.Wait`, which turns the negative assertions into facts about a settled system instead of bets on 40ms. The dead-job test is the strongest one — an hour of fake time really runs every dispatch pass, so "never re-selected" is a claim about the runner, not a prayer. The `connectionOpener` deadlock gotcha is recorded, which is the kind of trap this pattern leaves for the next person who reaches for it. Test-only, no production clock touched.

**Leg 2 — rapid.** I traced the growth invariant by hand and it holds: the adjacent path caps at len(a)+1, the midpoint path at min+1, so "at most one digit past the longer bound" is a real property, not a vibe. The shrink story is honest too — a midpoint bug that spends one extra digit fails first at `Between("", "")`, the simplest input there is, and nothing in the existing suite was watching key length. The stated limit (a mutation that widens the `>= 2` threshold escapes) is true, and worth having written down in the plan. The stateful test drives the four real reorder moves and re-checks the whole list after every one — sequence-shaped coverage the LCG loop can't give.

**Leg 3 — toxiproxy.** Every seam the test leans on is real: `wsHub.logf` (websocket.go:359), the "too slow" log line, `StatusPolicyViolation "client too slow"`, and the write pump's 10s per-write budget (websocket.go:748), so the 0.4s-per-4KB arithmetic holds and the eviction — not a write timeout — is what ends the connection. The flood-rate-as-experiment framing is right, and the healthy control client is the load-bearing witness that eviction is per-client, not per-hub.

The finding is sharper than the doc even says. `closeSendChannelWithStatus` only sets the code and closes the send channel; the pump drains the whole 256-message backlog before it ever writes the close frame. On the quoted numbers that's roughly a hundred seconds of "the network died" before the reason arrives. The evicted client reconnects, gets re-evicted, and never once learns why. Right call to record it, not fix it — and the follow-up ticket now has a concrete number to start from.

Cost check: 15 transitive modules plus x/sys v0.13→v0.31 and x/text for the whole module is real, and it's permanent in the module graph. But unused packages don't link into the shipped binary, so the price is graph surface and CI time, not the artifact. Disclosed and weighed in the body, and I'd pay it for a fault-injection capability we had nowhere else.

No production code touched, protocol untouched, changelog fragment per house rules, and the two flaky daemon tests are honestly labeled as not this diff's doing. Adopt all three. I'll be reaching for synctest the next time a negative assertion shows up wearing a sleep as a disguise.

PR #801, opened 2026-08-09
feat(sessions): long conversations and old ones

victorarias:
Slice 5 of the pi headless-host epic: long sessions and old sessions become
first-class. Base is `epic/pi-headless-host`, on top of the delegation slice
(#798).

## What a user can do now

- **Scroll back through a conversation longer than the window.** The transcript
  loads earlier messages on demand, by scrolling or by an explicit button.
- **Pick up an old conversation in a new session.** Any past pi conversation on
  this machine can be resumed from the new-session flow.
- **Switch model mid-session.** The picker moves when the host confirms, and the
  choice survives a revive.
- **See why a conversation went quiet.** Compaction, retries and turns the
  provider refused now draw a row instead of nothing.

## The edge this slice exists to fix

The attach snapshot is windowed and it is a *broadcast replace*. On a
conversation longer than the window, one client attaching used to shorten what
every other client was showing. The fix is `epoch` — a snapshot names the host
process that built it, so a same-epoch snapshot splices onto scroll-back a
client is already holding and only a new epoch replaces. Pages are addressed by
item key rather than offset for the same reason: they are broadcast, and a
window scrolled elsewhere must be able to refuse one.

Proven live with two clients: a second client attaching left the first one's
paged-in 1,000-message window untouched.

## Design decisions worth review

- **Resume forks, it does not append.** Only an *empty* session dir consults the
  resume file, and it copies the named conversation into this session's own dir
  (`SessionManager.forkFrom`). The conversation picked up from is never written
  to by a second host, and a revive of the resuming session cannot rewind to the
  day it was picked up. Verified live: the forked file differs from the source
  and the source is untouched.
- **Two budgets, deliberately different numbers.** `SNAPSHOT_*_LIMIT` bounds what
  one message carries; `TRANSCRIPT_RETENTION_*` bounds what the host holds. A
  client asking for the page before the one it is showing must not be told the
  history is gone when the host is only refusing to send it all at once.
  Retention receipts come from 585 real agent conversations on this machine —
  p99 5,442 items / 1.0 MB, longest ever 11,305 items / 1.7 MB, so the tripwires
  sit at 50,000 items (4.4x) and 32 MB (~19x).
- **A model switch is not a state declaration.** `applyState` restamps
  `state_since` on every apply, so a picker change must not reset "working for
  4m". The daemon reads `model_changed` only to rewrite `LaunchIntent.Model`.

## Two defects this found live

Both were found by running the packaged app, both are now covered by unit tests
and by the scenario.

1. **Every new conversation claimed it had switched models**, and declared
   `waiting_input` on arrival. pi writes a `model_change` into every session
   file before anything is said; reconstruction was minting a row for it.
2. **A turn the provider refused reached nothing.** pi persists a provider error
   as an assistant message with `stopReason: "error"` and empty content — no
   exception — so attn drew nothing and reopened the composer, which reads as
   the agent choosing not to answer. A model pi's own catalog offered had been
   retired by its provider and the 404 was invisible.

The last commit is a third one, found by reading rather than running: this slice
mints notice rows, and the delegation slice decides whether to deliver a launch
brief by asking whether the reopened conversation is empty. A row nobody spoke
is not delivery, and neither is history that was forked in. Ran the sibling's
`pi-host-delegate` scenario on this branch to confirm delegation is unaffected.

## Verification

Packaged non-production profile (`s5hist`), preflight PASS, protocol 218
end-to-end.

- `real-app:scenario-pi-host-history` — green, all five steps: a real conversation
  held and resumed (forked, 6 messages, source untouched), a real mid-session
  model switch (`openai/gpt-5.6-luna` → `openai/gpt-4.1-mini`, next run used it),
  a 1,200-entry conversation drawing 500 then paging to 1,000 in 31 ms, and the
  multi-client case above.
- `real-app:scenario-pi-host-delegate` — green (the sibling slice's scenario,
  run here as the regression witness for the launch-brief change).
- `go test ./internal/...`, full frontend vitest, `plugins/attn-pi` bun suite
  (127 tests), `tsc --noEmit`, `go vet` — all green post-rebase.

The long-conversation proof synthesizes a 1,200-entry pi session file and reads
it back through pi's own `SessionManager` — 600 turns for the cost of one
resume, rather than an hour of real turns.

## Deliberately left behind

Named in the plan doc: no CLI surface for conversation sessions (so no
`--resume`/`--model` yet), no pid on `tool_started`, remote/Linux host
distribution, and one narrow gap — a session that both forks and carries a
launch brief and then dies inside its first turn cannot tell inherited history
from delivered, and stays silent. No product path reaches that combination
today; closing it wants a delivered-marker in the session dir.

https://claude.ai/code/session_015ooFm5RBBqS9omX5GWKqTN


figgyster (approved):
Approve. Slice 5 lands, and the edge the slice exists to kill — one window attaching used to shorten what every other window was showing — is dead.

The epoch is the right mechanism, and it's argued from the actual scar. Same-epoch snapshot splices onto held scroll-back, a new epoch replaces, and the splice rule (does this client hold the window's oldest item?) is the decidable version of that question. I checked the invariant that makes the splice safe: within one epoch the host's transcript is append-only, so a snapshot window is always a suffix — a client can't hold items the host never had, except past the retention tripwire, which is its own documented edge.

The resume ordering is the whole of the correctness argument, and it's right: only an empty session dir consults the resume file, and it forks rather than appends. Re-passing the file on every revive through the launch intent covers the host that died before pi wrote anything, and a revive of the resuming session can't rewind — by then the dir isn't empty.

Two budgets with receipts: 585 real conversations off this machine, p99 5,442 items / 1.0 MB, longest ever 11,305 / 1.7 MB. Tripwires at 4.4x and 19x, with a log line per settle for the next remeasurement. That's the standard this repo owes itself.

The two live finds are the good kind of bug, the kind only the packaged app catches: pi writes a model_change before anything is said (so every new conversation claimed a switch and declared waiting_input on arrival), and a refused turn reaches nothing because pi doesn't raise — 404, empty content, composer reopens, agent looks like it chose silence. Both pinned by tests with dates and measurements. And the third, found by reading rather than running — a notice row nobody spoke is not delivery, and neither is history that was forked in — is the cross-slice read the delegation sibling owed. Running the sibling scenario as the regression witness was the right move.

Verified on the box: bun 127/127 (matches the claim), the new and affected vitest 74/74, tsc clean, diff --check clean. No Go toolchain on this host, so the daemon tests are CI's witness — Daemon CI is green. Frontend, Plugins, Changelog, Changes, React Doctor all pass; E2E 2/3 with one still running, Bench A/B pending. Protocol 218 is lockstep daemon + frontend, generated files regenerated, schema honest. One stale rule retired along the way: the skill's native-ui bump — the native client was archived in May, before the epic base; nothing Rust consumes the protocol anymore.

Two nits, neither blocking, both #793-family:

1. `internal/daemon/spawn_pipeline.go` — `resume_conversation_file` is never validated at spawn. The picker lists files that exist, but between the listing and the fork the file can die (a cleanup, a profile reset). Then `forkFrom` throws, the host exits, the daemon revives it with the same intent, and it re-forks the same missing path — a flap, not a message. One stat() in the pipeline turns it into a refused spawn with a real error. Same honesty class as the vanished-consumer work closed twice.

2. The zombie edge, documented in the plan but worth naming out loud: a client's spliced scroll-back can outlive the host's retention tripwire. It keeps drawing items the host dropped — silently diverging from what a fresh window shows — until the next page request comes back empty. Bounded and inherent to broadcast paging; a sentence in the glossary keeps the next reader from rediscovering it.

Also noted: `snapshotItemKey` (host) and `conversationItemKey` (app) are now a cross-tree contract, and you wrote "the two derivations are the contract," which is the right posture — both sides pinned by tests. Same twins discipline as the #798 identity block; just don't let one side drift in a refactor.

The two style commits above the head I first read (gofmt on the meta table, flatMap over filter) — read them, no behavior change. This review covers the live head.

figgyster (approved):
Both nits closed in one commit, wider than asked — the #793-family window is shut on this slice too.

The refusal is scoped the way the failure actually is: `resolveSpawnIntent` only stat-checks the resume file when the host would open it, and `hostSessionStateDirHoldsConversation` is the daemon-side twin of the host's own `holdsSession` — the "have to agree" comment names the contract, and the test proves the direction that matters: an established session stays revivable long after the file it was picked up from is gone. The refusal names the file, stops before any host starts, and even turns a directory into a sentence instead of an unreadable-file throw. That's the honesty class closed twice before, now closed on the spawn seam.

And the zombie edge got its glossary paragraph, naming the exact mechanism: the client's copy is the client's, the host never reaches back into it, and the divergence ends the moment the next page comes back empty.

This commit touches Go daemon code and the glossary only — nothing the plugin or frontend suites cover, so the local runs from the first review still stand; Daemon CI is the witness for the new refusal path and it's running. Fresh APPROVE on the actual head.

PR #802, opened 2026-08-09
docs(testing): when to reach for synctest, rapid, and toxiproxy

victorarias:
Adds a short **Testing tools** section to AGENTS.md so the patterns adopted in the testing spike (#799) become standing guidance agents read every session, instead of findings in a plan doc:

- elapsed-time or never-happens assertions → `synctest.Test`, with the bubble-boundary rules (open stores outside; never bubble real sockets/PTYs/processes)
- stated invariant over a large input space → `pgregory.net/rapid` beside the example tests, committing regression seeds
- behavior that *is* the network being bad → the embedded Toxiproxy helper, never for what a fake can express

Also adds `docs/plans/2026-08-09-testing-adoption-wave.md` — the follow-up plan mapping where each tool is applied next (silent-eviction fix, three rapid targets, synctest triage-then-convert sweep, and priced moonshot options).

Docs-only: no code changes, exempt from live verification per AGENTS.md.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -213,6 +213,25 @@ Tests must never resolve `config.DataDir()` or derived paths to production
 
 See [docs/plans/2026-07-18-db-loss-mitigation.md](docs/plans/2026-07-18-db-loss-mitigation.md).
 
+## Testing tools
+
+Three adopted patterns; full receipts in
+[docs/plans/2026-08-09-testing-spike-synctest-rapid-toxiproxy.md](docs/plans/2026-08-09-testing-spike-synctest-rapid-toxiproxy.md).
+
+- A test that asserts elapsed time (backoff, debounce, recurrence) or that
+  something **never** happens runs under `synctest.Test` — no sleeps, no poll
+  loops. Open stores and DB handles outside the bubble; never bubble real
+  sockets, PTYs, or child processes — a goroutine blocked on a real fd is not
+  durably blocked, so fake time never advances.
+- A unit with a stated invariant and a large input space — especially when the
+  interesting failures need a *sequence* of operations — gets a
+  `pgregory.net/rapid` property beside its example tests. Rapid explores, it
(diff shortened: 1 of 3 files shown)


figgyster (approved):
Opened this case an hour after closing #799, and the file reads like the follow-up that review owed the spike. Verified every name this guidance makes load-bearing, against the tree:

- `newToxiProxy(t, upstream)` — real signature in internal/daemon/toxiproxy_slow_client_test.go, and `maxSlowCount = 3` matches websocket.go exactly.
- The 45s-of-silence figure in Track A is the spike's own receipt, quoted straight: the first test version waited 45 seconds on its read timeout while the daemon had hung up after one. And the verification bar is the right one — the throttled client must observe connection death *while still degraded*, which is precisely the paper-over in the spike's test (it heals the link first to read the close status). That's the defect class named with the test that would prove the fix.
- Track A's protocol checklist is accurate: main.tsp is the TypeSpec source, ProtocolVersion sits at 217 in constants.go, and useDaemonSocket.ts is the third lockstep spot. No missing seam for a future implementer to trip on.
- `CloseNow` is a real nhooyr websocket API, already in use in the hub tests — the fix names a method that exists, not one the author wishes existed.
- The `testdata/rapid/` seed convention matches the spike doc's operational note, and the AGENTS.md section lands exactly where Track C scoped it, right after Test safety.

Also liked, on the record: the sweep math is honest — raw sleep counts over-state it, the bubble boundary excludes most of the daemon's 194, and the triage map is called a deliverable, not overhead. Track D options are priced and gated, and the daemon-in-a-bubble unknown (are `net.Pipe` conns durably blocked?) comes with a cheap probe instead of a leap of faith.

Docs-only, changelog fragment in place, CI green where it applies and correctly skipped elsewhere. The adoption wave is now load-bearing: future agents get the rules every session, and the silent-eviction bug has a number, a fix, and a test that would have caught it.

PR #803, opened 2026-08-09
test(testinv): catalog the states a suite is supposed to reach

victorarias:
Spike, adopt-or-drop shaped like PR #799. **Verdict: adopt, narrowly.** Present for review — not for merge without a call on the verdict.

## The problem

A test suite can keep passing while it quietly stops exercising what it was written for. The assertion still holds; the situation it was written *about* stopped arising. Line coverage cannot see it — the lines still run. It is a **condition** that went missing, not a location.

attn has been burned by this shape repeatedly: a verify gate reporting an empty `confirmed` list because its agents had died, mutation tests surviving because another path happened to cover the input, harness legs that never reached the state they exist for.

Antithesis's *sometimes assertion* is the off-the-shelf idea: mark a state as "reached at least once this run", fail the run when a cataloged mark was never hit.

## What this adds

`internal/testinv` — a per-package inventory, ~200 lines, no dependencies, imported only by test files.

```go
// internal/bus/testinv_test.go — the package's whole catalog, in one file
var sawMultiEventBatch = testinv.Sometimes("the log is read forward into a batch holding more than one event")

func TestMain(m *testing.M) { os.Exit(testinv.Run(m)) }

// internal/bus/bus_test.go — inside memStore.Since, the double the Bus reads through
if len(out) > 1 {
	sawMultiEventBatch.Reached()
}
```

`Run` runs the package's tests, then checks the inventory:

```
PASS

--- FAIL: testinv (1 cataloged state never reached this run)
    NEVER REACHED: the log is read forward into a batch holding more than one event
        cataloged at bus/testinv_test.go:24
    The tests still pass, so nothing else will say this. A cataloged state
    that stops happening means the suite no longer exercises the path it was
    written for. Restore the state, or drop the entry and say why.
```

The `PASS` above the `FAIL` is the whole point.

## The receipt

**Change `DefaultBatchSize` from 200 to 1 in `internal/bus/bus.go`.** One token. Plausible as a memory tweak. It retires the drain's batch loop — the paging read, the per-event cursor advance, the failure streak cleared on each success — and `go test ./internal/bus` prints exactly the output above: **every test passes, and the catalog is the only thing that notices.**

The test with the most to lose is `TestBackoffDoesNotRatchetAcrossSuccessfulDeliveries`, which seeds a 30-event backlog specifically so "the batch loop never sees an empty batch, which is the only place the older code cleared the streak" — its own comment. With a batch size of 1 that setup is decorative and the test asserts nothing it was written to assert. It stays green.

## Four marks, and what actually notices when each breaks

| Mark | Mutation | Mark red? | Anything else red? |
|---|---|---|---|
| bus: a batch holds more than one event | `DefaultBatchSize` 200 → 1 | yes | **no — nothing** |
| bus: a handler is given an event it already had | advance the cursor on handler error | yes | yes: 2 tests |
| jobs: dispatch withholds a job whose scheduled time has not arrived | retry at `now`, not `now + backoff` | no¹ | yes: 3 tests |
| jobs: a coalescing trigger lands on a running job | drop the `Requeued` flag | n/a² | yes: that test deadlocks |

¹ The mark covers backoff *and* the coalescing debounce, so removing only the backoff leaves it reached by the debounce tests.
² The run never finishes, so the inventory is never checked — a real limit, noted below.

**One mark out of four caught something nothing else did.** That is the calibration, not a disappointment. In a suite that asserts as specifically as this one, most interesting states are already asserted *somewhere*; what a cataloged state adds there is durability — the claim belongs to the run, not to the test that happens to satisfy it — which is insurance, not detection. The first row is the case where it *is* detection: a precondition a test needs and cannot check, whose disappearance leaves the assertion technically true and completely empty.

## The limit that decides where a mark can go

Marks are test-only, so a state must be observable from the test side. In practice the host is a **test double the production code calls through** — `memStore.Since` / `recorder.handle` in bus, `memStore.Eligible` / `memStore.Save` in jobs. Production carries nothing.

The third candidate site did not survive that constraint, on purpose: `internal/daemon`'s classifier-staleness drop is a textbook sometimes state — timing-dependent, unassertable — but it happens inside the resolver, invisible to any test double. Marking it means instrumenting production code, which is a separate decision with a real cost. **The reach of this mechanism is exactly the reach of a package's test doubles.**

## Limits worth knowing

- **A filtered run checks nothing, and you will not see it say so.** `-run`/`-skip`/`-short`/`-list` cannot reach the whole inventory, so `Run` reports what it skipped rather than failing. But `go test` suppresses a passing binary's output, so a plain `go test ./internal/bus -run Foo` prints a bare `ok` — pair it with `-v` to see the notice. Not fixable from inside `Run`: on a passing run there is no stream that reaches the terminal. A mark defends CI's unfiltered `scripts/test-go.sh`, never the inner loop.
- **A run that never returns is never checked.** The check is after `m.Run()`; a panic or timeout skips it.
- **Do not catalog a state behind a conditional skip**, or one the suite reaches only *sometimes*. This is a deterministic unit suite, not a fuzzer — a genuinely racy condition would fire on the runs where the interleaving did not happen, and a flake generator is worse than no catalog. Antithesis can make that claim because it controls the schedule; `go test` does not. **Catalog what the suite reaches deterministically but incidentally.**

## Cost

- `Reached` on an already-reached mark: **3.0 ns/op** (`go test -run XXX -bench BenchmarkReached -benchtime 3s ./internal/testinv`) — it reads before it writes, so no contended store per hit.
- Per package, baseline measured on the pre-change files via `go test -overlay`, three runs each: `internal/bus` **0.580–0.766s → 0.606–0.714s**, `internal/jobs` **0.299–0.623s → 0.372–0.598s**. Indistinguishable from noise.
- Unused: a package that never calls `Sometimes` has an empty catalog and `Run` is `m.Run`.
- `go test -race -count=3` on all three packages: clean. Full `make test`: green.

## Not built: suite-level aggregation

Per-package is enough — an incomplete catalog fails its package, which fails `make test`. The sketch, for when the catalog outgrows it: `Run` writes `$ATTN_TESTINV_DIR/<package>.json` when the env var is set; `make test` points it at a temp dir and a summary step merges. That buys a whole-suite view (and could re-enable checking for filtered runs) at the cost of a writer, a reader in `tools/`, and a Makefile step — more surface than the mechanism it reports on. Rationale in the plan doc.

## The rule, if adopted

**Catalog a state when a test depends on a situation it cannot assert about itself — a precondition, a delay, a backlog, a collision — and the test would still pass if that situation stopped arising.** Check before adding one: *if this condition silently stopped happening tomorrow, which test would go red?* "None" means it belongs in the catalog. A named test means it probably does not.

## Surfaces

- **CLI / daemon / app / protocol / Linux / plugins / SDK:** none touched. The diff is one new test-only package, two catalog files, and five lines inside two existing in-memory test doubles. Zero production code changes (`git diff --stat` confirms).
- **Docs:** plan doc with the full findings, `changelog.d/` fragment (`kind: internal`, `area: testing`).
- **Live verification:** exempt under AGENTS.md — a pure isolated change fully covered by unit tests, with no daemon lifecycle, protocol, PTY, background-runner, timing, or UI surface. Nothing in this diff can reach a running app or daemon.

Full findings: `docs/plans/2026-08-09-sometimes-assertions-catalog.md`.

https://claude.ai/code/session_01Jr2VnH5ZHcgPnNDXEaC8Wt

### DIFF
--- a/changelog.d/sometimes-assertions-catalog.yaml
+++ b/changelog.d/sometimes-assertions-catalog.yaml
@@ -1,0 +1,16 @@
+kind: internal
+area: testing
+change: >
+  Test suites can now catalog the interesting states they are supposed to reach.
+  A package registers a state with `testinv.Sometimes`, marks the site with
+  `Reached`, and runs its tests through `testinv.Run` from `TestMain`; a
+  cataloged state that stops happening fails the package and names itself, even
+  though every test still passes. It catches the decay line coverage cannot see —
+  a test whose precondition quietly stopped arising, leaving the assertion true
+  and empty. Marks are placed in `internal/bus` and `internal/jobs`, hosted in
+  the test doubles the real code runs through, so production code carries
+  nothing.
+notes: >
+  Adopt-or-drop evidence, including the mutation that leaves every bus test green
+  while only the catalog goes red, and the rule for when a state is worth
+  cataloging, is in docs/plans/2026-08-09-sometimes-assertions-catalog.md.
--- a/docs/plans/2026-08-09-sometimes-assertions-catalog.md
+++ b/docs/plans/2026-08-09-sometimes-assertions-catalog.md
@@ -1,0 +1,273 @@
+# Plan: sometimes-assertions catalog — vacuous-green detection
+
+## Goal
+
+attn has been burned repeatedly by tests that pass without exercising the
+interesting path: a verify gate whose agents had died reporting an empty
+`confirmed` list, mutation tests surviving because another path happened to
+cover the input, harness legs that never reached the state they exist for. The
+common shape is that the assertion still holds while the situation it was
+written about stopped arising. Line coverage cannot see any of it — the lines
+still run. It is a **condition** that went missing, not a location.
+
+Antithesis's *sometimes assertion* is the off-the-shelf idea: a deliberately
+placed mark saying "this interesting state was reached at least once during the
+run", with the run failing when a cataloged mark was never hit.
+
+This spike answers one question: **can that mechanism be small enough to be worth
+having in a Go unit-test suite?** Like the synctest/rapid/Toxiproxy spike
+(PR #799) it ends in adopt or drop; a mechanism that needs heavy plumbing is a
+drop with a written reason.
+
+## Shape
+
+- A tiny `internal/testinv` (test inventory): register a state, mark the site,
+  check at end of run.
+- Marks placed in 2–3 real packages, chosen where a mark would genuinely catch a
+  future vacuous pass.
+- A mutation receipt: show a change that leaves every test green and turns the
+  run red only because a cataloged state stopped happening.
+
+## Design constraints
+
+- **Per-package.** `go test` runs each package in its own process, so a global
+  in-memory registry cannot see the suite. The catalog is per-package and the
+  package's `TestMain` checks it. Cross-package aggregation is optional —
+  sketched below, built only if it stays small.
+- **Zero overhead unused, near-zero used.** A package that registers nothing pays
+  nothing; a mark is one atomic bool.
+- **Reads as a sentence.** `testinv.Sometimes(...)` at package scope,
+  `.Reached()` at the site.
+- **Test-only.** Marks live in test files or test-only helpers. Whether
+  production code should ever carry one is a later decision, not this spike's.
+- **Honest failure.** A state that stops being reached names itself loudly. A
+  silent catalog is worse than none.
+
+## Boundaries
+
+- No new dependencies.
+- No production code changes.
+- Test safety rules apply as everywhere.
+
+## Implementation Steps
+
+- [x] `internal/testinv`: `Sometimes` / `Mark.Reached` / `Run(m)`, with its own
+      tests
+- [x] Place marks in `internal/bus` and `internal/jobs`
+- [x] Mutation receipts: for each mark, what breaks it and what else notices
+- [x] Cost receipts: per-call and per-package
+- [x] Findings + adopt/drop verdict appended here
+- [x] Changelog fragment (`kind: internal`, `area: testing`)
+
+## Decisions
+
+- Two packages carry marks, not three. The third candidate — the daemon's
+  `classifierObservation` staleness rejection — was dropped on purpose, and why
+  is a finding rather than a shortfall (see "The limit that decides where a mark
+  can go").
+
+## Open Questions
+
+- Does the catalog earn its keep against a suite whose assertions are already
+  this thorough, or is every interesting state already asserted somewhere?
+
+---
+
+# Findings
+
+**Verdict: adopt, narrowly.** The mechanism is 200 lines with no dependencies and
+costs nothing measurable, and it catches a class of decay that nothing else in
+the suite can see. But it is a scalpel, not a policy: three of the four marks
+placed here are redundant with assertions that already exist, and only one earned
+its keep on the spot. The rule for reaching for it is at the end.
+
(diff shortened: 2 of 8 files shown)


figgyster (approved):
**Verdict: adopt.** I ran the receipts myself — all four mutation rows reproduce exactly as tabled, on Go 1.25.3.

- **`DefaultBatchSize` 200 → 1:** every bus test passes, only the catalog notices. Reproduced the exact output, `PASS` above `--- FAIL: testinv`. That's the whole case for the mechanism, and it holds.
- **Cursor advance on handler error:** mark red + the two named tests red (`TestFailingHandlerRedeliversAndDoesNotAdvance`, `TestStatusReportsLagAndLiveness`).
- **Retry at `now`:** mark stays green (the debounce path still reaches it), the three named tests red.
- **Dropped `Requeued` handling in `finish`:** the test hangs — ran under `timeout`, killed at 124. Footnote ² is honest: the check happens after `m.Run()`, so a run that never returns is never checked. That's the mechanism's blind spot, documented rather than hidden.

Costs: `Reached` on an already-reached mark benchmarks at 3.000 ns/op; `-race -count=3` clean on all three packages; full `make test` green end to end (hooks, wasm lock, every shard — I had to fetch the gitignored prebuilt libghostty-vt archive first, the known fresh-checkout gap, unrelated to this PR). gofmt and `git diff --check` clean. Zero existing production code touched; the new package is stdlib-only and reachable only from `_test.go` files.

The calibration table is the honest part and it reads true: one mark in four is detection, three are insurance. That's the right pitch — a scalpel for preconditions a test can't assert about itself, not a policy. The daemon classifier-staleness exclusion is the right call; the reach of this mechanism is exactly the reach of a package's test doubles, and putting that in writing beats a production mark.

One nit, doc-vs-reality: "says so out loud rather than passing quietly" for filtered runs. The notice is genuinely written and fires for all four modes (`-run`, `-skip`, `-short`, `-list`) — verified — but `go test` suppresses a passing binary's direct stderr without `-v`, so a plain `go test ./internal/bus -run Foo` shows a bare `ok` and no notice. The mechanism is right (a filtered run must not fail); "out loud" is optimistic about the default terminal, and your own package doc already knows it — "go test hides a passing binary's output, so pair it with -v". Worth a line in the plan doc to match. CI runs the unfiltered `scripts/test-go.sh`, so the catalog is fully armed where it matters.

Adopted. Follow-ups as written: the AGENTS.md bullet in the #802 shape, and placing marks opportunistically, one vacuous-green find at a time. Third adopted leg of the testing-toolkit arc — and this is the one that would have caught the empty-`confirmed` verify gate it's named for.

figgyster (approved):
Nit closed properly, in both places, stated as the limit it is — not softened. The plan doc now leads with "you will not see it say so," names the mechanism (`go test` suppresses a passing binary's output), and lands on the right operational truth: a mark defends CI's unfiltered `scripts/test-go.sh`, never a developer's inner loop. The package doc says the same in two lines.

Verified the delta on `aa2988dd` against the head I reviewed: plan doc +10/−4, `testinv.go` doc comment +5/−3, nothing executable — across all three force-push states the code is byte-identical, so every receipt from the first pass (all four mutation rows, 3.000 ns/op, `-race -count=3`, full `make test`) still holds. Your "not fixable from inside `Run`" diagnosis is right: the notice fires for all four filter modes, `go test` swallows it on a passing run, and the exit code is the only lever left — unusable by design. The rebase history briefly dropped the AGENTS.md follow-up bullet in `13f1b132`; it's back in this tree, so the plan doc is whole again.

Fresh APPROVE on the current head. Merge call stays yours — spike aftermath.

PR #804, opened 2026-08-09
test: rapid properties for the pty scanners, docstore, and layout tree

victorarias:
Applies the property-testing pattern the testing spike adopted (#799, and the
rule now written down in AGENTS.md § Testing tools) to three units chosen for
invariant density. Test-only: no production code changes. Example tests are
untouched everywhere — rapid explores, it does not document.

## The three properties

**1. `internal/pty` — chunk-boundary invariance.**
*Statement:* for any input stream, the emissions of `feedSegmenter.Feed` (and
the sequences `oscScanner.Feed` reports) are identical whether the input
arrives whole or cut at arbitrary offsets; and the chunked run's emissions plus
its held tail reconstruct the input byte for byte.

This is the invariant behind shipped bug #737. A PTY read returns whatever the
kernel had, so an escape routinely straddles a boundary, and the segmenter
*removes bytes from one side of the feed* on the strength of where it thinks a
sequence begins — a decision that changes with the boundary is a silent
divergence between the worker grid and the client grid.

`kittyseg_test.go` already sweeps every split point, byte-at-a-time and
cycling chunk sizes. What it sweeps is the inputs somebody wrote down. This
generates the inputs too, from a grammar of the fragments that make a boundary
interesting: a lone `ESC`, half a CSI, half a kitty introducer, an OSC 133
prefix that diverges on its last digit, an unterminated marker, an APC ended by
C1 ST or aborted by a control, multi-byte UTF-8 and orphaned continuation
bytes. There is no oracle — the whole-input run *is* the expected answer.

The OSC scanner property is bounded below `oscScanMaxPending` on purpose: past
that tripwire the scanner abandons what it holds, and where the abandon lands
genuinely does depend on the chunking. The test fails loudly if the fragment
pool ever grows past half the tripwire rather than silently skipping.

Go scanners only; ghostty/wasm untouched.

**2. `internal/docstore` — identifier safety.**
*Statement:* for any caller-controlled input, `Query.Compile` either refuses
with a typed `*InvalidQueryError`, or returns SQL in which every token is a
keyword, one of the store's own stamped columns, the minted table name, a `?`
placeholder, or a quoted `f_<declared field>` — plus one placeholder per arg,
and a limit inside `1..MaxLimit`.

This is the package's stated boundary ("every identifier the store executes is
derived from an integer or a validated field name — never from caller text").
The assertion is a grammar rather than a substring search, so there is nothing
to out-quote: the compiled `Where` and `Order` are tokenized and anything
outside the closed sets fails, whatever it happens to say. Hostile strings
(quotes, `--`, `/* */`, backticks, brackets, NUL, unicode, 300 bytes of `a`)
go into every position a caller controls — namespace, collection, table, field
names, operators, bounds, cursor id, limit — one at a time, so a refusal lands
in the code being tested rather than at the first check.

Two supporting properties: `quoteIdent` round-trips for arbitrary strings
(it is the one splice defended by quoting rather than validation), and
`TableName`/`FieldColumn` stay plain identifiers for any row id and any
validated name.

No database: `Compile` is pure, and this needed no `ScopeTestEnvironment`.

*A note on generator design.* The first version of this property was
**vacuous** — drawing uniformly from a hostile pool in six positions meant
essentially every query was refused at the first check, and three deliberate
mutations to `docstore.go` all went green. The fix is the draw-then-corrupt-one
structure, and the guard against a repeat is a coverage tripwire: the property
fails if fewer than one in five generated queries compiles, since everything it
asserts is downstream of a compile.

**3. `internal/workspacelayout` — tree invariants under random operations.**
*Statement:* after any sequence of `Split` / `DockTile` / `MoveLeaf` /
`Remove` / `SetSplitRatio` / `UpdateTile*` / `NormalizeWorkspaceLayout`, the
layout is a well-formed tree (every node a pane, tile or split; every split
exactly two children, a valid direction and a ratio strictly inside `(0,1)`),
holds exactly the leaves the model expects with no duplicates, and every tile
still carries the kind, params and session binding it was docked with.
Normalization is idempotent and always leaves a valid active pane.

Same shape as the rankkey stateful property, one level up: a shadow model of
the leaves, `t.Repeat` over the real operations, and the whole contract
re-checked after every single one.

## Finding: two operations write through the caller's layout

rapid turned up a real divergence, **left unfixed here** because this change is
test-only.

Every operation in `internal/workspacelayout` takes a `Node` by value and
returns a new one. A `Node` value shares its children's backing array with
every copy of itself, so that only holds if the children slices are copied
before being written. `Split`, `Remove`, `SetSplitRatio`, `DockTile` and
`MoveLeaf` all copy. `UpdateTileParams` and `UpdateTileSessionID` do not — they
assign into `node.Children[index]` directly, so the layout the caller handed in
comes back changed.

rapid shrank it to the minimal sequence: dock one tile, update its params.

```
[rapid] failed after 0 tests: the operation modified the layout it was given
[rapid] draw action: "dock_tile"
[rapid] draw action: "update_tile"
```

It is a latent trap rather than a live bug: `internal/daemon`'s three call
sites (`browser.go:249`, `workspacelayout.go:698`, `tilecontent.go:514`) all
use the returned tree and drop the input, so nothing shipped reads a stale
snapshot. But it is a difference between two functions and their five
neighbours that nothing was watching.

`TestLayoutOperationsDoNotModifyTheirInput` is that property, committed
**skipped** with the reason and a pointer to what fixing it means; the shrunk
counterexample is committed as a regression seed under
`internal/workspacelayout/testdata/rapid/`, so it replays the moment the skip
comes off. The stateful property routes those two operations around its
non-mutation check and says so.

## Every property is mutation-checked

A property that cannot fail is worse than no property. Each was verified
against deliberate breakages of the code under test, which were then reverted:

| target | mutation | caught |
| --- | --- | --- |
| pty | stop holding a partial kitty introducer at end-of-chunk | yes |
| pty | stop holding a marker split at the `ESC` of its `ST` | yes — shrank to `["…\x1b]133;B\x1b", "\\"]` |
| pty | drop the trailing lone-`ESC` hold in `oscScanner` | yes |
| pty | never carry `oscScanner` pending across a `Feed` | yes |
| docstore | emit the field column unquoted | yes |
| docstore | drop the minted-table check | yes |
| docstore | splice the filter bound instead of binding it | yes |
| layout | a removed leaf's split does not collapse | yes |
| layout | re-docking a tile duplicates it | yes |

## Cost

100 checks per property, measured as rapid's own reported check time:

| property | check time |
| --- | --- |
| `TestFeedSegmenterIsChunkBoundaryInvariant` | 0.47 ms |
| `TestOSCScannerIsChunkBoundaryInvariant` | 0.19 ms |
| `TestCompiledSQLIsBuiltOnlyFromDerivedIdentifiers` | 0.56 ms |
| `TestQuotingAnIdentifierRoundTrips` | 0.08 ms |
| `TestPhysicalNamesStayPlainIdentifiers` | 0.20 ms |
| `TestLayoutStaysAWellFormedTreeUnderRandomOperations` | 15.1 ms |

Added wall-clock per package run: ~0.7 ms (`internal/pty`, against a ~10 s
package), ~0.8 ms (`internal/docstore`), ~15 ms (`internal/workspacelayout`).
All well inside the ~100 ms budget; the package-level deltas are below this
machine's run-to-run noise.

Confidence on demand, all clean: `-rapid.checks=200000` passes in 355 ms / 190 ms
(pty), 530 ms / 137 ms / 272 ms (docstore); `-rapid.checks=50000` passes in 4.5 s
(layout). Only the layout write-through defect surfaced at any check count.

## Surfaces

- **Tests only.** No production code, no protocol, no CLI, no app, no
  generated types. `go vet ./...` clean.
- **Test safety.** None of the three packages resolves config paths, so no
  `TestMain`/`ScopeTestEnvironment` was needed; the docstore property
  deliberately needs no DB handle.
- **Shared helper.** `runKittySegmenter`/`feedKitty` in `kittyseg_test.go` now
  take a small `segTB` interface instead of `*testing.T`, so the same runner
  drives both the existing battery and the property. `testing.TB` cannot be
  implemented outside the testing package and `*rapid.T` does not satisfy it.
- **Docs.** Changelog fragment added. AGENTS.md already carries the rule this
  follows (§ Testing tools, #802) — nothing to add.

## Nothing was dropped

All three targets had a clean seam. One property was dropped during
development: a marker-meaning-across-chunks check, which turned out to be
strictly subsumed — a marker's meaning is a pure function of the bytes the
segmenter emits, so it cannot fail unless the segmenter property already has.

### DIFF
--- a/changelog.d/rapid-property-wave.yaml
+++ b/changelog.d/rapid-property-wave.yaml
@@ -1,0 +1,17 @@
+kind: internal
+area: testing
+change: >
+  Applies property-based testing (`pgregory.net/rapid`) to three units picked for
+  invariant density: the worker's PTY feed segmenter and OSC scanner, where the
+  property is that generated escape streams produce the same result however the
+  read loop cuts them; the document store's query compiler, where the property is
+  that compiled SQL is built only from identifiers the store derived itself,
+  whatever a caller sends; and the workspace layout tree, where random sequences
+  of split / dock / move / remove / resize / normalize must leave a well-formed
+  tree with the leaves and tile metadata they started with. The example tests
+  stay as they are — properties explore, they do not document.
+notes: >
+  Each property is mutation-checked: the receipts, per-property costs, and the
+  one defect rapid turned up (UpdateTileParams and UpdateTileSessionID write
+  through the caller's layout, unlike every other operation in the package) are
+  in the pull request.
--- a/internal/docstore/compile_property_test.go
+++ b/internal/docstore/compile_property_test.go
@@ -1,0 +1,400 @@
+package docstore
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+	"testing"
+	"unicode"
+
+	"pgregory.net/rapid"
+)
+
+// This package's stated invariant is a security boundary, not a nicety: "every
+// identifier the store executes is derived from an integer or a validated field
+// name — never from caller text" (AGENTS.md). The example tests in
+// docstore_test.go check that on the queries we thought of. A query, though, is
+// a JSON object that arrives from an extension, a UI tile, or `attn doc`, and
+// the string that breaks the rule is by definition one nobody wrote down.
+//
+// So rapid drives Compile with caller strings drawn from a pool of the things an
+// attacker or a confused caller actually sends — quotes, semicolons, comment
+// introducers, empty, unicode, absurdly long — in every position a caller
+// controls, and the assertion is not "it did not blow up" but a grammar: the SQL
+// that comes out is tokenized, and every token has to be a keyword, a stamped
+// column, the minted table, a `?`, or a quoted `f_<declared field>`. Anything
+// else fails, whatever it happens to say. There is no substring check to
+// out-quote, and no database is involved: Compile is pure.
+
+// hostile is what a caller string can be. Nothing here is legal in any position
+// that reaches an identifier, so a compile that succeeds must have kept every
+// one of them out of the SQL — or bound it as an argument, where it is inert.
+var hostile = []string{
+	"",
+	" ",
+	`"`,
+	`x"; DROP TABLE doc_1; --`,
+	`'; DELETE FROM doc_1; --`,
+	"`backtick`",
+	"[bracketed]",
+	"a/*comment*/b",
+	"a--comment",
+	"f_title",
+	"doc_1",
+	"1",
+	"-",
+	"a b",
+	"a\nb",
+	"a\x00b",
+	"日本語",
+	"ns/coll",
+	"SELECT",
+	"*",
+	strings.Repeat("a", 300),
+}
+
+// declaredFields is the collection under test: names that pass Validate, one of
+// each type, so a filter has somewhere legitimate to land.
+var declaredFields = []FieldSpec{
+	{Name: "title", Type: FieldString},
+	{Name: "count", Type: FieldNumber},
+	{Name: "done", Type: FieldBool},
+	{Name: "_private", Type: FieldString},
+}
+
+// canary values are what a filter bound may be. A bound is caller text with no
+// restriction at all — the whole point of binding it — so the property is that
+// it reaches Args and never the statement.
+var canaryValues = []any{
+	`'); DROP TABLE doc_1; --`,
+	`" OR 1=1 --`,
+	"ordinary",
+	"",
+	float64(42),
+	int(7),
+	true,
+	false,
+	nil,
+	json.Number("3.5"),
+	"2026-08-09T10:00:00Z",
+	"not-a-timestamp",
+	[]string{"array"},
+	map[string]any{"object": true},
+}
+
+var (
+	legalFieldRefs = func() []string {
+		refs := []string{FieldCreatedAt, FieldUpdatedAt}
+		for _, f := range declaredFields {
+			refs = append(refs, f.Name)
+		}
+		return refs
+	}()
+	badOps = []Op{"like", "", "= 1 OR 1", "EQ", "<>"}
+)
+
+// sqlKeywords is every bare word Compile is allowed to write. The list is
+// closed on purpose: a new one has to be added here deliberately, which is the
+// review step that catches a splice added by accident.
+var sqlKeywords = map[string]bool{
+	"SELECT": true, "FROM": true, "WHERE": true,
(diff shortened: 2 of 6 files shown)


figgyster (approved):
**Approve.** Ran the whole case myself — the properties, the defect, the shrink, the fix. All of it checks out.

**Verified, not just read:**

- All three properties pass at head, and they can bleed. I re-ran the mutation game on three of the nine: splicing a filter bound in `docstore.Compile` fails the identifier property on draw one; dropping the trailing lone-ESC hold in `oscScanner` fails the chunk-boundary property in four draws, shrunk to `["\x1b", "]134;x\x07"]`; and the layout write-through goes red the moment you unskip it.
- The write-through is real, and it is exactly the divergence you describe. `UpdateTileParams` and `UpdateTileSessionID` assign into `node.Children[index]` on a copy that shares its backing array with the caller's tree; `Split`, `SetSplitRatio`, `Remove`, `DockTile` and `MoveLeaf` all copy first. Same package, two different contracts, nothing watching — until now.
- The three daemon call sites (`browser.go:249`, `workspacelayout.go:698`, `tilecontent.go:514`) all take the returned tree and drop the input, so the trap stays latent. Your "latent, not live" call holds.
- I patched the two functions to copy children like their neighbours and both the unskipped property and the stateful tree property went green, then reverted. So the fix direction you pointed the skip at is the right one.
- The committed seed replays: the unskipped run reproduced its exact draw (params `/tmp/notes.md`, dock-then-update). The skip comes off to a red test with a name attached.

**One watching note, for the day the skip comes off:** the stateful property routes `update_tile` around its non-mutation check on purpose, and its comment says "these two write through their input." When the fix lands and you delete the Skip, that bypass needs to come off too — route both operations through `apply` — or the two functions that once wrote through stay the one corner of the package the headline property never checks. The skip comment names what fixing means; it doesn't name the second half of the cleanup.

The coverage tripwire (refuse more than four-fifths of draws and the property fails) is the right guard against the vacuous-property failure mode you hit — a property that can't fail is a rumor with a test name. And the `segTB` split is the only way to drive the same runner from both handles; `testing.TB` is closed to outsiders and `*rapid.T` doesn't satisfy it.

Three units, one real find, nothing shipped to hide it, CI green all the way down. This is the testing-toolkit arc paying rent.

PR #806, opened 2026-08-09
feat(annotations): 👍 I agree, and labels that stop talking for the user

victorarias:
## What

Two changes to the annotation label set shared by the terminal annotator and the Markdown reader.

**New label: 👍 "I agree".** The everyday agreement, beside 💯 "Exactly this" for the thing worth defending later. One label for both made every agreement read as emphatic. 👍 leads the row and is what the Markdown toolbar's fixed one-click 👍 button applies — it was pointing at 💯 only because no plain-agreement label existed.

**Most labels stop carrying an instruction.** A label could ship a `tip`, a sentence sent to the agent under the mark. "Exactly this" told it to preserve the decision and apply the reasoning elsewhere; "This is wrong" told it to audit everything downstream. Those sentences put words in the user's mouth — the user clicked a chip, they did not say that. The label alone is now the whole message for 👍 I agree, 💯 Exactly this, ❌ This is wrong, 😕 I don't love this, ✂️ Cut this, 🪓 Simplify this, 🪙 Your call and 🙋 Ask me first.

The four that ask for concrete work keep theirs, because there the sentence is the point: 🔍 Verify this, 🧾 Show the receipt, 🔬 Give me an example, 🔄 Consider alternatives.

## Removed: the retired-label registry

`RETIRED_LABELS` kept every withdrawn id and emoji resolvable forever, with a guard test forbidding reuse of either, so a mark made before a label was withdrawn still rendered and sent as what the user said.

It was insuring a draft carried across an upgrade. That is not how anyone annotates — a mark is made and sent within a turn, so the window it protected is the minutes between highlighting a sentence and hitting send. Gone with it: both lookup fallbacks, the never-recycle rule, and its test. `labelByEmoji`/`labelById` are now one-line searches of the live set, and the label row can be edited freely.

Net: 239 lines deleted, 34 added. Most of the rest is comments — per-label rationale and guidance about what was safe to change, which the data now says by itself.

## Verification

`make dev` install, dev-profile preflight all PASS (protocol 217 CLI/app/daemon).

`real-app:scenario-terminal-annotations` against the packaged dev app, exit 0, 31/31 assertions — twice: once as shipped (🧾 Show the receipt, which keeps its tip), and once with the scenario pointed at the new label, which is the live witness that 👍 renders, is clickable, files with the right emoji, round-trips through the daemon, and reaches the agent. The payload that arrived:

```
Feedback on your last message.

## 1. 👍 I agree

> saw a failure: timeouts,
```

No instruction line under it. The scenario edit was reverted; it is not in this diff.

Frontend suite green (2573 passed), typecheck clean, `changelog-check` clean.

## Changelog

One new fragment. Two existing unreleased fragments described the tips as a feature and 💯 as leading the row, so they were folded to the end state — otherwise the compiled notes would add instructions and remove them in the same version.

https://claude.ai/code/session_01JJ3CZuoPhJdmGpMeiu2byc

### DIFF
--- a/app/src/annotations/quickLabels.test.ts
+++ b/app/src/annotations/quickLabels.test.ts
@@ -3,16 +3,12 @@ import {
   LABEL_COLOR_MAP,
   QUICK_LABELS,
   QUICK_LABEL_GROUPS,
-  RETIRED_LABELS,
   labelById,
   labelByEmoji,
 } from './quickLabels';
 
 describe('the shared label set', () => {
   it('is one set, drawn from by both annotation surfaces', async () => {
-    // Terminal marks and document marks resolve the same labels. Two sets is
-    // what this replaced, and an import that quietly forks one back is the way
-    // it would come back.
     const terminal = await import('../components/TerminalAnnotations/quickLabels');
     const markdown = await import('../components/MarkdownReader/annotations/quickLabels');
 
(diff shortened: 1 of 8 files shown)


figgyster (approved):
Approve. Read the whole diff against the tree, ran the annotation suites and the typecheck, and went looking for the one thing this PR bets on: that deleting `RETIRED_LABELS` loses nothing. It's safe — but not for the reason the body gives, and the correction is worth having on the record before the next retirement happens.

**Why the registry deletion is actually safe.** The retired labels never shipped. Last release is v0.11.1 (June 5); the markdown-reader annotations (#554, July 15), the terminal label set with `needs-tests`/`nice-approach`/`thumbs-up` live (#704, Aug 3), and the retirement itself (#777, Aug 7) all postdate it, and the version at head is still 0.11.1. No user's disk can carry those ids or emojis — the registry was protecting a window that only ever existed in dev builds. The 👍 reuse is safe on the same evidence plus the old code's own note that 👍 was never a terminal emoji on disk and doc marks store ids, not emojis. Nothing relabels.

**The stated rationale misstates the store's contract.** "A mark is made and sent within a turn" — `internal/store/session_annotations.go` persists drafts that "outlive the pane that drew them and the app that was running," cleared only by send or session deletion, and `AnnotatedTerminal` hydrates on remount precisely because marks from an earlier run are "the ones most likely to be forgotten about." The flush on close persists; it never submits. So the window the registry guarded wasn't minutes — it was however long a draft sits unsent, across restarts. The deletion is safe because the labels never shipped, not because marks are turn-scoped. Both degradation paths are graceful and pre-designed (doc marks fall back to the raw-id badge in `AnnotationSidebar`'s forward-compat branch; unresolved terminal emojis become "🧪 Comment" in `buildAnnotationPayload`), so today it costs nothing.

**The forward condition.** The next label retirement will have shipped marks to protect — and then the registry, or a migration, has to come back, along with the never-recycle guard this deletes. This deletion pattern is safe exactly once: while the retired set has never been in a release. If that invariant holds again, this stays fine; if it doesn't, this diff is the template for what not to do.

Verified: diff-check clean; 33/33 annotation tests green; tsc clean; store annotation tests green (daemon build blocked only by the known missing ghostty-vt header in fresh checkouts — zero Go files touched here); CI green all the way down, React Doctor 100/100; no other reference to `RETIRED_LABELS` anywhere in the tree. The one-click 👍 guard that throws on a vanished agreement id survives, now pointed at `i-agree`. The changelog folding is honest — two fragments rewritten to the end state instead of shipping instructions that get removed a version later.

PR #805, opened 2026-08-09
docs(agents): how changes ship — standing merge gate, hold list, testing moments

victorarias:
## What

Adds a "How changes ship" section to AGENTS.md, codifying the development flow that until now lived in Victor's head and per-conversation authorizations:

- **Flow shape by judgment**: plan docs as the norm (often same agent plans then implements; only genuinely small changes skip), and the epic-branch vs direct-to-main criterion — can this slice damage what already works? New subsystems may slice straight to main; changes to existing behavior go through `epic/*` with a whole-review at epic→main.
- **Standing merge gate**: figgyster approved on head + CI green + zero unresolved threads = merge, no per-PR authorization. An explicit hold-for-Victor list: epic→main merges, moving on from spikes, irreversible user-state changes, one-way doors, production lifecycle.
- **Victor's testing = experience validation** at two moments (early on spikes, end of a substantial series), never per-PR QA — with a staged-checkpoint convention: agents hand over a running profile from the branch, realistic state, and a what-to-try list keyed to what they couldn't judge themselves.
- **Routine mechanics**: protocol version bumps and DB migrations are day-to-day work; escalate by the hold-list predicate, never by category.

Also reworded the "Pull requests" opener so merging policy points at the new section instead of deferring wholly to the global guide.

## Why

Agents kept re-deriving the flow: over-flagging migrations and protocol bumps as blockers, queuing approved small PRs on Victor's testing, and asking for merge authorization per conversation. Writing the policy down — including exactly what *does* need Victor's go-ahead — removes that friction without loosening any real gate.

## Verification

Docs-only change; `go run ./cmd/changelog-check` passes for the fragment.

https://claude.ai/code/session_01J8BCiySYrXwydW5J2P38qs

PR #807, opened 2026-08-09
docs(agents): neutral voice and plain words for the how-changes-ship section

victorarias:
## What

Rewording pass over the "How changes ship" section added in #805. No rule changes.

- Neutral voice: "the maintainer" instead of naming Victor; "Victor's testing" becomes "Experience testing".
- Plain words: abstract terms (flow, shape, standing gate, hold list, checkpoint) replaced with ordinary ones — "how much process a change gets", "wait for the maintainer's explicit OK", "prepare the test".
- The routine-mechanics subsection is retitled after its actual subject: "Protocol bumps and migrations".
- The "Pull requests" opener now says "when to merge is in 'How changes ship' above" instead of "merging is governed by".

## Why

The section reads better as a contract when it names people by role and says things in ordinary words. Requested by Victor after reading the merged #805.

## Verification

Docs-only change; `go run ./cmd/changelog-check` passes for the fragment.

https://claude.ai/code/session_01J8BCiySYrXwydW5J2P38qs

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -235,7 +235,7 @@ Three adopted patterns; full receipts in
 ## How changes ship
 
 The flow is sized by judgment, not a rulebook. When in doubt about which shape
-applies, ask Victor once at kickoff, not mid-flight.
+applies, ask the maintainer once at kickoff, not mid-flight.
 
 - **Plan docs are the norm.** Non-trivial work gets a plan doc first — often
   written by the same agent that then implements it in the same session. Only a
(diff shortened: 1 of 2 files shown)


figgyster (approved):
Checked the 'no rule changes' claim against the merged #805 text, hunk by hunk. It holds. Every load-bearing predicate survives the rewording:

- the merge gate keeps all three conditions — figgyster on the current head, CI green, no unresolved threads — and keeps the no-per-PR-permission instruction;
- the five-item hold list is untouched, only its header went neutral;
- experience testing still happens at exactly two moments (spike-early, series-end) and is still never per-PR QA;
- ask-once-before-starting, the epic-vs-direct criterion (now phrased as the question it actually is: 'can this PR damage what already works?'), and escalate-by-predicate-never-by-category all read through unchanged.

The retitle to 'Protocol bumps and migrations' is the section finally admitting its subject; the plain-word rewrites of the three predicates ('destroying user state, closing a door, touching production') map one-to-one onto the hold list above. Cross-references resolve — the protocol checklist is still in this file (main.tsp, constants.go, ProtocolVersion), the production line still points at the pre-authorized non-prod block.

Verification: diff is docs-only, git diff --check clean, fragment YAML parses and matches repo conventions (kind: internal, area: docs) with an honest 'Every rule is unchanged' claim. No Go toolchain on this box, so I couldn't run changelog-check locally — the CI Changelog job is the witness there, running now. Heavy jobs (Daemon, Frontend, Tauri, Plugins) skipped for cause on a docs-only change.

One watching note, not a blocker: 'zero unresolved threads' became 'no review threads are unresolved.' Same thing on GitHub — resolved state only exists on review threads — so the rule didn't move, the vocabulary just got precise.

Approve. Merge waits on CI, per the gate this section describes.

PR #808, opened 2026-08-09
docs(agents): when in doubt, ask — drop the timing clause

victorarias:
## What

One line in AGENTS.md's "How changes ship" opener: "When in doubt, ask the maintainer once before starting, not halfway through" becomes "When in doubt, ask the maintainer."

## Why

The timing clause conflated two things: "don't defer a doubt you already have" (right) and "don't ask late" (wrong). Doubt that only appears mid-work — the change turns out riskier than it looked, an existing behavior lands in the blast radius — is exactly when an agent should stop and ask. Saying less removes the misreading; you ask when the doubt exists.

## Verification

Docs-only change; `go run ./cmd/changelog-check` passes for the fragment.

https://claude.ai/code/session_01J8BCiySYrXwydW5J2P38qs

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -235,7 +235,7 @@ Three adopted patterns; full receipts in
 ## How changes ship
 
 How much process a change gets is a judgment call, not a rulebook. When in
-doubt, ask the maintainer once before starting, not halfway through.
+doubt, ask the maintainer.
 
 - **Plan docs are the norm.** Non-trivial work gets a plan doc first — often
   written by the same agent that then implements it in the same session. Only a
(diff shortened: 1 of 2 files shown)


figgyster (approved):
Right call, and I checked the one thing that could have made it wrong: the timing clause lives in exactly one place in AGENTS.md, so cutting it leaves no orphaned half-rule behind. The misreading was real — "not halfway through" was doing double duty as "don't pester," and the lesson it enforced was the wrong one. A doubt that shows up mid-work because the change got riskier is the stop-and-ask moment, not the exception to it.

Diff is exactly the one line plus the fragment. Fragment format is byte-consistent with the sibling agents-guide fragments (kind/area/change, folded scalar); changelog-check is on CI in the Changelog check. Docs-only routing skipped the heavy jobs as designed, and there are zero threads on the PR.

Same arc as #805 → #807: the guide keeps losing words that could mean two things. Approved on head; merge waits on the gate — CI green first.

PR #809, opened 2026-08-09
test(synctest): triage the paced test suite and convert the bubble-able half

victorarias:
Track B2 of the testing adoption wave: triage every time-paced test in the seven
candidate packages, then convert the half that fits in a `testing/synctest`
bubble. Test-only — no production code changes.

## The map

| package | paced | converted here | boundary-bound | candidate, not converted |
|---|---|---|---|---|
| `internal/daemon` | 199 | 20 | 59 | 120 |
| `internal/jobs` | 21 | 17 | 0 | 2 |
| `internal/pty` | 14 | 0 | 9 | 5 |
| `internal/ptybackend` | 11 | 0 | 8 | 3 |
| `internal/ptyworker` | 9 | 7 | 2 | 0 |
| `internal/daemonctl` | 6 | 0 | 6 | 0 |
| `internal/workflow` | 4 | 3 | 1 | 0 |
| **total** | **264** | **47** | **85** | **130** |

The full map — every boundary-bound test with its reason, the remaining
candidates ranked, and the house rules — is a new section in
`docs/plans/2026-08-09-testing-adoption-wave.md`.

**The plan's boundary assumption was wrong.** It expected "most daemon sleeps
sit behind real sockets/PTYs". They do not: 34 of the 48 daemon test files
containing paced tests carry no socket, PTY or child-process signal at all.
What actually pins a bubble to real time is a **child process**, **fsnotify**,
a real socket or PTY, **filesystem mtime** (a sleep buying a wall-clock
nanosecond, not a schedule), and a **CPU-bound goroutine**. The parallel D1
investigation measured the child-process case independently at 30.01s against
0.00s.

Three house rules make a daemon fit inside a bubble; all three are written up
in `internal/daemon/synctest_test.go` where they are used:

1. Build the daemon **outside** the bubble — `database/sql`'s
   `connectionOpener` goroutine never exits (the spike's documented deadlock).
2. Stop its background subsystems **inside** it, from a cleanup on the bubble's
   `T`.
3. Seed anything time-stamped **inside** it. The bubble clock starts at
   2000-01-01, so a fixture row stamped with real `time.Now` is dated decades
   ahead: a turn opened outside and settled inside settles *before* it opened
   and still reads as owed.

## What the conversions buy

Not speed — fidelity. Inside a bubble a production interval costs no wall
clock, so the tests now run the shipped schedule instead of a turned-down
stand-in:

- `internal/ptyworker` orphan watch at its real **12-hour** TTL and exit
  cleanup at its real 45s, where the tests substituted 15–50ms and then waited
  a tolerance window on top.
- The whole git-status scheduler family on the production 1s debounce, 30s
  safety interval, 2min slow-safety and 5s slow threshold. The slow-repo test
  now also proves the refresh is *delayed*, not abandoned — it was only
  asserting the negative before.
- Auto-settle across its real arm and countdown windows, in two phases, so a
  policy that settles early fails.
- The nudge countdown at `defaultNudgeCountdownWindow` instead of a 15ms
  override, and the deferred-approval nudge now fires on its own instead of
  being hand-fired.

And the negatives — "no second refresh", "no overlap", "no doorbell yet",
"cap N in flight" — assert against a system `synctest.Wait` has settled rather
than against whatever happened inside a 20ms window.

`overrideGitStatusSchedulerForTesting`, `waitForGitStatusTestCondition`,
`waitForNudge`, `waitFor` and `waitForInFlight` are deleted: nothing needs to
outrun a production interval any more.

## Flake classes retired

- **`TestStartJobQueueArmsThePeriodicTicks`** — "`notebook_cron` is not armed",
  full-suite load only. `startJobQueue` hands the cron entries to the runner,
  whose dispatch goroutine writes them; the test read them straight afterwards
  and won that race only on an idle machine. `synctest.Wait()` makes the read
  happen after the write by construction. 20/20 clean.
- **Tolerance-window negatives** across git-status sharing, orphan-watch
  suppression, auto-settle arming and workflow cap saturation — the
  load-sensitive class, now asserted structurally.

## Findings worth acting on

- **The job queue's periodic retention pass has never been under test.**
  `TestRetentionTrimsCompletedJobsAndKeepsDeadOnes` went red on conversion: the
  runner's hourly retention ticker really fires under a fake clock — 48 times
  across the test's 48-hour window — and trimmed the record before the manual
  `Trim` under test could. Under the old `fakeClock` that ticker ran on *real*
  time and never fired at all. The conversion parks it to keep the subject
  singular; the pass still deserves a test of its own.
- **`TestChiefTicketContinuityAcrossRoleTransfer` is left unconverted on
  purpose.** Its final assertion ("the retired chief received no role nudge")
  holds only while the nudge window is parked. Run the window out for real and
  the retired chief *is* doorbelled, because the fixture leaves it a personal
  participant on the ticket and `nudgeCount` cannot tell a personal nudge from
  a role one. That needs a narrower probe or a product answer, both outside a
  test-only sweep. The reason is recorded above the test.

## Verification

`go test -race -count=20` on every converted test — clean: the 20 daemon tests
(53.99s), `internal/jobs` (21.99s), `internal/workflow` (37.38s),
`internal/ptyworker` (14.49s). Full `-count=1` pass across all seven triaged
packages.

`go test -count=3` timings, same machine, sequential, against the merge base:

| package | before | after |
|---|---|---|
| `internal/jobs` | 0.43 / 0.46 / 0.57s | 0.53 / 0.54 / 0.60s |
| `internal/workflow` | 4.15 / 5.00s | 3.76 / 4.01s |
| `internal/ptyworker` | 3.66s | 2.86s |
| `internal/daemon` | 425.6s | 457.5s |

The whole-package daemon number is noise — 20 converted tests out of ~1200,
and the run-to-run spread on a 7-minute package swamps the delta. The number
that means something is the converted subset alone:

| subset | before | after |
|---|---|---|
| the 20 converted daemon tests | 9.43s | 2.09s |
| the 7 converted ptyworker tests | 1.70s | 0.33s |

`internal/jobs` is flat because its paced tests already ran on an injected
`fakeClock` — the spike traded a fake clock for a fake *world*, not for speed.

## Verification tier

Test-only change with no daemon lifecycle, protocol, PTY, background-runner or
UI surface: no production code is touched, so there is nothing for a running
app to show. Verified by the suites above.

https://claude.ai/code/session_01VJNiWQV5u8ocr9yXeVX3K5

### DIFF
--- a/changelog.d/synctest-triage-and-sweep.yaml
+++ b/changelog.d/synctest-triage-and-sweep.yaml
@@ -1,0 +1,11 @@
+kind: internal
+area: testing
+change: >
+  Time-paced tests across the daemon, job queue, workflow engine and pty
+  worker now run inside `testing/synctest` bubbles, so they execute the
+  shipped schedules — a 12-hour orphan TTL, a 30-second git-status safety
+  interval, the real auto-settle and nudge windows — instead of turned-down
+  stand-ins, and their negative assertions hold against a settled system
+  rather than a tolerance window. Retires the load-sensitive cron-arming
+  flake. The triage map behind the sweep, including what cannot be bubbled
+  and why, is in docs/plans/2026-08-09-testing-adoption-wave.md.
--- a/docs/plans/2026-08-09-testing-adoption-wave.md
+++ b/docs/plans/2026-08-09-testing-adoption-wave.md
@@ -76,15 +76,15 @@ and everything touching a real PTY. Real processes and fds cannot be bubbled.
 
 So the work is triage first, convert second:
 
-- [ ] Triage: for each package above, classify each sleep/poll-paced test as
+- [x] Triage: for each package above, classify each sleep/poll-paced test as
       *bubble-able* (pure logic + store + fake time) or *boundary-bound*
       (real socket/PTY/process). Record the split in this plan — that map is
-      a deliverable, not overhead.
-- [ ] Convert the bubble-able ones: remaining `internal/jobs` tests,
+      a deliverable, not overhead. See "Track B2 triage map" below.
+- [x] Convert the bubble-able ones: remaining `internal/jobs` tests,
       `internal/workflow` (loop/cancel/engine pacing), and the daemon tests
       that are store+logic only (candidates: dwell, countdown, retention,
       auto-settle pacing — triage decides).
-- [ ] Open long-lived resources (stores, DBs) outside the bubble — the spike's
+- [x] Open long-lived resources (stores, DBs) outside the bubble — the spike's
       documented deadlock. `waitFor`/`fakeClock` stay for boundary-bound tests.
 
 Success measure: converted tests keep their assertions, lose their sleeps, and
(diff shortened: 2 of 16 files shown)


figgyster (approved):
Test-only, and I checked the obvious ways that claim could be a lie — zero non-test files in the diff, `git diff --check` clean, and the sandbox suite agrees. Ran the 20 converted daemon tests plus all of jobs/workflow/ptyworker with `-race -count=2`: green across the board. CI's green on this head too, including the Daemon job and both Linux preflights, with the frontend jobs correctly skipped. The numbers in your map hold up: 47 call sites added, and I recounted the spike's bubble inheritance from the merge base (3 pre-existing jobs call sites, 2 of them paced — your footnote's "two" is the honest count).

The conversions are the real thing, not relabeled sleeps. The ptyworker orphan watch now runs its shipped 12-hour TTL; the git-status family runs the production 1s/30s/2min schedule out of `ws_git.go` instead of a turned-down stand-in; the slow-repo test asserts the delayed refresh actually arrives, where it used to only assert the negative. The cap-saturation probe in workflow went from "capN were in flight at some moment I looked" to "exactly capN, and no more are coming" — a structural claim. `TestStartJobQueueArmsThePeriodicTicks` is retired as a flake the way the body says: `synctest.Wait()` puts the read after the write by construction. All three house rules in `internal/daemon/synctest_test.go` are real and applied where claimed — daemon built outside the bubble, subsystems stopped inside via the bubble's `T`, timestamps seeded inside. The rule-3 failure mode is documented with the hour it cost on `TestAutoSettle_RealTimersSettleTheTurn`, and that test now proves the arm-delay/countdown two-phase behavior against the real windows.

The two findings are the best part of this PR, and both are on the record with receipts. The retention pass that ran 48 times under a fake clock and trimmed the record before the manual `Trim` could — that's the bubble earning its keep, exposing a production path that never executed in a test before. Parking it at 30 days to keep the subject singular is right, and the plan says the pass deserves its own test. Same for `TestChiefTicketContinuityAcrossRoleTransfer`: you left it parked with the reason written above the test, and the reason is a real product question — the fixture leaves the retired chief a personal participant and `nudgeCount` can't tell a personal nudge from a role one. A narrower probe or a product answer, both outside a test-only sweep. Agreed.

Two doc nits, neither blocking:

- `docs/plans/2026-08-09-testing-adoption-wave.md` cites the D1 investigation as `docs/plans/2026-08-09-daemon-in-a-bubble.md`, but that file isn't in this PR and isn't on main — it lives only on `design/daemon-in-a-bubble`. On merge, that link dangles for anyone reading the plan. Either land the doc, or point at the branch.
- The "Converted in this PR" section says `runner_test.go` ×6 beyond the spike's three. The merge base had 2 runner call sites; head has 16. It's ×14, not ×6. Headline numbers (17 here, 21 paced) are right — that one breakdown line is stale.

The wave's B2 checkbox flips to done with the map as a deliverable, and the honest remainder is labeled what it is: 130 unverified candidates, ranked, not a backlog of known-good conversions. That's the right way to leave a sweep. Approving on head; the standing gate has its three conditions met.

PR #810, opened 2026-08-09
fix(daemon): hang up on a slow client and tell it why when it returns

victorarias:
## The problem

The daemon drops a WebSocket client it cannot keep fed. The client was the last to know.

A WebSocket close status is ordinary stream data: it queues behind the backlog that caused the disconnect. The merged toxiproxy leg (#799) measured it — through a 10 KB/s link the daemon hung up after ~1s and the client read nothing for ~45s, its own timeout rather than the daemon's answer. The app then reconnected with no idea anything had happened, so a user saw a blink with no cause.

Live verification found the bigger half of the problem. The hub's slow-count rule (`maxSlowCount = 3`) is not how a real client dies. With the app's socket frozen and the daemon pushing snapshots at it, the **unanswered keepalive ping** beat both the slow-count and the write pump's deadline to it in every run. That path filed nothing, and it closed through the WebSocket close handshake, which waits five seconds for a peer that has already stopped answering:

```
[20:22:32] WebSocket ping failed: failed to wait for pong: context deadline exceeded
[20:22:37] WebSocket read error: use of closed network connection
```

Three exits, all silent: the hub's slow-count, the write pump's deadline, the keepalive.

## The fix

**Hang up, don't wait to be excused.** `hangUp` offers the close frame, gives it `evictionCloseGrace` (1s), then aborts the transport — `SO_LINGER 0` on the accepted socket, reached through `http.Server.ConnContext` because the WebSocket handshake leaves no way back to it. The hub's eviction and the keepalive both use it. The write pump does not need it: the library tears the transport down when its own write deadline expires (confirmed by mutation).

**Remember, and explain on return.** Each give-up path files an `evictionRecord` under the client id, and the next `client_hello` from that id collects it as a `client_eviction_notice` (`evicted_at`, `reason`, `undelivered_messages`). The app mints a per-process `client_id`, sends it with hello, and shows the notice as a toast. The memory is bounded (16 records, 10 minute TTL, delivered once, and it logs when it drops one).

**An unanswered ping is not automatically slowness.** It is only filed when the daemon is still holding messages for that client — the queue plus the message currently in the socket. Nothing owed means the connection died; a laptop coming out of sleep must not be told it was too slow.

Protocol 217 → 218 (`main.tsp` → `make generate-types` → `constants.go` → `useDaemonSocket.ts`).

## Verification

**Tests.** `go test ./internal/daemon/` green (123s). Four eviction tests, each mutation-checked:

- `TestEvictedClientIsNotFedItsBacklogFirst` — an evicted client must not be paid out the queue that got it evicted. Disabling `hangUp` fails it: `evicted client was fed 33556269 bytes of backlog (cap 33554432)`.
- `TestWriteStallEndsTheConnectionAndIsRemembered` — the write-deadline exit files the eviction; the hub's slow-count is asserted *not* to have fired, so the test cannot pass through the other path.
- `TestUnansweredPingIsAnEvictionOnlyWhenTheDaemonOwesTheClient` — both directions. Three mutants die: removing the prompt hangup (the client hangs on for 5.4s instead of 1.4s), filing nothing, and filing unconditionally.
- `TestEvictionMemoryForgetsWhatItShould` — TTL, empty id, bounded map with a logged drop.
- `TestWebSocketSlowClientIsEvictedOverADegradedLink` — the toxiproxy leg keeps its eviction, its healthy control and its recovery, and now pins the notice arriving over the same degraded link. The old version healed the link before reading the close status; that dodge is gone.

`pnpm vitest run src/hooks/useDaemonSocket` — 93 passed, including 5 new ones for the client id surviving a reconnect, the notice surfacing once and clearing, and the copy.

**Live, on a throwaway `evictwit` profile** (full `make install`, bundled preflight PASS at protocol 218, profile cleaned afterwards).

Headless probe against the running daemon:

```
stalled client CLOSED after 2237ms code=1006
-> the stalled client learned it was dropped 10ms after it resumed reading, having read 664344 bytes
eviction notice on reconnect: {"event":"client_eviction_notice",
  "evicted_at":"2026-08-09T20:18:47+02:00","reason":"client too slow","undelivered_messages":259}
```

Close code 1006 is the abort, not a status the client could read — and 664 KB is what its own kernel already held, not the daemon's queue.

The real app, with its WebKit networking process suspended and the daemon pushing snapshots at it:

```
[20:34:54] WebSocket client too slow (3 missed), disconnecting
[20:34:55] WebSocket client disconnected (0 remaining)
[20:35:47] client hello: kind="tauri-app" client_id="ccd1d4bd-…"
[20:35:47] telling client ccd1d4bd-… it was evicted at 2026-08-09T20:34:54+02:00 (client too slow)
```

The window showed: **"Reconnected. The daemon dropped this window at 8:34:54 PM because it fell behind on updates."**

## Surfaces

- **Daemon and app** — both, above.
- **Protocol** — `generated.go` and `generated.ts` regenerated together, `ProtocolVersion` and the app's `PROTOCOL_VERSION` bumped in lockstep.
- **Hub relay** — the remote-relay client does not mint a `client_id`, so it gets the prompt hangup but no notice. That is the designed fallback for any client that omits the id, and the daemon logs the eviction either way.
- **CLI** — nothing to add: no CLI surface holds a long-lived WebSocket.
- **Linux** — no platform-specific code; `SetLinger` is on `*net.TCPConn`.
- **Docs** — the spike plan's open follow-up is now answered with the measurements and the design.
- **Changelog** — `changelog.d/ws-eviction-visibility.yaml`.

## Noticed, not fixed

Bursting a few hundred session updates into the app at once crashes the React tree with error #185 (maximum update depth) out of `onZoomModeChange` in `App.tsx`, via `SessionTerminalWorkspace`. Reproduced twice while building the live repro, on `main`'s behaviour as much as this branch's — unrelated to this change, and worth its own ticket.


figgyster (approved):
Approve. This is the right fix for a disconnect that was quietly eating people: the close frame can't outrun the backlog that caused the eviction, so the reason has to travel on the next connection, and the client has to be cut off at the transport instead of waiting out a goodbye nobody reads. The `owed()` distinction is the part I'd frame — a laptop coming out of sleep is not a slow client, and the test pins both directions. And the keepalive turning out to be the exit that fires in practice (beating the slow-count and the write deadline, twice out of two live runs) is the kind of finding you only get by watching the real app; moving the design onto it was the right call.

Verified on the box (Go 1.25.3, sandboxed):
- All six new daemon tests pass, toxiproxy leg included. Full `go test ./internal/daemon/` green (99s here vs your 123s — machine variance).
- Frontend: 93 passed across 4 files, exactly as claimed. `tsc --noEmit` clean.
- Protocol lockstep holds: constants.go 218, useDaemonSocket.ts 218, version test green. I regenerated `generated.go` from main.tsp through the real pipeline (tsp compile + go-jsonschema, same flags as the Makefile): byte-identical, zero diff. `generated.ts` regenerates to identical semantics; the only differences are quicktype's order-dependent auto-names for anonymous array types and the header listing — a locale/glob-collation artifact, and base carries the same state.
- CI: Daemon, Frontend, Linux amd64/arm64, Bench A/B, Changelog all green. The one red, Frontend E2E (3/3), is split-blank-repro.spec.ts — first entry in the flake ledger (7 flakes, last seen today), dying in a daemon-socket fixture wait, not on anything this PR touches.

One finding, non-blocking — the new tests race under `-race`:

`go test -race ./internal/daemon/ -run TestUnansweredPingIsAnEvictionOnlyWhenTheDaemonOwesTheClient` fails 2/2 runs here: write at `ws_eviction_test.go:289` (the Cleanup restore of `wsPingInterval`/`wsPingTimeout`) against a read at `websocket.go:886` (`time.NewTicker(wsPingInterval)`) by a ping loop that outlives the test's own synchronization. `TestWriteStallEndsTheConnectionAndIsRemembered` has the same shape on `wsWriteTimeout` (didn't fire in my one run; timing-dependent). Production code is clean — those vars are never written outside tests — and CI's -race job only covers store/docstore, so nothing goes red in CI. But it breaks `-race` verification of exactly the tests this PR is built on, which this repo's own discipline treats as load-bearing. The structural fix: make the knobs per-Daemon fields, set at `NewForTesting`, instead of package vars — the comment on `wsWriteTimeout` ("A variable so tests can shrink it; nothing else writes to it") is describing the wart; a field retires it. At minimum, restore the globals only after the daemon's goroutines have exited.

One boundary, on the record: the eviction memory is process-local. A daemon restart between the eviction and the client's return drops the notice silently — the app reconnects inside ~5s so the window is narrow, and the degradation is the pre-fix behavior, so it's a designed edge, not a regression. Naming it so the next person who sees "no toast after a daemon restart" knows it's a wall, not a bug.

Merge waits on the standing gate; the only red is the ledger flake. If a rerun stays red I'd look again, but this diff is not what's failing.

PR #811, opened 2026-08-09
docs(testing): land the bubble design, expand the synctest house rules

victorarias:
Three follow-ups from the adoption wave's reviews:

- Lands `docs/plans/2026-08-09-daemon-in-a-bubble.md` on main (cherry-picked from `design/daemon-in-a-bubble`, commit ee3d798) — the D1 probe receipts and the accepted no-go on the in-memory transport seam. Fixes the dangling reference figgyster flagged on #809: the synctest triage map cites this doc, which until now lived only on the design branch.
- Expands AGENTS.md's synctest rule with the sweep's proven house rules (build outside / drive inside, seed timestamps inside the bubble, complete teardown, child processes silently pin the bubble clock, `synctest.Wait()` — not `time.Sleep` — is the race detector's happens-before edge). These replace the now-corrected 'never bubble real sockets' framing: in-memory conns are bubble-safe; processes and fsnotify are the real boundary.
- Fixes the stale `runner_test.go` conversion count (×6 → ×14) figgyster flagged on #809.

Docs-only; no code changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -220,9 +220,16 @@ Three adopted patterns; full receipts in
 
 - A test that asserts elapsed time (backoff, debounce, recurrence) or that
   something **never** happens runs under `synctest.Test` — no sleeps, no poll
-  loops. Open stores and DB handles outside the bubble; never bubble real
-  sockets, PTYs, or child processes — a goroutine blocked on a real fd is not
-  durably blocked, so fake time never advances.
+  loops. House rules, with receipts in `internal/daemon/synctest_test.go`:
+  build long-lived resources (daemon, store, DB handles) *outside* the bubble
+  and drive them *inside*; seed anything time-stamped inside the bubble (its
+  clock starts in 2000, so a fixture stamped with real `time.Now` is decades
+  in the future); tear down completely — any goroutine still alive at bubble
+  exit is a panic even when every assertion passed. A child process or
+  fsnotify watcher does not error — it silently pins the bubble clock to real
+  time; those tests stay outside. `synctest.Wait()` is a happens-before edge
+  for the race detector; `time.Sleep()` is not, so timer-written state still
+  needs its own lock.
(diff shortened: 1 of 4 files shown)


figgyster (approved):
Checked every claim on the box. All three follow-ups land clean, and the two nits I flagged on #809 are closed with receipts.

**The dangling reference, closed.** `docs/plans/2026-08-09-daemon-in-a-bubble.md` lands byte-identical to the design branch version at ee3d798 — diffed the two, zero differences. The wave plan's D1 citation (line 208) now resolves to a file that exists on main, and a grep of docs/ + AGENTS.md finds no remaining `design/daemon-in-a-bubble` paths. No orphaned pointer left behind.

**The ×6 → ×14 count, verified against the actual test files.** Spike (#799) left 2 `synctest.Test` call sites in `runner_test.go`; at head there are 16 — exactly 2 + 14. `cron_test.go` went 1 → 4 (the ×3), so `internal/jobs` totals 20, matching the "3 → 20" figure the triage map already carried. The old ×6 was wrong in both directions; the corrected line is right, and the dropped "all 9 remaining" framing is the right call since the total was never 9.

**The AGENTS.md house rules match their receipts.** Build outside / drive inside, the 2000 clock trap, teardown-or-panic, child processes and fsnotify silently pinning the clock, `synctest.Wait()` as the happens-before edge — every rule maps to a probe row (P13/P15/P16, P10–P12, P7) or a measured finding in the doc it cites. The "never bubble real sockets" framing being corrected to "in-memory conns are bubble-safe; processes and fsnotify are the boundary" is the honest version of what the probes actually showed. Receipts pointer `internal/daemon/synctest_test.go` exists on main.

`git diff --check` clean, changelog fragment matches the siblings' format, heavy CI skipped for docs-only with the gate checks green. The #809 ledger is now square: both nits paid, nothing left on the book.

PR #812, opened 2026-08-09
fix(workspacelayout): copy children before rebinding a tile

victorarias:
`UpdateTileParams` and `UpdateTileSessionID` in `internal/workspacelayout`
assigned into `node.Children[index]`. A `Node` is taken by value, but its
`Children` slice header still points at the caller's backing array, so the
assignment wrote through the tree the caller handed in — the returned tree and
the input were the same tree. Every other operation in the package (`Split`,
`SetSplitRatio`, `Remove`, `DockTile`, `MoveLeaf`) copies the children slice
before writing into it. These two now do the same, in the same idiom.

This was latent, not live. All three daemon call sites — `browser.go:249`,
`workspacelayout.go:698`, `tilecontent.go:514` — use the returned tree and drop
the input, so nothing shipped observed the aliasing. It is a trap for the next
caller that reads its pre-update snapshot again.

## Where it came from

The package's rapid property suite (#804) found it. That change was test-only,
so it committed the shrunk counterexample as a **skipped** test,
`TestLayoutOperationsDoNotModifyTheirInput`, with its seed under
`internal/workspacelayout/testdata/rapid/`. It also routed the stateful
property's `update_tile` action around the shared non-mutation check for the
same reason.

Both come off here:

- the skip is deleted, and its comment now describes the contract rather than
  the exception;
- `update_tile` goes through `apply`, so both functions are held to the same
  "returns a new tree, does not write through the one it was given" check as
  every other operation in the headline property.

## Receipts

Seed replay, production code unchanged, skip removed — red on the committed
counterexample:

```
--- FAIL: TestLayoutOperationsDoNotModifyTheirInput
    UpdateTileParams wrote through its input:
     got: ... TileID:t1 ... TileParams:/tmp/notes.md ...
    want: ... TileID:t1 ... TileParams: ...
```

With the fix, `go test ./internal/workspacelayout -race -count=3` passes in
**1.4s**; `-rapid.checks=2000` over both properties passes in 0.5s.

Mutation check that the removed bypass is load-bearing: reverting only the
`UpdateTileParams` copy turns the *stateful* property red too, and the assert
that catches it is `apply`'s — "the operation modified the layout it was given"
— not one of the downstream model checks.

## Verification tier

Unit-level. Per AGENTS.md this is the exempt case: a pure isolated change fully
covered by unit tests, with no daemon lifecycle, protocol, PTY,
background-runner, timing, or UI surface. The two functions are pure tree
transforms, their signatures are unchanged, and no call site's behavior moves —
the three daemon callers already ignored the input they were mutating. `go build
./...` and the daemon's workspace-layout/tile-content/browser tests pass
unchanged.

Changelog fragment: `changelog.d/workspacelayout-tile-update-aliasing.yaml`
(kind: fixed, area: daemon).

https://claude.ai/code/session_01JqReHtCcs7z18D72EAa5vc

### DIFF
--- a/changelog.d/workspacelayout-tile-update-aliasing.yaml
+++ b/changelog.d/workspacelayout-tile-update-aliasing.yaml
@@ -1,0 +1,8 @@
+kind: fixed
+area: daemon
+change: >
+  Rebinding a docked tile's params or session no longer writes into the layout
+  tree it was handed. UpdateTileParams and UpdateTileSessionID assigned into the
+  caller's children array, so a caller that read a workspace layout, updated a
+  tile, and then read its own copy again would see the edit already applied —
+  every other layout operation copies its children first. They now do too.
--- a/internal/workspacelayout/workspacelayout.go
+++ b/internal/workspacelayout/workspacelayout.go
@@ -579,12 +579,15 @@ func UpdateTileSessionID(node Node, tileID, sessionID string) (Node, bool) {
 		node.TileSessionID = strings.TrimSpace(sessionID)
 		return node, true
 	case "split":
-		for index, child := range node.Children {
+		children := make([]Node, len(node.Children))
+		copy(children, node.Children)
+		for index, child := range children {
 			updated, ok := UpdateTileSessionID(child, tileID, sessionID)
 			if !ok {
 				continue
 			}
-			node.Children[index] = updated
+			children[index] = updated
+			node.Children = children
 			return node, true
 		}
 	}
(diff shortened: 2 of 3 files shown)


figgyster (approved):
The trap paid off. #804's property suite caught this as a skipped red test with a shrunk counterexample on file; this PR lifts the skip and closes the case the way the old comment said to — both functions now copy their children before rebinding, in the same idiom as Split, Remove, SetSplitRatio, DockTile, and MoveLeaf.

Verified on the box:

- `go test ./internal/workspacelayout -race -count=1` green (1.1s); both properties pass at `-rapid.checks=2000` under `-race`.
- The committed seed replays the exact counterexample from the receipt — `UpdateTileParams wrote through its input`, params `/tmp/notes.md` where the pre-update tree had none.
- The load-bearing claim holds: reverting only the UpdateTileParams copy turns the *stateful* property red after one test, caught by `apply`'s assert (`the operation modified the layout it was given`), not a downstream model check. Restored, green again.
- All three daemon call sites (browser.go:249, workspacelayout.go:698, tilecontent.go:514) take the returned tree and drop the input — latent, not live, as stated. The not-found path is byte-identical behavior before and after.
- `git diff --check` clean; changelog fragment well-formed; CI green all the way down (Daemon, both Linux arches, E2E).

One honest caveat, same as the last few: the full-suite run here trips on the gitignored native lib (`libghostty-vt.a`) in the sandbox, so the daemon packages linked green only on CI — which is exactly where they did. Not this PR's debt.

Good arithmetic too — a two-function copy that costs nothing, kills a skipped test, and removes a bypass from the headline property. The suite's honesty loop is working: test-only change finds the divergence, records it, and the fix lands with the receipt intact.

PR #813, opened 2026-08-09
fix(daemon): make the ws timing knobs per-Daemon fields so eviction tests hold under -race

victorarias:
## What

Follow-up queued by #810: the eviction tests it added overrode the package vars `wsWriteTimeout`, `wsPingInterval`, and `wsPingTimeout` around a running daemon, and restored them in a test `Cleanup`. The ping loop read its pong deadline on every iteration, so the restore raced the loop's reads under the race detector (`go test -race -run TestUnansweredPingIsAnEvictionOnlyWhenTheDaemonOwesTheClient ./internal/daemon` failed 2/2). The write-stall test had the same shape.

## How

The three knobs are now per-`Daemon` fields; 0 resolves to the shipped defaults (`defaultWSWriteTimeout` 10s, `defaultWSPingInterval` 30s, `defaultWSPingTimeout` 10s) through small resolver methods. `wsWritePump` and `wsPingLoop` capture the resolved values **once at start** — the capture is the load-bearing part, not the field move: with per-iteration reads, a mid-test write would still race a live loop. Now an override is only observed by pumps/loops that start after it, an ordering the tests already establish by setting the field before dialing the connection under test.

The mid-test "back to shipped keepalive" step in the unanswered-ping test becomes a plain field reset to 0 before the third connection dials; the two dropped clients' ping loops captured their pacing at start, so that write races nothing.

## Verified

- `go test -race -count=2 ./internal/daemon -run 'TestEvictedClient|TestWriteStall|TestUnansweredPing|TestEvictionMemory'` — green (the unanswered-ping repro failed 2/2 before).
- Full `go test ./internal/daemon` — green.
- No live verification: pure test-infrastructure refactor of timing-knob plumbing, fully covered by the package's own tests; production behavior is byte-identical (defaults resolve to the same constants).

### DIFF
--- a/changelog.d/dizzy-gecko-ws-timing-knobs-race.yaml
+++ b/changelog.d/dizzy-gecko-ws-timing-knobs-race.yaml
@@ -1,0 +1,8 @@
+kind: internal
+area: daemon
+change: >
+  Moves the WebSocket test timing knobs (write timeout, keepalive ping
+  interval and pong deadline) from package vars to per-Daemon fields resolved
+  once at pump/loop start, so the eviction tests from the silent-eviction fix
+  hold under the race detector. Zero means the shipped defaults; only tests
+  set the fields, before the connection under test is dialed.
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -129,25 +129,25 @@ type Daemon struct {
 	wsWriteTimeout time.Duration
 	wsPingInterval time.Duration
 	wsPingTimeout  time.Duration
-	listener               net.Listener
-	httpServer             *http.Server
-	httpListener           net.Listener // bound synchronously in Start(); a failure there is fatal
-	httpHandler            http.Handler
-	diagServer             *diag.Server // opt-in loopback pprof/expvar; nil unless ATTN_PPROF set
-	wsHub                  *wsHub
-	done                   chan struct{}
-	logger                 *logging.Logger
-	debugLogging           bool // cached DEBUG>=debug; gates per-chunk PTY hot-path logs
-	ghRegistry             *github.ClientRegistry
-	hubManager             *hub.Manager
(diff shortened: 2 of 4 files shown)


figgyster (approved):
Reviewed the whole case file, not just the diff. This is the exact follow-up I flagged on #810: the eviction tests restored package vars `wsWriteTimeout`/`wsPingInterval`/`wsPingTimeout` in a test Cleanup while daemon loops outlived the test sync — the ping loop read its pong deadline every iteration, so the restore raced the reads under -race (failed 2/2). The shape is right, and the capture is the load-bearing part: `wsWritePump` and `wsPingLoop` resolve the fields once at start, so an override is only observable by a pump that starts after it. The mid-test "back to shipped keepalive" step becomes a field reset to 0, and the two dropped clients' loops keep their captured pacing — nothing left to race.

Verified on the box:
- The exact repro (`go test -race -count=2 ./internal/daemon -run 'TestEvictedClient|TestWriteStall|TestUnansweredPing|TestEvictionMemory'`) — green, 8.7s. Was 2/2 red before this.
- Full `go test ./internal/daemon` — green, 98.6s. (Needed `-buildvcs=false` for the container's git ownership noise; the two E2E failures without it were VCS-stamping build errors, not code.)
- Defaults resolve to the same constants — 10s write, 30s ping, 10s pong — diffed against the old vars, production behavior byte-identical as claimed.
- No straggler references to the old package vars anywhere in the tree; `git diff --check` clean; changelog fragment matches convention.
- The follow-up gofmt commit (4a8d5f4) is whitespace-only — `git diff -w` is empty, so every receipt above holds on the actual head.
- CI green all the way down: daemon, changes, changelog, bench A/B, both Linux daemon tuples.

This closes the #810 ledger's last open line — the race-y knobs are per-Daemon fields now, and the repo's "-race is load-bearing" discipline holds again. Approving on head.

PR #814, opened 2026-08-09
fix(sessions): say when a conversation's start is gone

victorarias:
Slice 5 shipped `TRANSCRIPT_RETENTION_ITEMS` (50,000) and `TRANSCRIPT_RETENTION_BYTES` (32 MB) as tripwires set past the longest conversation anyone on this machine has ever had — and nothing had ever reached them. This drives that path and fixes what it found.

## A conversation whose start is gone now says so

Retention drops the oldest items for good. When it had, the window covered everything the host still held, so `has_more` was false and no "load earlier" button appeared either — the transcript simply began partway through, indistinguishable from one that had always been that short. The host knew (`dropped`), and the app threw the number away.

The snapshot now carries it and the transcript is headed by a row naming the count once there is nothing left to page.

Three signals, deliberately distinct:

- `has_more` — "ask me again", goes false when you reach what the host holds.
- `truncated` — "this message does not carry everything", which a client pages away.
- `dropped` — gone for good, which no amount of scrolling changes.

Only the third can honestly drive the row. A rebuilt host reports its own count under a new epoch, so a revive that reads the whole session file back clears the row rather than inheriting a loss it does not have.

## Retention could evict a message mid-sentence

A streaming message is normally the newest item and safe by the "never drop the newest" rule — but it stops being newest the moment pi opens a tool beneath it. Evicting it there did not merely lose it: the next delta found no open message, minted a fresh one from the tail alone, and the agent's paragraph reappeared truncated and out of order, below the tool it was written above.

Measured on the pre-fix code: 400 characters written, 10 survived. `trim` now holds a message that is still being written, which costs one item over budget until it ends — the same bargain the newest item already gets.

## Making the tripwire reachable

`ATTN_PI_HOST_RETENTION_ITEMS` and `ATTN_PI_HOST_RETENTION_BYTES` override the compiled-in budgets, following `ATTN_KITTY_STORAGE_LIMIT`'s shape including its "a typo must not silently change behavior" rule — a non-positive or unparseable value is reported and treated as absent, meaning the default rather than zero. They are how this was verified live, and the escape hatch for a machine where a host's memory matters more than a month of scroll-back.

## Verification

Live, on a throwaway profile with the item budget lowered to 40: a 400-entry conversation resumed into a new session, host retained 40 and dropped 360, and the app reported `historyDropped: 360` with no load-earlier button. The transcript is headed by "360 earlier items are no longer kept for this conversation." and begins at "Filler question 181" — the 361st of the 400 entries, the first one left after 360 were dropped. Profile reaped afterwards.

The eviction path itself is covered by unit tests rather than a committed scenario: the real tripwire is 50,000 items, and a scenario that depended on a tuned daemon would be testing a configuration no user runs.

- 13 new host tests driving retention: what it drops and keeps, byte accounting given back on a drop, the byte tripwire firing where the item count would not, one oversized item kept rather than leaving nothing, paging past what was dropped, and both streaming cases.
- 6 new pane tests including the StrictMode replaying-updater check this repo requires of any new surface, plus the splice and epoch-change paths.
- Every new rule mutation-checked — including one guard (`Math.max` on the splice) that survived its mutation because seq-dedup makes a decreasing count unreachable; it was removed rather than kept as a guard that cannot fail.
- Full frontend suite (2,612), full plugin suite (140), `tsc --noEmit`, `go build` — green.

Base is `epic/pi-headless-host`. This closes watch-list item (5) from the slice-5 report and narrows (6): the zombie scroll-back edge remains, but the fresh-window half of it is no longer silent.

https://claude.ai/code/session_015ooFm5RBBqS9omX5GWKqTN

### DIFF
--- a/app/src/components/ConversationPane/ConversationPane.css
+++ b/app/src/components/ConversationPane/ConversationPane.css
@@ -94,6 +94,14 @@
   color: var(--color-error);
 }
 
+/* The top of a conversation whose start the host no longer keeps. It is not a
+   notice from the agent's side of the exchange, so it sits apart from the first
+   item rather than between two of them. */
+.conversation-pane-dropped {
+  flex: 0 0 auto;
+  margin-bottom: 8px;
+}
+
 .conversation-pane-empty {
   color: var(--color-text-secondary);
   font-size: var(--font-size-sm);
--- a/app/src/components/ConversationPane/ConversationPane.history.test.tsx
+++ b/app/src/components/ConversationPane/ConversationPane.history.test.tsx
@@ -185,3 +185,135 @@ describe('ConversationPane history', () => {
     expect(screen.getByTestId('conversation-model')).toHaveTextContent('local/experimental');
   });
 });
+
+/**
+ * Retention reached: the host's transcript budget dropped the start of the
+ * conversation for good. Nothing can page it back, so the only honest thing the
+ * pane can do is say so — a transcript that silently begins mid-thought is the
+ * failure this row exists to prevent.
+ */
+describe('ConversationPane dropped history', () => {
+  beforeEach(() => {
+    useConversationsStore.setState({ conversations: {} });
+  });
+
+  function dropped(count: number, hasMore: boolean, seq = 2) {
+    apply('conversation_snapshot', {
+      epoch: 'e1',
+      items: [{ kind: 'message', id: 'm900', role: 'assistant', text: 'nine hundred', streaming: false }],
+      has_more: hasMore,
+      dropped: count,
+      running: false,
+      queue: { steering: [], followUp: [] },
+    }, seq);
+  }
+
+  it('says how much of the conversation is gone once there is nothing left to page', () => {
+    renderPane();
+    apply('session_ready', { state: 'idle' }, 1);
+    dropped(1_240, false);
+
+    const row = screen.getByTestId('conversation-history-dropped');
+    expect(row).toHaveTextContent('1,240 earlier items are no longer kept');
+    expect(row.getAttribute('data-dropped')).toBe('1240');
+  });
+
+  it('stays quiet while the way back is still offered', () => {
+    // Retention has dropped items AND the host still holds pageable ones. The
+    // row would be premature: what the user can reach has not run out yet, and
+    // the button is the honest surface until it does.
+    renderPane();
+    apply('session_ready', { state: 'idle' }, 1);
+    dropped(1_240, true);
+
+    expect(screen.queryByTestId('conversation-history-dropped')).toBeNull();
+    expect(screen.getByTestId('conversation-load-earlier')).toBeInTheDocument();
+  });
+
+  it('draws nothing for a conversation that simply reached its start', () => {
+    // The ordinary short conversation: a window covering everything, nothing
+    // dropped. This is why the row is driven by the dropped count rather than by
+    // "the window did not carry it all".
+    renderPane();
+    apply('session_ready', { state: 'idle' }, 1);
+    dropped(0, false);
+
+    expect(screen.queryByTestId('conversation-history-dropped')).toBeNull();
+  });
+
+  it('does not double-count under StrictMode', () => {
+    // The bar this repo holds for any new surface: React replays state updaters,
+    // so a count folded in with anything other than an idempotent rule reads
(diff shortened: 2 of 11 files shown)


figgyster (approved):
Approve — with three small notes that don't block anything.

The three-signal split is the right call, and I mean structurally: `has_more` is "ask me again", `truncated` is "this message doesn't carry everything", `dropped` is "gone for good". Only the third can honestly drive that row, and the row only renders once paging is exhausted — the button stays the honest surface until it stops being one. The epoch rule is the part I like most: a rebuilt host that read the whole session file back reports its own count, so a revive doesn't inherit a loss it doesn't have.

The streaming guard fixes a real corruption-class bug, not a theoretical one. I traced the delta path: evict the open message, the next delta finds no open message and mints a fresh one from the tail alone — paragraph split, first half filed under the tool that ran beneath it. Your measured 400 chars → 10 survived is the kind of receipt that makes a bug real. And the guard is placed right: the only streaming item trim can ever reach is `items[0]`, so `return` is correct, not a lazy `continue`.

Verified on the box, on this head:
- Plugin suite 140/140, host-history 41/41 — the 13 retention tests are witnesses, not mocks: byte accounting given back on drop, the byte tripwire firing where the item count wouldn't, the oversized item kept rather than an empty transcript, paging past a drop answering with an empty page instead of an error.
- Pane history tests 15/15, including the StrictMode replay check and the splice-keeps-count path.
- Full frontend suite green (2607 passed here; your 2,612 count is env-conditional skips), tsc clean.
- Mutation-checked three of your rules myself: removing the streaming guard fails 2 retention tests; dropping `droppedBefore > 0` from the row condition fails 2 pane tests; making the splice keep the client's count instead of the snapshot's fails the splice test. Every rule I touched is pinned. The "Math.max survived its mutation so we removed it" story is consistent with what's in the tree.
- The daemon never reshapes `conversation_snapshot` — the host emits it straight onto the stream, so `dropped` rides through untouched. No protocol-version concern; this is the plugin envelope layer, not the native contract.

CI: everything green except the ledger-documented split-blank E2E flake on shard 3/3 (`split-blank-repro-agent-pa-10f34-after-opening-a-shell-split`) — terminal-blocks territory, nothing this PR touches; shards 1/3 and 2/3 passed. Daemon/Tauri correctly skip (no Go or Rust in the diff).

Notes, all cosmetic:

1. `conversation-notice--info` (ConversationPane/index.tsx:219) has no CSS definition anywhere — only `--warn` and `--error` exist in ConversationPane.css. The row renders fine on the base notice style, so either the modifier was meant to do something and never landed, or it's dead weight. Define it or drop it.

2. `retentionBudget` is the one new rule without a witness. The whole point of it is "a typo must not silently change behavior" — that's a claim about parsing, and it's the single new function with no test. Pure, five lines to pin, and this repo's discipline is mutation-checked rules.

3. Arithmetic in the PR body: "begins at 'Filler question 181' — which is exactly entry 360 of 400". Under any consistent reading that's off by one — 360 dropped of 400 leaves the first retained entry at 361; entry 360 is the last one dropped. The row's "360" is right; the parenthetical isn't.

The remaining zombie edge — a client holding scroll-back the host has since evicted keeps drawing the old items above the row — is honestly scoped in the body, and the fresh-window half of the silence is genuinely dead. Slice-5 watch item (5) closes, (6) narrows. Good night's work.

figgyster (approved):
Both remaining nits closed at d444511, and the retention one the better way: instead of just adding a test around the env read, `retentionBudget` moved into envelope.ts as a pure function — raw string in, warn callback out — with env reading and the host-log prefix left behind a thin wrapper in index.ts. That makes the typo rule directly testable, and the six new tests pin every clause: empty → default silently, "40" → taken (the reachable-tripwire case), "1" → a legitimate ask, the typo corpus ("fourty", "0", "-1", "1.5", "40 items", "1e3px", "Infinity", "NaN") → all default, an exact warning string, and "1e3" → 1000 so the exponent edge is on the record rather than implicit. The dead `--info` hook is gone from the className; AGENTS.md now documents the overrides.

Verified at the new head: plugin suite 146/146 (140 + the six), pane tests 15/15, full frontend 2607 passed / 0 failed, tsc clean. CI fast jobs green, E2E shards spinning. Re-approving on head — #814's ledger is now fully square.

PR #815, opened 2026-08-09
fix(tests): let the plugin fixture read attn's messages in any order

victorarias:
attn's longest-standing CI flake was the plugin-driver end-to-end test timing out
on a `driver.session_closed` notification that normally reaches the plugin in
under 50ms. The daemon-log instrumentation added for it caught the failure on
PR #800's CI (run 31313203551) and named the link: the daemon reached the send,
and the write came back `EOF` because the plugin process was already gone.

## The bug

The daemon answers a plugin's `session.report_*` call on that connection's read
loop, and `applyState` broadcasts the state change *before* the answer is
written. Anything that pauses the daemon between those two steps — a loaded CI
runner descheduling that goroutine is enough — leaves the plugin waiting while
the test, which watched for the broadcast, moves on and kills the session's PTY.
The close notification then goes out on a separate goroutine and reaches the
socket ahead of the answer.

That is legal JSON-RPC and exactly what both shipped plugin clients expect: the
SDK (`sdk/plugin`) and attn-pi's own client each route a message by shape — a
method is a request to handle, a bare id resolves a pending call. The end-to-end
test's fixture did not. It treated any request arriving while it awaited a
response as fatal:

```
unexpected plugin request "driver.session_closed" while waiting for response id=24
```

So the fixture died on the very notification the test was waiting for, its
connection went with it, and the notification the daemon had already written came
back `EOF` with nobody left to receive it. The 5s tripwire then fired, the session
sat at `idle`, and the diagnosis printed `driver_registered=false` — the fixture's
supervised restart, whose `hello` raced test teardown and was rejected as a stale
generation. That stale-generation line is teardown noise, not the cause; the
ordering above is.

Why it was rare: the gap between broadcast and answer is normally microseconds,
and the PTY kill it has to lose to takes milliseconds.

## The fix

The fixture now has one reader that routes by shape, the same way the SDK does.
There is no unexpected-request branch left to regress to — an interleaved request
is answered wherever it lands, including in the middle of the fixture's own call.
No product code changes; the daemon was behaving correctly throughout.

## Receipts

Forcing the ordering — a temporary local hook pausing 400ms between the daemon's
broadcast and its answer, not part of this diff — reproduces CI exactly:

| | before | after |
|---|---|---|
| provoked interleave (400ms) | **3/3 fail** | **4/4 pass** |
| 32 loaded parallel runs (4 copies × 8 rounds, 4 spinners) | pass | pass |

The pre-fix failures printed the CI text verbatim: `unexpected plugin request
"driver.session_closed" while waiting for response id=24`, `close records
recorded: 0`, `plugin session close notification failed: … err=plugin
driver.session_closed: EOF`. Load alone never hit the window on this machine,
before or after — the race is too narrow, which is the point.

## The test that pins it

`TestPluginFixturePeerAnswersRequestArrivingBeforeItsResponse` scripts the daemon
side over a `net.Pipe`: read the peer's report, send `driver.session_closed`
first, answer the report only after the close is acknowledged. Deterministic by
construction — no sleeps, no polling, no load — so the ordering that used to be a
rare race is now exercised on every run. It asserts the peer recorded the close
and still got its response.

## Surfaces

- **Plugins and SDK** — checked both shipped clients (`sdk/plugin`,
  `plugins/attn-pi`, `plugins/attn-opencode`); all three already route by shape.
  The fixture was the only offender.
- **Docs** — the SDK README now states the full-duplex rule instead of leaving
  plugin authors to discover it as a flake.
- **Daemon, protocol, CLI, Linux** — untouched; the diff is test code plus a
  README and a changelog fragment.

## Known limitation, deliberately not fixed here

`notifyPluginDriverSessionClosed` is fire-and-forget: when the send fails, the
run has already been ended in the store, so the notification is lost and a
reconnecting plugin will not see that run in `driver.register`'s `active_runs`
either. In practice the only way that send fails is a dead connection, and the
plugin process that owned the run's in-memory state is being restarted with it —
attn-pi's `sessionClosed` only drops two map entries. Adding a redelivery ledger
would be machinery for a leak I have no evidence of, and it would also have
masked this fixture bug. Happy to file it as a ticket if it is worth closing.

## Verification tier

No product code changed, so a profile install would exercise nothing this diff
touches. The integrated evidence is the end-to-end test itself — a supervised
plugin process, the real daemon socket, a real worker PTY — passing under the
provoked ordering that reliably failed before, plus the full `internal/daemon`
package green.

### DIFF
--- a/changelog.d/plugin-close-notification-flake.yaml
+++ b/changelog.d/plugin-close-notification-flake.yaml
@@ -1,0 +1,14 @@
+kind: internal
+area: tests
+change: >
+  Closed attn's longest-standing CI flake: the plugin-driver end-to-end test's
+  missing driver.session_closed notification. The daemon answers a plugin's
+  state report on its read loop after broadcasting the state change, so a
+  session closing in that gap puts driver.session_closed on the wire ahead of
+  the answer. The test fixture treated an unexpected request as fatal and died,
+  taking its connection with it, and the daemon's close notification came back
+  EOF. The fixture now routes messages by shape the way the shipped SDK does —
+  a method is a request to answer, a bare id is the response being waited for —
+  and a scripted-socket test pins that order on every run instead of leaving it
+  to a scheduling race. The SDK README states the full-duplex rule plugin
+  authors have to honour.
--- a/internal/daemon/plugin_driver_e2e_test.go
+++ b/internal/daemon/plugin_driver_e2e_test.go
@@ -4,10 +4,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"net"
 	"os"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"syscall"
 	"testing"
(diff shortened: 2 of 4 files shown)


figgyster (approved):
Case closed, and it was a good one. The fixture was the killer, not the daemon — the daemon was doing legal full-duplex JSON-RPC the whole time, and the test peer was the only thing on the force that couldn't handle a message out of turn. That's a hell of a root cause to pin down through instrumentation, and the fix is the right shape: one reader, route by shape, no unexpected-request branch left to regress to.

Verified on the box (sandbox, real native lib):

- Full `internal/daemon` package green under `-race` — 109s, clean.
- The new pin, `TestPluginFixturePeerAnswersRequestArrivingBeforeItsResponse` — ×2 under `-race`. Deterministic by construction, and it actually pins the whole ordering: close lands mid-request, peer records it, acks it, still gets its response. The scripted side checks the ack id, so a peer that dropped the close would fail the test, not just the record count.
- The E2E that used to flake — `TestPluginDriverEndToEnd_InstalledProcessLaunchReportAndResumeThroughWorkerPTY` — green under `-race`, 4.4s. The exact race it used to lose, now losing nowhere.
- Checked the "only the fixture was the offender" claim against the tree: `sdk/plugin` routes by shape (`"method" in message` → handle, else resolve pending), and so do attn-pi (`attn-rpc.ts`, `relay.ts`) and attn-opencode (`attn-rpc.ts`). All three were already speaking the language. The fixture was the only cop who couldn't.
- The old fixture's fatal branch is byte-for-byte the CI failure text — `unexpected plugin request "driver.session_closed" while waiting for response id=24`. Good receipts.
- No product code touched, no orphaned helpers left behind, `git diff --check` clean, changelog fragment in the right place with the right shape.

Two things worth saying out loud:

1. The `default` branch in the peer's `handle` is a quiet improvement over the old silent `continue`: answering unknown methods with `-32601` instead of ignoring them keeps the daemon's caller off its 30s timeout and says why. Matches what the SDK does. Good instinct.

2. The known limitation — `notifyPluginDriverSessionClosed` stays fire-and-forget. Agreed with the call. A redelivery ledger is machinery for a leak with no evidence, and it would have masked exactly this fixture bug. If it ever shows up in the wild, ticket it then. No ticket needed now.

One cosmetic nit, non-blocking: `newPluginFixturePeer`'s comment says ids 1 and 2 belong to the hello and driver.register handshake, but register now goes through `call()` which bumps `nextID` to 3 first — so register is id 3 and id 2 never appears. The comment describes the old numbering; the conclusion (first generated id starts after the handshake) still holds. One-line comment fix if you're ever in the file.

CI's green on the jobs that matter — Daemon, Linux daemon both arches, Changelog, Changes. Bench A/B is still running; if the gate counts it, merge waits on it.

figgyster (approved):
The nit's closed, and you fixed it the better way — `nextID` 1 instead of just editing the words, so the ids run contiguous from hello (1 → register 2 → reports onward) and the comment's honest now. Re-verified on the new head, sandbox with the real native lib:

- Full `internal/daemon` package green under `-race` — 1219 tests, 112s, zero failures.
- The E2E that used to flake, ×4 under `-race` — clean every run.
- CI all green on this head: Daemon, Linux daemon amd64/arm64, Changelog, Changes, Bench A/B (which was still pending at the first review — now passed).

One honest data point: my first package run at this head failed once in the sandbox, unidentified, and never repeated across two clean full-suite runs and the E2E ×4 after it — under 3 CPUs with the E2E building its own binary mid-test, I'm calling it sandbox noise, not a repo signal. If it ever shows up on CI, it gets a ticket.

Approval stands on the new head.

PR #816, opened 2026-08-09
feat(activity): say what each agent is doing right now

victorarias:
One short present-tense line per session, saying what that agent is doing right
now — "Rebasing brisk-toucan onto main, resolving migration numbering" instead of
a state chip that says `working`. It is generated from the session's own
transcript by a headless agent the user picks, and it is off by default.

Design and the measurements behind every number:
[docs/plans/2026-08-07-session-activity.md](docs/plans/2026-08-07-session-activity.md).

## What it does

- A background job scans sessions on a cadence. For each one whose transcript
  has moved since the last look, it renders a window of the new events —
  thinking, assistant prose, tool calls, tool results — and asks a headless
  agent for a single line.
- The line is persisted on the session beside the transcript cursor that
  produced it, broadcast through a `session.activity.changed` bus fact, and
  rendered under the session's name on the dashboard.
- `attn activity` shows the config, the presence tier, and every session's
  current line from a terminal; `attn activity clear` wipes them.

## Generation follows the user, not the agents

A line nobody can see has no value, so the daemon derives a presence tier from
what connected clients report and gates on it: **watching** (the dashboard is on
screen) refreshes every 120s, **present** (input in the app, dashboard not
showing) every 300s, and **away** generates nothing at all.

Away is a hard stop rather than a slow rate, which is safe because leaving away
is always an action that restores a higher tier — the staleness heals itself the
moment it would matter. Measurement forced it: real state transitions run at 13
per session-hour against 30/hr for watching, so generating through away would
have cost nearly half of always-on for an empty room.

`set_client_presence` is a heartbeat, not a latch. A client that is force-quit
mid-watching would otherwise pin generation on forever with nobody looking, so an
unheard-from client expires to away — it fails toward off. Watching also expires
after ten minutes of an untouched dashboard, which is the case the heartbeat
alone cannot see: a left-open window on a second monitor keeps reporting
`dashboard_visible` truthfully and would otherwise generate all night.

## Cost

The cursor is the precondition: a session whose transcript has not moved has
nothing new to say, and costs nothing. Two run stamps keep that honest —
`ObservedAt` advances on every pass and gates transcript movement, `SpentAt`
advances only on a pass that actually called the agent and gates the interval —
because the stored line only records successes, and without the split a failing
session would be retried on every scan.

The prompt is split at a `{{USER}}` marker so the invariant half rides
`--system-prompt` and replaces the CLI's own, which is where nearly all the
per-run cost lived.

## No default agent

`activity.config` ships with no agent selected. Claude and Codex differ enough in
speed, price, and which account pays that picking one would be choosing how the
user's money gets spent. An enabled feature with no agent selected is a reported
error, not a silent fallback.

A run that fails keeps the last good line on screen and surfaces the failure
underneath it — `attn activity` prints `last run failed: …`, and the status
carries it on the wire. A line that stops moving should say why.

## Surfaces

- **CLI** — `attn activity`, `attn activity clear`.
- **Daemon and app** — settings section, presence reporting from the app, the
  dashboard line with its own staleness window derived from the configured
  intervals.
- **Protocol** — `ProtocolVersion` 219; `generated.ts` and `generated.go`
  regenerated together.
- **Docs** — design plan, live-verification record, changelog fragment.

## Verification

`make install PROFILE=actrev7` on a throwaway profile, preflight PASS, protocol
219 end to end:

- lines generated twice from a real Claude session and rendered on the dashboard;
- a deliberately broken model surfaced as `last run failed: headless agent MCP
  tool server failed: exit status 1` beneath the retained last good line, and
  cleared on the next good run;
- `watching` held through 9m08s of an untouched dashboard and fell to `away` at
  10m08s after the last input; one click restored `watching`.

`go test ./...`, 2604 frontend tests, `tsc --noEmit`, and `changelog-check` are
green on this head.

https://claude.ai/code/session_017ZqiJukMRfyq9WrrmEraRX


figgyster (approved):
Big feature — 6.9k lines, and the first one I've seen where every number in the plan has a receipt before the code landed. The cost model, the window sizes, the thinking-block rescue rate (76% of prose-less windows), the 13/hr vs 30/hr state-transition measurement that killed the always-on design, the `--system-prompt` prefix trim from 46,745 to 33,955 tokens — all measured through the real seam, all written down. And the honest-remainder discipline from the testing wave carries over: `presence_idle_seconds` is flagged UNMEASURED in the code, the state-consistency check's known blind spots are documented, the subject-changes-within-working gap ships unsolved with a named cheap fix. That's the culture that made #809 worth reading, applied to a product feature.

The adversarial-review commit closed real gaps — I confirmed each in the code rather than taking the commit message's word. The ObservedAt/SpentAt split is what stops a failing session re-running at the 30s tick rate (the queue was resetting the attempt count, so backoff and the dead-job notification never fired). The 10-minute watching expiry kills the "cheapest tier is the only one that never ends" hole. The rolling-tail window read fixes the worst one: a long delta used to return its oldest page and leave the cursor mid-backlog, so a busy session got a line about its ancient history and the next window was stale too. The transcript-path cache kills a 235–489ms Codex rollout walk per session per tick. All present, all tested.

Verification on the box (new head 5414dd00):

- daemon suite green at 99s including the plugin E2E leg, store green, activity/transcript/protocol/agent/client/cmd green; -race green on the activity/presence/store/transcript/protocol surface; vet clean; `cmd/activity-bench` builds.
- frontend: full suite 232 files / 2599 passed, tsc clean; the new presence/settings/dashboard tests are real witnesses — the tier table, the watching-expiry boundaries, the heartbeat-collapse, the post-disable write guard, the projection carrying the line on the pushed snapshot.
- protocol lockstep: 219 in both spots, enforced by `TestFrontendProtocolVersionMatchesDaemon`; `generated.go` regenerates byte-identical from main.tsp (go-jsonschema v0.24.1); `generated.ts` differs only in quicktype's environment-dependent auto-name ordering, same noise #810 documented.
- CI all green on this head — all three E2E shards, Bench A/B, both Linux arches. Even the usual split-blank E2E flake stayed home tonight.
- `git diff --check` clean.

The last commit, `fix(activity): keep presence reporting out of render`, is the repo's own recurring risk #2 catching itself: the refs were being written during render, so a discarded render could leak a stale presence report into committed work. Moved into an effect — right call, and the parsers moving to `utils/activitySettings` so the component file stays Fast-Refresh-clean is the right shape too.

One nit on the record, non-blocking: the interval inputs don't actually keep their promise. `type="number" min/max` doesn't stop a typed out-of-range value, `save()` doesn't clamp, and the daemon clamps on read — so the pane can show `10` while the daemon runs `30`. The comment says the inputs "say so rather than silently accepting a number that comes back different", but they don't; the mismatch just isn't loud. Clamp in `save()` or live with the lie knowingly.

Approved on head. The line is display-only, the money is gated on somebody looking, and the failure modes all fail toward off — this is the right shape for a feature that spends per refresh.

PR #817, opened 2026-08-09
feat(app): apply pipeline, scaffold, and codegen

victorarias:
A4 slice 4: the pipeline that turns an app directory into an installed,
versioned app. `attn app new` scaffolds one, `attn app apply` installs it,
`attn app rollback` moves it back to a version that already ran, and
`attn app dev` re-applies on save.

Slice 3 shipped the registry and the read-only CLI; this is the half that
writes to it.

## The rule

**Apply never evaluates app code.** Parse, generate, typecheck, bundle, hash,
record. Nothing in the app is imported and nothing is run, so every way an
apply can fail happens before the pointer moves, with the previous version
still serving.

That is what makes `attn app dev` safe to leave running: a save that does not
compile costs you an error message, not a running app.

## What a version is

`attn-app.toml` declares the app's name, the events it subscribes to, and the
document collections it keeps. Codegen turns that declaration into
`src/generated.ts` — a `Handlers` type with one member per subscribed pattern —
and the scaffold's entrypoint ends in `satisfies Handlers`, so the compiler is
what keeps manifest and code in step. Declare an event with no handler and the
apply fails with the file and the line.

The content hash covers **the declaration and the bundle**, not the bundle
alone. A manifest-only edit changes what the version means; hashing only the
code would reuse a row whose frozen declaration is stale. Applying the same
directory twice therefore records nothing the second time — same hash, same
version — and a description change is a new version, which the dev-loop
receipt below shows both of.

## Where the build runs, and why the daemon still owns the flip

The build runs **CLI-side**, in `internal/appbuild`. bun lives on a developer
PATH (`~/.asdf/shims/bun` here) that `pathutil.EnsureGUIPath()` does not
reconstruct, so a daemon spawned by the app cannot reliably find it — and a
build is foreground work with output someone is waiting on, not daemon work.

The daemon keeps the atomicity it must own. `app_apply` carries
`(name, content_hash, declaration)` and **never a path**: the daemon re-derives
the artifact path from the name and the hash, re-hashes the bytes on disk, and
refuses the apply if they disagree. A client cannot point the registry at a
file the pipeline did not produce. The insert and the pointer move are one
transaction (`Store.CommitAppVersion`, from slice 3), the artifact is staged
and renamed into place inside the store — `apps/.staging`, because a rename out
of `/tmp` crosses filesystems on Linux — and the flip publishes
`app.version.changed` carrying the previous id so slice 5's runtime need not
race the pointer.

## The two product-shaped decisions

**Typechecking.** A pinned TypeScript (5.8.3) installed lazily on the first
apply under `<data-dir>/apps/toolchain`, shared by every app, behind an `flock`
so concurrent applies install it once. Invoking `tsc` directly rather than
through `bun x` was measured at 0.77s against 2.1s; the install is ~30MB, once
per machine. A scaffolded app has no `node_modules` and no `package.json` of
its own.

**The SDK.** It ships as an ambient module declaration written into the app
(`src/attn-app.d.ts`, declaring `@victorarias/attn-app`). Handlers get a typed
surface with nothing published and nothing installed, and if the package is
ever published it takes over the same specifier without touching app code. This
is the answer to the brief's "if the SDK's package shape needs a product
decision, stop and ask" — it did not need one, because nothing about this
forecloses publishing later.

## Receipts

Live, on a throwaway profile installed from this branch
(`make install PROFILE=appsforge`, preflight `PASS`, CLI/app/daemon all on
protocol 219).

<details>
<summary><code>new</code> → <code>apply</code> → <code>apply</code> → edit → <code>apply</code></summary>

```
$ attn app new .../standup-digest
created app standup-digest in .../standup-digest
  edit src/index.ts, then: attn app apply .../standup-digest
  AGENTS.md (and CLAUDE.md, a symlink to it) is the whole brief — an agent needs nothing else

$ find standup-digest
standup-digest/.gitignore
standup-digest/AGENTS.md
standup-digest/attn-app.toml
standup-digest/CLAUDE.md
standup-digest/src/attn-app.d.ts
standup-digest/src/generated.ts
standup-digest/src/index.ts
standup-digest/tsconfig.json

$ attn app apply .../standup-digest          # zero edits to the scaffold
  installing TypeScript 5.8.3 into /Users/victor/.attn-appsforge/apps/toolchain (once per machine)
  typechecking src/index.ts
  bundling src/index.ts
applied app standup-digest: version 1 (2968be01eba9), new, 278 bytes
  artifact .../apps/standup-digest/2968be01.../bundle.js

$ attn app apply .../standup-digest          # nothing touched
applied app standup-digest: version 1 (2968be01eba9), unchanged — byte-identical to the version it already was

$ attn app status standup-digest | grep history
  history:    1 version(s), 0 invocation(s)

$ # add `at: event.published_at` to the handler
$ attn app apply .../standup-digest
applied app standup-digest: version 2 (a56e25577de8), new, 306 bytes
  was on version 1
```
</details>

<details>
<summary>A broken apply over an installed app changes nothing</summary>

```
$ # add "ticket.created" to the manifest, no handler for it
$ attn app apply .../standup-digest
  typechecking src/index.ts
app apply: app "standup-digest" does not typecheck against its manifest:
src/index.ts(20,3): error TS1360: Type '{ "session.state.changed": (event: AppEvent, ctx: Ctx) => Promise<void>; }' does not satisfy the expected type 'Handlers'.
  Property '"ticket.created"' is missing in type '{ "session.state.changed": ... }' but required in type 'Handlers'.
exit=1

$ attn app status standup-digest | grep -E 'version:|history:'
  version:    2 (a56e25577de8...)
  history:    2 version(s), 0 invocation(s)

$ ls apps/standup-digest ; ls -A apps/.staging | wc -l
2968be01...  a56e2557...
0
```
</details>

<details>
<summary>Nothing in the app is evaluated</summary>

An app whose module body writes a sentinel file and then throws at import time.
It applies cleanly, the sentinel never appears, and the `throw` is sitting in
the installed bundle — compiled, never run.

```
$ cat src/index.ts
declare function require(m: string): any
require("node:fs").writeFileSync("<sentinel>", "evaluated\n")
throw new Error("this module must never be evaluated by an apply")
export default { "session.state.changed": async (_e, _c) => {} } satisfies Handlers

$ attn app apply .../hostile-app
applied app hostile-app: version 3 (64cb8b305d84), new, 436 bytes

sentinel absent: nothing in the app was evaluated
the throw is in the bundle (compiled, never run): 1
```
</details>

<details>
<summary><code>attn app dev</code></summary>

Three edits: a code change, a comment-only change, a manifest description
change. The comment is byte-identical after bundling; the description is not,
because the declaration is in the hash.

```
$ attn app dev .../standup-digest
watching .../standup-digest — every change is parsed, typechecked, bundled and applied
shows apply results and build errors. Handler invocations are not streamed: nothing dispatches to apps yet — `attn app status <name>` is where that will appear.
ctrl-c to stop

22:45:05  applied app standup-digest: version 1 (2968be01eba9), ...   in 179ms   # code reverted
22:45:08  applied app standup-digest: version 1 (2968be01eba9), ...   in 221ms   # comment only
22:45:11  applied app standup-digest: version 4 (6d90b34811f9), new, 278 bytes
          was on version 1                                            in 182ms   # description changed
```
</details>

<details>
<summary><code>rollback</code></summary>

```
$ attn app rollback standup-digest          # no version named
rolled app standup-digest back to version 2 (a56e25577de8)
  was on version 4

$ attn app rollback standup-digest 4        # naming a newer one is not a rollback
moved app standup-digest forward to version 4 (6d90b34811f9)
  was on version 2

$ attn app rollback standup-digest 4        # already there
app rollback: daemon error: app rollback standup-digest: it is already on version 4;
  its versions, newest first: 4 (6d90b34811f9) — current, 2 (a56e25577de8), 1 (2968be01eba9)
exit=1
```
</details>

The second commit is a wording fix the dev-loop receipt caught: re-applying
content that already had a row while the app sat on a different version said
"unchanged" directly above "was on version 2" — the row was unchanged, the
pointer was not.

## Surfaces

- **CLI** — `new`, `apply`, `rollback`, `dev`, with help text; `list`/`status`
  from slice 3 read the new rows unchanged.
- **Daemon** — `app_apply`, `app_rollback`; `app.version.changed` joins the
  other app facts in having no projection (nothing on the wire renders apps
  yet).
- **Protocol** — 219, via `main.tsp` + `make generate-types`, bumped in
  `constants.go` and `useDaemonSocket.ts`.
- **Linux** — `GOOS=linux go build ./...` clean. The pipeline shells out to
  `bun` and a pinned `tsc` with nothing darwin-only in the path. **I did not
  run an app build on Linux**; the live receipts above are darwin, and CI now
  runs the real-toolchain tests on ubuntu (see below), which is the Linux
  witness.
- **CI** — the `Daemon` job had no bun, so every real-toolchain test in
  `internal/appbuild` would have skipped itself and the job would have gone
  green on a pipeline it never ran. Added the bun setup step, pinned from
  `.tool-versions`.
- **Docs** — glossary gains "applying" and the never-evaluate rule; the plan
  doc's slice 4 entry carries the as-builts; changelog fragment.

Not touched: the frontend (no UI for apps yet) and the plugins/SDK
directories.

## Tests

`internal/appbuild` covers the manifest (including a differential test that
holds its name rule to `apps.ValidateName` rather than restating it), codegen,
and the pipeline: scaffold-applies-with-no-edits, the tsc error carrying
`filename(line,col)`, never-evaluates, identical-content-same-version,
manifest-only-change-is-a-new-version, and failure leaving an untouched store
with an empty staging dir.

`internal/daemon` covers the handlers against a fake artifact store — hash
mismatch, missing artifact, a declaration naming another app, rollback's
default target and its refusals — plus two tests that run the **real** builder
into the daemon's artifact directory and apply what came out, which is the only
place the two halves' agreement on hash and path is checked.

Green: full `./scripts/test-go.sh`, `-race` on `internal/appbuild`,
`internal/store`, `internal/daemon`, `gofmt`, `go vet ./...`, and the frontend
typecheck after the protocol regeneration.

## Next

Slice 5 gives apps a runtime: dispatch, handler context, the invocation log,
per-app consumers wired to slice 1, auto-disable, `attn app logs`. That is when
`attn app dev` gets something to stream.

---
https://claude.ai/code/session_018Cva287WpGCubmwozVpEo9


cmd/attn/app_build.go
```diff
@@ -0,0 +1,343 @@
...
+		},
+	})
+	if err != nil {
+		return nil, appbuild.Result{}, err
+	}
+	abs, _ := filepath.Abs(dir)
+	result, err := appClient().AppApply(res.Manifest.Name, res.ContentHash, res.Declaration, abs)
+	if err != nil {
+		if res.ArtifactWritten {
+			_ = os.RemoveAll(filepath.Dir(res.ArtifactPath))
```
figgyster: This cleanup assumes a daemon refusal means nothing was recorded. Two interleavings break that:

- The daemon commits the row, then the response is lost (connection drop, daemon restart mid-apply). The CLI removes the artifact a committed version row now points at.
- Two concurrent applies of identical content — the race this PR's own toolchain-lock comment contemplates: a dev loop and a manual apply. One places the artifact and commits; the other's daemon call fails for a transient reason and removes the shared directory.

Rollback doesn't stat the artifact (right for the normal case, where a version's bundle is on disk by construction), so it would happily move the pointer onto a version whose bundle this cleanup just deleted, and slice 5's runtime would fail to load it. Narrow window, and the next apply of the same content repairs it — not a blocker. But "a failed apply leaves the store exactly as it found it" holds only modulo this race. Two ways out: have the daemon own the removal (it alone knows whether the row was written), or only remove when the error is one of the artifact-refusal classes the daemon already distinguishes.

figgyster (approved):
Slice 4 closes the loop the plan doc opened: the write half of the registry, and the rule the whole stage is arranged around — apply never evaluates app code — is held together by ordering, not by hope. Parse, generate, typecheck, bundle, hash, and only then does anything move, and the compiler is the cop: a declared subscription with no handler is a tsc error with a file and a line before a byte lands in the store. The hostile-app test (module top level writes a sentinel and throws; sentinel absent, apply green) is the right witness for the rule, and the body's receipts match it.

The two product calls both land. The pinned TypeScript behind an flock, installed once per machine, has a measured receipt (0.77s vs 2.1s for bun x) — and it's the right call for a dev loop that pays per keystroke pause. And the ambient-module SDK answers the brief's "stop and ask" by not needing one: a .d.ts that declares @victorarias/attn-app forecloses nothing, and the day the package publishes it takes over the same specifier without touching app code. Clean.

I read the whole diff (4059 lines) and there's no Go toolchain on this box, so CI is the witness — and it's a real one now: the Daemon job got bun 1.3.14 (read from .tool-versions), and internal/appbuild ran 30.6s on ubuntu with the real toolchain, no skips. That's the Linux app-build witness the body was honest about not having run locally, now on record. Store and daemon suites green across the repeated (race) runs, frontend typecheck green, all three E2E shards, Tauri, Bench A/B. Protocol 219 is in lockstep on every surface that carries it — constants.go, generated.go, app/src/types/generated.ts, useDaemonSocket.ts — and the frontend/daemon version test holds them together; there's no Rust protocol constant in this tree, so nothing else to bump.

Seams I specifically checked: the daemon re-derives the artifact path and re-hashes the bytes (the tampered-artifact test pins that a row can never name content the daemon didn't read), the declaration's inner name must match the app, rollback's default target is the highest id below the pointer (a "newest non-current" default would roll forward after a rollback — this one doesn't), the dev loop has two independent guards against waking its own watcher (the mtime-skip in WriteGenerated, pinned by test, plus the devRelevant filter), and staging lives inside the store so the final rename never crosses filesystems on Linux. The manifest-only-edit-is-a-new-version call is the right one, and the test spells out why: generated types are erased at build time, so the bundle alone would let a stale declaration ride along.

One comment on the failed-apply cleanup in applyApp, and two cosmetics: the package comment on cmd/attn/app.go still says "the three verbs that change it" when there are now seven, and shortHash is implemented twice (cmd/attn and internal/daemon). Neither is worth a revision round.

cmd/attn/app_build.go
victorarias: Took the third way out: the CLI stops removing anything.

Both of your interleavings are the same fact — only the daemon knows whether the row was written, and "refused" and "committed, response lost" are indistinguishable from here. Gating on refusal classes would still get the lost-response case wrong, and it would add a taxonomy the next reader has to trust.

The leftover it leaves behind is inert: nothing lists it, the path is content-addressed so re-applying the same content lands on that directory and reuses it rather than accumulating, and the daemon re-hashes the bytes before any row can name them. An unreferenced artifact costs a few hundred bytes; a missing one is the broken app you described. 440ece81.

Also folded shortHash into appbuild.ShortHash and fixed the "three verbs" comment on cmd/attn/app.go.

figgyster (approved):
Re-approved on head 440ece81 — the fix commit closes the loop, and the third way out is the right one. Gating on refusal classes would still guess wrong on the lost-response case, and the arithmetic is honest: an inert artifact costs a few hundred bytes and gets reused by the next apply of the same content, a missing bundle is the broken app. The removal was the only part of the pipeline that could delete something a committed row names, and now nothing can. The refactor is complete too: no lowercase shortHash survives anywhere, appbuild.ShortHash is the one source, and the package comment now describes the whole surface instead of three verbs.

CI on the new head: all green, including the Daemon job with the real-toolchain appbuild tests and all three E2E shards, Tauri, and Bench A/B. The thread on the cleanup is resolved. Ship it — slice 5's runtime is next.

PR #818, opened 2026-08-09
chore(comments): compress comment prose to constraints and receipts

victorarias:
## What

Comment density in hand-written non-test Go tripled since May (6% → 17% of lines), concentrated in recent work: package headers retelling their plan docs, per-field essays, narration of the following line, and justification written for the reviewer. This PR:

- compresses the 67 densest files (>22% comment lines): invariants, trap warnings, and measured receipts stay inline at a line or two; narration, design-doc retellings, and history go. Net −4,758 lines; the swept set drops from ~9,900 comment lines to ~4,700, and the overall ratio from 17.1% to 12.7%.
- states the policy in AGENTS.md: comments say what the code cannot show — a constraint, an invariant, a measured receipt — in one or two lines; package headers link to the design doc instead of retelling it; a comment addressed to the reviewer is a defect.

## Verification

Every touched Go file is token-stream identical to its previous version (go/scanner comparison ignoring comments — comments and gofmt whitespace are the only change), verified mechanically for all 69 files after the rebase. `gofmt -l` is silent, `go build ./...` and `go vet ./...` pass, and `go test` was run for the heaviest touched packages (daemon, pty, jobs, workflow, docstore). No `//go:` directives, `//nolint`, or cgo preambles were touched; `docs/plans/` links are preserved.

Live verification: exempt as a trivial comments/docs-only change (token-identical code, verified mechanically).

https://claude.ai/code/session_01AdM1oGRTFk3c75NPQQjPt2

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -336,8 +336,12 @@ forgets that position and `--since <RFC3339>` replays from an instant.
 - No continuously repainting animations. attn renders GPU terminals, often on
   high-refresh displays, beside agents that run all day — a permanent repaint
   loop is a battery and thermal bug no test will catch.
-- Comments describe how a thing is used, and move when the code moves. Mostly on
-  functions and seams, not annotating every line.
+- Comments state what the code cannot show — a constraint, an invariant, a
+  measured receipt — in one or two lines, and move when the code moves. Godoc
+  on an exported symbol is one line. A package header is a few lines plus a
+  link to the design doc, never a retelling of it. Never narrate the next line
+  or argue that the change is correct: that talk belongs in the PR, and a
+  comment addressed to the reviewer is a defect.
 - Conventional commit titles with a scope, in plain language:
   `fix(queue): hand over the next agent however a turn closes`.
 
--- a/changelog.d/feral-ibex-comment-sweep.yaml
+++ b/changelog.d/feral-ibex-comment-sweep.yaml
@@ -1,0 +1,8 @@
+kind: internal
+area: docs
+change: >
+  Comment sweep across the Go source: design-doc retellings, narration, and
+  review-directed justification are cut; invariants and measured receipts stay,
+  compressed to a line or two. AGENTS.md now states the policy — comments say
+  what the code cannot show, package headers link to the design doc instead of
+  retelling it.
--- a/cmd/attn/db.go
+++ b/cmd/attn/db.go
@@ -17,9 +17,8 @@ import (
 	"github.com/victorarias/attn/internal/store"
 )
 
-// runDB routes `attn db <command>`: operator-facing database maintenance that
-// sits alongside the daemon's own automatic rotating backups (BackupNow /
-// backupPreMigration in internal/store).
+// runDB routes `attn db <command>`: operator-facing maintenance beside the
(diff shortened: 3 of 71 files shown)


figgyster (approved):
Approve. This is the rare diff where the strongest verification is the right one, and I re-ran it independently rather than taking the receipts on faith.

**Token-identical, checked by me.** I compared all 69 touched Go files against `origin/main` with my own go/scanner pass (comments excluded, mode 0): 69/69 identical token streams. That's the load-bearing claim and it holds — the code did not move, so behavior cannot have changed. Also verified: `gofmt -l` silent on all 69, `git diff --check` clean, no `//go:` / `//go:build` / `//nolint` changes, no cgo preamble touched, and `go build ./...` + `go vet ./...` pass with the real libghostty-vt fetched per the repo's own script. Tests on the heaviest touched packages are green here (pty, jobs, workflow, docstore, and the daemon suite minus two plugin tests that fail identically on `origin/main` in this sandbox — the fixture's `worktree.create` handler never registers; that's this box's spawn chain, not the PR. CI's Daemon job is green, 2m35s.)

**The arithmetic is honest.** My independent comment-line count on the swept set: 9,102 → 4,342, Δ −4,760. You said −4,758. Close enough that the difference is measurement, not spin.

**The quality of what remains is the real story, and it's good.** Sampled a dozen files across pty, daemon, workflow, docstore, config. The trap warnings and receipts survive compression intact: kittyseg keeps the byte-for-byte concatenation invariant and the "extraction is only safe from GROUND with a reached terminator" warning verbatim; ws_eviction keeps the 10 KB/s-link measurement (~1s hangup, ~45s blind client); ordinal keeps "phase is a sequence number, never the title — a rename must not invalidate cached calls"; config keeps the init-vs-TestMain ordering trap; docstore keeps the int64 overflow math (~7 years at 10 writes/s). What got cut is what the policy says should be cut — design-doc retellings, the gateway→here migration table, scaffolding history. I did not find a single compressed comment that misstates the code.

**docs/plans links:** every removed reference is a reflow or consolidation; package headers now carry the pointers, so nothing is orphaned. The AGENTS.md policy reads like a real rule a reviewer can enforce, and the changelog fragment is the right shape.

Two nits, neither blocking:

1. The sweep dropped one-line godoc from two exported types — `docstore.Filter` ("one comparison against a declared or reserved field") and `sessionstate.Observation` ("one recorded piece of evidence") — and the policy this PR adds says godoc on an exported symbol is one line. These two are the sweep being overzealous against its own rule; restore the one-liners.
2. Bench A/B was still pending when I looked. Moot — token-identical code can't move benchmarks — but if it finishes red, it's the harness, not this diff.

The direction is right: the codebase now says what it means in the places that matter, and the policy makes sure it stays that way. Good sweep. The godoc nit is the only loose thread.

figgyster (approved):
Both one-liners restored verbatim, and verified at the new head — 69/69 files still token-identical to `origin/main` with my own scanner, gofmt clean, build OK. The fix is exactly the policy this PR writes: godoc on an exported symbol, one line. Loop closed.

(Bench A/B and the daemon jobs were still spinning up when I looked; token-identical code can't move them. Changes and Frontend E2E already green.)

PR #819, opened 2026-08-09
test(synctest): take the paced-test remainder to zero candidates

victorarias:
## What

The previous sweep (#809) left 130 tests classified by static analysis alone —
"no boundary signal found", never actually run in a bubble. This takes that
remainder to zero: every one now has a verdict from running it.

- **87 more conversions** into `testing/synctest` — 86 in `internal/daemon`,
  1 in `internal/ptybackend`.
- **52 reclassifications** as boundary-bound, each with its reason recorded.
- **2** turned out not to be paced at all (a `time` reference elsewhere in the
  file, not in the test).

Plus the test #809 asked for: the job runner's periodic retention pass, which
had **never executed in any test** because its hourly ticker ran on real time
under the old `fakeClock`.

Test-only. No production code changes anywhere.

## The conversions

The big clusters the triage map predicted, confirmed by running them:

- `notebook_narration_test.go` (18), `notebook_daily_narrate_test.go` (10),
  `notebook_tasks_enabled_test.go` (6) — the executor stub never execs, so every
  poll over job-queue state collapses to a `synctest.Wait()`.
- `plugin_rpc_test.go` (9), `plugin_supervisor_test.go` (7) — the open question
  was whether the plugin runtime spawns a real interpreter. It does not: these
  drive the RPC over `net.Pipe`, which is channel-backed and bubble-safe, read
  deadlines included.
- `transcript_watcher_abort_test.go` (5), `workspace_keeper_test.go` (5), and
  smaller sets in `browser_test.go`, `reload_test.go`, `daemon_test.go`,
  `spawn_characterization_test.go`, `stop_terminality_test.go`,
  `ticket_buffer_test.go`, `tilecontent_test.go`, `ws_tasks_test.go`.

Converted tests keep their assertions and run the **shipped** intervals rather
than turned-down copies. `bus_test.go` no longer drops `PollInterval` from 5s to
5ms; the plugin supervisor tests now exercise the real restart backoff.

Two helpers land in `internal/daemon/synctest_test.go`: `requireOutbound` and
`requireNoOutbound`, replacing `select { case <-client.send: … case
<-time.After(2s): }`. The negative is the one that changes meaning — outside a
bubble, "nothing was sent" is only ever "nothing yet".

`waitForNudgeDeadline`, `waitForNudgeDeadlineBefore` and `waitForBus` are gone.

## Three new boundary rules

Each found by converting a test and watching it hang:

1. **A claim about a goroutine parked on a lock.** `go doc testing/synctest` is
   explicit that locking a `sync.Mutex` is not durable blocking. A test whose
   subject is "this writer is stuck behind that lock" — `doorbellMu`,
   `autoSettleFireMu`, the spawn lock, the evidence and trace tables — has no
   state a bubble can settle into, so `synctest.Wait()` hangs instead of
   returning. 9 tests.
2. **An assertion on real elapsed wall-clock.** Under a fake clock
   `time.Since(start) > 100ms` reads zero; converting would quietly delete the
   test's only claim.
3. **A goroutine with no exit path.** `go d.wsHub.run()` loops forever over its
   channels, and a bubble ends only when its last goroutine exits. Giving the
   hub a quit channel is a production change; this sweep is test-only.

## Two corrections to #809's map

- `reload_test.go` drives a `fakeReloadBackend`, not `ptybackend` worker
  processes. 3 of its 6 convert; the other 3 are blocked by an fsnotify watcher
  that *setting the notebook root* starts.
- `internal/pty` has no convertible tests — all 14 go through `manager.Spawn` or
  an `os/exec` helper, not the 9/5 split the map recorded.

## The retention pass test

`TestThePeriodicRetentionPassTrimsOnItsOwnInterval` leaves `TrimInterval` at the
shipped `DefaultTrimInterval` — one hour, the value that had never executed —
with a 24-hour retention policy. Through 23 ticks it trims nothing; the tick
after the policy window trims exactly the aged completed job and leaves the
young completed job and the dead one alone. The runner's own log line is the
receipt that the pass ran once rather than zero or twice:
`jobs: trimmed 1 completed job(s) older than 24h0m0s`.

Mutation-checked by setting `DefaultTrimInterval` to 24h in production code:
three assertions fail. That change was reverted.

## Numbers

`go test -count=3`, same machine, this branch against `39c3c8c4`:

| subset | before | after |
|---|---|---|
| the 86 converted daemon tests | 94.6s | 6.0s |
| `internal/daemon` whole package | 403.6s | 327.0s |
| `internal/ptybackend` whole package | 6.46s | 5.48s |
| `internal/jobs` whole package | 0.68s | 0.91s |

The subset number is the one to read; 88.6s of tolerance windows and poll loops
is gone. The whole-package figure moves this time — 86 tests is enough to show
through — but a 6-minute package has minutes of run-to-run spread, so treat it
as directional, not as a receipt. `internal/jobs` grows because it *gained* a
test that runs 25 hours of fake time.

Every converted test is green at `-race -count=20`; all three packages are green
at `-count=1` on the rebased branch.

## One finding worth flagging

`TestStartJobQueueArmsThePeriodicTicks` — the cron-arm flake #809 reported as
retired ("20/20 clean") — **failed once** in a full-suite `-count=3` run at
`39c3c8c4`, i.e. on merged main, unmodified by this PR. It has not reproduced
since: 300 isolated runs and five further full-suite runs are clean, and the
failure message was not captured. The class is not as dead as #809 claimed;
recorded in the plan doc rather than left as a passing-by-luck belief.

## Also

The triage map in `docs/plans/2026-08-09-testing-adoption-wave.md` is updated:
the candidate column is zero, both scans are shown with their methodology, the
new boundary rules and corrections are written down, and the closing state names
its population (tests that landed on main after `39c3c8c4` are not in it).

Changelog fragment: `changelog.d/synctest-remainder-and-retention-pass.yaml`.

### DIFF
--- a/changelog.d/synctest-remainder-and-retention-pass.yaml
+++ b/changelog.d/synctest-remainder-and-retention-pass.yaml
@@ -1,0 +1,11 @@
+kind: internal
+area: testing
+change: >
+  The paced-test triage from the previous sweep now ends at zero unverified
+  candidates: 87 more tests run inside a `testing/synctest` bubble on their
+  shipped intervals, and every test still paced carries a recorded reason —
+  a child process, a real socket, fsnotify, real file mtimes, a goroutine
+  parked on a mutex, or an assertion on real elapsed wall-clock. The daemon's
+  converted subset drops from 94.6s to 6.0s. The job runner's periodic
+  retention pass, which had never executed in any test because its hourly
+  ticker ran on real time, gets one of its own.
--- a/docs/plans/2026-08-09-testing-adoption-wave.md
+++ b/docs/plans/2026-08-09-testing-adoption-wave.md
@@ -147,7 +147,8 @@ Named per the boil-the-ocean rule; each needs an explicit go.
       fallback to fix 1 alone); extended toxiproxy test; live verification
 - [ ] Track B1: three rapid property files (pty scanners, docstore,
       workspacelayout)
-- [ ] Track B2: triage map, then bubble-able conversions
+- [x] Track B2: triage map, then bubble-able conversions (two passes; candidate
+      column at zero)
 - [ ] Track C: AGENTS.md testing-tools note
 - [ ] Track D: none until explicitly picked
 - [ ] Changelog fragments per PR
@@ -172,12 +173,18 @@ Named per the boil-the-ocean rule; each needs an explicit go.
 
 ## Track B2 triage map (2026-08-09)
 
-Measured on `test/synctest-sweep`. A test counts as **paced** when its body
-sleeps, waits on `time.After`, or spins a deadline-bounded poll loop. Tags are
-resolved through the fixtures a test calls, two levels deep, so a test inherits
-the boundary its helper crosses.
+A test counts as **paced** when its body sleeps, waits on `time.After`, or spins
+a deadline-bounded poll loop.
 
-| package | paced | converted here | boundary-bound | candidate, not converted |
+The sweep ran in two passes. The first pass converted what was already known to
+be safe and left a remainder of unverified candidates; the second pass took that
+remainder to zero, either by converting a candidate or by giving it a reason.
+
+**First pass** (the triage itself), measured on `test/synctest-sweep` against
+its merge base. Tags are resolved through the fixtures a test calls, two levels
+deep, so a test inherits the boundary its helper crosses.
+
+| package | paced | converted | boundary-bound | candidate, not converted |
 |---|---|---|---|---|
 | `internal/daemon` | 199 | 20 | 59 | 120 |
 | `internal/jobs` | 21 | 17 | 0 | 2 |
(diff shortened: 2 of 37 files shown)


figgyster (approved):
The remainder is zero and I ran the receipts myself. Verdict: approve.

**What I verified independently**

- Test-only claim holds at the file level: every changed file is `_test.go`, the plan doc, or the changelog fragment. No production bytes anywhere.
- The retention test's mutation check reproduces: set `DefaultTrimInterval` to 24h and the test fails exactly as described — the aged-out job survives, the store holds 3 records instead of 2, and because the `jobs: trimmed` log line only fires when n > 0, the third assertion (exactly one pass) fails on the zero-trim run. That's the third failure, and it's the same three the PR body names. Reverted after.
- The converted clusters run green under `-race -count=3` in 1.7s total — the 94.6s→6.0s shape is real, and it's the shipped intervals doing the work: `bus_test.go` sleeps `bus.DefaultPollInterval` now instead of substituting 5ms, and the supervisor tests exercise the real restart backoff.
- No new skips anywhere in the diff. The removed `time.After` tolerance windows weren't deleted, they were re-homed into `requireOutbound`/`requireDone`, which are strictly stronger claims — after `synctest.Wait()` there's nothing left that could still send, so "nothing arrived" stops meaning "nothing yet".
- Boundary reclassifications carry checkable reasons (autoSettle's `autoSettleFireMu`, delegation's real wall-clock, the fsnotify trio in reload_test). The two "not paced after all" corrections are in the doc with their scanner artifact.

**One failure in my full run — not this PR**

`TestDaemon_StartInstalledPlugins_SpawnsProviderPlugin` fails in my sandbox (5s timeout on the helper handshake). It's not in this PR's file list, and it fails 4/4 on origin/main in the same box — same environment class as #818's two plugin tests. The whole three-package run is otherwise green here, and CI's Daemon + both Linux daemon jobs are green on this head.

**The crown piece**

The retention pass had never executed in any test — its hourly ticker ran on real time while tests ran on an injected clock. The new test runs 25 hours of bubble time on the shipped interval and pins the exact boundary: 23 ticks trim nothing, the tick after the policy window takes precisely the aged completed job, and the runner's own log line is the receipt. That's a witness for a crime nobody had ever seen committed.

The honest flake flag on `TestStartJobQueueArmsThePeriodicTicks` — one failure on merged main, unreproduced since, recorded rather than papered over — is the right way to carry that. Good sweep. This is the #809 ledger closing square.

PR #821, opened 2026-08-09
fix(tickets): let a revived agent's first report land

victorarias:
## The bug

A delegated agent that dies and is revived could not report on its ticket on the first attempt — deterministically, every time.

Dying and coming back writes two events on the bound ticket, both authored by `attn` itself: the crash stamp (`agent process ended mid-flight without reporting`) and the crashed→working flip (`session was reloaded and is running again`); reconciliation usually adds a verdict comment as a third. The CLI mutation gate (`store.withTicketMutation`) refused *any* write while unread activity existed on the ticket — it printed the activity, marked it read, and exited 1 so the agent reads before it writes. So a revived agent's first `attn ticket status` / `ticket comment` / `ticket attach` always failed, on activity describing its own death.

The retry would have worked (the refusal clears the unread), but agents do not retry. Observed live on 2026-08-09: a revived delegated agent was refused once, told its user "done", and its report never landed.

## The contract shipped

**Only another participant's word gates a write.** attn's own bookkeeping about a ticket never does.

- Unread activity authored by the user, the chief, or a sibling agent still refuses the mutation and exits 1 — their word may change what the agent was about to write. The message now names the contract explicitly: *"unread ticket activity is shown above and is now marked read; the requested update did NOT run — retry the same command to apply it."*
- Unread activity authored by `attn` (crash stamp, revive flip, reconciliation verdict, an automation's failure note) is delivered as catch-up **beside a successful mutation**. It is news to the agent — it was dead when the events landed — but there is nothing to answer.
- A mixed batch is gated: the peer event decides, and everything consumed is shown.

Cursor advance, attention repair, and nudge-countdown cancellation are unchanged and now run on both paths, so delivering catch-up alongside a success does not leave a stale nudge armed.

### Why not the alternatives

- **Revive pre-clears activity up to the revive point.** It loses real messages: a chief comment posted while the agent was down would be marked read without ever being shown, and a revived agent's context replays its *conversation*, not the ticket thread. It is also revive-specific — the same trap exists wherever attn stamps a ticket about a session that is not alive.
- **Only improve the error text.** Shipped anyway (see the refusal message above), but it is the floor, not the fix. A deterministic first-attempt failure is still a trap when it is labeled.

## What changed

- `internal/store/ticket_mutation.go` — `TicketMutationOutcome` splits "what was consumed" (`CatchUp`) from "did it cost the write" (`Blocked`); `blocksTicketMutation` is the one-line predicate that decides.
- `internal/daemon/ticket_{status,comment,attach}.go` — carry the catch-up on both paths; report `applied`.
- `cmd/attn/main.go` — the three commands print catch-up either way, exit 1 only when refused, and both notices go to stderr so `--json` stdout stays parseable.
- Protocol 220: `applied` on `TicketStatusResult`, `TicketCommentResult`, `TicketAttachResult`.
- `internal/agent/attn_skill/references/tickets.md` — a "Read before you write" section, so an agent knows to retry.

## Verification

**Tests.** `internal/store`: an attn-only unread batch applies and returns its catch-up with the cursor advanced; a batch carrying one peer comment blocks and leaves the ticket alone. `internal/daemon`: `TestRevivedSessionReportsOnFirstAttempt` drives a real crash (`handlePTYExit` mid-flight) and a real respawn, then asserts the first report applies with both records delivered; `TestRevivedSessionStillGatedByPeerActivity` adds a peer comment during the downtime and asserts refusal-then-retry. Mutation-checked: reverting `blocksTicketMutation` to the old always-gate behavior turns both new tests red on their load-bearing assertion.

**Live witness** (throwaway profile `revivetrap`, packaged app + daemon installed from this branch, protocol 220 on both, cleaned up afterwards). A real `claude` agent, ticket bound and working, mid-turn; its pty-worker SIGKILLed by a pid captured from the daemon's own registry; ticket stamped `crashed`; daemon restarted and the app reopened; session revived through `reload_session` and the ticket flipped back to `working`. Then the revived session's **first** report:

```
$ attn ticket status ready_for_review --session <id> --comment "..."
revive-fixture-7-876z
  [22:26:14Z] status_changed by attn (working → crashed)
    agent process ended mid-flight without reporting
  [22:26:32Z] commented by attn
    🩺 Reconciliation verdict — session … the agent process exited on its own …
  [22:28:02Z] status_changed by attn (crashed → working)
    session was reloaded and is running again
ticket revive-fixture-7-876z → in_review
(stderr) ticket status: attn's own updates on this ticket are shown above and are now marked read; your update ran.
exit 0
```

All three attn records delivered; the report landed on the first attempt.

And the gate still holds, on the same live session:

```
$ attn ticket comment revive-fixture-7-876z -m "peer: scope changed, do not close yet" --session <other-session>
$ attn ticket status in_progress --session <id> --comment "back on it"
  [22:28:20Z] commented by <other-session>
    peer: scope changed, do not close yet
(stderr) ticket status: unread ticket activity is shown above and is now marked read; the requested update did NOT run — retry the same command to apply it.
exit 1
$ attn ticket status in_progress --session <id> --comment "back on it"
ticket revive-fixture-7-876z → working
exit 0
```

`--json` on a refused mutation returns `applied: false` with the catch-up in the payload and nothing but JSON on stdout.

## Surfaces walked

- **CLI** — all three mutating commands.
- **Daemon** — the three handlers; the app's optimistic-concurrency path (`expected_event_seq`) never produces catch-up and always reports `applied: true`.
- **Protocol** — `main.tsp` → `generate-types` → `generated.go` + `generated.ts`, `ProtocolVersion` 219→220 in `constants.go` and `useDaemonSocket.ts`.
- **Linux** — pure Go, no platform code.
- **App/plugins/SDK** — no consumer reads ticket `catch_up`; the added field is additive.
- **Docs** — changelog fragment; attn skill reference.

The epic branch's `pi-host-delegate` scenario has a revived-agent leg that becomes the packaged e2e regression for this once that epic merges. I do not own that scenario; this PR does not touch `epic/pi-headless-host`.

## Noticed, not fixed

After a hard worker SIGKILL with the daemon alive, `reload_session` cannot revive the session: `Remove` cannot reach the dead worker's socket, so the backend registry keeps the id and `Spawn` returns `session already exists`. The session only becomes reloadable after a daemon restart clears the stale entry. Pre-existing and unrelated to this change — flagging it, not fixing it here.

https://claude.ai/code/session_019UmUszqz2JEhp6S4g3Fgzd

### DIFF
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -923,14 +923,16 @@ func runTicketStatus(args []string) {
 		os.Exit(1)
 	}
 	if result.CatchUp != nil {
-		printTicketMutationCatchUp("ticket status", result.CatchUp, parsed.JSON, result)
-		os.Exit(1)
+		printTicketMutationCatchUp("ticket status", result.CatchUp, result.Applied, parsed.JSON)
 	}
 	if parsed.JSON {
 		printJSON(result)
-		return
+	} else if result.Applied {
+		fmt.Printf("ticket %s → %s\n", result.TicketID, result.Status)
 	}
-	fmt.Printf("ticket %s → %s\n", result.TicketID, result.Status)
+	if !result.Applied {
+		os.Exit(1)
+	}
 }
 
 type stringListFlag []string
@@ -1021,16 +1023,18 @@ func runTicketAttach(args []string) {
 		os.Exit(1)
 	}
 	if result.CatchUp != nil {
-		printTicketMutationCatchUp("ticket attach", result.CatchUp, parsed.JSON, result)
-		os.Exit(1)
+		printTicketMutationCatchUp("ticket attach", result.CatchUp, result.Applied, parsed.JSON)
 	}
 	if parsed.JSON {
 		printJSON(result)
-		return
+	} else if result.Applied {
+		fmt.Printf("attached %d artifact(s) to ticket %s → %s\n", len(result.Artifacts), result.TicketID, result.State)
+		for _, artifact := range result.Artifacts {
+			fmt.Printf("  %s\n", artifact.Path)
+		}
 	}
-	fmt.Printf("attached %d artifact(s) to ticket %s → %s\n", len(result.Artifacts), result.TicketID, result.State)
-	for _, artifact := range result.Artifacts {
-		fmt.Printf("  %s\n", artifact.Path)
+	if !result.Applied {
+		os.Exit(1)
 	}
 }
 
@@ -1310,23 +1314,32 @@ func runTicketComment(args []string) {
 		os.Exit(1)
 	}
 	if result.CatchUp != nil {
-		printTicketMutationCatchUp("ticket comment", result.CatchUp, parsed.JSON, result)
-		os.Exit(1)
+		printTicketMutationCatchUp("ticket comment", result.CatchUp, result.Applied, parsed.JSON)
 	}
 	if parsed.JSON {
 		printJSON(result)
-		return
+	} else if result.Applied {
+		fmt.Printf("commented on ticket %s\n", result.TicketID)
 	}
-	fmt.Printf("commented on ticket %s\n", result.TicketID)
+	if !result.Applied {
+		os.Exit(1)
+	}
 }
 
-func printTicketMutationCatchUp(command string, bundle *protocol.TicketEventBundle, jsonOutput bool, result any) {
-	if jsonOutput {
-		printJSON(result)
-	} else {
+// printTicketMutationCatchUp renders ticket activity the caller had not read, which
+// the daemon consumed on its behalf, and says whether the write still ran. Both
+// notices go to stderr so --json keeps stdout parseable. The retry sentence is
+// load-bearing: the events are marked read now, so the same command a second time
+// goes through, and an agent told only "did not run" stops there.
+func printTicketMutationCatchUp(command string, bundle *protocol.TicketEventBundle, applied, jsonOutput bool) {
+	if !jsonOutput {
 		fprintTicketInbox(os.Stdout, &protocol.TicketInboxResult{Bundles: []protocol.TicketEventBundle{*bundle}})
 	}
-	fmt.Fprintf(os.Stderr, "%s: unread ticket activity was shown above; the requested update did not run — review it and retry\n", command)
+	if applied {
+		fmt.Fprintf(os.Stderr, "%s: attn's own updates on this ticket are shown above and are now marked read; your update ran.\n", command)
+		return
+	}
+	fmt.Fprintf(os.Stderr, "%s: unread ticket activity is shown above and is now marked read; the requested update did NOT run — retry the same command to apply it.\n", command)
 }
 
 // ticketIDArgs is a single ticket-id positional plus the common session/json flags —
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -1332,14 +1332,19 @@ func runTicketComment(args []string) {
 // load-bearing: the events are marked read now, so the same command a second time
 // goes through, and an agent told only "did not run" stops there.
 func printTicketMutationCatchUp(command string, bundle *protocol.TicketEventBundle, applied, jsonOutput bool) {
-	if !jsonOutput {
+	// --json leaves the events in catch_up rather than rendering them, so the
+	// notice must not point at output that is not there.
+	where := "shown above"
+	if jsonOutput {
+		where = "in this result's catch_up"
+	} else {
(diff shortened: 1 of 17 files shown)


cmd/attn/main.go
```diff
@@ -1310,23 +1314,32 @@ func runTicketComment(args []string) {
...
+// notices go to stderr so --json keeps stdout parseable. The retry sentence is
+// load-bearing: the events are marked read now, so the same command a second time
+// goes through, and an agent told only "did not run" stops there.
+func printTicketMutationCatchUp(command string, bundle *protocol.TicketEventBundle, applied, jsonOutput bool) {
+	if !jsonOutput {
 		fprintTicketInbox(os.Stdout, &protocol.TicketInboxResult{Bundles: []protocol.TicketEventBundle{*bundle}})
 	}
-	fmt.Fprintf(os.Stderr, "%s: unread ticket activity was shown above; the requested update did not run — review it and retry\n", command)
+	if applied {
+		fmt.Fprintf(os.Stderr, "%s: attn's own updates on this ticket are shown above and are now marked read; your update ran.\n", command)
```
figgyster: Both notices say "shown above", but on `--json` the inbox never renders — the events live in the payload, which prints after this notice. The machine signal is right (`applied: false` + `catch_up` in the payload, stdout parseable); the sentence just points at nothing on that path. Suggest wording that fits both, e.g. "in the payload" for JSON consumers. Not a blocker — the substance is all there.

figgyster (approved):
The case file on this one is a death and a resurrection, and the PR fixes exactly that shape of bug.

**Approve.** The trap was real — a revived agent's first report died on attn's own bookkeeping about its death, and the live witness shows the actual damage: refused once, told the user "done", report never landed. The contract shipped is the right one: only another participant's word gates a write; attn's own records ride along as catch-up beside a successful mutation; a mixed batch is gated by the peer event. Clean line to draw, easy for an agent to learn.

**Checked this turn, fresh checkout at e0919c23:**

- Store + protocol + daemon ticket suites green locally (full store run 117s, daemon Ticket set 2.5s; both new revive tests and the updated catch-up/attach tests pass).
- Mutation-check reproduced by hand: reverted `blocksTicketMutation` to always-gate. `TestTicketMutationAppliesThroughAttnAuthoredCatchUp` goes red on `updated=<nil> … Blocked:true`; `TestRevivedSessionReportsOnFirstAttempt` goes red on "want applied". Restored after. The body's claim checks out — both load-bearing tests die with the old predicate.
- Protocol 220 lockstep complete: `constants.go`, `useDaemonSocket.ts`, `main.tsp`, both generated files; `TestFrontendProtocolVersionMatchesDaemon` gates the live handshake. One stale worry retires: the old invariant about bumping `native-ui/crates/attn-protocol/src/lib.rs` is moot — the native client was archived at 0c3a38fdf, nothing in the tree pins a second protocol version.
- The trust boundary on `TicketAuthorAttn` holds: only daemon-internal paths write that author (crash stamp, revive flip, reconciliation verdict, automation failure notes); client comments are authored as the session. No path for a peer to forge the non-blocking identity.
- `afterTicketMutationCatchUpLocked` no-ops on an empty slice, so the success-path call is safe, and when catch-up rides along it cancels the nudge countdown — the "no stale nudge armed" claim is real.
- CI green across the board; Plugins/Tauri skipped on a non-release branch, as usual.

**One nit on record, non-blocking** (inline): the "shown above" wording in the stderr notices assumes the terminal path. On `--json` the inbox never renders — the events sit in the payload, printed after the notice. The machine signal is right; the sentence just points at nothing on that path.

The "noticed, not fixed" `reload_session` stale-entry flag is a fair pre-existing head-scratcher, and I agree it doesn't belong in this diff.

Same loop as the last few: trap on the table, surgical contract, receipts. Good work.

cmd/attn/main.go
victorarias: Fixed in 070e6c14. The notice now names where the events actually are: "shown above" on the terminal path, "in this result's catch_up" under --json.

figgyster (approved):
Loop-closer on 070e6c14: the one nit is answered the way this repo answers nits — surgical, and pointed at the machine consumer. `printTicketMutationCatchUp` now names where the events actually are per path: "shown above" on the terminal path, "in this results catch_up" under `--json`. Checked the delta myself: 8 lines, one function, no test pins the old wording, gofmt clean, diff-check clean, and the JSON branch still skips the inbox render so stdout stays parseable. Thread resolved on your side.

Same loop as #814/#818/#820 — finding, fix with receipts, closed. Six-for-six this week; the ledger stays square. Head is at 070e6c14, re-approved there.

PR #822, opened 2026-08-09
fix(jobs): retry a refused cron arm and say which kind and why

victorarias:
`armCron` logged its enqueue error and moved on. A cron kind that fails to arm
leaves **no record at all** — dispatch never selects it and `finish()` never
re-arms it — so that duty was gone until the next daemon restart, and the only
evidence was work that silently never happened. The failing tripwire test
(`TestStartJobQueueArmsThePeriodicTicks`) said `notebook_cron is not armed` and
nothing more, which is why the flake behind it survived a triage.

## The failure that actually happened

Instrumenting `AcquireDirLock` and running the daemon suite reproduced it:

```
BEFORE (unmodified main, go test ./internal/daemon -count=3 -v)
  --- FAIL: TestStartJobQueueArmsThePeriodicTicks   (2 of the 3 passes)
  72 × AcquireDirLock REFUSED at <pkg data dir>/.runner.lock (held by pid 15192, ours 15192)

AFTER (same command, this branch)
  PASS, 0 refusals
```

The queue's single-instance lock was one file in `config.DataDir()`. A test
binary builds many daemons against one package-scoped data dir, so one live test
daemon's runner refused every other test daemon's `Runner.Start()` — with *our own
pid* in the lock file, which `lockHolderAlive` correctly reads as live. `Start()`
returned `ErrAlreadyRunning`, `armCron` never ran, and the entry was missing.
Both diagnostics were swallowed twice over: `armCron` dropped its error, and
`NewForTesting` sets `logger: nil`, so `d.logf` is a no-op in tests.

## What changed

- **Arming retries.** A refused arm is parked on the queue's own backoff (1m, 2m,
  … capped at 1h) and re-attempted by the dispatch loop until the record lands. A
  cron entry never dies, so there is no attempt cap — but there is no hot loop
  either, since every retry is at least one backoff apart.
- **Every failure names itself.** The log carries the kind, the store's error, the
  attempt, and the next try; a successful retry logs the recovery. `CronEntry`
  returns *why* a kind has no record — a parked arm's cause, or a `Start` that
  failed before arming ran — instead of `(nil, nil)`, which reads exactly like a
  healthy entry nobody has written yet. That is what the red test now prints.
- **`rearmCronLocked`'s sibling failure states its consequence**: the claim write
  left the row `RUNNING`, which dispatch never selects, so only the orphan
  recovery at the next `Start` puts it back in the rotation.
- **The lock moves to the daemon's own data root**, beside the `attn.pid` that
  already marks one daemon per root. The same file in production —
  `ValidateDaemonIsolation` refuses to start a daemon whose data root is not the
  profile data dir unless its database is separately isolated, so a root of its
  own always means a store of its own. In a test process it is what stops one
  daemon's runner from blocking every other one's queue.

## Verification

Live, on a real headless daemon (own profile, own data dir, real SQLite), with a
trigger forcing the `notebook_cron` insert to fail:

```
00:30:13 jobs: CRON notebook_cron IS NOT ARMED (attempt 1): enqueue cron entry:
         store: upsert job 96d69a3a…: database is locked (simulated)
         — it will not fire until it is; retrying at 2026-08-09T22:31:13Z
   (trigger dropped)
00:31:14 jobs: cron notebook_cron is armed after 1 failed attempt(s)

sqlite> select kind, unique_key, state from jobs;
notebook_cron|cron|queued
```

Before this change that heartbeat stayed dead until someone restarted the daemon.

Automated: two new bubbled tests in `internal/jobs` (a store that refuses the
arming write, and a runner refused the single-instance lock), both
mutation-checked — dropping the retry or the `CronEntry` reason turns each red on
its load-bearing assertion. `go test ./internal/jobs -race`, the full Go suite,
and `./internal/daemon -count=3` are green.

Tier: background-runner change with no protocol, PTY, or UI surface (nothing on
the wire reads cron entries), so the witness is a real daemon driven headlessly
rather than the packaged app.

### DIFF
--- a/changelog.d/cron-arming-retries-and-says-why.yaml
+++ b/changelog.d/cron-arming-retries-and-says-why.yaml
@@ -1,0 +1,11 @@
+kind: fixed
+area: jobs
+change: >
+  A periodic tick whose recurring record the database refused to write is no
+  longer lost until the next daemon restart. Arming a cron kind now retries on
+  the queue's own backoff until the record lands, and while it has not, the
+  daemon log and any caller reading the entry both name the kind and the
+  underlying error instead of reporting nothing at all. The job queue's
+  single-instance lock also moves next to the daemon's own attn.pid, so one
+  daemon means one queue — the same file in production, and no longer one lock
+  shared by every daemon a test process builds.
--- a/internal/daemon/job_store.go
+++ b/internal/daemon/job_store.go
@@ -18,8 +18,11 @@ var _ jobs.Store = (*sqlJobStore)(nil)
 // (which imports both internal/jobs and internal/store) so neither of those
 // packages depends on the other.
 //
-// The single-instance lock is a file lock in the profile data dir — one queue per
-// profile, matching one daemon per profile.
+// The single-instance lock is a file lock in the daemon's own data root, beside
+// the attn.pid that already marks one daemon per root — one queue per daemon.
+// config.ValidateDaemonIsolation refuses to start a daemon whose data root is not
+// the profile data dir unless its database is separately isolated, so a root of
+// its own always means a store of its own.
 type sqlJobStore struct {
 	store   *store.Store
 	lockDir string
@@ -27,9 +30,13 @@ type sqlJobStore struct {
 }
 
 // newSQLJobStore builds the adapter over the daemon's store, locking under the
-// profile data dir.
+// daemon's data root.
 func (d *Daemon) newSQLJobStore() *sqlJobStore {
-	return &sqlJobStore{store: d.store, lockDir: config.DataDir(), log: d.logf}
+	lockDir := d.dataRoot
+	if lockDir == "" {
+		lockDir = config.DataDir()
+	}
+	return &sqlJobStore{store: d.store, lockDir: lockDir, log: d.logf}
 }
 
 // jobSubject names the entity a background job acts on — the workspace, session,
--- a/internal/daemon/job_store_testing_test.go
+++ b/internal/daemon/job_store_testing_test.go
@@ -7,11 +7,9 @@ import (
 )
 
 // newTestJobStore builds the queue's persistence over the test daemon's own
-// SQLite store, with the single-instance lock in a per-test temp dir.
-//
-// The lock dir is NOT config.DataDir() (what production uses): a test binary
-// builds many daemons in one process against one scoped data dir, so a shared
-// lock would let the first started runner block every later one.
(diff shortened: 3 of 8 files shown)


figgyster (approved):
Approve. Read the whole case file, then verified the claims myself.

**The autopsy is the review.** The old comment called arming asynchronous and the flake a race the test lost under load. Neither was true: `armCron` runs synchronously in `Start()`, before the dispatch loop launches, and the missing entry was two swallowed diagnostics — one package-scoped `.runner.lock` in `config.DataDir()` refusing every other test daemon's `Start()` with our own pid in the file, and `NewForTesting`'s nil logger eating both errors. The triage filed it as a race; it was a lockfile and a mute button. That's the right shape of a flake fix: a cause, not a reclassification.

**What I verified myself:**

- Structural: `armCron` before `go r.loop()`, so "synchronous" holds. The lock moves to `d.dataRoot` — the dir that already owns `attn.pid`, with `ValidateDaemonIsolation` guaranteeing a root of its own means a store of its own, so production keeps the same file and tests stop colliding. `Eligible` selects only Queued/Failed, and `RecoverOrphans` runs only in `Start`, so the re-arm message's "it will not fire again until the daemon restarts" is exact, not dramatic. The retry lives in memory because a failed arm leaves no durable record to retry from — the dispatch loop is the only engine that can bring it back, and the retryAt gate plus capped backoff (1m→1h, no cap) makes it a retry, not a hot loop. No production code outside `internal/jobs` reads `CronEntry`, so the new missing-record error has no consumer to surprise.
- Empirically: diff-check/gofmt/vet clean; `./internal/jobs -race` green; tripwire 3/3; the full daemon suite at `-count=3` — the exact command your body shows failing 2-of-3 with 72 refusals — green in 210s; both mutation checks red (drop the retry → the store-recovery assertion fails; drop the startErr reason → the never-started test fails), restored tree green after. CI green top to bottom, Daemon and both Linux jobs included.

**One observation, for the record, not a change:** the new "a missing record names its reason" contract has one case the runner can't explain — a cron kind registered *after* `Start()`. `armCron` snapshots handlers at Start and nothing else arms a late registration, so such a kind reads `(nil, nil)`: healthy-looking, unarmed, silent. Pre-existing — nothing in the daemon registers late today — and out of this PR's scope, but the `CronEntry` doc now draws its line exactly where that case sits.

#819's triage ran the paced-test remainder to zero candidates. This is the one that got away, closed with a real cause and receipts. The evidence culture keeps compounding.

PR #823, opened 2026-08-09
fix(workspace): keep zoom mode in one place

victorarias:
Bursting a few hundred session updates at the app could take the React tree
down with minified error #185 (maximum update depth exceeded), reported out of
`onZoomModeChange` in `App.tsx` via `SessionTerminalWorkspace`. Found during
the WS-eviction work (#810) and reproduced twice there, unrelated to that
change.

## Root cause

`SessionTerminalWorkspace` kept its own `zoomActive` flag and pushed it up to
`App` from an effect that listed the parent's callback in its dependencies:

```ts
useEffect(() => {
  onZoomModeChange?.(Boolean(effectiveZoomedPaneId));
}, [effectiveZoomedPaneId, onZoomModeChange]);
```

`App` creates that callback inline inside the workspace map, so its identity
changes on every `App` render, and the callback writes what it receives back
into `App`'s own `zoomModeBySessionId` state. Every parent render therefore
produced a notification, and every notification produced another parent
render. React's eager bail-out on a state write that changes nothing normally
cuts the cycle after one turn, which is why this stayed invisible: the
bail-out is only available when the fiber has no pending update. Under a
sustained stream of daemon session snapshots there always is one, so the cycle
keeps scheduling commits until React's nested-update limit trips and it throws
away the tree.

## The fix

`App` already kept a copy of the flag — `zoomModeBySessionId`, which the dock's
zoom chip reads — so the whole readback existed to keep a duplicate in sync.
The fix is to keep only `App`'s copy: the workspace takes `zoomActive` as a
prop and writes it through `onSetZoomActive` from the shortcut. That removes
the effect, the callback, and the duplicated state, so the child no longer
writes to the parent outside an event handler and the cycle is structurally
impossible rather than merely broken.

One behavior shifts slightly. `App` used to receive the *effective* value,
which went false for a frame whenever the focused leaf left the layout (a pane
closing), so the dock chip flickered and settled back. `App` now holds the
requested mode and the workspace still derives what it renders, so a pane
closing under an active zoom no longer blinks the chip.

An earlier revision of this PR fixed the dependency array with a ref instead.
React Doctor flagged that correctly — the ref write happened during render, and
the effect-to-parent pattern itself was the smell — so it was replaced with
this.

## Scope

A sweep of every `useEffect`/`useLayoutEffect` in `app/src` that both invokes a
prop callback and lists it as a dependency turned up three siblings —
`MarkdownReader`'s `onAnnotationsCountChange`, `PresentTour`'s
`onAnnotationAnchorsChange`, `WorkspaceDockTile`'s `onRequestContent`. All
three receive a stable identity from their parent (a `useState` setter or a
`useCallback`), so none of them can churn. This callback was the only inline
one, and no daemon-side behavior changed.

## Tests

`SessionTerminalWorkspace.zoomMode.test.tsx` mounts the workspace under a
parent shaped like `App` — re-rendering off an external store, holding the zoom
flag, handing down a fresh inline setter each render — and covers three things:
the workspace never writes zoom mode back on its own (including across 300
external store updates), the shortcut toggles it through the parent and the
rendered layout follows, and an active zoom survives a burst.

Two existing zoom tests rendered the workspace standalone and relied on it
owning the flag; they now render through a small host that holds it, which is
what `App` does.

Full frontend suite, macOS 26.5.2 arm64: 233 files, 2607 passed / 15 skipped
(2622). CI's Frontend job on ubuntu-latest reports 2602 passed / 20 skipped
over the same 2622 — five specs skip on Linux and run here.

## Live verification

Isolated profiles installed from this branch; bundled preflight PASS. A probe
created shell sessions in the running app, zoomed a split pane, and then drove
the daemon WebSocket directly with 1200 `rename_session` commands — each
publishes `session.renamed`, whose projection re-pushes the session snapshot to
every client.

On the reshaped build (profile `r185zoom`, 6 sessions, 1200 messages):

```
CONTROL maximize after toggle: pane-6e71cdba…    CONTROL maximize after untoggle: null
zoom before toggle: null
zoom after toggle: pane-6e71cdba…
zoom after 1200-message burst: pane-6e71cdba…
zoom after untoggle: null
ping={"frontendReady":true,"pong":true}
```

Zoom arms, holds through the burst, and releases; the app stays responsive.
The maximize line is a control — it shares the same shortcut gate and this
change does not touch it.

The loop itself was measured earlier against the unmodified code on profile
`r185burst`, by counting calls into `App`'s `onZoomModeChange` through a
temporary bridge field: **1674** notifications while creating six sessions and
**189** during the burst, against **33** and **0** with the fix. The 33 are the
genuine ones — mount and unmount across six workspaces. With the state lifted
the count is now structurally zero, since nothing calls back.

Worth stating plainly: neither build actually crashed under this burst — the
app stayed responsive both times. What the runs show is the runaway feedback
loop the crash reports, measured live, and its removal. The original crash was
hit while the app was simultaneously being pushed into WebSocket eviction,
which this probe does not recreate.

Both profiles were removed with `attn profile clean` after the runs.

### DIFF
--- a/app/src/components/SessionTerminalWorkspace/index.tsx
+++ b/app/src/components/SessionTerminalWorkspace/index.tsx
@@ -61,6 +61,10 @@ function suppressTerminalMouseDuringResize(durationMs = RESIZE_MOUSE_SUPPRESSION
   );
 }
 
+// The tile's content request is an effect dependency, so the no-op stand-in
+// has to keep one identity.
+const noRequestContent = () => {};
+
 function zoomLayoutTowardLeaf(node: TerminalLayoutNode, leafId: string): TerminalLayoutNode {
   if (node.type !== 'split') {
     return node;
@@ -1134,7 +1138,7 @@ export const SessionTerminalWorkspace = forwardRef<SessionTerminalWorkspaceHandl
                 onUpdateTile?.(tileLeaf.tileId, tileLeaf.tileParams ?? '', sessionId)
               )}
               onHeaderPointerDown={(event) => beginLeafDrag(tileLeaf.tileId, event)}
-              onRequestContent={onRequestTileContent ?? (() => {})}
+              onRequestContent={onRequestTileContent ?? noRequestContent}
               bodyRef={tileBodyRefFor(tileLeaf.tileId)}
             />
           </div>
--- a/app/src/components/SessionTerminalWorkspace/index.tsx
+++ b/app/src/components/SessionTerminalWorkspace/index.tsx
@@ -186,7 +186,11 @@ interface SessionTerminalWorkspaceProps {
   // daemon use the selected session).
   onOpenMarkdown?: (path: string, sessionId: string) => void;
   onTerminalModelRecovered?: () => void;
-  onZoomModeChange?: (zoomed: boolean) => void;
+  // Zoom is a mode, not a target: it always widens whichever leaf is focused,
+  // so it is a flag rather than a leaf id that has to chase focus. App owns it
+  // because the dock's zoom chip reads it too.
+  zoomActive?: boolean;
+  onSetZoomActive?: (active: boolean) => void;
   onNavigateOutOfSession: (direction: TerminalNavigationDirection) => void;
   onResizeSplit?: (splitId: string, ratio: number) => Promise<unknown> | void;
   // Move an existing leaf (terminal pane or docked tile) beside an anchor leaf,
@@ -241,7 +245,8 @@ export const SessionTerminalWorkspace = forwardRef<SessionTerminalWorkspaceHandl
     onOpenPresentation,
     onOpenMarkdown,
     onTerminalModelRecovered,
-    onZoomModeChange,
+    zoomActive = false,
+    onSetZoomActive,
     onNavigateOutOfSession,
     onResizeSplit,
     onMoveLeaf,
@@ -258,9 +263,6 @@ export const SessionTerminalWorkspace = forwardRef<SessionTerminalWorkspaceHandl
     onRequestTileContent,
   }, ref) {
     const [maximizedLeafId, setMaximizedLeafId] = useState<string | null>(null);
-    // Zoom is a mode, not a target: it always widens whichever leaf is focused,
-    // so it is a flag rather than a leaf id that has to chase focus.
-    const [zoomActive, setZoomActive] = useState(false);
     // The docked tile that currently owns workspace focus, or null when a
     // terminal pane does. `activePaneId` is derived from the focused session, so
     // it can only ever name an agent pane; this is the other half of the focus
@@ -598,10 +600,6 @@ export const SessionTerminalWorkspace = forwardRef<SessionTerminalWorkspaceHandl
       }
     }, [paneIds, ticketOverlayPaneId]);
 
-    useEffect(() => {
-      onZoomModeChange?.(Boolean(effectiveZoomedPaneId));
-    }, [effectiveZoomedPaneId, onZoomModeChange]);
-
     // Focusing the terminal is a leaf handoff, not just a DOM focus call: it
     // has to release any focused tile. `activePaneId` does not change here (the
     // pane was already the session's active one), so without the release the
@@ -767,14 +765,14 @@ export const SessionTerminalWorkspace = forwardRef<SessionTerminalWorkspaceHandl
     }, [activeLeafId, activeLeafIsTile, handleClosePane, onUndockTile]);
 
     const toggleMaximizeActivePane = useCallback(() => {
-      setZoomActive(false);
+      onSetZoomActive?.(false);
       setMaximizedLeafId((current) => (current ? null : activeLeafId));
-    }, [activeLeafId]);
+    }, [activeLeafId, onSetZoomActive]);
 
     const toggleZoomActivePane = useCallback(() => {
       setMaximizedLeafId(null);
-      setZoomActive((current) => !current);
-    }, []);
+      onSetZoomActive?.(!zoomActive);
+    }, [onSetZoomActive, zoomActive]);
 
     // Focus a leaf by id: a tile takes DOM focus in place (the session's pane
     // focus is left alone), a pane goes through the session-level focus path.
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -3916,11 +3916,12 @@ function AppContent({
                         setActiveSession(paneSessionId);
                       }
                     }}
-                    onZoomModeChange={(zoomed) => {
+                    zoomActive={Boolean(zoomModeBySessionId[workspace.id])}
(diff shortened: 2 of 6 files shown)


app/src/components/SessionTerminalWorkspace/index.tsx
```diff
@@ -298,9 +298,16 @@ export const SessionTerminalWorkspace = forwardRef<SessionTerminalWorkspaceHandl
...
+    const onZoomModeChangeRef = useRef(onZoomModeChange);
 
     activePaneIdRef.current = activePaneId;
     isActiveSessionRef.current = isActiveSession;
+    // App re-creates this callback every render and writes what it receives
+    // back into its own state, so depending on its identity turns each parent
+    // render into another notification and another parent render. Under a burst
+    // of daemon session snapshots that cycle never settles and React aborts the
+    // tree with error #185.
+    onZoomModeChangeRef.current = onZoomModeChange;
```
figgyster: React Doctor flags this write as `no-ref-current-in-render` (the error in its 79/100). The concern is theoretical here: the ref is only read post-commit by the notification effect, and a commit implies this render body ran and wrote it — a discarded concurrent render can't leak a stale callback into committed work. Same pattern as the three sibling refs in this block, which is why I'm not blocking on it.

If you want the doctor's score back: a useLatest-style hook (ref updated in an effect) silences it with identical behavior, as long as the hook is called above the notification effect — effects flush in declaration order, so the notification effect always sees the fresh value. The doctor's other three warnings at L609 are the inherent cost of notifying a parent from an effect; no fix shape avoids those.

figgyster (approved):
Verified the case on the branch, not just the write-up.

The mechanism holds up. App hands down a fresh inline callback on every render and writes the readback into its own state. That's a feedback loop with the throttle off; the eager state bail-out usually cuts it after one turn, but under a fiber with a pending update it can't, so a burst runs it straight into #185. The ref readback is the right shape: fire on the value, read the callback through a ref.

Ran the witnesses myself:

- The new test is an eyewitness, not a bought one. Ran it against the pre-fix code: all three fail with exactly the outputs you listed (2 initial calls, 300 spurious, `[[false],[true],[true]]`).
- Full frontend suite on the branch: 233 files / 2602 green, matching your count to the test. tsc clean, `git diff --check` clean.
- The scope sweep checks out. I re-ran it mechanically (every useEffect/useLayoutEffect in app/src that invokes a prop callback and lists it in deps, direct calls only): the three siblings you named are the only ones in this class, and all three get stable identities from their parent — two useState setters, one useCallback. The only near-miss my broader net caught (LocationPicker's `onGetRecentLocations`) is a useCallback too, and it writes local state, not the parent's — no loop shape. The sweep is airtight.
- The live numbers are a solid receipt, and the honesty in them is worth stating: you measured the loop, not the crash, and said so plainly. That's the difference between a receipt and a boast.

CI: Frontend and 2/3 E2E green, Changelog/Changes pass, Daemon/Plugins/Tauri skipping as expected for a frontend-only change. One E2E shard still running.

Two nits, neither a blocker:

1. Line 1146, `onRequestContent={onRequestTileContent ?? (() => {})}` — that fallback creates a fresh identity every render whenever the prop is absent. Inert today (App always passes the stable useCallback, and the fallback is a no-op anyway), but it's the exact shape this PR just removed, sitting in the same file. Hoist it to a module constant and the sweep stays airtight even in the dead branch. Detail on the React Doctor error in the inline note.

Approved on head.

figgyster (approved):
Re-verified at the rewritten head (3cc2924). The force-push replaced the ref-based fix with a better one, and I'd rather say that plainly: lifting zoom mode into App as a controlled prop is stronger than the useLatest I suggested. App already held a mirror of the flag for the dock chip, so the mirror becomes the source — the workspace's local copy, the notification effect, and the callback ref are all deleted, and the child writes the flag only from discrete shortcut handlers. The #185 loop is now structurally impossible, not throttled: there is no render-effect write path left to loop on.

Receipts at head, all mine:

- Full frontend suite: 233 files / 2602 green, matching the claim. tsc clean, `git diff --check` clean.
- Workspace cluster green (161 tests), including the two adapted test files (ZoomHost wrapper is a clean way to keep standalone renders honest) and the new `zoomMode.test.tsx`. Ran the new test against the pre-fix component: the toggle-through-parent assertion fails (`[]` vs `[[true]]`); the other two pass because the old component ignores the new props entirely. That's the right shape — the loop-pin moved from the test to the architecture, which is the stronger place for it. The old test was 3-for-3 red on pre-fix code; the new one pins the contract itself.
- React Doctor at head: no new issues, 0 errors, 0 warnings. The ref-write error and all three notify-via-effect warnings are gone, not silenced. Resolved my own stale thread on the ref write accordingly.
- CI: Changelog/Changes pass, Frontend + E2E spinning up fresh, Daemon/Plugins/Tauri skipping as expected for a frontend-only change.

Two nits, neither a blocker:

1. The PR body now describes a build that doesn't exist. It says the notification "reads the callback through a ref" and names `zoomModeNotify.test.tsx`, but head ships the controlled-prop design and `zoomMode.test.tsx`. The root-cause section is still accurate and the scope sweep is untouched, but the fix narrative, the test-section file name, and the live table's post-fix column (measured on the intermediate build) need a pass to match what's on the branch. The commit message is the accurate story.

2. Carried over from my first review, still open: line 1137, `onRequestContent={onRequestTileContent ?? (() => {})}` — the fallback creates a fresh identity every render whenever the prop is absent. Inert today (App always passes the stable useCallback), but it's the exact shape this PR just eliminated, sitting in the same file. Hoist it to a module constant and the sweep stays airtight even in the dead branch.

Approved on head.

figgyster (approved):
Both nits closed, verified at head 80fe9ca1.

- The fallback fix is exactly right: module-level `noRequestContent`, one identity in the absent-prop branch, and the comment says why. +5/−1, surgical. tsc clean, `git diff --check` clean, full suite 2602 passed / 20 skipped on my Linux run — matching the CI figure you now quote in the body.
- The platform qualifier is the #820 move done properly: macOS 26.5.2 arm64 2607/15 vs ubuntu-latest 2602/20, same 2622 total, five Linux skips named. The figure reconciles; the ledger stays honest.

And your framing of the test trade is better than mine. The counting test pinned the loop, which only works while the notification exists — it would have died with the mechanism it counted. `never writes zoom mode back to the parent on its own` pins the invariant, so it fails exactly when someone reintroduces the child-owned flag. That's the test that outlives the fix.

Context, not a finding: React Doctor's two warnings at this head are on ConversationPane and LocationPicker — files this PR never touched. Its base-detection, not this diff; the check passes.

CI: Frontend, Changelog, Changes green; E2E 1/3 and 2/3 green, 3/3 still running; Daemon/Plugins/Tauri skipping as expected. #823 is square — re-approved on head.

PR #824, opened 2026-08-09
fix(tickets): attach a chief's delegations to the role, not to its session

victorarias:
## What

When the chief of staff delegates, the ticket now attaches to the chief **role**
alone. Handing the role to another agent hands the ticket notifications with it:
the retired chief goes quiet, the new chief is nudged.

Before, the ticket attached twice — to the durable role, and to the session that
happened to hold the role, as a personal subscription. Only the role moved on a
handover, so an agent that had been chief kept being nudged about every ticket it
had ever delegated, for as long as it ran, while the new chief received the same
activity.

Genuinely personal participation is untouched: an explicit `attn ticket subscribe`
still works, and an ordinary session delegating a ticket the chief role happens to
own stays attached as itself.

## Which representation, and why

Two ways to make the attachment role-bound: record it at **creation time**, or
retarget the creator's rows at **transfer time**. Creation time, because the ticket
already had the right vocabulary and transfer time would need to guess which of a
retired chief's rows were role work.

Concretely, `ticket_events` gains `author_role`: the durable role the author was
acting as. `author` stays the session for audit provenance; participation follows
`author_role`. The delegation paths (mint and adopt) stamp it and stop subscribing
the acting chief personally.

That replaces migration 82's approximation — *"a `created` event on a role-owned
ticket belongs to the role"* — which had two problems the stamp does not:

- it missed the **adopt** path entirely, where the chief writes `assigned` and
  `status_changed` rather than `created`, so the acting session stayed attached;
- it fired on tickets it should not have. A backlog ticket filed by session X and
  later adopted by the chief for a delegation silently dropped **X's**
  participation, because X's `created` event was on a now-role-owned ticket.

**One deliberate behavior change beyond the chief.** Fixing the second point means
the filer of an adopted backlog ticket keeps its participation, where today it
loses it. Measured before the change: filing a ticket as `someone-else` and adopting
it as the chief left participants `[assignee, role:chief_of_staff]` — the filer was
gone. Call it out if you would rather keep the old carve-out; it is one `WHERE`
clause.

## Migration 99

Carries every row. It recognizes the delegating transaction by its timestamp — the
role owner row is written in the same transaction as the events beside it, so they
share a second — stamps those events with the role, and deletes only the acting
session's personal subscription. Restricted to `created`/`assigned`/`status_changed`
so a comment landing in the same second is not swept up, and the subscription it
deletes must carry the delegating second itself, so a subscription the chief made by
hand at some other time survives and a bystander who happened to subscribe in that
second keeps theirs.

Dry run against a copy of a real `~/.attn/attn.db` (97 → 99, 41 tickets, 239 events):

```
events stamped with a role:  5 (all chief-minted, all `created`)
ticket_subscriptions:        66 -> 66   (nothing to remove on this database)
ticket_events:               239 -> 239
ticket_event_cursors:        73 -> 73
tickets:                     41 -> 41
ticket_role_owners:          5 -> 5
```

`TestMigration99DetachesPastChiefSessionsFromTheirDelegations` covers the case that
database does not have, seeded from a populated pre-99 schema (old column dropped,
migration 82's view restored) with both shapes of delegation.

## The parked test

`TestChiefTicketContinuityAcrossRoleTransfer` was the one test the synctest sweep
(#809, #819) could not convert: its nudge count could not tell a role nudge from a
personal one, so it only passed with parked windows and a hand-fired timer. With the
distinction real it runs bubbled, and the windows run out instead — the retired
chief's zero nudges is now a fact about attachment rather than a timer that had not
come due.

## Verification

Mutation-checked both halves of the fix (restore the personal subscription; drop the
`created` stamp) — each turns three tests red, including the bubbled one at its
role-nudge assertion. Dropping the migration's `DELETE` turns the migration test red
on both ticket shapes.

Live, on a throwaway profile (`rolefix`, preflight PASS, cleaned up after): promoted
a session to chief, delegated from it, transferred the role to a second session, and
had the delegated agent report `ready_for_review`.

```
minted ticket: ticket_events.author_role = chief_of_staff
               ticket_subscriptions      = (empty)
               participants              = [assignee, role:chief_of_staff]

chief A (retired):  no unread ticket activity
chief B (current):  attn-fix-chief
                      created by chief-a-live
                      status_changed by <agent> (working → in_review)
```

## Surfaces

CLI (`attn ticket inbox`, exercised live), daemon and store, the shipped delegation
reference (participation now stated as role-bound), a changelog fragment. No protocol
change — `author_role` is server-side routing and never crosses the wire — so no
version bump and no frontend change. Pure Go and SQL, so Linux is unaffected.

### DIFF
--- a/changelog.d/chief-delegation-participation-follows-the-role.yaml
+++ b/changelog.d/chief-delegation-participation-follows-the-role.yaml
@@ -1,0 +1,14 @@
+kind: fixed
+area: daemon
+change: >
+  A delegation the chief of staff starts now attaches to the chief role rather
+  than to the session holding it, so handing the role to another agent hands over
+  the ticket notifications with it. The retired chief goes quiet and the new one
+  is nudged.
+symptom: >
+  The chief was attached to the tickets it delegated twice — once as the durable
+  role and once as itself. Only the role moved on a handover, so an agent that had
+  been chief kept being nudged about every ticket it ever delegated, for as long
+  as it ran, while the new chief received the same activity. Existing tickets are
+  corrected on first launch; a ticket someone filed and the chief later picked up
+  for a delegation also stops dropping the filer from its notifications.
--- a/internal/agent/attn_skill/references/delegation.md
+++ b/internal/agent/attn_skill/references/delegation.md
@@ -25,7 +25,9 @@ session — the brief is its description, the session is its assignee — and yo
 participant on the ticket you created. That ticket is the channel in both
 directions: the agent reports its work state onto it, and you reach the agent with
 `attn ticket comment <ticket-id> -m "<note>"`. The chief of staff is also a
-participant on every delegation ticket, whoever started it.
+participant on every delegation ticket, whoever started it — and when the chief is
+the one delegating, that participation belongs to the ROLE. It moves to whoever
+holds the role next, instead of following the session that delegated.
 
 Follow-up: all runtimes receive the same ticket nudge when activity remains unread;
 Claude may also arm a Monitor on `attn ticket inbox --watch` to consume updates
--- a/internal/daemon/delegate_ticket.go
+++ b/internal/daemon/delegate_ticket.go
@@ -34,22 +34,14 @@ func ticketSlug(label string) string {
 // session, starting in Working. The chief of staff is always a participant via
 // its durable ROLE identity — owner when it initiated (which is also what marks
 // the session `delegated_from_chief`), subscriber otherwise — so its view
-// survives the session filling the role changing. A creator that IS the chief
-// holds both identities; ticketnotify.ConsumeAll merges them by event seq.
+// survives the session filling the role changing.
 func (d *Daemon) createDelegatedTicket(creatorSessionID string, ownedByChiefRole bool, session *protocol.Session, brief, label, agent string) (string, error) {
 	if d.delegationTicketCreateHook != nil {
 		if err := d.delegationTicketCreateHook(); err != nil {
 			return "", err
 		}
 	}
-	ownerRole := ""
-	chiefRoleIdentity := store.TicketRoleIdentity(store.TicketRoleChiefOfStaff)
-	subscribers := []string{creatorSessionID}
-	if ownedByChiefRole {
-		ownerRole = store.TicketRoleChiefOfStaff
-	} else {
-		subscribers = append(subscribers, chiefRoleIdentity)
-	}
+	ownerRole, subscribers := delegationTicketAttachment(creatorSessionID, ownedByChiefRole)
 	created, err := d.createTicketWithUniqueSlug(store.Ticket{
 		Title:       label,
 		Description: brief,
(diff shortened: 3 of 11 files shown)


figgyster (approved):
Read the whole file, then ran the witnesses. This one holds up.

**The fix, verified.** Mint and adopt now route through one attachment decision — chief delegation attaches the role and nothing else, ordinary delegation attaches the session and pulls the role in as subscriber. `author_role` stamps the delegation-transaction events; the participant view attributes them to the role; `author` stays for audit. The migration's same-second recognition is sound because the code really does write the role owner row and the events in one transaction off one `now` (RFC3339, second precision), and the kind restriction keeps comments out of the sweep.

**The claims, checked.** Full store suite green (127s). Daemon ticket/chief/nudge/delegation suites green, full daemon suite green. Mutation-checked all three halves myself:
- restoring the chief's personal subscription → red on exactly `TestCreateDelegatedTicketParticipants`, `TestChiefCreatedTicketAttachesTheRoleAndNotTheSession`, and the bubbled `TestChiefTicketContinuityAcrossRoleTransfer` (which now runs for real — zero nudges is a fact about attachment, not a parked timer);
- dropping the `created` stamp → red on the same three, with the failure output literally showing `chief-a:true` leaking back into the participants map — the ghost, on screen;
- dropping the migration's DELETE → red on both ticket shapes.
New tests stable at -count=3, diff-check and gofmt clean, CI green on head. The two plugin-driver tests fail here too, but they fail identically on origin/main — same environmental spawn-handshake class as #819, not this diff. And the gofmt self-fix on the fixture is the loop closing itself before I even got here.

**Two notes, neither blocking.**

1. The acknowledged hand-made-subscription loss is narrower than it needs to be. You wrote that a chief's hand-made subscription to its own delegation is indistinguishable from the automatic row — but they're distinguishable by `created_at`: the automatic row is written in the same transaction as the role owner row, so it shares the second; a hand-made one is a separate transaction at a different time. Scoping the DELETE keeps the cleanup and spares the hand-made row:

```sql
DELETE FROM ticket_subscriptions
WHERE EXISTS (
    SELECT 1 FROM ticket_events e
    JOIN ticket_role_owners ro
        ON ro.ticket_id = e.ticket_id AND ro.created_at = e.created_at
    WHERE e.ticket_id = ticket_subscriptions.ticket_id
        AND e.author_role != ''
        AND e.author = ticket_subscriptions.identity
        AND ticket_subscriptions.created_at = e.created_at
);
```

It also narrows the residual same-second race on the adopt path — a non-chief participant who status-changes in the same second the chief adopts would otherwise get both a wrongly stamped event and their subscription swept. The seeded test passes unchanged under this shape. One WHERE clause, same spirit as your kind restriction.

2. The deliberate behavior change — the filer of an adopted backlog ticket keeping their participation — keep it. Dropping a real participant's attachment silently was the bug; the old carve-out was the approximation this whole PR exists to retire.

Approving on head.

figgyster (approved):
Cleaner than my version, that WHERE clause — one condition on the subscription row, no JOIN, and it buys both wins: the hand-made subscription survives, the same-second bystander keeps theirs. The new `watched` shape pins the hand-made case properly, and the auto-row cleanup still lands because the automatic subscription shares the delegating second with the stamped events. Verified: migration tests green at -count=3, daemon ticket suites green, diff-check and gofmt clean, amd64 CI green. That closes the loop at seven-for-seven.

PR #825, opened 2026-08-09
feat(sessions): nisse, attn's own agent

victorarias:
attn gets its own agent. **nisse** is a conversation session: pi's SDK runs the
loop and the model, and everything around that loop — the host process and its
lifetime, the envelope stream, the verbs, the delegation, the pane you read it
in — is attn's. It is not a terminal with an agent in it. There is no PTY, no
screen scraping, and no classifier guessing at what the agent is doing: the
agent says what it is doing, on a wire attn designed.

This merges `epic/pi-headless-host` into `main`. Nine slices' worth of work
already reviewed one at a time (#775, #779, #780, #786, #792, #798, #801, #814),
plus the rename to nisse, one product fix, and the merge pass's own evidence.

## The name

pi is the engine. **nisse** is the agent. A **host** is the process any
conversation agent runs in — agent-agnostic machinery that would run a second
conversation agent unchanged. The three words are kept apart deliberately, and
`docs/glossary.md` now says so.

A nisse is the Scandinavian household spirit that keeps a house going while
everyone sleeps, and works tirelessly for as long as you leave out its porridge.
attn is the house, nisse lives in it, and your attention is the porridge.

The rename is complete rather than layered: driver and registry name, the
`ATTN_NISSE_*` launch environment, the host executable (`bin/attn-nisse`),
scenario names and package scripts, changelog fragments, glossary, plan doc.
No compatibility shim and no dual-accept, because nothing ever shipped under the
old name — `pi-host` lived only on this branch. The plugin package keeps its
`attn-pi` directory: it registers two agents, the PTY-backed `pi` and `nisse`,
and naming the package after either one would be wrong.

## What nisse can do

- **Hold a conversation.** Prompt, streamed reply, steer mid-run, follow-up that
  drains only when the run would otherwise settle. The composer shuts for
  exactly one window — between sending a prompt and the host confirming a run
  exists — and reopens on every way a run can end, including a prompt that
  reached no host at all.
- **Say what it is doing.** `session_ready` / `run_started` / `run_settled`
  carry the attn state directly, applied through the normal `applyState` path.
  A conversation session takes turns, nudges and the queue like any other.
- **Show its work.** Tool cards carry a name, a line and a status; what a tool
  read, wrote or printed is fetched only when a card is opened. The receipt is
  the attach-snapshot corpus — p99 11.6 MB transcript, ~0.4% message text —
  eagerly inlining tool output is how transcripts get that way.
- **Die and come back.** `kill -9` parks the session as `recoverable` with the
  transcript intact and a Reload in the pane. The replacement host reopens the
  same pi session file, mints a new epoch, and the conversation continues.
- **Take a delegation.** A brief arrives as the conversation's first user
  message. The agent works in its own worktree and reports on its own ticket —
  attributed to the delegated session, which means the session id survived
  daemon → host → pi → tool subprocess.
- **Remember a long conversation.** Windowed snapshots, scroll-back paging by
  item anchor, an honest row when retention has dropped history for good,
  resume of an existing session file (by forking, so the source is never
  written to), and a mid-session model switch that moves on the host's answer
  rather than on the click.

## Also in this PR

**A delegated conversation agent gets attn's skill without codex.** pi reads
skills from `~/.agents/skills`, and until now the only thing writing there was
the codex driver's launch preparation. On a machine without codex configured, a
delegated conversation agent got the repository's own guidance but never attn's
skill — no idea it was working a ticket or how to report on one. The host spawn
now installs it directly. Best-effort: a missing skill is a poorer agent, not a
session that refuses to start.

## Endgame watch-list

Every item the epic accumulated, with what happened to it.

| # | Item | Disposition |
|---|---|---|
| 1 | A revived delegated agent's first ticket mutation fails | **Closed.** Fixed by #821 on `main` and picked up in this branch's sync. The `nisse-delegate` early-crash leg is now its packaged e2e regression test: the brief asks for one report, not a retry, and it lands after three attn-authored entries. |
| 2 | The attn skill installs only as a side effect of codex | **Fixed here**, with a mutation-verified test (`TestConversationHostSpawnInstallsTheAttnSkill` fails when the spawn's install is removed). |
| 3 | Identity env block duplicated between PTY and host spawn | **Checked-good.** Both spawn paths write the same block and each is test-pinned separately; `buildSpawnEnv` now carries the back-reference to its twin so the next reader finds both. |
| 4 | Fresh live pass and full CI on the merged head | **Done** — see Verification. |
| 5 | A conversation whose start is gone | **Closed by #814**, on this branch. Retention drops are drawn as their own row rather than a transcript that silently begins mid-thought. |
| 6 | A stale client can hold half a scroll-back | **Accepted.** A page is taken only when the epoch matches and the anchor is the client's own oldest item; a window scrolled elsewhere ignores it rather than splicing a hole. Documented in the glossary. |
| 7 | `snapshotItemKey` and `conversationItemKey` are twins that share no code | **Checked-good.** Both are test-pinned; the host side now names its twin so a refactor of one cannot quietly break paging. |
| 8 | Fork + brief + die inside the first turn | **Accepted**, recorded in the plan doc. The two halves — resume from the picker, briefs from delegation — meet on no product path today; closing it means a delivered-marker in the session dir. |
| 9 | No CLI surface for conversation sessions | **Accepted.** There is no CLI spawn path at all, so resume/model/paging have nothing to hang off. `--resume` and `--model` are the shape when one lands. |
| 10 | `tool_started` carries no pid for the tool subprocess | **Accepted/unowned.** Reaching a tool a hard kill stranded stays open. |
| 11 | `ConversationPane` size and `LocationPicker`'s useState count | **Accepted** — already ruled on in slice review. |

## Verification

Packaged app installed from a throwaway profile (`make install PROFILE=…`, never
`dev`), preflight PASS with protocol 221 across CLI/app/daemon, on the merged
head.

Whole scenario matrix green, serially (the harness is single-tenant):

```
PASS  nisse-conversation   9.9s     PASS  nisse-history    9.5s
PASS  nisse-nudge         30.5s     PASS  nisse-delegate  11.6s
PASS  nisse-revive        10.7s
```

Every step, not just the verdicts: round trip and a second prompt after settle,
close leaves no orphans, working declared on run open, steer at the turn
boundary, idle owes a turn, nudge on idle starts a run, `kill -9` mid-tool-run
parks the session, reload resumes with history, the dead host is gone and the
record names the live one, a cold client draws the whole conversation, killed
before its first word relaunches clean, model switch on the host's word, a new
session picks up an old conversation, a long transcript pages the rest in, the
brief is the conversation's first message, the agent reports onto its own
ticket, a brief survives a crash before the first word.

Plus a hand-driven session, prompt through resume:

```
STREAMING  "A Scandinavian nisse is a small household spirit from the folklore
            of Denmark and Norway…"                        (sendLabel: Steer)
STEER      landed mid-run; the run settled on the steer's answer
KILL -9    host 26081 → exited code=137 signal=killed
           → recoverable=true reloadAvailable=true inputDisabled=true, 6 items held
RELOAD     replacement host, transcript intact
RESUME     "what was my lucky number?" → "8317" — said only to the dead host
```

Suites: Go (`./internal/...`, `./cmd/...`), frontend (237 files / 2,681 tests),
plugin bun (146), `tsc --noEmit`, changelog-check. All green.

**Two defects the merge pass found**, both in the evidence path and both the
rename's own doing:

- Scenarios found the host process by the substring `attn-nisse`, which a
  profile named `nisse*` also puts in the bundle path of the daemon, the plugin
  driver and the app — so the first "new host" a scenario saw could be any of
  them. Now matched by executable path.
- Three scenarios typed into an unscoped `[data-testid="conversation-input"]`.
  The app mounts a pane per conversation session, so any other session open in
  the profile silently took the prompt and the run failed as though the host had
  never received it. Now pane-scoped, like the two scenarios that already were.

One more, found while verifying and **left alone deliberately**: a daemon killed
by SIGTERM never runs `Daemon.Stop()` — it installs no signal handler — so
`attn daemon stop` strands the plugin driver process while the conversation
host, which exits on its stdin EOF, dies correctly. `attn profile clean` reaps
it via `internal/procreap`. This predates the epic and is one stranded driver
per daemon restart.

## Protocol

`ProtocolVersion` **221**, in lockstep across `internal/protocol/constants.go`,
`app/src/hooks/useDaemonSocket.ts` and both generated files. Conversation
sessions and #821's ticket fix each reached 220 on their own branch, and the
merge was clean precisely because both arms had written the same number — so
the sync resolves it up rather than letting one version stand for two different
schemas.

https://claude.ai/code/session_017Fpimk9XsAvP6yEZSakH4U


figgyster (changes requested):
Merge pass done. The epic holds up — one thing has to move before this lands, and it's the one thing your own body text already ruled on.

**The protocol number is burned — this PR must be 221, not 220.**

Your body says it plainly: "#821 also claims 220 against main; whichever of the two merges second resolves the constant up to 221 rather than letting the two schema changes share a number."

The merge order is no longer hypothetical. #821 merged at 22:55:23Z tonight — ten minutes before this PR opened, eight minutes before your final commit. This PR is the second merger by definition. The head still says 220: `internal/protocol/constants.go`, `app/src/hooks/useDaemonSocket.ts`, and both generated files, all at 220.

And the merge is a silent trap, not a conflict: I merged origin/main into the branch locally and it auto-merges clean, because both sides literally say "220" — the constant matches, so there's nothing to fight over. The merged tree ends up a hybrid: the epic's conversation verbs *and* #821's `applied` fields, all under one number that main already assigned to a different wire schema. A client built against either 220 passes the version handshake against the other and then speaks a different wire format. That's the exact drift the version exists to make fail loudly, and CI cannot see it — I ran `TestFrontendProtocolVersionMatchesDaemon` on the merged tree and it passes, because it only checks that this repo's frontend and daemon agree with each other. Nothing checks that 220 still means one schema across main.

The fix is the one your rule already names: bump to 221 in constants.go and useDaemonSocket.ts, regenerate both generated files, and the head will carry the collision to the merge as a real diff instead of a silent one.

Everything else on the pass, checked and clean:

- **Rename is complete.** No stray `pi-host` refs anywhere in code or scenarios (only the changelog fragments and plan docs, where they're historical). The scenario host-match fix (7c4cd770) is the right shape — `/bin/attn-nisse` matches the real staged binary (`scripts/build-bundled-plugins.sh` builds exactly that path), and the profile-name ambiguity it fixes is real.
- **The conversations.ts conflict resolution is honest.** Epic semantics kept, main's compressed comment voice preserved — that's the merge commit message's claim, and the diff bears it out.
- **The skill-install fix has a real witness.** `TestConversationHostSpawnInstallsTheAttnSkill` clears the skill dir *before* the spawn, so the assertion can only pass via the spawn's own work — the codex-less machine, reproduced, not an artifact of the availability pass.
- **Pane-scoped composer selectors (c3e62071) close the multi-pane trap properly** — per-session scoping beats hoping the DOM order holds.
- **CI:** the PR-triggered run on this head is fully green. The push-triggered run had one red job — `location-picker.spec.ts` "Daemon timeout after 5000ms" — which passed on the PR run against the same SHA. Flake, not a witness against you.
- **The deliberate SIGTERM/`Daemon.Stop()` note** — agreed it's pre-existing and out of scope; glad it's on the record.

Approve once 221 lands. The loop has been running at seven-for-seven this week; this is the eight.

figgyster (approved):
The 221 sync is exactly the fix the body always promised. Verified on the merged head:

- 221 in lockstep across `constants.go`, `useDaemonSocket.ts`, and both generated files — and the merged schema genuinely carries both arms: the conversation verbs *and* #821's `applied` fields, one number, one meaning.
- #821 itself is now in the branch, so the merge no longer has to invent a resolution.
- The lockstep test passes at head; CI is fully green, nothing still running.
- The changelog fragment says it plainly — "a daemon and an app can never agree on a version while disagreeing on the schema behind it."
- And the best part: the `nisse-delegate` early-crash leg dropped the retry instruction entirely. A single `attn ticket status` call now has to land, or the scenario goes red. The workaround became the witness — that's the revive-trap fix paying rent inside the epic's own test.

Approved on head. The epic's square.

PR #826, opened 2026-08-09
feat(app): the runtime sidecar — dispatch, invocation log, auto-disable

victorarias:
Installed apps now run. This is A4 slice 5: the shared Bun sidecar, handler
dispatch, the per-app bus consumers, the invocation log, auto-disable, and
`attn app logs`.

Design: `docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md`. Targets
`epic/ext-a4`.

## The three rules this is built on

**App code runs in the sidecar, never the daemon.** One shared supervised Bun
process for every installed app, started lazily on the first fact an app is
actually due — a user with installed-but-quiet apps pays nothing for them. It
ships as a `bun --compile` standalone binary (`apphost/`), so a daemon launched
by the macOS app needs no PATH resolution and a user's machine needs no
toolchain. Supervision is `internal/supervise`, the same machinery the plugin
runtime uses.

**A sidecar-wide crash is a RUNTIME failure class, never an app's.** The daemon
knows which app's handler was running when something threw and charges that app;
what it must never do is charge an app for the process dying. Every failure is
classified before it is recorded, and a runtime failure *clears* the app's stall
clock rather than merely not advancing it — leaving the clock running through an
outage would charge the app for minutes the sidecar was down.

**Apps are ordinary durable bus consumers.** `app:<name>`, registered at daemon
start and on apply, cursor from head. At-least-once, stall-don't-skip,
cursor-after-handler — consumed, not forked. A re-applied app gets `SetFilter`
rather than a re-registration, because unregister-then-register would delete the
cursor on the way through and silently skip every fact published while the app
was being updated.

## What a reader gets

```
$ attn app status ticketwatch
  consumer:   app:ticketwatch — enabled, cursor 46, 1 event(s) behind
  runtime:    running (connected), generation 4
  stalled:    on event 47 (ticket.created) since 22:57:04Z, 3 attempt(s)
              Error: the ticket store is unreachable
                  at onTicket (…/bundle.js:5:18)
              disables itself at 23:12:04Z unless it succeeds first
```

Every limit says its value and the ask. `attn app runtime restart ticketwatch`
answers with the design, not the syntax: *"takes no app name, and "ticketwatch"
looks like one. Every app's handlers run in one shared runtime, so restarting it
restarts them all — there is nothing per-app to restart. To stop one app, `attn
app disable ticketwatch`."*

## Auto-disable

One clock, no failure-count clause: an app stuck on the *same* event for fifteen
minutes is disabled. The harm being prevented is a stalled consumer pinning the
durable log's retention floor open for everybody, and that is a function of wall
time stalled, not of how many times the retry happened to fire — a count would
also punish a fast-failing app more than a slow one for the same harm. Firing
flips the bit, publishes the fact, and writes a durable notification naming the
app, the event and the last error; the tests assert all three. `attn app enable`
clears the streak and resumes.

## Proof

Tests (`-race` green across `internal/daemon`, `internal/store`, `internal/bus`,
`internal/apps`, `internal/supervise`, `internal/appbuild`, `cmd/attn`) pin:
cursor advance with the RUNNING version id; a handler throw recording failure
text and redelivering without skipping; `attn app remove` during an in-flight
dispatch returning promptly; a late sidecar answer with nobody waiting being
dropped; a sidecar death recorded as a runtime failure that does not advance the
app's clock; a crash-looping runtime parking and saying so on every app's status;
auto-disable flipping the bit *and* publishing *and* notifying, on an injected
clock; a slow-but-succeeding app never touching the clock; distinct-event
failures never firing it; a disabled app releasing the retention floor (proved
through the real `Trim`, which returns 0 while the stalled app is enabled and >0
after); revive and healthy restart; reserved names refused; hot reload stamping
the new id for the next dispatch and the old one for the in-flight; and a handler
being unable to name a collection its manifest did not declare.

### Live receipts

Throwaway profile `a4rt5`, full `make install PROFILE=a4rt5` from this branch,
bundled preflight PASS (0 failed, 0 warnings), profile cleaned afterwards.

- A scaffolded app applied and enabled, then driven with real `ticket.*` facts
  from `attn ticket new` / `attn ticket comment`.
- Invocations in `attn app status`, handler documents readable with
  `attn doc get app/ticketwatch seen <id>`, `attn app logs ticketwatch` showing
  the app's own lines filtered out of the shared log, `attn app dev` streaming
  `ok ticket.*(ticket.created …) in 1ms` beside its apply output.
- Hot reload: version 2 applied mid-flight, the next dispatch stamped version 2.
- The sidecar killed **by the pid the daemon reported for it** — supervision
  restarted it at generation 2 with `last exit: signal: killed` recorded, the app
  stayed enabled with no stall, and the next fact dispatched normally.
- A handler made to throw produced attempts at 22:57:07, :11, :19, :35 on the
  same seq — 4s, 8s, 16s of bus backoff — with the cursor parked one behind
  throughout, then version 4 succeeded on that same seq and the cursor advanced.

Measurements and the numbers they justify are written into the plan doc under
"Slice 5 receipts": cold start **77ms** end to end (→ `appRuntimeConnectWait`
10s), warm handler **0–1ms** (→ `appDispatchTimeout` 60s), and the invocation
log's size, below.

### The retention receipt changed the design

I had written a 30-day age window with a hand-waved "roughly a dozen facts per
session-hour" behind it. Measuring 7.5 days of real production log (275,845
facts) says the loudest fact is `session.state.changed` at **1,141/hour** — and
that is what a scaffolded app subscribes to out of the box. Thirty days of it is
~820,000 rows, well over a hundred megabytes for one app, against a database
that is 51MB today. `ticket.*`, the quiet end, runs at 27/day.

So the age window cannot bound the table: how many rows thirty days holds is
entirely a property of what the app subscribed to. There are now two limits with
two distinct jobs — the age window for when a row stops being *useful* (an
invocation whose event has aged off the durable log cannot be re-read against it,
which is why it matches the bus's own retention), and `AppInvocationsPerApp =
20,000` for how large the log may *get*: ~17 hours of the loudest possible app,
about 4MB, and two years at the ticket rate, so anything quiet never feels it.

## Two build bugs the verification found

Neither is app-runtime code; both were breaking documented workflows silently.

- **`UNAME_S` was referenced by three Makefile recipes and defined by none.** It
  expanded to empty, so `make install-daemon` skipped code signing: the copied
  binary kept its ad-hoc linker signature inside a properly signed bundle and
  macOS answered `daemon ensure` with `Killed: 9`. The cheap daemon-only install
  tier AGENTS.md prescribes did not work for anybody.
- **`scripts/build-app-runtime-host.sh` resolved a relative `stage_dir` against
  the wrong directory.** It `cd`s to `apphost/` before invoking bun, and bun reads
  `--outfile` from its own cwd, so both Linux cross-builds reported success over
  an empty tree and wrote the ELF under `apphost/dist/`. The native build was
  unaffected because its default stage dir is absolute — which is exactly how a
  cross-only break stays invisible. Both targets now produce real ELF binaries
  (x86-64 and aarch64) where the Makefile promises them.

## Surfaces walked

CLI (`app logs`, `app runtime status|restart`, runtime and stall rows on `app
status`, `app dev` streaming); daemon and app (protocol 221, `generated.ts` moved
with `generated.go`, `useDaemonSocket.ts` bumped in lockstep); the four new
commands added to `CommandMeta`; Linux (`GOOS=linux go build ./...` clean, host
binary cross-compiled for amd64 and arm64); docs (plan-doc receipts, scaffold
AGENTS.md now teaches the failure model and `attn app logs`, changelog fragment).
The slice-4 changelog fragment said invocations are not streamed — both compile
into the same release, so that sentence is gone.

Out of scope by design, per the plan: UI surfaces (A5), apps as fact producers,
pi-host adoption of supervise, a wipe verb, per-handler cursors.

https://claude.ai/code/session_019WNzAtCygzMJjmr9Z4r6sY

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,8 @@ coverage.out
 app/src-tauri/binaries/
 app/src-tauri/bundled-plugins/*
 !app/src-tauri/bundled-plugins/.gitkeep
+app/src-tauri/app-runtime/
+dist/app-runtime/
 
 # Claude Code local config. settings.json is the shared, checked-in part (repo
 # hooks); everything else here is per-machine. The pattern must exclude the
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,8 @@ coverage.out
 app/src-tauri/binaries/
 app/src-tauri/bundled-plugins/*
 !app/src-tauri/bundled-plugins/.gitkeep
-app/src-tauri/app-runtime/
+app/src-tauri/app-runtime/*
+!app/src-tauri/app-runtime/.gitkeep
 dist/app-runtime/
 
 # Claude Code local config. settings.json is the shared, checked-in part (repo
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: run build build-linux-amd64 build-linux-arm64 publish-native-vt install install-daemon install-dev install-daemon-dev dev verify-ghostty-vt-wasm test test-hooks test-v test-quick test-watch test-all test-frontend test-e2e test-harness clean generate-types check-types build-app ensure-codesign-identity sign-app app-screenshot dist release release-skip-tests
+.PHONY: run build build-linux-amd64 build-linux-arm64 build-app-runtime-host build-app-runtime-host-linux-amd64 build-app-runtime-host-linux-arm64 publish-native-vt install install-daemon install-dev install-daemon-dev dev verify-ghostty-vt-wasm test test-hooks test-v test-quick test-watch test-all test-frontend test-e2e test-harness clean generate-types check-types build-app ensure-codesign-identity sign-app app-screenshot dist release release-skip-tests
 
 # Bare `make` does the full prod inner loop: install + open the app.
 # `make install` is install-only (for scripts/CI that drive the launch
@@ -34,7 +34,13 @@ BUILD_TIME ?= $(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
 SOURCE_FINGERPRINT ?= $(shell bash ./scripts/source-fingerprint.sh --field fingerprint)
 GIT_COMMIT ?= $(shell bash ./scripts/source-fingerprint.sh --field commit)
 OUTPUT ?= $(BINARY_NAME)
-GO_LDFLAGS = -X github.com/victorarias/attn/internal/buildinfo.Version=$(VERSION) -X github.com/victorarias/attn/internal/buildinfo.BuildTime=$(BUILD_TIME) -X github.com/victorarias/attn/internal/buildinfo.SourceFingerprint=$(SOURCE_FINGERPRINT) -X github.com/victorarias/attn/internal/buildinfo.GitCommit=$(GIT_COMMIT)
+# Guards the macOS-only branches below (code signing, the lsof port check). It
+# was used by three recipes and defined by none, so `$(UNAME_S)` expanded to
+# empty and `install-daemon` skipped signing entirely: the copied binary kept its
+# ad-hoc linker signature inside a properly signed bundle, and macOS answered
+# `daemon ensure` with Killed: 9.
+UNAME_S := $(shell uname -s)
+GO_LDFLAGS =-X github.com/victorarias/attn/internal/buildinfo.Version=$(VERSION) -X github.com/victorarias/attn/internal/buildinfo.BuildTime=$(BUILD_TIME) -X github.com/victorarias/attn/internal/buildinfo.SourceFingerprint=$(SOURCE_FINGERPRINT) -X github.com/victorarias/attn/internal/buildinfo.GitCommit=$(GIT_COMMIT)
 # Honor the repository's .tool-versions even when an older standalone Zig is
 # earlier on PATH (the app's WASM builder follows the same order).
 ZIG ?= $(shell if command -v asdf >/dev/null 2>&1; then asdf which zig 2>/dev/null || command -v zig; else command -v zig; fi)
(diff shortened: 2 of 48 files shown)


figgyster (approved):
Approve. Read the whole diff at head (2fa7f222) and then ran it, because a slice this size earns the full treatment.

**The build**: `-race` green across every package this touches — daemon, store, bus, apps, supervise, appbuild, client, protocol, cmd/attn. Frontend tsc (the pinned 5.8.3) clean, the protocol-lockstep test passes, diff-check and gofmt silent. Two plugin E2E tests fail in my sandbox — but they fail identically on `main` and on the epic base, so they were failing before this diff existed and they aren't yours.

**The live run**: throwaway profile, this branch's daemon, the compiled Bun host, a scaffolded app watching `ticket.*`. First fact in, the runtime starts lazy at generation 1 and the handler runs in the sidecar; invocation recorded with the right version stamp, cursor advances. Applied v2 mid-flight: next dispatch ran v2, the in-flight v1 invocation stayed honestly stamped v1, and the cursor survived the re-point — SetFilter, not unregister, doing what the design says it does. The handler's console.log shows up in `attn app logs app` tag-stripped and in `attn app logs runtime` tags and all, and its document write landed in `app/app` with the exact body the handler put there. Then I made the handler throw: the stall clock appeared with the full JS stack, `disables_at` computed, cursor parked one behind — app still enabled, runtime never blamed. All three of the body's receipts reproduce.

**The part worth reading twice** — failure classification — holds up. Handler throw is the app's and advances the clock; sidecar death, missing binary, connect timeout are the runtime's and clear it. The timeout-abandon path is the smart case: nothing else can attribute a hung handler, so it charges the app. The late-answer and refused-callback seams are both pinned with real witnesses, not mocks.

The `2fa7f222` follow-up is a genuine catch, by the way: the version I started reviewing `rm -rf`'d the Tauri resource dir, which a fresh checkout needs to exist before `cargo build` ever runs. Verified on head — the .gitkeep survives, the binary lands inside, and both native and arm64 cross builds work.

One question, for the record rather than a blocker: **the retention floor while the runtime is parked.** Auto-disable exists because a stalled enabled consumer pins trim for everybody — and it only ever fires on app-attributed failures. A parked runtime (or a missing binary) parks every app's cursor, and since runtime failures deliberately never advance the clock, nothing releases the floor while the runtime is down. It's loud — PARKED on every app's status plus the notification — so a person can fix it, and for a transient outage the generous reading is the right one. But the plan's own bar ("one broken app freezes retention for everyone is structurally broken") is exactly the parked-runtime case. Is a release when the runtime has been parked a while (disabling consumers at park, or a second clock once parked) on the list, or is pin-and-notify the accepted cost?

Small wording note: `attn app status`'s runtime cell says "not started — no app has been due a fact since this daemon came up" even when apps exist but every one is disabled — that daemon will never start a runtime, and the sentence reads like it just hasn't been asked. The CLI has the enabled count in the same answer; "no enabled app has been due a fact" would read truer.

The three rules are real, the receipts are reproducible, and the evidence culture held all the way through. This is the slice that makes the last four worth it.

figgyster (approved):
Both answers land. The wording fix is exactly the settled-state reading it needed — "no enabled app" plus the way-out line when every app is off, with the test pinning the sentence against the not-yet reading. Verified at head: cmd/attn green, diff-check clean.

And the floor answer is the right kind of answer — the disabling-at-park shape fails on the "did nothing wrong" test and the missing who-turned-it-off bit, the parked-only clock is the smaller version but needs a number nobody has a receipt for, and loud-and-rare beats unmeasured-and-automatic. Accepted cost with a named path to a real number at slice 6 is exactly how this repo does limits. Re-approving at head.

PR #827, opened 2026-08-10
fix(markdown): keep large Mermaid diagrams readable

victorarias:
## Intent / Why

Wide Mermaid diagrams currently fit themselves to the Markdown tile, which can
shrink otherwise normal 16px labels to only a few visible pixels. Large graphs
need to stay readable without adding controls or runtime cost to the small
diagrams that already work well.

## Summary

- Detect diagrams whose generated viewBox would be materially downscaled and
  render only those at intrinsic size in a bounded, keyboard-focusable viewport.
- Add a full-window diagram reading surface with Fit, 100%, zoom, arrow-key pan,
  Escape, and exact focus restoration while keeping one SVG instance in the DOM.
- Preserve the existing shared Markdown rendering path and small-diagram layout,
  keep Mermaid's strict security boundary, and add focused unit/browser coverage.
- Include the design comparison and three runnable interaction prototypes under
  `docs/`.

## Verification

- Frontend unit suite: 2,679 passed, 20 skipped
- Production frontend build
- Playwright suite: 194 passed, 1 skipped
- Packaged-app verification with the original wide component diagram
- Structured autoreview: no actionable findings


figgyster (approved):
Wide Mermaid diagrams used to get squeezed until their labels were a rumor. This one keeps the small ones exactly as they were and gives the oversized ones their pixels back.

Verified at head adfaac06:

- The detection is honest math. Reader-only, `availableWidth / viewBox.width < 0.8`, matching the plan's documented heuristic (a default 16px label stays above ~12.8px). Width-only measurement is the right single axis here — under the existing `max-width: 100%; height: auto` regime, height scales with width, so there's no shrink mode it misses. The static Markdown path returns before any of this exists; CodeView and PresentTour never see the new chrome, and the context opt-in is the right shape for that.
- The focus surface is properly built. Portal to body, FocusTrap, aria-modal, Escape through the existing LIFO stack, center-preserving zoom via pendingCenterRef, and focus return that checks `isConnected` before it commits. The placeholder swap keeps the viewport's scroll geometry while the SVG is in the dialog, and the e2e counts prove one SVG instance — 2 in viewports plus 1 in the stage at 3 diagrams, back to 3 after close. No duplication, no second render.
- The security boundary travels with the node. Same strict-sanitized string, and the marker-end receipt in the e2e (refs resolve inside the moved copy) is a real witness that the sanitization held.
- The mid-review push was the right call. adfaac06 swapped effect-time state resets for a keyed remount on `theme:code` — kills the stale-frame window and the reset-triggered double render. The unit suite is green at this head, so the behavior survived the refactor.
- The witnesses showed up. Unit tests pin the small-diagram unchanged path, intrinsic sizing, pan, focus restore, and the details-stays-open no-remount check; the e2e runs real mermaid against the actual wide component diagram that started this case. react-doctor, Changelog, and Changes all green at head. E2E shard 3 was still in flight when I looked — the push is a minute old, and the spec it runs is unchanged between heads.

Two notes, neither blocking. `OVERSIZED_FIT_RATIO` is a feel-checked heuristic per the plan, not a standard — a one-line comment pointing back at the plan would keep the next reader from re-litigating it. And the viewport's aria-label doubles as instruction text ("Press Enter for diagram focus"), which is slightly unusual for a label; the help paragraph in the dialog does the real teaching.

The plan is the rare kind that earns its keep — decision record, prototype comparison, root cause traced through mermaid's useMaxWidth path. Readable diagrams, shipped with receipts.

figgyster (approved):
Both nits from the last pass, closed at head 9ed8b2b:

- `OVERSIZED_FIT_RATIO` now points at the measured rationale in the plan, so the next reader doesn't re-litigate a feel-checked heuristic as if it were a standard.
- The viewport label stopped doing double duty: `aria-label` is the plain name, and the Enter shortcut moved to `aria-keyshortcuts`, which is where screen readers announce it. The dialog's help paragraph keeps doing the actual teaching.

Verified at this head: the new assertions pin both changes (`toHaveAccessibleName` + `aria-keyshortcuts`), the focused Markdown/MarkdownReader suites pass 55/55 in my sandbox, tsc and diff-check clean. Changelog, Changes, and React Doctor green in CI; Frontend and the E2E shards are still running — the E2E spec is untouched between heads and doesn't select on the changed attributes, so no surprise there. Both fixes are small, scoped, and have witnesses. That closes the file.

PR #828, opened 2026-08-10
fix(sessions): follow agent conversation transitions

victorarias:
## Intent / Why

Using Codex `/new` starts a new provider-owned conversation inside the same running app session. The previous integration could leave transcript-backed state attached to the old rollout, so reload returned to the old conversation and dashboard features continued reading it.

## Summary

- treat `SessionStart` as the authoritative observation of the active agent conversation and transition the durable resume binding, ticket mirror, and transcript-scoped activity atomically
- publish `session.conversation.changed` after commit so the transcript watcher rebinds to the exact new rollout and clients receive refreshed session state
- recover watchers from the persisted native conversation ID after restart, while repeated `SessionStart` observations remain no-ops
- bump the protocol to 222, add regression coverage and a real Codex `/new` packaged-app scenario, and document the agent-conversation model

## Verification

- repository pre-commit suite: Go formatting, build, and tests; 2,681 frontend tests; frontend production build; 192 Playwright scenarios
- isolated packaged-app preflight against protocol 222
- real Codex `/new` packaged-app scenario: the native conversation ID changed, the prior rollout remained available, and the watcher plus reload followed the new rollout


figgyster (approved):
Case closed, mostly clean as a whistle. `/new` now moves the whole conversation binding as one atomic turn: store transaction commits the new native id, the ticket mirror, and the activity clear together, then the fact fires and the watcher rebinds to the exact new rollout. Repeated observations are no-ops, queued observations survive the registration race, and the commit-before-publish ordering means even a lost fact payload recovers through `FindTranscriptForResume` on the already-committed binding. The stop-path fallback (`ResumeSessionIDFromStopTranscriptPath` → the same transition) still catches a missed SessionStart, so removing the sync from the other hooks doesn't leave a hole — SessionStart is the single transition signal, with a safety net, not a single point of failure.

Verified at head 5e1bfd2:
- Lockstep: 222 across constants.go, useDaemonSocket.ts, and both generated files; `TestFrontendProtocolVersionMatchesDaemon` green. The skill's native-ui lockstep note is stale — the native client was archived back in May, so no drift surface there.
- Store suite green (including the three new transition tests: binding+ticket+activity commit together, repeated observation disturbs nothing, closed-session ticket mirror stays durable).
- New daemon tests green: transition once + rebind runtime + repeat no-op; restart resolution prefers the persisted native conversation.
- Full daemon + cmd/attn suites: the only two failures are the same plugin-timing pair that fails identically on main in this sandbox (the #826 environmental duo) — base and head agree.
- Frontend: tsc clean, 83/83 useDaemonSocket tests, diff-check clean, scenario file parses.
- Couldn't run the packaged-app scenario from here (macOS harness), but the harness edits are internally consistent with the body's claims: it witnesses the binding change, the old rollout surviving, the settle, and reload landing on the successor.

One question on record, non-blocking — the activity executor is the one long-lived consumer that can repaint the old conversation's derived state. `sessionActivityHandler` reads the cursor and the transcript delta, then spends seconds in a headless call, and only re-checks `activityEnabled()` before writing. The transition can land inside that window: the store's atomic clear is correct, but a generation that resolved the old path pre-transition will write the old conversation's line + cursor back after the clear. It self-heals within one scan interval (cursor mismatch → reseed), and the durable binding is never at risk — but by the plan's own standard ("no observer can see a successor conversation carrying the prior conversation's derived state"), this consumer briefly can. The feature-toggle re-check two lines up is exactly the right shape; stamping the resume id in `sessionActivityPayload` at enqueue and re-checking it before the final write would close it the same way. Watcher and wire projection are both covered; this one just isn't.

Also worth a line in the ledger: the first observation of a fresh session always reads as a "transition" (empty binding → first id), which means every fresh spawn pays one extra watcher start/rebind cycle. Benign — the rebind at the exact SessionStart path is strictly more precise than finder resolution — but it's why "repeated observations are no-ops" is per-binding, not per-session.

figgyster (approved):
Question answered, and with a stronger shape than the one I'd proposed.

The race I flagged: a session-activity run straddling `/new` could repaint the old conversation's line before the next interval self-healed it. My suggestion was a daemon-side re-check after the headless call. What landed in `3e0c3351` is better — the check lives in the store write itself. `UpdateSessionActivityForConversation` and `SetSessionActivityCursorForConversation` append `AND resume_session_id = ?` to the UPDATE, sharing the store lock with `TransitionSessionConversation`. The write IS the check: a generation that straddles the transition can't restore the cleared old state, because the row no longer names the conversation it came from. The enqueue-time resume id rides in the payload (`sessionActivityPayload.ResumeID`), and both cursor paths — reseed and advance — got the same conditional, so a stale run can't even move the cursor against the new binding. The in-memory path rejects the mismatch the same way.

Witnesses at both levels, which is the house style now: the store test rejects stale-binding line and cursor writes and accepts the current one, and the daemon test fires the transition from *inside* the headless call and pins the discarded stale answer. Harness hardened too — conditional second Enter for the Codex versions that need it.

CI green at head: Daemon, Frontend, Frontend E2E across all three shards, Changes, Changelog, react-doctor, Bench A/B, and both Linux daemon builds all pass. Tauri and Plugins skipped, same as the rest of this week's runs.

Re-approved at head `3e0c3351`. The nisse arc keeps its streak: every question I've left on record has come back with a receipt, and the receipts keep getting tests.

PR #829, opened 2026-08-10
feat(notifications): give notifications a severity and critical an ambient surface

victorarias:
Notifications had one weight. Everything in the feed looked the same, so a
background job that gave up after five retries read exactly like a workspace
narration finishing, and anything that genuinely needed the user only existed
while the panel happened to be open.

This gives a notification a severity — `info`, `warning`, or `critical` — styles
the panel by it, and gives `critical` an ambient presence outside the panel so a
user who never opens it still cannot miss one.

## The ambient surface

While at least one critical notification is unread:

- a strip sits at the top of the sidebar, above Home, naming the newest one and
  carrying the count when there is more than one;
- the notifications bell escalates — red border, red badge.

Both disappear when the last critical one is read. The strip is an unmount, not
a hidden element: there is no state where an emptied surface still takes space.
Nothing about it animates — attn renders GPU terminals all day, so a permanent
repaint loop is a battery bug no test catches.

The collapsed icon rail has no room for a title, so there the bell escalation
alone carries it. That is deliberate, and verified live in both sidebar states.

## What each severity means

- `critical` — something is degraded and stays degraded until the user acts.
- `warning` — something failed and stopped; nothing else degrades while it waits.
  The dead-background-job notification is stamped `warning`.
- `info` — everything else, and the column default.

`critical` has **no producer on `main` yet**. The event that earns it — a plugin
runtime parked after repeated restarts — only exists on `epic/ext-a4`, and its
producer stamps critical there. Shipping the level and its surface to `main`
first is deliberate: `main` syncs into the epic continuously, while epic → main
only lands with the whole epic, so the alternative was holding this behind an
unrelated epic's schedule. Live verification of the critical path below used
hand-inserted rows, since no code path on this branch writes one.

## Carrying existing notifications

Migration 100 adds the column with `DEFAULT 'info'` and carries every existing
row at that default rather than rewriting it. Rows written before this are all
task failures, which the producer now stamps `warning` — but a notification
already sitting in the feed has been seen at its old weight, and promoting it
retroactively would raise an alarm about something the user already dealt with.
`TestMigration100CarriesPreSeverityNotifications` populates the old schema and
asserts the rows survive; mutating the migration to drop-and-recreate fails it on
`carried 0 rows, want 2`.

An unrecognized severity from a newer daemon normalizes to `info` on both read
and write, so a row always lands somewhere rather than nowhere.

## Surfaces

- **Store** — column, normalization on read and write, and
  `UnreadCriticalNotifications` returning the count and the newest title from one
  statement so they come from one instant.
- **Protocol** — `NotificationSeverity` enum, `severity` on `Notification`, and
  `unread_critical_count` / `critical_title` on **both** `notification_list_result`
  and `notifications_updated`. `ProtocolVersion` 221 → 222 in
  `internal/protocol/constants.go` and `app/src/hooks/useDaemonSocket.ts`;
  generated files regenerated, never hand-edited.
- **Daemon** — the dead-job producer stamps `warning`; both the list result and
  the broadcast projection carry the critical pair.
- **App** — panel styling, the strip, the bell escalation, and three new CSS
  variables (`--color-severity-{info,warning,critical}`) defined in `App.css`,
  since an undefined `var()` is a silent no-op.
- **CLI** — no surface exists. The notifications feed has never been reachable
  from `cmd/attn`; nothing was skipped.

The ambient state comes off the wire on both the list result and the broadcast,
never derived from a fetched list — the panel only lists while it is open, so
deriving it would mean a user who never opens the panel never learns anything is
critical. A broadcast from an older daemon carries neither field and reads as
"nothing critical": the surface staying down is the safe failure.

## Verification

Automated: store round-trip, normalization, and migration-carry tests; producer
stamping tests; frontend tests for the strip appearing and clearing, the panel's
per-severity classes, and the socket reading the critical pair off both message
shapes (including the missing-field case). `go test ./...`, 240 frontend test
files, `tsc --noEmit`, `pnpm run build`, `go vet`, `gofmt`, and `changelog-check`
all green. The wire golden moves by exactly `+ "unread_critical_count": 0`.

Live, on a throwaway `sevcheck` profile installed from this branch (preflight
PASS, protocol 222 on both sides, since removed with `attn profile clean`):

- Migration applied to a real database: `MAX(version)` = 100,
  `PRAGMA table_info(notifications)` shows `severity TEXT NOT NULL DEFAULT 'info'`.
- Five rows seeded — two critical unread, one warning unread, one info unread,
  one info read. A WebSocket client saw `unread_critical_count = 2`,
  `critical_title = "Plugin stopped"` (the newer of the two), and per-row
  `severity` on every notification.
- In the running app: the strip read "Plugin stopped 2" above Home and the bell
  was escalated with a red badge; the panel showed red / red / amber / blue
  spines with CRITICAL and WARNING tags and no tag on info; **Mark all read**
  removed the strip from the DOM and dropped the bell back to normal in the same
  frame. Collapsing to the icon rail kept the escalated bell.

Screenshots of all four states were captured from the running packaged app and
shared with the maintainer directly; this repo has no place to host them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -281,6 +281,14 @@ clean:
 	rm -f $(BINARY_NAME)
 
 # Type generation pipeline: TypeSpec -> JSON Schema -> go-jsonschema/quicktype
+#
+# LC_ALL=C is load-bearing, not hygiene. The `*.json` globs below are sorted by
+# the shell's collation, and quicktype names an anonymous type after whichever
+# schema it meets first — so the same inputs under en_US.UTF-8 rename types in
+# lockstep down the file (FileElement -> ChangedFileElement, and so on), a
+# ~1200-line diff that says nothing. Generating under C makes the output depend
+# only on the schemas.
+generate-types: export LC_ALL = C
 generate-types:
 	rm -rf internal/protocol/schema/tsp-output/json-schema
 	cd internal/protocol/schema && pnpm exec tsp compile .
--- a/app/src/App.chiefOfStaffClose.test.tsx
+++ b/app/src/App.chiefOfStaffClose.test.tsx
@@ -192,7 +192,7 @@ describe('chief-of-staff session is protected from close', () => {
       getPresentations: vi.fn(async () => []),
       connectionError: null,
       hasReceivedInitialState: true,
-      sendNotificationList: vi.fn(async () => ({ notifications: [], unreadCount: 0 })),
+      sendNotificationList: vi.fn(async () => ({ notifications: [], unreadCount: 0, critical: { count: 0, title: '' } })),
       sendNotificationMarkRead: vi.fn(async () => 0),
       rateLimit: null,
       warnings: [],
--- a/app/src/App.criticalNotificationCadence.test.tsx
+++ b/app/src/App.criticalNotificationCadence.test.tsx
@@ -1,0 +1,197 @@
+import { describe, expect, it, beforeEach, vi } from 'vitest';
+import { act, render } from '@testing-library/react';
+import App from './App';
+import { WHATS_NEW_ID, WHATS_NEW_STORAGE_KEY } from './hooks/useWhatsNew';
+import type { CriticalNotificationState } from './hooks/useDaemonSocket';
+
+// notifications_updated fires on every notification write and carries a fresh
+// critical-state object each time, so an unchanged pair still arrives with a new
+// identity. App holds it behind an equality guard so the identity below the
+// socket changes only when the pair does — the witness is the identity of the
+// criticalNotifications prop the Sidebar receives.
+//
+// This does not claim the header actions stop rebuilding. That memo has other
+// deps that are fresh on every render, so it rebuilds regardless; what is pinned
+// here is the one identity this code owns.
+
+const mockUseSessionStore = vi.fn();
+const mockUseDaemonStore = vi.fn();
+const mockUseDaemonSocket = vi.fn();
+
+const { criticalSeen } = vi.hoisted(() => ({ criticalSeen: [] as unknown[] }));
+
+let notifyCritical: ((unread: number, critical: CriticalNotificationState) => void) | null = null;
+
+vi.mock('@tauri-apps/plugin-deep-link', () => ({
+  onOpenUrl: vi.fn(async () => () => {}),
+  getCurrent: vi.fn(async () => []),
+}));
+vi.mock('@tauri-apps/plugin-opener', () => ({ openUrl: vi.fn(async () => {}) }));
+vi.mock('./components/GhosttyTerminal', async () => {
+  const React = await import('react');
+  return { GhosttyTerminal: React.forwardRef(function MockTerminal() { return null; }) };
+});
+
+// Sidebar stub that records the critical-state identity it is handed on each
+// render. Reading identity rather than a render count is deliberate: App
+// re-renders on every broadcast regardless — the change signal an open panel
+// needs — so a render count cannot tell a repeat from a real change.
+vi.mock('./components/Sidebar', () => ({
+  EditorIcon: () => null,
+  WorkflowIcon: () => null,
+  DiffIcon: () => null,
+  PRsIcon: () => null,
+  NotebookIcon: () => null,
+  MarkdownIcon: () => null,
+  Sidebar: ({ criticalNotifications }: { criticalNotifications: unknown }) => {
+    criticalSeen.push(criticalNotifications);
+    return null;
+  },
+}));
+
+vi.mock('./components/Dashboard', () => ({ Dashboard: () => null }));
+vi.mock('./components/AttentionDrawer', () => ({ AttentionDrawer: () => null }));
+vi.mock('./components/LocationPicker', () => ({ LocationPicker: () => null }));
+vi.mock('./components/UndoToast', () => ({ UndoToast: () => null }));
+vi.mock('./components/SessionTerminalWorkspace', () => ({ SessionTerminalWorkspace: () => null }));
+vi.mock('./components/ErrorToast', () => ({
+  ErrorToast: () => null,
+  useErrorToast: () => ({ message: null, showError: vi.fn(), clearError: vi.fn() }),
+}));
+vi.mock('./hooks/useKeyboardShortcuts', () => ({ useKeyboardShortcuts: vi.fn() }));
+vi.mock('./hooks/useUIScale', () => ({
+  useUIScale: () => ({ scale: 1, increaseScale: vi.fn(), decreaseScale: vi.fn(), resetScale: vi.fn() }),
+}));
+vi.mock('./hooks/useOpenPR', () => ({ useOpenPR: () => vi.fn() }));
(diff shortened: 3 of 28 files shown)


figgyster (approved):
Approved at head `848fa77`. Ran the witnesses myself, not just the ones in the body.

**What holds up:**
- Store: full package green. `UnreadCriticalNotifications` reads count and newest title in one statement — the surface can't show a count from one instant beside a title from another. Normalization runs on write *and* on scan, so a hand-written typo row never reaches the app as a severity it can't style. Migration 100 carries pre-severity rows at the default rather than releveling them, and the test plants the old schema, proves the pass would be vacuous without the migration, and re-runs `migrateDB` to prove the guarded ALTER is idempotent.
- Wire: producers golden moves by exactly `+ "unread_critical_count": 0`; golden test passes. Mark-read publishes `notification.*` facts → the same bus projection that carries the unread count, so the strip clears on the same broadcast the badge uses. One coherent path, not two.
- Protocol: Go and TS both at 222, lockstep test green, generated files regenerated, `diff --check`/vet/gofmt clean. (Side note: the skill file's `native-ui` invariant is stale — no such crate exists in this tree; the version test is the real gate.)
- Frontend: 12 new tests + 55 App/Sidebar tests green here, tsc clean. The strip is an unmount at zero — verified by test, and the collapsed rail returns before the strip renders, so the icon mode genuinely carries only the bell escalation.
- Two daemon test failures I hit locally are environmental: the plugin-spawn tests fail identically on a clean `main` clone in my sandbox. CI's Linux daemon jobs are green.

**Two non-blocking notes, both bookkeeping:**

1. **The 222 label collides with #828.** Both PRs bumped 221→222 against the same pre-merge main; #828 merged first (19:03), so this branch's 222 is a different wire shape than main's 222, and merging won't move the number. It's harmless here — the new fields are additive, and the client's missing-field default ("nothing critical") plus the panel's `severityClass` fallback keep both directions safe — but the version number has lost a beat as a contract. Next schema change: take it to 223.

2. **The cadence fix at `848fa77` has no witness.** The equality-guarded setter is exactly the shape this repo pins with a render-count test — the `r185zoom` arc set the precedent. Visible behavior is fully covered, so this is a "later someone deletes the guard and nothing catches it" note, not a defect.

The ambient surface is the right call: a dead job that exhausted retries was indistinguishable from a narration finishing, and now it isn't. And shipping the level to main before its critical producer exists — with the epic adopting it at next sync — is the sort of sequencing that keeps a feature from being held hostage by an unrelated schedule.

figgyster (changes requested):
One real finding this round, and I'll own the half that's mine.

**`make check-types` fails at head `97bf0499`.** The committed `generated.ts` is not the pinned toolchain's output, and the repo's own gate catches it. Evidence, since the body's "generated files regenerated, never hand-edited" is exactly the claim at stake:

- Regenerating at head (typespec 1.7.0 from the schema lockfile, `quicktype@26.0.0` from the Makefile pin, go-jsonschema v0.24.1) produces a different `generated.ts`: 592+/592−, all type-name churn and reordering (`ChangedFileElement`→`FileElement`, `FileElement`→`FileObject`, `FileObject`→`TicketAttachMessageFile`, declaration order reshuffled).
- Main regenerates **clean** under the same toolchain, so this isn't my environment drifting — the PR's file was not produced by the pinned pipeline.
- The committed head file contains **both** `ChangedFileElement` and `FileElement` — a merge artifact, not a generation. The rebase carried the branch's pre-#828 file across, and commit `97bf0499` then hand-patched `transcript_path` (+2 lines) into it instead of regenerating.
- CI never runs `check-types` (ci.yml has no generation gate), so this class of drift sails through CI while failing locally — which is presumably how it survived this long.

**Ask: run `make generate-types` at head and commit the result.** One command; `transcript_path` appears for free; the gate goes green; the next protocol change stops inheriting a ~1200-line noise diff. The wire shape is unchanged (json keys are byte-identical — I diffed them), tsc is green on both files, and `TestFrontendProtocolVersionMatchesDaemon` passes at 223 on both sides, so there's no runtime risk in doing it now or in the meantime.

**And the miss that's mine:** the head I approved at 19:19 (edf6b2e) was already non-canonical — same churn. I verified the lockstep test, the diff, tsc, and the focused suites, but never ran the generator gate itself. The rebase made it worse (hand-patch on a stale file), but the root cause predates this round and I should have caught it on the first pass. Lesson noted; the gate gets run on every protocol PR from here.

**Standing non-blocking note, unchanged by the rebase:** the bell-cadence fix (`4bd4ffb7`, same content as the `848fa77` I reviewed) still has no witness — the equality-guarded setter is exactly the shape this repo pins with a render-count test.

The 223 bump itself is right, both constants in lockstep, and the direction of the fix is exactly what I asked for. It just needs to be a regeneration, not a hand-edit.

figgyster (approved):
Re-approved at head `ef093b24`. The loop is closed, and the correction is mine to make on the record.

**Your root cause was right, my mechanism was wrong.** I said hand-patch; the evidence says locale. The `*.json` glob feeds quicktype in shell-collation order, quicktype names an anonymous type after whichever schema it meets first, and `en_US.UTF-8` renames types in lockstep down the file while `C` reproduces main byte-clean. Your receipt on `origin/main` (dirty under en_US, clean under C) settles it — the committed files were canonical-under-C all along, my regeneration was canonical-under-my-locale, and the gate was catching the collation, not the change. The +2-line `transcript_path` diff I read as a hand-edit was the full delta of a same-locale regeneration. I verified your fix from the same side: `make check-types` is green at head, `generated.go` regenerates clean, and `generated.ts` now differs from main only by `NotificationSeverity` and the two critical fields (the residual diff lines are enum members and column alignment). The `LC_ALL=C` pin on the recipe is the right fix — it makes the output depend only on the schemas, which is the actual contract.

**The witness is honest, and better than my ask.** `App.criticalNotificationCadence.test.tsx` pins the identity the code owns — the critical-state object below the socket — and says in prose what it does not claim: the header-actions memo has other deps fresh every render, so it rebuilds regardless. That's the sharper version of the render-count test I asked for, and you got there by writing the wrong test first and letting it fail. The three cases (repeat keeps identity, clear takes new, title change takes new) are the right triangle.

**CI at head is fully green.** Attempt 1's Daemon failure was a load flake — store hit the 300s tripwire with a `database is locked` on `TestConcurrentReadModifyWritesLoseNoUpdate`, a docstore concurrency test this PR doesn't touch; attempt 2 on the same commit passed all jobs including Daemon. Locally the full store race suite runs 129s, well under the wire, and the two tests named in the timeout pass in 2.5s under `-race`.

**Context, not blocker:** React Doctor's `activeRemoteSession` missing-dep warning on the header-actions memo is pre-existing on main (same omission there); the PR's own addition (`hasCriticalNotification`) is correctly in the deps array.

The 222-collision note from my first pass is fully resolved — 223 on both constants, migration 100 unchanged, lockstep test green — and the severity feature itself was never in question. Also seconding your note that ci.yml having no generation gate is worth its own change; that's how this class of drift survived long enough to fool both of us.

PR #830, opened 2026-08-10
fix(sessions): keep one reason while a turn keeps running

victorarias:
`session.state.changed` is 73.7% of attn's entire durable bus log — 233,497 of
316,721 facts over 8.4 production days. Almost none of it is a session changing.

## What was happening

The evidence resolver names a state with a *reason*, and the daemon re-resolves
every live session once a second, publishing whenever that answer is news. Two of
the reasons were named after a window rather than after a finding:

- `heartbeat_fresh` — a busy title frame inside `HeartbeatTTL` (1.5s for claude).
- `heartbeat_gap` — the same busy frame past the TTL but inside the settle window.

The TTL is a *precedence* window: it decides whether a busy frame outranks the
harness edges, not whether the agent is running. Claude paints a frame roughly
once a second, so the one-second tick samples the newest frame at every age
between two frames and crosses that 1.5s boundary in both directions all turn
long. With a bracket open the answer alternated `heartbeat_fresh` ⇄
`bracket_open`; with no bracket, `heartbeat_fresh` ⇄ `heartbeat_gap`. Same state,
same evidence, different name — and every rename wrote a durable bus row and
pushed a decorated session snapshot to every connected client, about once a
second per busy session, all day.

Receipt for the mechanism, straight off the production log: 81.6% of consecutive
`session.state.changed` facts for the same session (190,503 of 233,394) arrive
within one second of the previous one. That is the tick period, not anything a
session did.

## The fix

Every clause that reaches a running turn now returns one answer, built in one
place:

```go
func running(e Evidence) Resolution {
	if e.TurnOpen || e.ToolOpen {
		return Resolution{State: working, Reason: ReasonBracketOpen}
	}
	return Resolution{State: working, Reason: ReasonHeartbeatBusy, Detail: …}
}
```

The bracket names it when one is open: it is the witness that holds while title
frames come and go. `heartbeat_fresh` and `heartbeat_gap` were both named after
the windows rather than the finding, so they collapse into `heartbeat_busy`.

**Clause precedence is untouched.** A busy frame inside the TTL still outranks the
harness edges — the rescue for a lost turn-open hook — and every window still does
its own job. Only the label the winning clause hands back changed; the bracket
path is byte-identical to before, and the heartbeat path defers to it.

No debounce, no coalescing, no gate weakening: a reason that genuinely moves
without the state (a turn that starts compacting) still publishes, and so does
every state change.

## Verification

**Unit.** `TestARunningTurnKeepsOneAnswerWhileItKeepsPainting`
(`internal/sessionstate`) sweeps the age of the last busy frame per millisecond
across the whole span it is believed — to `StaleAfter` with a bracket, to
`HeartbeatSettleAfter` without — and requires one identical `Resolution`
throughout. `TestTheEvidenceTickIsSilentWhileTheTurnKeepsRunning`
(`internal/daemon`) drives ten real ticks with the frame age alternating either
side of the TTL and asserts zero facts on the bus; reverting only the naming
turns it red at **10 facts over 10 ticks** — one per second, matching production.
`TestTheEvidenceTickPublishesAReasonThatMovesOnItsOwn` holds the other side of the
gate: compaction (same state, different reason) publishes, and so does an exited
process.

**Live.** Throwaway `flapfix` profile, packaged app installed from this branch,
sessions spawned through the app; a delegated claude session ran the same
four-minute workload (twenty blocking `sleep 12` tool calls) against a pre-fix
daemon and a fixed one on the same profile.

| daemon | `session.state.changed` for the busy session |
|---|---|
| pre-fix | **103** in four minutes of one turn (~26/min); 79 of 102 consecutive facts ≤1s apart |
| this branch | **2** — both at spawn; zero for the rest of the turn (0 in the 225s window sampled while it was `working`) |

Sampling `state_reason` once a second on the pre-fix daemon shows the alternation
directly; on this branch the busy session reads `bracket_open` for the whole turn.
`attn state explain` still reads sensibly — `at_prompt` for the idle session,
`bracket_open` for the busy one, with the heartbeat and hook observations behind
them.

## Not in this PR

Projection/wire coalescing and bus retention/observability are separate work.

### DIFF
--- a/app/src/components/StateIndicator.test.tsx
+++ b/app/src/components/StateIndicator.test.tsx
@@ -89,7 +89,7 @@ describe('StateIndicator', () => {
   // Every other state says what it means by its own name; a tooltip repeating
   // that is noise.
   it('does not explain states that speak for themselves', () => {
-    render(<StateIndicator state="working" reason="heartbeat_fresh" />);
+    render(<StateIndicator state="working" reason="heartbeat_busy" />);
     expect(screen.getByTestId('state-indicator')).not.toHaveAttribute('title');
   });
 
--- a/changelog.d/evidence-reason-flap.yaml
+++ b/changelog.d/evidence-reason-flap.yaml
@@ -1,0 +1,9 @@
+kind: fixed
+area: sessions
+change: >
+  A running session no longer renames itself once a second. The reason behind a
+  state named the freshness window the tick happened to sample the agent's last
+  title frame in, so a busy turn crossing that window — which a turn painting a
+  frame roughly once a second does constantly — was reported as a change, and
+  every one of those wrote a durable event and pushed the session to every open
+  client. A turn that keeps running now has one reason for as long as it runs.
--- a/internal/attention/turn_test.go
+++ b/internal/attention/turn_test.go
@@ -51,7 +51,7 @@ func TestBreaksSnooze(t *testing.T) {
 		{"a session sitting at its prompt", protocol.SessionStateIdle, string(sessionstate.ReasonAtPrompt), false},
 		{"a question", protocol.SessionStateWaitingInput, string(sessionstate.ReasonQuestionOpen), false},
 		{"an approval", protocol.SessionStatePendingApproval, string(sessionstate.ReasonApprovalOpen), false},
-		{"working", protocol.SessionStateWorking, string(sessionstate.ReasonHeartbeatFresh), false},
+		{"working", protocol.SessionStateWorking, string(sessionstate.ReasonHeartbeatBusy), false},
 		{"scheduled", protocol.SessionStateScheduled, string(sessionstate.ReasonCronPending), false},
 		{"recoverable, which the daemon revives unattended", protocol.SessionStateRecoverable, "", false},
 
--- a/internal/daemon/session_evidence_test.go
+++ b/internal/daemon/session_evidence_test.go
@@ -183,7 +183,7 @@ func TestTheResolveTickPublishesTheResolution(t *testing.T) {
 	}
 	// The reason names the clause that won, which is the whole diagnostic value
 	// of the row: "working" alone never explains a wrong color.
-	if !strings.Contains(got.Detail, string(sessionstate.ReasonHeartbeatFresh)) {
+	if !strings.Contains(got.Detail, string(sessionstate.ReasonHeartbeatBusy)) {
 		t.Fatalf("detail %q does not name the winning clause", got.Detail)
 	}
 }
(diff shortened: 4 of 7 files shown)


figgyster (approved):
Approve. The flap dies, and the witnesses hold up.

Traced the full resolver before trusting the "precedence untouched" claim. It is — with one label shift that IS the fix. In the fresh-heartbeat rescue clause, an open bracket now names the reason. Before, `heartbeat_fresh` won that clause before the bracket clauses were ever reached, so a busy claude with a bracket open flipped `bracket_open` ⇄ `heartbeat_fresh` every time a frame age crossed the 1.5s TTL. Now all three running-turn clauses funnel into `running()`, and the bracket is the name whenever one is open. The gap clause is only reachable with brackets closed, so its change is label-only (`heartbeat_gap` → `heartbeat_busy`), detail preserved. Byte-identical where it matters, different exactly where the flap lived.

Witnesses, all in the box:

- `internal/sessionstate` suite green; the four daemon evidence tests green with cgo (the real toolchain — pulled the prebuilt libghostty-vt to run them honestly; my first cgo-off run failed on the store's SQLite refusing to open, which is the sandbox, not this diff).
- Revert-witness, run for real: reverted only the naming, kept everything else. `TestTheEvidenceTickIsSilentWhileTheTurnKeepsRunning` goes red at **10 facts over 10 ticks** — one per second, exactly the production cadence. The test catches the bug it claims to catch.
- Full daemon suite: two plugin tests fail, but they fail identically on main — the sandbox can't spawn plugin runtimes. Not this diff.
- `StateIndicator` 13/13. No stale `heartbeat_fresh`/`heartbeat_gap` anywhere in go/ts/rs/docs, and the frontend only names unknown-reach reasons, so no coupling to the working-state labels.
- CI green all the way down at head, react-doctor 100/100.

One bookkeeping nit, non-blocking: the receipts disagree on the headline number. Body says 73.7% of the bus log (233,497/316,721 over 8.4 days), the daemon test comment says 74.7% over 7.5 days, the unit test comment says 66% of the entire bus log was this rename, and "two thirds of it was this" in the daemon comment works out to about 49% of the whole log, not 66%. The mechanism is proven regardless — the live A/B (103 facts vs 2) and the revert-witness both stand. But if one number goes into the permanent record, pick the canonical one and make the comments agree. Somebody will quote the wrong one back at you someday.

The scope call is right, too: no debounce, no coalescing, no gate weakening. The state never moved, so the publisher shouldn't have. Wire coalescing and bus retention stay separate work — 233k rows over eight days is the debt side of this ledger, and now it's on the record where it belongs.

figgyster (approved):
The receipts agree now. 73.7% (233,497/316,721 over 8.4 days) and the 81.6% consecutive-within-a-second figure are quoted verbatim in both test comments — one canonical measurement, no more 74.7%/66%/"two thirds" drift. Comments-only delta; focused suites green at this head. Re-approving at head.

PR #831, opened 2026-08-10
fix(build): define UNAME_S so install-daemon signs the binary on macOS

victorarias:
## Intent / Why

`make install-daemon` (and `install-daemon-dev`) ends with macOS killing the freshly installed daemon: `daemon ensure` reports `Killed: 9`. The recipe's codesign block is guarded by `if [ "$(UNAME_S)" = "Darwin" ]`, and `UNAME_S` was referenced by three Makefile recipes but defined by none — it expanded to empty, the guard never matched, and signing was silently skipped, leaving the sidecar with its ad-hoc linker signature inside a properly signed bundle. This forced everyone off the cheap daemon-only install tier onto full app builds.

## Summary

- define `UNAME_S := $(shell uname -s)`, with a comment explaining what it guards and what the absence broke

This is the same one-line fix that already landed on `epic/ext-a4` in #826, brought to main so main-based worktrees get working daemon-only installs before the epic lands. The epic sync will merge cleanly (identical line).

## Verification

- root-caused and fixed on the epic in #826 with a live install receipt; re-hit independently on a main worktree in #830's live verification, where hand-codesigning confirmed the diagnosis
- `make -n install-daemon` parses; changelog fragment validated with `go run ./cmd/changelog-check`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,12 @@ BUILD_TIME ?= $(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
 SOURCE_FINGERPRINT ?= $(shell bash ./scripts/source-fingerprint.sh --field fingerprint)
 GIT_COMMIT ?= $(shell bash ./scripts/source-fingerprint.sh --field commit)
 OUTPUT ?= $(BINARY_NAME)
+# Guards the macOS-only branches below (code signing, the lsof port check). It
+# was used by three recipes and defined by none, so `$(UNAME_S)` expanded to
+# empty and `install-daemon` skipped signing entirely: the copied binary kept its
+# ad-hoc linker signature inside a properly signed bundle, and macOS answered
+# `daemon ensure` with Killed: 9.
+UNAME_S := $(shell uname -s)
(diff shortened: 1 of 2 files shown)


figgyster (approved):
Verified this one from the crime scene up. The bug's real: `UNAME_S` is referenced by three recipes on main — `test-e2e`'s lsof kill, the `install-daemon`/`install-daemon-dev` codesign block, and `sign-app` — and defined by none. Empty expansion, Darwin guard never fires, signing silently skipped, sidecar keeps its ad-hoc linker signature inside a signed bundle, and macOS answers `daemon ensure` with `Killed: 9`. Exactly the corpse you described.

The fix is the right shape. One `:=` definition sitting with the other shell-derived vars, and a comment that explains what it guards and what the absence cost — that last part earns its keep, because this class of bug (guard reads an empty variable, silently does nothing) is exactly the kind that rots for months. The comment is also honest about the blast radius: this repairs all three recipes, not just `install-daemon` — `sign-app` would've told a macOS user their signing tool "is only needed on macOS" too. That's a bonus, not a nit.

Checks from my side: `make -n install-daemon` and `make -n install-daemon-dev` both parse, `git diff --check` clean, and the changelog fragment conforms to the validator schema (fixed/build/change, no unknown fields) — CI's Changelog gate agrees. The epic claim holds: comment and line are byte-identical to `epic/ext-a4`, so the sync merges clean. The one pre-existing `GO_LDFLAGS` spacing drift between epic and main is someone else's case, not this PR's.

CI's green where it can matter: Changes, react-doctor, both Linux daemon builds, Changelog. The Daemon test suite was still running when I looked, but this diff can't reach Go tests — it doesn't touch any.

Clean port, right timing: main-based worktrees get working daemon-only installs before the epic lands. Approve.

PR #833, opened 2026-08-10
feat(skill): make the CLI the syntax authority and add attn skill

victorarias:
## What

Guidance patterns adopted from studying herdr's agent skill (a terminal multiplexer for coding agents with a similar CLI-driven skill), applied to attn's bundled agent skill, plus one new CLI command.

**Skill guidance (`internal/agent/attn_skill/`):**
- The installed binary is now the stated authority for command syntax: the Bootstrap section tells agents to discover flags via `attn --help` and each group's own help, and the references keep only the rules and concepts the help cannot carry. `tickets.md` and `delegation.md` drop their hand-maintained flag mirrors — a second source of truth that drifts from the CLI with no test to catch it.
- Probing discipline: never explore with bare `attn` (it launches or attaches a session — a real footgun for an agent poking around), and never probe a mutating command by omitting its arguments.
- The skill description gains a negative trigger: do not use the skill merely because a task could benefit from delegation or parallel agents — the known opportunistic-delegation failure mode.
- New shared rule: read identifiers and state from command output instead of predicting them.

**New command:** `attn skill` prints the bundled, release-matched skill; `--list` names the references and `--reference <name>` prints one. This is the manual path for agents attn does not auto-install the skill for, and a version-skew diagnostic across profiles. An unknown reference name fails naming the ask and the bundled names.

## Verification

- `go test ./cmd/attn ./internal/agent` green; `go vet` and `changelog-check` clean.
- Ran the built binary directly: `attn skill`, `--list`, `--reference delegation`, and the unknown-name error path. `attn skill` is a self-contained CLI command that never talks to the app or daemon, so per AGENTS.md the built binary plus unit tests is the matching verification tier; the skill-install path that copies these files at session launch is unchanged and covered by existing `internal/agent` tests.

https://claude.ai/code/session_01QbZebKEamJpHThxfWwxrUR

## Command output

`attn skill --help`:

```
usage: attn skill [--reference <name>] [--list]

print the agent skill bundled with this binary — the instructions attn
installs for supported agents, release-matched to the running version.

  (no flags)            print SKILL.md
  --reference <name>    print one bundled reference, e.g. tickets
  --list                list bundled reference names
```

`attn skill` (start of the full SKILL.md print):

```
---
name: attn
description: "Operate attn capabilities from an agent, including user-steered delegations, tickets, workflows, shared workspace context, the Notebook, Present reviews, markdown, and the in-app browser. Use when the user explicitly asks for an attn capability or delegation, or when acting as attn's chief of staff. Do not use merely because a task could benefit from delegation, parallel agents, or a background terminal."
---

# attn
…
```

`attn skill --list`:

```
browser
delegated-agent
delegation
markdown
notebook
present
tickets
workflow
workspace-context
```

`attn skill --reference tickets` (start):

```
# Writing a good ticket

A ticket's `description` is the brief — the literal prompt an agent receives when the
ticket is delegated, and the cold-start spec when the work is picked up later. It is
…
```

Error paths:

```
$ attn skill --reference nope; echo $?
skill: no bundled reference "nope"; bundled references: browser, delegated-agent, delegation, markdown, notebook, present, tickets, workflow, workspace-context
1

$ attn skill --list --reference tickets; echo $?
skill: --list and --reference are mutually exclusive; pass one
2

$ attn skill --reference; echo $?
skill: --reference requires a name; run `attn skill --list` for the bundled names
2
```

### DIFF
--- a/changelog.d/skill-cli-authority.yaml
+++ b/changelog.d/skill-cli-authority.yaml
@@ -1,0 +1,16 @@
+kind: added
+area: cli
+change: >
+  `attn skill` prints the agent skill bundled with the binary —
+  release-matched instructions for driving attn — with `--list` and
+  `--reference <name>` for the per-capability references. The skill itself
+  now treats the CLI's own help as the authority for command syntax: the
+  references keep the rules and concepts and drop their hand-maintained
+  flag mirrors, and the skill warns agents never to explore with bare
+  `attn` (it launches a session) or probe mutating commands by omitting
+  arguments.
+notes: >
+  Guidance patterns adapted from herdr's agent skill: binary-as-authority
+  for syntax, probing discipline, a negative trigger in the skill
+  description ("do not use merely because a task could benefit from
+  delegation"), and a read-IDs-from-output-not-prediction shared rule.
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -216,6 +216,8 @@ func main() {
 		runList()
 	case "presence":
 		runPresence()
+	case "skill":
+		runSkill()
 	case "activity":
 		maybePrintProfileBanner()
 		runActivity()
@@ -648,6 +650,7 @@ commands:
 	  daemon ensure|stop                ensure the daemon is running, or stop it
   profile <status|resolve|list>     show / resolve the active profile's resources
   profile-env <profile|--unset>     print shell commands for selecting a profile
+  skill [--reference <name>]        print the bundled agent skill and its references
   version                           print version information
 `)
 }
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -650,7 +650,7 @@ commands:
 	  daemon ensure|stop                ensure the daemon is running, or stop it
   profile <status|resolve|list>     show / resolve the active profile's resources
   profile-env <profile|--unset>     print shell commands for selecting a profile
-  skill [--reference <name>]        print the bundled agent skill and its references
+  skill [--reference <name>|--list] print the bundled agent skill and its references
   version                           print version information
 `)
 }
--- a/cmd/attn/skill.go
+++ b/cmd/attn/skill.go
@@ -39,6 +39,10 @@ func writeSkill(stdout, stderr io.Writer, args []string) int {
 		}
 	}
 
+	if list && reference != "" {
+		fmt.Fprintln(stderr, "skill: --list and --reference are mutually exclusive; pass one")
+		return 2
+	}
 	if list {
 		for _, name := range agent.SkillReferenceNames() {
 			fmt.Fprintln(stdout, name)
--- a/cmd/attn/skill.go
+++ b/cmd/attn/skill.go
@@ -1,0 +1,76 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/victorarias/attn/internal/agent"
+)
+
+// runSkill prints the bundled agent skill so any agent or human can read the
+// release-matched copy without locating an installed skill directory.
+func runSkill() {
+	os.Exit(writeSkill(os.Stdout, os.Stderr, os.Args[2:]))
+}
+
+func writeSkill(stdout, stderr io.Writer, args []string) int {
+	reference := ""
+	list := false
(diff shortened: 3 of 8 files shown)


figgyster (approved):
Approve. This one's the same class of fix as the last few, aimed at the same villain: a second source of truth drifting from the CLI with no test to catch it. The flag mirrors in the references are gone, and I went and checked that the authority they defer to actually exists — it does.

**The premise holds, verified against the built binary:**

- `attn skill` prints SKILL.md (frontmatter included), `--list` names all 9 references, `--reference tickets` (and `tickets.md` — the TrimSuffix path works) prints the reference, an unknown name fails naming both the ask and the bundled names (exit 1), bare `--reference` exits 2 pointing at `--list`, unknown args exit 2 with help, `--help` exits 0. All exactly as the body describes.
- The skill now tells agents "discover flags via help" — so I checked the help actually carries what the references still lean on. `attn ticket` bare prints full syntax for every subcommand including the four "reach any ticket" commands and their flags (`--status`, `--all`, `--json`, `--ticket`, `--confirm`, `-m`). `attn delegate --help` covers every flag the delegation reference still mentions — `--agent`, `--model`, `--effort`, `--workspace`, `--new-workspace`, `--cwd`, `--worktree`, `--repo`, `--from`, `--source-session`, `--worktree-path` included. `attn workflow` and `attn browser` bare group helps exist. The `-m` shorthand tickets.md recommends is real (`fs.StringVar(message, "m", ...)`), and the bare-`attn` footgun warning is accurate — `runWrapper` launches an agent or deep-links the app.
- The two worktree rules the delegation reference *keeps* are genuinely not in `delegate --help` (help says only "branch or ref to start from" and "--repo ... defaults to the repository the target workspace's sessions are in" — nothing about `origin/<default>` precedence or the multi-repo failure). So the references kept exactly the concepts help can't carry and dropped exactly the syntax help can. The line holds.
- Single embed source (`//go:embed attn_skill`), no duplicate copies anywhere in the tree — the byte-identical-across-agents invariant is preserved by construction, and `attn skill` reads the same embed the installer copies, so what it prints is what gets installed.

**Witnesses:** build clean (needed the prebuilt libghostty-vt fetch — the sandbox lacks the C headers, same as #830, unrelated to this diff), `go vet` clean, gofmt clean, diff-check clean, changelog gate OK. `go test ./cmd/attn ./internal/agent` green including all five new skill tests — which cover the happy paths and the two error paths (unknown name, bare `--reference`). CI green at head: Daemon, Changes, Changelog, react-doctor, Linux daemon both arches, Frontend E2E all success; frontend/Tauri/plugins skipped as expected for a Go+docs-only diff.

Two nits, neither blocking:

1. `--list` and `--reference` together: the list silently wins and the reference is ignored (`attn skill --list --reference tickets` prints the list, no complaint). If a probe combines them, it gets one of two intents with no signal which. Either reject the combination as a usage error or document the precedence in the flag help — silent ignore is the one behavior that'll confuse the exact audience this command exists for.
2. Top-level `attn --help` lists `skill [--reference <name>]` but omits `--list`, while the command's own help documents both. Cosmetic, but it's the surface the skill now tells agents to trust as authoritative — worth keeping the two help texts in lockstep, since that's the whole point of the PR.

The arc keeps its shape: #829 killed generated-file drift, #830 killed the reason-flap, this kills the hand-maintained flag mirrors. Good deletion, well-earned.

figgyster (approved):
Both nits closed, cleanly. Re-approving at 595d96ad.

1. `--list` + `--reference` now rejects as a usage error (exit 2) with the conflict named, in either flag order — and the new test is the honest kind: it pins the exit code, the error text, *and* that stdout stays empty on a usage error, so a silent partial write can't sneak back in.

2. Top-level help now reads `skill [--reference <name>|--list]`, matching the command's own help — the authority surface keeps itself honest.

Verified at head: build, `go test ./cmd/attn ./internal/agent` (all six skill tests green), vet, gofmt, diff-check, changelog gate all clean; ran the binary directly for the combined-flags rejection in both orders and confirmed list/reference still work standalone. CI on the new head is still spinning (Daemon/Changelog/Linux daemon pending), but the delta is cmd/attn-only with a matching test — nothing for those jobs to trip on. Full loop: two nits, two commits, zero arguments.

PR #834, opened 2026-08-10
ci(protocol): fail when the generated types drift from the schema

victorarias:
## What

CI regenerates the protocol types from `internal/protocol/schema` and fails on any
difference. A stale or hand-edited `internal/protocol/generated.go` /
`app/src/types/generated.ts` can no longer merge.

## Why

Both files are compiled output, and nothing verified that the committed copies match
what the schema produces. The existing jobs cannot do it:

- **Frontend** runs `tsc`, but quicktype emits an `any` index signature on every
  generated interface, so a missing wire field is not a type error.
- **Daemon** compiles Go against `generated.go`, which is only a check of the file
  against itself.

Measured, on this branch, by deleting `transcript_path?: string;` from `SetSessionResumeIDMessage`
in `generated.ts` and committing it:

```
tsc exit=0            # the field is simply gone, and nothing objects
check-types exit=2    # +    transcript_path?:  string;
```

So today a dropped field reaches the app as a runtime-only absence — the wire carries
it, the type says it does not exist, and the code that would have read it was never
written. This is also the class of drift that produced the ~1200-line phantom diff in
#829 and had both the author and the reviewer chasing the wrong mechanism, so it is
worth a gate rather than vigilance.

## What changed

**`.github/workflows/ci.yml`** — a `Generated types` job behind a new `generated`
path filter. The filter lists the pipeline's inputs *and* both of its outputs, so a
hand-edit is gated even when no schema moved; that edit is exactly what this catches.

**`Makefile`** — two things the recipe needed before it could be trusted off one
machine:

- `go-jsonschema` was the only generator input with no version pin. The recipe ran
  whatever `go install` had last left behind, so two machines on different versions
  could disagree about `generated.go` the same way collation made them disagree about
  `generated.ts` (fixed in #829). Pinned to `v0.24.1` — the version that produces the
  committed file today; v0.23.0 also reproduces it byte-clean, so the pin is a
  deliberate choice rather than a rescue. It installs only on a version mismatch, so
  the usual run costs one `go version` and no network.
- The binary was resolved as `$(HOME)/go/bin/go-jsonschema`, correct only when `GOBIN`
  is unset. Under asdf the install lands elsewhere: `make check-types` on a clean
  checkout failed here with `No such file or directory` / `Error 127` before this
  change.

## Verification

- `make check-types` green on a clean tree, from a checkout where `go-jsonschema` was
  absent (the pin installed it, second run reinstalled nothing).
- The drift probe above: hand-edit committed, `tsc` passes, the gate fails and names
  the missing line. Probe commit dropped; not in this branch.
- `go run ./cmd/changelog-check` clean.

## Two things a reviewer should know

- `npx quicktype@26.0.0` fetches from npm at run time, so this job inherits npm-outage
  flakiness. Vendoring it is a bigger change than this one, and the version pin is the
  part that matters for reproducibility.
- A skipped path-filtered job produces no check, same as `Daemon` and `Frontend`. If
  this becomes a required check in branch protection it needs the `frontend-e2e-gate`
  treatment; it is not one today.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       daemon: ${{ steps.filter.outputs.daemon }}
+      generated: ${{ steps.filter.outputs.generated }}
       frontend: ${{ steps.filter.outputs.frontend }}
       tauri: ${{ steps.filter.outputs.tauri }}
       plugins: ${{ steps.filter.outputs.plugins }}
@@ -47,6 +48,15 @@ jobs:
               - 'app/vite.config.ts'
               - 'internal/protocol/schema/**'
               - '.github/workflows/ci.yml'
+            # Every input to the type pipeline plus both of its outputs. The
+            # outputs are listed so a hand-edit is gated even when no schema
+            # moved — that edit is exactly what this catches.
+            generated:
+              - 'internal/protocol/schema/**'
+              - 'internal/protocol/generated.go'
+              - 'app/src/types/generated.ts'
+              - 'Makefile'
+              - '.github/workflows/ci.yml'
             tauri:
               - 'app/src-tauri/**'
               - '.github/workflows/ci.yml'
@@ -71,6 +81,42 @@ jobs:
 
       - name: Gate
         run: ./scripts/changelog-gate.sh "${{ github.base_ref }}" "${{ github.head_ref }}"
+
+  # generated.go and generated.ts are compiled from internal/protocol/schema, and
+  # nothing was checking that the committed copies match. The other jobs cannot:
+  # the generated TS interfaces carry an `any` index signature, so `tsc` accepts a
+  # file with a wire field deleted from it — measured, exit 0. This regenerates
+  # from the schema and fails on any difference.
+  generated-types:
+    name: Generated types
+    needs: changes
+    if: needs.changes.outputs.generated == 'true'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.22'
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: 9
+
+      # The TypeSpec compiler the recipe invokes as `pnpm exec tsp`.
+      - name: Install schema toolchain
+        working-directory: internal/protocol/schema
+        run: pnpm install --frozen-lockfile
+
+      - name: Regenerate and diff
+        run: make check-types
 
   backend:
     name: Daemon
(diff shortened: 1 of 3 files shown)


figgyster (approved):
**Approved.**

Ran the gate the way CI runs it. Clean checkout at head, `make check-types`: regeneration lands byte-identical, exit 0. The committed files are canonical, which is the premise this whole thing rests on.

Then I tried to break it. Committed a probe that deleted `transcript_path?: string;` from generated.ts — the gate died exactly as advertised, exit 2, and the diff named the missing line. That is the property that matters, and it holds.

The blindness half of the argument reproduces too. Every generated interface carries `[property: string]: any`, and `tsc --strict` exits 0 on the file before and after the field is deleted. The field vanishes and nothing objects. Your measurement, re-run on this branch, same answer.

The details hold up:

- The filter lists both outputs as inputs, so a hand-edit to generated.ts alone trips the job even when no schema moved. That is the case this gate exists for.
- The pin completes the reproducibility trio: collation was pinned in #829, quicktype@26 was already pinned, go-jsonschema is now the third. And v0.24.1 is the honest pin — it is what produces the committed file, and the mismatch check no-op'd in my run, so "one `go version`, no network" is real.
- GOBIN resolution is a genuine fix: with GOBIN unset the recipe resolved to ~/go/bin in my run, and under asdf the old hardcode dies with 127. Your exact failure.
- LC_ALL=C rode into the gate, so the #829 collation fix cannot be lost by a future edit to the recipe.
- Changelog gate passes locally and in CI, and the Generated types job itself is green in CI at head (1m7s) — the part no local run can prove: the job actually wires up and runs.

1. The gate's contract is *committed* drift, not dirty-tree drift. `check-types` regenerates in place, so an uncommitted hand-edit gets clobbered before the diff runs. My first probe (uncommitted) passed silently; the committed one failed. In CI that is exactly right. Locally it means `make check-types` on a dirty generated file silently overwrites instead of flagging — worth a line in the recipe comment so nobody learns it the way I did.

2. The probe in the body names `SessionInfo`, but the optional field actually lives on `SetSessionResumeIDMessage` — the other three `transcript_path`s in the file are required. The mechanism reproduces exactly and the probe was dropped, so it is cosmetic. But the receipts in this repo are exact by convention, and the record should be too.

The arc lands where #829 pointed: that lesson was "run the generator gate on every protocol PR," and now it is a job, not a reviewer's memory. Same shape as #833 — kill the drifting source of truth, give it a witness. Only here the artifact is compiled output and the witness is the pipeline itself.

figgyster (approved):
Both observations, closed with one commit. The recipe comment now says the quiet part out loud — check-types flags *committed* drift because generate-types rewrites in place, and it even tells you how to reproduce a drift failure locally (commit the edit). That is the exact lesson from my probe, minus the hour of staring at an exit 0. And the body's probe now names the right type: `SetSessionResumeIDMessage`, the one interface where `transcript_path` is actually optional.

Verified at the new head: diff-check clean, `make check-types` green (regenerate byte-identical, exit 0), Changelog pass in the fresh CI run. Comments-only delta, so the earlier witnesses carry over unchanged.

Re-approved. The gate's contract is now documented, and the receipts are exact.

PR #835, opened 2026-08-10
docs(agents): publish on change, not on tick — producer-rate discipline

victorarias:
## Intent / Why

A recent producer bug did recurring work on every sampling tick while nothing changed, and quietly dominated the durable bus log. The defect class is general — publishes, pushes, scans, repaints — but the guide only warned about its repaint instance.

## Summary

One bullet added to AGENTS.md's Change discipline, directly above the no-repainting rule it generalizes:

- idle systems must be idle: recurring work (timers, watchers, changed-gates) does real work only when something moved;
- the named trap: a gate comparing an unstable value (clock-derived, sampled) fires forever while nothing changes;
- the check: new recurring work is verified silent on a quiet live session.

Deliberately loud producers stay untouched — the rule targets unintentional volume, not volume.

## Verification

Docs-only; changelog fragment validated with `go run ./cmd/changelog-check`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -466,6 +466,15 @@ carries the enumerated exception list.
 - Retention trims past the age window but never past an **enabled** consumer's
   cursor. Disabled consumers do not pin the log; they resume at head with a
   logged gap.
+- A producer publishes when the fact changed, not when a timer ticked. Every
+  fact costs a durable row plus, through its projection, a push to every
+  client, so a producer's steady-state rate is part of its design: before
+  shipping a publisher inside a periodic loop, know its facts/hour idle and
+  busy, and make sure a tick where nothing moved publishes nothing. The
+  canonical failure is the 2026-08 evidence flap — a reason string that
+  renamed itself every second and silently wrote 66% of the whole log.
+  Measured healthy rates and the wire-load reconsider threshold live in
+  [docs/plans/2026-08-10-projection-coalescing.md](docs/plans/2026-08-10-projection-coalescing.md).
 - Operator surface: `attn bus status`, `attn bus disable|enable <consumer>`.
   The enabled bit is database-only on purpose — the kill switch must not depend
   on the daemon it kills.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -334,6 +334,11 @@ forgets that position and `--since <RFC3339>` replays from an instant.
 - Complexity belongs at the boundary. The daemon owns orchestration and stays
   authoritative (`applyState`, `wireProjections`); the frontend renders what it
   is told.
+- Idle systems must be idle. Anything recurring — a timer, a watcher, a "did
+  it change?" gate — does real work (publishes, pushes, scans, repaints) only
+  when something moved; a gate comparing an unstable value (clock-derived,
+  sampled) fires forever while nothing changes. Verify new recurring work
+  stays silent on a quiet live session.
 - No continuously repainting animations. attn renders GPU terminals, often on
   high-refresh displays, beside agents that run all day — a permanent repaint
   loop is a battery and thermal bug no test will catch.
@@ -466,12 +471,6 @@ carries the enumerated exception list.
 - Retention trims past the age window but never past an **enabled** consumer's
   cursor. Disabled consumers do not pin the log; they resume at head with a
   logged gap.
-- Publish freely on real change — the bus is built for it, and a fact a
-  consumer needs must never be skipped to save traffic. The one rule: a
-  publish reports a change, so a periodic loop where nothing moved this tick
-  publishes nothing. Sanity ceiling: every healthy fact name measures under
-  500 facts/hour (loudest observed: 480/h); a producer sustaining thousands
-  per hour is republishing something that did not change.
 - Operator surface: `attn bus status`, `attn bus disable|enable <consumer>`.
   The enabled bit is database-only on purpose — the kill switch must not depend
   on the daemon it kills.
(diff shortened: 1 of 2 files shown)


figgyster (changes requested):
The discipline is right, the placement is right, and the number is wrong — the same number this PR exists to retire.

**Blocking: 66% is retired drift. The canonical figure is 73.7%.**

`AGENTS.md:475` and the changelog fragment both say the flap "silently wrote 66% of the whole log". That figure was measured, found wanting, and explicitly removed twelve minutes after #830 merged. The #830 loop-closer (79e1da32) retired 74.7%/66%/"two thirds" in favor of one canonical measurement, now quoted verbatim in two test comments on main:

> Measured over 8.4 production days, `session.state.changed` was 73.7% of the whole bus log (233,497 of 316,721 facts)

(`internal/daemon/session_evidence_test.go:659`, `internal/sessionstate/sessionstate_test.go:1042`.)

I flagged the drift on the record during #830 — "somebody will quote the wrong one back at you someday" — and here it is, eight hours later, in the very PR whose job is to stop the next author from doing exactly this. A producer-rate rule that ships a drifting number into the agent guide embeds the same defect class it warns about. Please swap 66% → 73.7% in both files (PR body too), ideally with the verbatim "233,497 of 316,721 facts over 8.4 days" receipt so the guide and the tests agree the way #830 made them agree.

**Non-blocking: the plan link dangles.** `docs/plans/2026-08-10-projection-coalescing.md` does not exist on main — it lives only on `design/projection-coalescing`, which has no open PR yet. The body acknowledges this, and a repo-relative 404 in a plan pointer is minor, but if the guide is the thing that survives, the link 404s for every agent until that PR lands. Consider landing the doc first or keeping the number self-contained in the bullet.

Verified: diff-check clean, guidance text consistent with the #830/#832 arc and the coalescing doc's measured 357/h post-fix rate, changelog fragment structurally sound (couldn't run changelog-check locally — no go in this sandbox; CI Changelog job is running it).

figgyster (approved):
Re-approving at head — the rework is better than the swap I asked for.

**The drift is gone, not just corrected.** 66% is nowhere: not in the bullet, not in the fragment, not in the body. Instead the rule now carries a measured ceiling — healthy fact names all run under 500 facts/hour, loudest observed 480/h — and that number has receipts: it's the healthy max over the 6h window in #832's own tripwire table (6h healthy 480/h vs 5,763/h flapping peak, 24h 357/h vs 1,075–2,265/h), with the 1000/h tripwire sitting at 2.08x headroom. A ceiling with a provenance beats a war-story percentage every time. And "publish freely on real change" keeps the #830 principle intact — a reason that genuinely moves still publishes; only the no-movement tick stays silent.

The dangling plan-doc link is gone too, which settles my other note cleanly. diff-check clean, Changelog gate green, CI all green at head.

One heads-up for the road, non-blocking: the plan doc on `design/projection-coalescing` still says the flap fix "removes 73% of the wire pushes and 66% of the durable bus log". The 73% matches its own push math (260,273 → 71,716), but the 66% is the same retired figure — the canonical is 73.7% of the whole log (233,497/316,721). When that doc lands via its own PR, swap it before it becomes the permanent record. The prophecy keeps trying to come true; this time it's a different document.

figgyster (approved):
Re-approving at the rebased head (4932977c) — the generalization is the right move, and it landed clean.

**What changed since my last look:** the PR was rebased over #832/#834, so its own diff vs main is now just the bullet plus the fragment — and the bullet was generalized out of the bus section into Change discipline proper: "Idle systems must be idle." That is the better home. The flap was one instance of a bigger class — any recurring timer, watcher, or "did it change?" gate that does real work when nothing moved — and stating it as a first-class principle next to "No continuously repainting animations" makes it the rule for every future producer, not just bus authors. The mechanism description is exact, too: "a gate comparing an unstable value (clock-derived, sampled) fires forever while nothing changes" is the heartbeat-TTL boundary crossing, named by its shape. And the closing instruction — verify new recurring work stays silent on a quiet live session — promotes the #830 live-test witness into a standing requirement. That is the #830 lesson, promoted from a fix to a law.

Verified at head: PR diff vs main is 12 insertions (bullet + fragment), diff-check clean, no 66% anywhere, fragment updated to match the generalized wording, CI green (Changes, react-doctor, Changelog, Frontend E2E; Generated types correctly SKIPPED — no protocol files in this PR).

One coverage note, non-blocking: the generalization dropped the paired guard the earlier wording carried — "publish freely on real change; a fact a consumer needs must never be skipped to save traffic." The new bullet keeps the silence direction only. The opposite failure — a future author over-suppressing real changes to save traffic, which is exactly the coalescing idea #836 records a measured "no" against — is now covered only by that plan doc, not by the guide. Consider one clause ("real change still publishes") so the rule reads both ways. The rest of the arc is complete: the flap is a footnote, the principle is a law, and the numbers have receipts.

figgyster (approved):
Re-approving at ecbff1d2 — the rule reads both ways now, one commit, exactly the clause I asked for.

"…stays silent on a quiet live session — and still fires on every real change; never suppress work a consumer needs to save traffic." The silence direction and the suppression guard now sit in the same sentence, so a future author gets the whole law from one bullet: idle work must be idle, and real change is never the thing you save on. That closes the one coverage gap in the generalization — the "no" from #836 is now in the guide, not just in the plan doc.

Verified at head: diff is still just the bullet plus the fragment (13 insertions vs main), diff-check clean, fragment unchanged from the head the Changelog gate already passed on, CI running green so far (Changes SUCCESS; react-doctor and Changelog in progress — docs-only, nothing else can move).

The arc is done. #830 fixed the flap, #832 gave it a witness, #835 made it a law — and the law has both edges. This one goes in the book as the rare case where the reviewer asked for one clause and got it, word for word, two minutes later.

PR #836, opened 2026-08-10
docs(bus): record the projection-coalescing decision and its measurements

victorarias:
Docs only. Adds `docs/plans/2026-08-10-projection-coalescing.md`, the design record for a question that was asked and answered with a "no": should wire projections coalesce their pushes per subject over a short window?

## What was asked

Every bus fact that matches a wire projection pushes to every WebSocket client immediately (`wireProjections`, `internal/daemon/bus.go`). Bulk operations collapse their whole-list pushes through `coalesceSnapshots`, but per-fact traffic does not coalesce at all. The concern was that under legitimate load — many working sessions, each honestly changing a few times a minute — the per-fact × per-client multiplication is unbounded: frontend wakeups and battery, and pressure on the 256-message WebSocket buffer whose overflow disconnects clients.

## What the measuring found

The mechanism is not needed, by three orders of magnitude.

A separate fix removes the evidence-reason flap, where two co-true resolver clauses alternate at 1 Hz and republish `session.state.changed` without the state ever moving. In the world after that fix, the entire projection layer pushes **357 messages/hour** — one every ten seconds — and **30 per session-hour**. Worst 60-second window: 1.7/s.

The 256-message buffer is also not the projection layer's buffer. It is shared with PTY output, which already coalesces at the source (`internal/pty/session.go:238`, a 5 ms window bounded at 256 KB) and can still reach **200 messages/second for a single session**. Eviction requires three *consecutive* full-buffer sends, and the production daemon log carries zero slow-client lines. Coalescing projections cannot meaningfully protect a buffer filled by something a thousand times louder.

The doc records the reconsider threshold as a number rather than a feeling: re-open when the projection layer's worst 60-second window exceeds 25 pushes/second at a client, which from the measured per-session rate is roughly 3,000 concurrently working sessions. Production peaks at 19.

## Receipts

Two independent read-only sources, both against production, both described in the doc so they can be re-run:

- **Fact rates** modelled over 201 hours of `bus_events` (315,073 facts) copied out with a read-only `.backup`, mapped through `wireProjections` to the wire message each fact produces, with snapshot coalescing applied where the producer wraps its publishes.
- **Payload sizes** read off the live socket by an observer connected as an ordinary client. This one cannot be skipped: fact counts say nothing about bytes. `tickets_updated` fires 1.1 times an hour and is 217 KB, the single largest thing attn puts on the wire.

## What it turned up instead

Two producers that publish when nothing changed — the same defect class as the flap, and neither fixable by coalescing downstream. Both are recorded in the doc with their measurements and are already tracked separately; this PR does not change any code.

- `tickets_updated` is 217 KB, bypasses `projectSnapshot` entirely (its key `snapshotTickets` sits declared and unreferenced beside the eleven that are wired), and 193 KB of it is ticket descriptions that no client reads.
- `plugin.health.changed` publishes unconditionally on every 15-second poll — 5,760 durable rows and 5,760 disk-scanning pushes a day per plugin, while idle.

## What is kept for later

The mechanism sketch, in case the threshold is ever crossed: leading-edge-immediate plus trailing flush (trailing-only would be a pure latency regression at these rates), why no window constant has a defensible receipt today, the ordering rules that make dropping intermediate pushes safe, and the traffic that must never be coalesced.

## Verification

Docs only — no code, no protocol, no daemon or app surface — so it is exempt from live verification. The changelog gate passes locally.

https://claude.ai/code/session_01V87zZ94FTdJaZp3uqFQBHe

### DIFF
--- a/docs/plans/2026-08-10-projection-coalescing.md
+++ b/docs/plans/2026-08-10-projection-coalescing.md
@@ -46,7 +46,7 @@ Two independent sources, both read-only against production:
 1. **The durable bus log.** `~/.attn/attn.db` copied out with a read-only
    `.backup` (never opened read-write). 315,073 facts spanning
    2026-08-02T10:24:55Z → 2026-08-10T19:26:58Z — 201 hours, 110 sessions,
-   1,461 session-hours. A model script maps every fact through
+   1,461.4 session-hours. A model script maps every fact through
    `wireProjections` to the wire message it produces, applies snapshot
    coalescing where the producer wraps its publishes in `coalesceSnapshots`,
    and reports the resulting push timeline.
@@ -93,7 +93,7 @@ below reports both worlds; the flap-fixed column drops exactly those.
 The last row is deliberately not `total ÷ session-hours` (which would read
 178.1 → 49.1). It counts only the pushes a session causes — the
 `session.state.changed` projection, 232,111 today and 43,554 post-fix — over the
-1,461 session-hours in the window. Dividing the total instead charges each
+1,461.4 session-hours in the window. Dividing the total instead charges each
 session a share of the 15-second plugin health poll and the PR refresh, which
 run whether or not a session exists. This is the row the threshold below is
 derived from, and it has to scale with session count to mean anything.
@@ -243,7 +243,7 @@ a number, so nobody has to re-derive it.
 ### What to do instead
 
 1. **Land the flap fix** (sibling work, ticket `a4-slice5-rt`). Over this
-   window it removes 188,557 facts — 72.4% of the wire pushes and 59.9% of the
+   window it removes 188,557 facts — 72.4% of the wire pushes and 59.8% of the
    durable bus log.
 2. **Give `plugin.health.changed` a delta gate.** Publish only when
    status or message actually moves. Same shape as the flap fix, ~5 lines. This
--- a/docs/plans/2026-08-10-projection-coalescing.md
+++ b/docs/plans/2026-08-10-projection-coalescing.md
@@ -59,23 +59,30 @@ Two independent sources, both read-only against production:
 The bus log gives rates over a realistic week; the socket gives byte sizes the
 log cannot know, because a fact's payload is not its projection's payload.
 
-One known inaccuracy, in the conservative direction: the model assumed
-`ticket.*` collapses inside `coalesceSnapshots` like `pr.*` does. It does not
-(see below), so the true push count is 13 higher than the table reports across
-201 hours. Nothing else in the model depends on it.
+Two traps in building that model, both of which produced wrong numbers on the
+first pass and are worth knowing before re-running it:
+
+- `ticket.*` does **not** collapse inside `coalesceSnapshots` the way `pr.*`
+  does (see below). Assuming it did undercounted by 13 pushes over 201 hours.
+- Deriving the gap between consecutive facts in SQL as
+  `CAST((julianday(b) - julianday(a)) * 86400 AS INT)` truncates 698 genuine
+  2-second gaps to 1, inflating the flap's measured share. Compute gaps on
+  integer epoch seconds. Every push figure below comes from the integer path;
+  the drop is **188,557**, which is exactly 260,286 − 71,729 and exactly
+  232,111 − 43,554.
 
 ### Post-flap-fix is the world this is designed for
 
 A sibling change fixes the evidence-reason flap: `runEvidenceResolveLoop`
 re-resolves every session each second and broadcasts whenever the resolver
 *reason* changes, and on a busy session two co-true clauses alternate at 1 Hz
-without the state ever moving. In the log, **81.5%** of `session.state.changed`
+without the state ever moving. In the log, **81.2%** of `session.state.changed`
 facts land within 1 second of the previous fact on the same subject. The model
 below reports both worlds; the flap-fixed column drops exactly those.
 
 |                          | today       | post flap-fix |
 | ------------------------ | ----------- | ------------- |
-| pushes over 201 h        | 260,273     | 71,716        |
+| pushes over 201 h        | 260,286     | 71,729        |
 | average rate             | 1,295 /h    | **357 /h**    |
 | seconds with any push    | 23.1 % of wall clock | 8.6 % |
 | pushes/sec p50 / p99 / max | 1 / 5 / 36 | 1 / 3 / 36    |
@@ -112,18 +119,42 @@ Measured on the wire, production daemon:
 | `tickets_updated`         | **217,391** | 1.1 /h (from the log) |
 | `initial_state`           | 239,486     | once per connect |
 
-Bytes are the live observation throughout. Rates are the live observation too,
-except `tickets_updated` — it fired once in the window, too rarely to rate from
-it, so its rate is the log's 217 pushes over 201 hours.
-
 Whole-socket throughput during observation: **3.0 KB/s**, or 2.3 KB/s
 discounting the one-off `initial_state` — with the flap still running.
 
-Carrying the post-fix rates in the table above through these sizes puts the
-whole socket at **206 B/s** on average. Two thirds of that is the two messages
-this document ends up recommending fixing: `prs_updated` at 72 B/s and
-`tickets_updated` at 66 B/s, the latter from a message that fires roughly once
-an hour.
+**The rate column above is a short-window sample; do not compute with it.** It
+is what a five-minute observation saw, which is fine for saying "this arrives
+often" and useless as a rate. `sessions_updated` is the clearest trap: two
+messages in 57 seconds reads as 127/h, and the log's 201 hours put it at 0.8/h.
+Rates to carry through the sizes come from the model, below.
+
+### The post-fix byte budget
+
+Model push counts over the 201-hour window, priced at the observed sizes above:
+
+| message                 | pushes | /h    | bytes   | B/s   |
+| ----------------------- | ------ | ----- | ------- | ----- |
+| `session_state_changed` | 43,769 | 217.8 | 720     | 43.6  |
+| `plugins_updated`       | 14,924 | 74.2  | 1,017   | 21.0  |
+| `prs_updated`           | 7,404  | 36.8  | 7,037   | 72.0  |
+| `workspace_state_changed` | 1,767 | 8.8  | 265     | 0.6   |
+| `tasks_changed`         | 1,642  | 8.2   | 25      | 0.1   |
+| `session_pty_resized`   | 1,042  | 5.2   | 720     | 1.0   |
+| `tickets_updated`       | 230    | 1.1   | 217,391 | 69.1  |
+| `sessions_updated`      | 160    | 0.8   | 6,009   | 1.3   |
+| everything else         | 791    | 3.9   | ≤1,000  | 0.4   |
+| **total**               | **71,729** | **356.9** | | **209.1** |
+
+**209 B/s** for the whole socket, post-fix. Two thirds of it is the two
+messages this document ends up recommending fixing — `prs_updated` at 72 B/s
+and `tickets_updated` at 69 B/s, the second from a message that fires about
+once an hour.
+
+The `session_state_changed` row counts 43,769, not the 43,554
+`session.state.changed` facts: seven other session facts (`registered`,
+`renamed`, `pin`, `cap`, `activity`, `conversation`, `workspace`) project the
+same wire message. The per-session-hour row above counts facts, this one counts
+pushes, and the 215 difference is why they are not the same number.
 
 The per-session push is 720 bytes. The daemon marshals it once and enqueues the
 same byte slice per client (`SendRawTextToMatchingClients`), so client fan-out
@@ -212,7 +243,7 @@ a number, so nobody has to re-derive it.
 ### What to do instead
 
 1. **Land the flap fix** (sibling work, ticket `a4-slice5-rt`). Over this
-   window it removes 189,255 facts — 72% of the wire pushes and 60% of the
+   window it removes 188,557 facts — 72.4% of the wire pushes and 59.9% of the
    durable bus log.
 2. **Give `plugin.health.changed` a delta gate.** Publish only when
    status or message actually moves. Same shape as the flap fix, ~5 lines. This
--- a/docs/plans/2026-08-10-projection-coalescing.md
+++ b/docs/plans/2026-08-10-projection-coalescing.md
@@ -1,0 +1,328 @@
+# Projection-side coalescing of snapshot pushes
+
+Status: proposal, for decision. Recommendation is **do not build it**.
+
+## The question
+
+Every bus fact that matches a wire projection pushes to every WebSocket client
+immediately (`wireProjections`, `internal/daemon/bus.go`). Bulk operations
+already collapse their whole-list pushes through `coalesceSnapshots`, but
+per-fact traffic does not coalesce at all. Under legitimate load — many working
+sessions each honestly changing a few times a minute — does that per-fact ×
+per-client multiplication need a per-(projection, subject) debounce, and what
+would it cost in latency?
+
+The answer is no, by three orders of magnitude, and the measuring turned up two
+producers that are worth fixing instead.
+
+## The short version
+
+After the evidence-reason flap is fixed, the entire projection layer pushes
+**357 messages per hour** — one every ten seconds — into a buffer that holds 256
+and is shared with PTY output that can legitimately produce **200 messages per
+second for a single session**. A debounce there is a mechanism with nothing to
+do.
+
+What the measuring did find:
+
+- `tickets_updated` is a **217 KB** message that bypasses the coalescing
+  machinery entirely, and **193 KB of it is ticket descriptions no client
+  renders**. One observed second in production pushed 1.09 MB of it.
+- `plugin.health.changed` publishes unconditionally every 15 seconds per
+  connected plugin — 5,760 durable rows and 5,760 disk-scanning pushes a day,
+  with attn idle. Once the flap is fixed it is the loudest wire message attn
+  produces.
+
+Both are the same defect class as the flap: a producer that publishes when
+nothing changed. Neither is fixed by coalescing downstream, and both are
+cheaper to fix than the mechanism this document was asked to design.
+
+## Receipts
+
+### How the numbers were taken
+
+Two independent sources, both read-only against production:
+
+1. **The durable bus log.** `~/.attn/attn.db` copied out with a read-only
+   `.backup` (never opened read-write). 315,073 facts spanning
+   2026-08-02T10:24:55Z → 2026-08-10T19:26:58Z — 201 hours, 110 sessions,
+   1,461 session-hours. A model script maps every fact through
+   `wireProjections` to the wire message it produces, applies snapshot
+   coalescing where the producer wraps its publishes in `coalesceSnapshots`,
+   and reports the resulting push timeline.
+2. **The live socket.** A throwaway WebSocket observer connected to the running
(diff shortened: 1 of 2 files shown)


docs/plans/2026-08-10-projection-coalescing.md
```diff
@@ -0,0 +1,328 @@
...
+
+|                          | today       | post flap-fix |
+| ------------------------ | ----------- | ------------- |
+| pushes over 201 h        | 260,273     | 71,716        |
+| average rate             | 1,295 /h    | **357 /h**    |
+| seconds with any push    | 23.1 % of wall clock | 8.6 % |
+| pushes/sec p50 / p99 / max | 1 / 5 / 36 | 1 / 3 / 36    |
+| worst 10 s window        | 62 (6.2 /s) | 42 (4.2 /s)   |
+| worst 60 s window        | 255 (4.2 /s) | **104 (1.7 /s)** |
+| per session-hour         | 158.8       | **29.8**      |
```
figgyster: The row the headline threshold hangs off. From the doc's own stated inputs: 260,273 ÷ 1,461 session-hours = 178.1 today, 71,716 ÷ 1,461 = 49.1 post-fix — not 158.8 / 29.8. The column ratio (5.33) also doesn't match the totals' ratio (3.63), so this isn't a rounding artifact. If the model computes per-session-hour as something other than pushes ÷ session-hours (an unweighted per-session mean?), the definition needs to be on the page — otherwise a reader doing the division lands on 49.1 and the 3,000-session threshold at line 220 becomes ~1,800. The decision stands either way; the record should say which number it means.

docs/plans/2026-08-10-projection-coalescing.md
```diff
@@ -0,0 +1,328 @@
...
+smooth traffic that already runs at one message per ten seconds. The best
+outcome of this investigation is discovering we do not need it.
+
+The condition under which this document should be re-opened is stated below, as
+a number, so nobody has to re-derive it.
+
+### What to do instead
+
+1. **Land the flap fix** (sibling work, ticket `a4-slice5-rt`). It removes 73%
+   of the wire pushes and 66% of the durable bus log.
```
figgyster: The same 188,557 dropped facts that are 72.4% of the wire pushes are 59.8% of the 315,073 durable facts (188,557 ÷ 315,073). "66%" doesn't reproduce from the doc's own stated inputs; should read ~60%.

docs/plans/2026-08-10-projection-coalescing.md
```diff
@@ -0,0 +1,328 @@
...
+   that no other client — the SDK, plugins — reads `description` off a list row,
+   and a protocol bump if the field leaves the wire type.
+
+Items 2 and 3 are small, independently shippable, and each has a measured
+before/after. None of them is this document's mechanism.
+
+### When to reconsider
+
+Re-open this when the projection layer's **worst 60-second window exceeds
+25 pushes/second sustained** at a client — 15% of one flooding session's PTY
```
figgyster: 25 ÷ 200 = 12.5%, not 15%. Either the threshold line is 30/s (30/200 = 15%) or the share is 12.5% — the doc should agree with itself.

docs/plans/2026-08-10-projection-coalescing.md
```diff
@@ -0,0 +1,328 @@
...
+| `session_state_changed`   | **720** (703–820) | 6,944 /h (flap running) |
+| `plugins_updated`         | 1,017       | 233–253 /h    |
+| `sessions_updated`        | 6,009       | 127 /h        |
+| `prs_updated`             | 7,037       | 47 /h         |
+| `tickets_updated`         | **217,391** | 1.1 /h        |
+| `initial_state`           | 239,486     | once per connect |
+
+Whole-socket throughput during observation: **3.0 KB/s**, or 2.3 KB/s
+discounting the one-off `initial_state` — with the flap still running. Post-fix
+the same traffic is roughly 100 B/s average.
```
figgyster: "The same traffic" reads as the whole socket, which from this doc's own table is ~0.5 KB/s post-fix: session_state_changed alone is 357/h × 720 B = 71 B/s, and adding plugins/prs/sessions/tickets/workspace at the table's rates lands near 514 B/s. ~100 B/s only works if this sentence means the session_state subject alone — one clarifying word either way.

docs/plans/2026-08-10-projection-coalescing.md
```diff
@@ -0,0 +1,328 @@
...
+#    map bus_events through wireProjections, collapse coalesced snapshot keys
+#    per (key, second), report pushes/sec and worst rolling windows, with and
+#    without facts whose gap to the previous fact on the same subject is <= 1s
+
+# 3. payload sizes: connect an observer to the live daemon as an ordinary
+#    client, send client_hello, never attach a PTY, log (arrival, event, bytes)
+```
+
+Step 3 is the one that cannot be skipped. Fact counts say nothing about bytes:
+`tickets_updated` is 1.1 events an hour and the single largest thing attn puts
```
figgyster: initial_state is 239,486 bytes in this same document's table — larger than tickets_updated's 217,391. "The single largest recurring thing" is the accurate phrasing; as written it contradicts the table two sections up.

docs/plans/2026-08-10-projection-coalescing.md
```diff
@@ -0,0 +1,328 @@
...
+list of snapshot keys (`internal/daemon/bus.go:561`) — the machinery was built
+with a slot for it and never wired. So `coalesceSnapshots` cannot collapse the
+board even where a producer wraps its publishes. Observed burst,
+2026-08-09T18:51:15Z: five `ticket.status_changed` facts in one second, which at
+today's board size is **1.09 MB** to every client.
+
+Of the 217 KB, `SELECT SUM(LENGTH(description)) FROM tickets` is **193,435
+bytes** across 45 tickets. The hook's own contract already says these are
+"bare rows … The detail view fetches full records itself"
+(`useDaemonSocket.ts:580`), and the detail panel reads `fullTicket` from a
```
figgyster: Citation drifted: the bare-rows contract comment is at useDaemonSocket.ts:597-598 ("Fired with the non-archived ticket board (bare rows)… The detail view fetches full records itself"), not :580. Everything else in this doc cited exactly, so this one stands out.

figgyster (changes requested):
Checked the receipts, every one I could reach. The mechanism claims all hold at head: the flap is exactly what the doc says it is (recordStateReason returns false when the reason hasn't moved, publishResolution broadcasts the 1 Hz reason-delta at session_evidence.go:516), projectTicketsUpdated broadcasts direct at ticket_board.go:97 with Description riding the bare rows at :147, snapshotTickets sits declared-and-unreferenced at bus.go:561 beside eleven wired keys, checkPluginHealth publishes on every branch at plugin_rpc.go:653, and the 256-slot client buffer with three-strike eviction is where it's said to be (websocket.go:434, :418). Changelog gate green locally, CI green, diff-check clean. The decision — do not build it — survives every correction below by three orders of magnitude.

But this document exists so nobody has to re-derive the numbers, and three of them don't reproduce from its own stated inputs. Details on the lines; the short list:

- Per-session-hour row (line 84): 260,273 ÷ 1,461 = 178.1, not 158.8; 71,716 ÷ 1,461 = 49.1, not 29.8. The column ratio (5.33) also doesn't match the totals' ratio (3.63), so this isn't rounding. If the model defines per-session-hour some other way, that definition needs to be on the page — it's the number the 3,000-session threshold hangs off.
- "66% of the durable bus log" (line 198): the same 188,557-fact drop is 59.8% of 315,073. The 73% wire-push figure checks (72.4%).
- "15% of one flooding session's PTY budget" (line 216): 25 ÷ 200 = 12.5%.

Requesting changes, not because the conclusion is in doubt — it isn't — but because a receipt that can't be re-derived from its own inputs is the one failure mode this doc was written to kill. Fix the three numbers (or pin the definition), and this is an easy approve.

docs/plans/2026-08-10-projection-coalescing.md
victorarias: Right that the definition was missing, and it is the one that mattered most. The row is session-attributable pushes only — the session.state.changed projection, 232,111 today and 43,554 post-fix, over 1,461 session-hours (ratio 5.33, which is why it does not track the totals ratio). Total ÷ session-hours would charge each session a share of the 15-second plugin health poll and the PR refresh, which run whether or not a session exists — and since the threshold asks "what if there were more sessions", the divisor has to be the part that scales with sessions. Written onto the page now, with that reasoning, and the threshold derivation restated as 25 ÷ (29.8 ÷ 3600) = 3,020.

docs/plans/2026-08-10-projection-coalescing.md
victorarias: Correct, and the cause is worth naming: 66% was carried over from the sibling investigation, which measured a different window (7.5 days, 275,845 facts). Against this document's own 315,073 it is 189,255 ÷ 315,073 = 60.1%. Now reads "over this window it removes 189,255 facts — 72% of the wire pushes and 60% of the durable bus log", so the fact count is on the page and both percentages divide out of numbers stated here.

docs/plans/2026-08-10-projection-coalescing.md
victorarias: My arithmetic error. Fixed to 12.5% — the threshold is the number with the reasoning behind it (the point where projections stop being noise beside PTY), so the share is what moves, not the line.

docs/plans/2026-08-10-projection-coalescing.md
victorarias: You are right that it reads as the whole socket and that ~100 B/s does not survive the table. Carrying the post-fix rates through the observed sizes gives 206 B/s, which is now on the page with the two lines that dominate it — prs_updated 72 B/s and tickets_updated 66 B/s, the latter from a message that fires about once an hour. Your 514 comes out if the live-observed rates are used instead of the post-fix ones; the sentence sits in the post-fix paragraph, so the model rates are the right input.

docs/plans/2026-08-10-projection-coalescing.md
victorarias: Agreed, and your phrasing is the accurate one. Now: "the largest recurring message attn puts on the wire — only initial_state, sent once per connect and carrying the same board, is bigger."

docs/plans/2026-08-10-projection-coalescing.md
victorarias: Fixed to :597. The drift is real: I read that line before rebasing onto main, and main moved under it. Re-checked every other citation in the doc at current head; they hold.

docs/plans/2026-08-10-projection-coalescing.md
```diff
@@ -0,0 +1,349 @@
...
+smooth traffic that already runs at one message per ten seconds. The best
+outcome of this investigation is discovering we do not need it.
+
+The condition under which this document should be re-opened is stated below, as
+a number, so nobody has to re-derive it.
+
+### What to do instead
+
+1. **Land the flap fix** (sibling work, ticket `a4-slice5-rt`). Over this
+   window it removes 189,255 facts — 72% of the wire pushes and 60% of the
```
figgyster: This drop (189,255) doesn't agree with the rest of the doc's own numbers. The totals imply 260,273 − 71,716 = 188,557; the session split implies 232,111 − 43,554 = 188,557. And the knock-ons: "81.5%" at line 72 only works with 189,255 (189,255 ÷ 232,111 = 81.54%; 188,557 ÷ 232,111 = 81.24%), while "72% of the wire pushes" only rounds to 72 with 188,557 (72.44 → 72; 189,255 → 72.71 → 73). The 60% of the durable log works either way (60.07% vs 59.84%). Three statements, one drop — pin the model's actual number and make the five places agree, or a reader doing the subtraction this commit invites lands on 188,557 and finds 81.5% reading 81.2%.

docs/plans/2026-08-10-projection-coalescing.md
```diff
@@ -0,0 +1,349 @@
...
+
+Bytes are the live observation throughout. Rates are the live observation too,
+except `tickets_updated` — it fired once in the window, too rarely to rate from
+it, so its rate is the log's 217 pushes over 201 hours.
+
+Whole-socket throughput during observation: **3.0 KB/s**, or 2.3 KB/s
+discounting the one-off `initial_state` — with the flap still running.
+
+Carrying the post-fix rates in the table above through these sizes puts the
+whole socket at **206 B/s** on average. Two thirds of that is the two messages
```
figgyster: 206 B/s doesn't come out of the table it says it's carrying through. Post-fix rates × the table's sizes: session_state 357/h × 720 = 71.4 B/s, plugins ~240 × 1,017 = 67.8, sessions 127 × 6,009 = 212.0, prs 47 × 7,037 = 91.9, tickets 1.1 × 217,391 = 66.4, plus tasks 1.6 and workspace 2.6 — ≈ 514 B/s. 206 is exactly plugins + prs + tickets (67.8 + 72 + 66.4 = 206.2), which omits session_state_changed (the doc's own headline post-fix traffic) and sessions_updated — neither of which the flap fix removes. And prs at 72 B/s needs either 36.8/h at 7,037 B or 5,515 B at 47/h; the table says 47/h and 7,037 B = 91.9 B/s. If the carrying used log-derived rates, annotate that the way tickets is annotated; otherwise the sum should come out of the table.

figgyster (changes requested):
Good loop-closer — most of it. The per-session-hour definition is the right fix, better than the one I asked for: session-attributable pushes only, so the row scales with session count, and the threshold formula (25 ÷ (29.8 ÷ 3600) = 3,020) is now explicit and checks. 66% → 60% (60.07% ✓), 15% → 12.5% ✓, citation 597 ✓, "largest recurring" ✓, and the provenance additions all verify: tickets at 217 pushes over 201 h = 1.08/h → 1.1 ✓, and the 0.3% with explicit run rates (6,944 vs 6,964 = 0.29%) ✓. Resolved my five fixed threads; the 100 B/s thread stays open because its replacement still doesn't derive.

Two figures still don't re-derive from the doc's own inputs — both in the payload this commit promises to make re-derivable. Details on the lines:

- The drop is stated as 189,255 facts (line 215), but the table's own totals and session split both imply 188,557: 260,273 − 71,716, and 232,111 − 43,554. The knock-ons: "81.5%" only works with 189,255 (189,255 ÷ 232,111 = 81.5%; 188,557 ÷ 232,111 = 81.2%), and "72% of the wire pushes" only rounds to 72 with 188,557 (72.44 → 72; 189,255 gives 72.71 → 73). Three statements, one drop — they can't all be true of the same number.
- The new whole-socket figure (line 123) is 206 B/s = plugins + prs + tickets (67.8 + 72 + 66.4 = 206.2), which leaves session_state_changed — the doc's own headline post-fix traffic — and sessions_updated out of "the whole socket" entirely. Carrying the table's rates and sizes through post-fix gives ~514 B/s (71.4 + 67.8 + 212.0 + 91.9 + 66.4 + 1.6 + 2.6). And prs at 72 B/s needs either 36.8/h at 7,037 B or 5,515 B at 47/h; the table says 47/h and 7,037 B, which is 91.9 B/s. If the carrying used log rates instead of the table's, say so the way you did for tickets.

The decision still stands — no version of these numbers moves the conclusion. But the receipt isn't re-derivable yet, and that's the one thing this doc promises. One more pass.

docs/plans/2026-08-10-projection-coalescing.md
victorarias: You found a real inconsistency and the cause is a measurement bug, not a typo. The drop was measured two ways. The push model computes gaps on integer epoch seconds and says 188,557. A separate SQL query derived them as CAST((julianday(b) - julianday(a)) * 86400 AS INT), which truncates a genuine two-second gap to one — 698 of them — and says 189,255. The model is the correct one, so: drop 188,557, share of session.state.changed 81.2% (not 81.5%), of the durable log 59.9%, of the wire pushes 72.4%. Same pass also corrected the model's ticket coalescing, which moves the totals to 260,286 and 71,729 — so 260,286 − 71,729 = 188,557 = 232,111 − 43,554, and the subtraction you invited now lands on the stated number from either direction. Both traps are written down beside the method so a re-run does not step on them.

docs/plans/2026-08-10-projection-coalescing.md
victorarias: Right again, and my reply made it worse rather than better: the number was carried through model rates while the sentence pointed at a table of live samples, so it derived from neither. It is now its own table — push counts, rates, sizes and B/s per message, summing to 209 B/s — with the everything-else row explicit so the column adds up.

On the two inputs you reconstructed: session_state_changed is 217.8/h, not 357/h (357 is the total across all messages), and sessions_updated is 0.8/h, not 127/h. That 127 is the sampled column, and it is the best illustration of why that column cannot be computed with — two messages in 57 seconds. The rate column is now labelled unusable for arithmetic with exactly that worked example. prs at 72 B/s is 36.8/h × 7,037 B, which is why it disagreed with 47/h.

One more number I owed you: the byte table counts 43,769 session_state_changed pushes against 43,554 session.state.changed facts, because seven other session facts project the same wire message. Called out on the page so the two rows are not read as the same quantity.

figgyster (approved):
The two figures that would not divide out now divide out — and the commit found root causes, not just replacement numbers. That's the better outcome.

The drop is 188,557, exactly 260,286 − 71,729 and exactly 232,111 − 43,554, both of which I re-derived. The 698-fact discrepancy I chased was real: the model script's julianday truncation collapsed 698 genuine 2-second gaps into the 1-second flap bucket — 189,255 − 188,557 = 698, exactly. The knock-ons all check: 81.2% (188,557 ÷ 232,111 = 81.24), 72.4% (188,557 ÷ 260,286 = 72.44), and the +13 on both totals is the ticket.* undercount now applied, which the traps section says out loud.

The byte budget is the right replacement for carrying the observed rates. I ran every row: 43,769 ÷ 201 = 217.8/h × 720 = 43.6 B/s; prs 7,404 ÷ 201 = 36.8/h × 7,037 = 72.0 B/s — there's the 72, and it's the log rate, not the table's. Rows sum to exactly 71,729 pushes and 209.1 B/s, and the 43,769-vs-43,554 = 215 fact/push split is documented instead of hidden. The rates-trap warning earns the commit by itself: two messages in 57 seconds read as 127/h while the log says 0.8/h — that column was a trap and now it's labeled as one.

One rounding nit, for the merge rather than another cycle: "59.9% of the durable bus log" — 188,557 ÷ 315,073 = 59.845%, which rounds to 59.8. (And the per-session-hour 158.8 is truncation of 158.87; 158.9 if you round.) Everything else in the record now divides out of stated inputs.

Approved. The record is re-derivable, the two model traps are on the page for whoever re-runs it, and the "no" to coalescing has receipts that will outlive the question.

docs/plans/2026-08-10-projection-coalescing.md
victorarias: Took the rounding nit rather than merging past it — in this document of all documents. Fixed by stating the input precisely instead of re-rounding the output: the window is 1,461.4 session-hours, so 232,111 ÷ 1,461.4 = 158.83 → 158.8 and 43,554 ÷ 1,461.4 = 29.80, both now dividing out of what is on the page. The threshold is unchanged at 3,020. Durable-log share is 59.8%.

figgyster (approved):
Both nits closed with one commit, and the session-hours precision is the better fix of the two — 232,111 ÷ 1,461.4 = 158.83, so 158.8 is the correct rounding, not truncation; the not-total-division comparison (178.1 → 49.1) still holds, and 188,557 ÷ 315,073 = 59.8%. Every figure in the record now divides out of stated inputs and rounds honestly. diff-check clean, changelog gate green. Re-approved — this one's ready to merge.

PR #837, opened 2026-08-10
feat(agent): let one agent list and peek at sessions without touching them

victorarias:
Agents living inside attn cannot see each other. A session knows its own id and
nothing else — not who else is running, not what they are doing. This is the
read-only half of the converse-and-observe vertical from
`docs/plans/2026-08-08-converse-and-observe.md`: the address book, and a way to
look without touching.

## What this adds

**`attn agent list`** — the address book. Every session on the daemon, with an
8-char short id, its name, workspace, state, and whether a turn is owed. Session
ids are exposed nowhere else, so without this there is no way for one agent to
name another. `--json` carries the full ids for anything that needs them.

```
ID        NAME                 AGENT   WORKSPACE        STATE     TURN
8010386f  peekwalk-a           shell   peekwalk-a       idle
97ffa61e  peekwalk-b           claude  peekwalk-b       working   owed
```

**`attn agent peek <id>`** — one session's state, todos, last assistant message,
and rendered screen. The id may be the short id, any unique prefix, or the full
id; an ambiguous prefix is refused by name rather than guessed at.

**Peek is passive by construction.** It assembles the answer from what the
daemon already holds — the store snapshot, the transcript file on disk, and the
worker's already-parsed terminal. Nothing in this path writes to the target's
PTY, generates input, or wakes its agent. Being watched costs the observed
session nothing: no tokens, no interruption, no trace in its transcript.

Everything degrades rather than fails. A backend with no snapshot provider, an
older worker, or a session that has not rendered a frame yet yields an absent
screen, not an error. Same for a session with no transcript or no todos.

## Shape

- `agent_peek` / `agent_peek_result` on the protocol (`ProtocolVersion` 225 → 226,
  lockstep across `main.tsp`, `constants.go`, and `useDaemonSocket.ts`).
- `internal/daemon/agent_peek.go` — the whole daemon side. Resolution by id or
  unique prefix, decoration through `sessionForBroadcast`, last message through
  the existing `inspectableTranscriptPath` identity ladder, screen through the
  optional `ptybackend.SnapshotProvider`.
- `cmd/attn/agent.go` — the CLI surface for both subcommands, plus `--json`.
- `client.AgentPeek` for the socket call.
- The message cap reuses the annotatable-window receipt: the largest prose block
  across 120 transcripts measured 18,713 chars, so 64KiB is a tripwire only a
  runaway transcript touches.

Both subcommands appear in the top-level `writeHelp`, so bare `attn` names
the address book. The plan filed that under slice 2's discoverability item
alongside the embedded-skill reference; the help half lands here because the
verbs it names exist here, and the skill reference still waits for `msg` so it
can advertise the pair.

`agent_peek` is scoped to a session and answered where it lands, like every
other command that arrives from an agent process over the unix socket — recorded
with that reason in `sessionCommandsAnsweredWhereTheyLand`.

## Verification

Automated: `go build ./...`, `go vet`, the full `go test ./...`, `pnpm test`
(2714 passed / 15 skipped), `pnpm exec tsc --noEmit`, `changelog-check`, and
`make build-linux-amd64` producing a verified ELF x86-64 binary — `cmd/attn` and
`internal/**` still cross-compile for the Linux remotes.

Live, on a throwaway `peekwalk` profile (never `dev`, never production), built
and installed from this branch:

- Bundled preflight passed every check, including `protocol.cli_app` and
  `protocol.app_daemon`. Run without the profile selected it correctly *failed*
  against the production app's older protocol — proof the profile scoping was
  real and the evidence came from this build.
- Two real sessions in separate workspaces: a shell and a Claude session.
- `attn agent list` rendered the table both from a plain shell and from inside a
  session's own pane, showing the short ids, agents, workspaces, states, and
  `owed` in the TURN column.
- `attn agent peek <b>` mid-turn returned `state: working (background_work)`,
  `turn: owed to this session`, the last assistant message
  (*"Waiting for the 90-second sleep to complete; I'll say done when it
  finishes."*), and the full 140x40 rendered screen. `--json` carried the same
  with `cols/rows: 140 40`.
- Error paths: an unknown id printed *no session matches "zzzz"; `attn agent
  list` names the sessions on this daemon*; a short id resolved to its full
  session.
- **Passivity proven at the source.** After repeated peeks at the Claude
  session, its transcript contained only the entries for the brief typed into it
  — no reaction of any kind to being watched.

One thing worth naming: peek's `todos` came back empty for the Claude session
even though its pane showed a task list. That is by design — the plan rules that
todos ship as the flat strings the hook already stores, and Claude Code's
in-pane list is not that payload. The populated-todos path is covered by unit
test rather than by that live walk.

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -230,7 +230,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '225';
+export const PROTOCOL_VERSION = '226';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 // Identifies this app process to the daemon across its own reconnects, so a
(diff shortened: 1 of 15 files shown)


figgyster (approved):
APPROVED. The read-only half of converse-and-observe, and it's the good kind of slice: it leans on machinery the daemon already had instead of inventing new, and the passivity is structural, not promised.

Checked at head 6723249a:

- Protocol 224→225 is lockstep across all five carriers — constants.go, generated.go, main.tsp, useDaemonSocket.ts, generated.ts — regeneration is faithful (Generated types + Changes green), and the daemon/frontend version-match test passes. The archived native adapter is gone from the tree, so there's no Rust version left to drift.
- Peek's passivity holds under the hood: the handler reads the store snapshot, the transcript file read-only, and the pty snapshot RPC. Nothing in the path writes to the target's PTY or wakes its agent. The 64KiB cap reuses the annotatable-window receipt, and the 3s snapshot timeout matches the model-capture budget.
- `agent list` rides the existing List RPC — the address book adds zero protocol surface.
- Prefix resolution is honest: exact match, then unique-prefix scan, and an ambiguous prefix is refused by name with the remedy printed. The error strings name the target and the fix — the right shape for a reader that must fix its own command line.
- Degrade-not-fail is everywhere it belongs: no snapshot provider, no frame yet, no transcript, no todos — absent, not error. The screen path even logs and returns nil on snapshot failure.
- Ran the touched-package suites (cmd/attn, client, protocol, transcript, store, daemon agent_peek tests) — all green; diff --check clean; CI green across Daemon, Frontend, Changelog, Bench A/B, and E2E. The one daemon test failing in my box (plugin spawn) fails identically on main — pre-existing, environmental.
- The live walk is the right witness: repeated peeks left the observed session's transcript containing only what was typed into it. That's the claim that matters, and it was proven at the source.

Two non-blocking notes:

1. The commit message says "ProtocolVersion 223 -> 224"; the code and the PR body say 224 → 225. This repo has a documented history of the wrong number getting quoted back from documents — the commit message is the next document in that line. Worth a reword before this lands.
2. `attn agent` is missing from the top-level `writeHelp`. The plan doc's own follow-up says the group joins writeHelp once both subcommands exist — they exist now, so that item is due. Bare `attn` hides the address book.

The error paths got real tests, not just the happy path — that's the part that makes peek safe to hand to agents.

figgyster (approved):
Re-approving on the rebased head (0c4199a) — my earlier APPROVED sat on the pre-rebase commit, and after the force-pushes it needs to live on the one that will merge.

Rebase over the new main (#834–#839) verified:

- Protocol collision handled right. #839 took 225; this branch now carries 226, lockstep across all five carriers — constants.go, generated.go, main.tsp, useDaemonSocket.ts, generated.ts. The daemon/frontend version-match test passes at head.
- The commit message followed the number this time: "ProtocolVersion 225 -> 226", matching the code. The rebase lesson from last round applied — the message moved with the constant.
- The PR's own delta vs main is unchanged from what I approved: same 15 files, same 973 insertions, nothing new beyond the rebase resolution. diff-check clean, build clean, cmd/attn and daemon agent-peek suites green.
- CI: Changelog, Changes, Linux daemon (amd64 + arm64), and React Doctor already green on this head; the rest is still running.

Approval stands. Peek stays passive through a rebase — the observed agent never knows the branch was touched.

PR #838, opened 2026-08-10
fix(plugins): report plugin health only when it moves

victorarias:
Every connected plugin is polled for health every 15 seconds, and until now every
poll published a `plugin.health.changed` fact whether or not the answer had
moved. That fact is not cheap: the hub projection rebuilds the plugin catalog,
which re-scans both plugin directories from disk, re-sends the whole plugin list
to every open window, and the poll writes a log line — 5,760 events per plugin
per day with nothing to say. On the measured machine it was the loudest producer
on an idle daemon.

This is what the measurement behind `docs/plans/2026-08-10-projection-coalescing.md`
found: 14,919 gaps of ~14s in the bus log, all from one plugin, all identical.
The conclusion there was to fix the producer rather than build coalescing
machinery in the projection layer, and that is what this does — the same rule
AGENTS.md now states as "idle systems must be idle".

## What changed

`setHealth` reports whether the observation moved, comparing status and message
only. The check timestamp is deliberately not part of that answer: it advances
on every poll and no client renders it, so gating on it would publish forever
while nothing changes. `reportPluginHealth` is the single seam every poll verdict
goes through, and it publishes (and logs) only the ones that moved.

`last_health_at` still advances every 15s in the daemon's own state, so
`attn plugin list` still shows when a plugin was last checked. It is just not a
reason to wake the wire.

## Real signals still surface

- **A plugin whose verdict changes** — healthy → unhealthy, or unhealthy with a
  new message — publishes on the very next poll, exactly as before.
- **A crash** does not depend on the poll at all: the process dies, the
  supervisor moves the plugin's phase (`connected` → `backoff` → `starting` →
  `connected`) and each transition publishes `plugin.health.changed` through the
  supervisor's `onChange`, alongside the connect/disconnect facts. Recovery
  arrives the same way, and the first poll after a restart publishes too,
  because a fresh connection starts from "has not been checked".
- The one case this cannot see is a plugin that goes unhealthy and back between
  two polls without its process dying — and a 15-second sampler never could;
  that is unchanged.

## Receipt: idle daemon, `plugin.health.changed` over 300s

Same profile, same machine, same two bundled plugins connected (`attn-pi`,
`attn-opencode`), nothing else running, counted from the `bus_events` table:

| | facts in 300s | per plugin |
|---|---|---|
| before (`main`) | **42** | 21 (one per 15s poll) |
| after | **0** | 0 |

Extrapolated: 5,760 facts/day/plugin → 0 while a plugin's health is steady.

## Verification

- `TestDaemon_PluginHealthPublishesOnlyWhenItMoves` (synctest, pipe-backed
  plugin) drives repeated healthy polls, an unhealthy verdict repeated, a
  changed unhealthy message, and a recovery, asserting the fact count after
  each. Mutation-checked: reverting the gate fails it.
- Live on an isolated profile installed from this branch, with the app running:
  Settings → Plugins renders `attn-opencode` as DEGRADED / UNHEALTHY with its
  message and `attn-pi` as healthy, and the daemon log shows publishes only at
  the connect handshake and the first poll verdict, then silence.

https://claude.ai/code/session_01XyiRXwGmqAZaEX8Dyy4E32

### DIFF
--- a/changelog.d/fix-plugin-health-delta-gate.yaml
+++ b/changelog.d/fix-plugin-health-delta-gate.yaml
@@ -1,0 +1,14 @@
+kind: fixed
+area: plugins
+change: >
+  A connected plugin whose health is not moving no longer costs anything. The
+  15-second healthcheck now reports only when the verdict changes, instead of
+  publishing an event, rebuilding the plugin catalog off disk and writing a log
+  line on every poll.
+symptom: >
+  With attn idle and nothing to report, each connected plugin produced 5,760
+  events a day, and every one of them re-scanned both plugin directories from
+  disk and re-sent the plugin list to every open window. It was the loudest
+  thing on the wire while the machine should have been asleep. Plugin status in
+  Settings still updates the moment a plugin becomes unhealthy, crashes, or
+  recovers.
--- a/changelog.d/fix-plugin-health-delta-gate.yaml
+++ b/changelog.d/fix-plugin-health-delta-gate.yaml
@@ -1,10 +1,10 @@
 kind: fixed
 area: plugins
 change: >
-  A connected plugin whose health is not moving no longer costs anything. The
-  15-second healthcheck now reports only when the verdict changes, instead of
-  publishing an event, rebuilding the plugin catalog off disk and writing a log
-  line on every poll.
+  A connected plugin whose health is not moving no longer wakes the rest of
+  attn. The 15-second healthcheck still asks, but now reports only when the
+  verdict changes, instead of publishing an event, rebuilding the plugin catalog
+  off disk and writing a log line on every poll.
 symptom: >
   With attn idle and nothing to report, each connected plugin produced 5,760
   events a day, and every one of them re-scanned both plugin directories from
--- a/internal/daemon/plugin_rpc.go
+++ b/internal/daemon/plugin_rpc.go
@@ -370,15 +370,23 @@ func (p *pluginConnection) request(ctx context.Context, method string, params in
 	}
 }
 
-func (p *pluginConnection) setHealth(status, message string, at time.Time) {
+// setHealth records a health observation and reports whether it moved. The
+// check timestamp is not part of that answer: it advances every poll and no
+// client renders it, so gating on it would publish a fact 5,760 times a day per
+// plugin with nothing to say.
+func (p *pluginConnection) setHealth(status, message string, at time.Time) bool {
 	p.healthMu.Lock()
 	defer p.healthMu.Unlock()
-	p.healthStatus = strings.TrimSpace(status)
-	if p.healthStatus == "" {
-		p.healthStatus = "unknown"
+	next := strings.TrimSpace(status)
+	if next == "" {
+		next = "unknown"
 	}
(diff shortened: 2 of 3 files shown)


figgyster (approved):
This is the fix the whole arc was pointing at. #835 made it law, #836 measured the crime scene and said "fix the producer, don't build coalescing machinery" — and here's the producer, quiet at last. The receipt matches the discipline: 42 facts per 300s down to 0, same machine, same profile.

Checked it like evidence, not just read it:

- **The seam is real.** All three poll branches (RPC error, unhealthy verdict, healthy verdict) now route through `reportPluginHealth`; `setHealth` has exactly those three callers and no others. There is no side door.
- **Crash recovery doesn't depend on the poll.** The supervisor's `onChange` path in `plugin_runtime.go` still fires `plugin.health.changed` on every phase transition — connected → backoff → starting → connected — plus the connect/disconnect facts. Untouched, verified.
- **First verdict after restart still lands.** A fresh connection starts at `"unknown"` (`newPluginConnection`), so the first poll is always a move. Nothing gets stuck silent on boot.
- **The timestamp is honestly excluded.** `setHealth` still advances `lastHealthAt` on every poll, so the daemon-side snapshot (`ws_plugins.go`, and the same `PluginInfo` the CLI reads) still shows when a plugin was last checked. What stopped is the fan-out, not the record.
- **Nothing downstream was leaning on the cadence.** The frontend renders health from the snapshot's `health_status`/`health_message`, not from the fact stream; no consumer keys off `last_health_at` staleness; the projection only rebuilds the catalog on facts, which now fire exactly when the picture changes.
- **The witness holds up under pressure.** Ran the new test — passes. Then removed the gate myself: fails exactly where it should, "three healthy polls published 3 facts, want 1". The mutation claim in the body is true; I watched it bleed.
- **The wider suite agrees.** Full daemon package green here (skipping two process-spawn/PTY tests that fail identically on `main` in this sandbox — environmental, not yours; CI's Linux daemon amd64+arm64 passed on this exact head). Changelog gate passed, Generated types correctly skipped, diff-check clean.

One loose thread, non-blocking: the changelog headline says a steady plugin "no longer costs anything" — the 15-second RPC still round-trips; what's gone is the publish path (fact, catalog rebuild off disk, window resend, log line). Your own detail sentence in the same fragment says it exactly right, so it's a headline-only stretch. Given the standard this repo just set for itself in #836, a future reader will trip on "anything" the way we'd trip on a figure that doesn't divide out.

The flap arc reads like a proper case file now: #830 fixed it, #832 witnessed it, #835 wrote the law, #836 priced the damage, and this closes it — the loudest producer on an idle daemon, silent, with a test that bleeds if anyone brings the noise back. Good work.

figgyster (approved):
Verified at head 6ef0700c: the only delta from the head I approved is the changelog fragment (4+/4-), and it says the right thing now — the healthcheck still asks, it just doesn't wake the rest of attn. The code is byte-identical to what the full daemon suite ran green on, and CI on this head is green all the way down (Daemon, Linux daemon amd64/arm64, Changes, Changelog). The one loose thread, closed in one commit, inside a minute. Approval stands.

PR #840, opened 2026-08-10
feat(notifications): make a parked runtime critical, and make all three arrive

victorarias:
Main shipped notification severity (info | warning | critical) with an ambient
surface for critical — a sidebar strip and a red bell outside the panel. This
stamps the A4 epic's own producers and fixes a delivery gap found while reading
them.

## What changes

- **Parked plugin runtime → critical.** Nothing retries a parked plugin. The
  notification is the only word the user gets that it stopped.
- **Parked app runtime → critical.** While it is parked, no installed app's
  handlers run at all, and the way back is a person typing
  `attn app runtime restart`.
- **App auto-disabled → warning.** One app is off; everything else still runs,
  and the notification names the way back (`attn app enable <name>`).
- **Two of the three now publish `notification.created`.** The app runtime's
  parked notification and the auto-disable notification were written to the
  database and never announced, so an open app learned about them only when
  something else happened to re-push the notification snapshot. The plugin one
  already published; these two now match it.

## Why these two are critical and the third is not

Severity is about how much the notification wants the user, and the dividing
line here is whether anything else will fix it. A parked runtime is not retried
by anything: it stays broken, silently, until a person acts, and while it is
broken every app in the system is inert. An auto-disabled app is a bounded
failure with the rest of the system healthy — it is something that failed and
stays failed until the user gets to it, which is exactly what warning means.

## Tests

Assertions added to the three existing tests that already cover these
notifications, rather than new ones — the behaviour is a property of the
notification each producer already writes.

- `TestParkedPluginRaisesADurableNotification`
- `TestParkedRuntimeIsVisibleOnEveryAppAndRevivable`
- `TestAppStuckOnOneEventIsDisabledAndSaysSoThreeWays`

Both mutations are red: removing a `Severity:` line fails with
`severity = "info", want critical/warning`, and neutralizing a `publishFact`
fails with `notification.created facts = [], want one for <id>`.

## Verification

Go suite green. The live witness for the critical surface — the strip and the
red bell appearing when the app runtime parks — is the A4 exit proof (slice 6),
which parks the runtime on a real profile; it is recorded on the epic→main PR
rather than duplicated here.

Part of A4 (`docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md`),
targeting `epic/ext-a4`.

### DIFF
--- a/changelog.d/parked-runtime-notifications-are-critical.yaml
+++ b/changelog.d/parked-runtime-notifications-are-critical.yaml
@@ -1,0 +1,16 @@
+kind: changed
+area: app
+change: >
+  A parked plugin runtime and a parked app runtime now raise critical
+  notifications, so they get the ambient surface — the sidebar strip and the red
+  bell — instead of sitting quietly in the feed. An auto-disabled app raises a
+  warning. All three now reach an open app the moment they happen.
+notes: >
+  Nothing retries a parked runtime, so the notification announcing it is the last
+  word the user gets: while it is parked no app's handlers run and no plugin is
+  restarted, and the only way back is a person running `attn app runtime restart`
+  or reinstalling the plugin. That is what critical is for. An auto-disabled app
+  is one app off with everything else still running, which is a warning. The app
+  runtime's parked notification and the auto-disable notification were also
+  written to the database without publishing `notification.created`, so they only
+  reached an open app whenever something else happened to re-push the feed.
--- a/internal/daemon/app_autodisable_test.go
+++ b/internal/daemon/app_autodisable_test.go
@@ -122,6 +122,14 @@ func TestAppStuckOnOneEventIsDisabledAndSaysSoThreeWays(t *testing.T) {
 	if !strings.Contains(notes[0].Body, "ticket.created") || !strings.Contains(notes[0].Detail, "TypeError") {
 		t.Fatalf("the notification does not say what failed: body=%q detail=%q", notes[0].Body, notes[0].Detail)
 	}
+	// One app is off; everything else still runs. That is a warning, and it
+	// reaches the app now rather than at whatever re-pushes the feed next.
+	if notes[0].Severity != store.NotificationWarning {
+		t.Fatalf("severity = %q, want warning", notes[0].Severity)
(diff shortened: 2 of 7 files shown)


figgyster (approved):
Approved. Checked the whole seam, not just the diff.

**The severity ladder is coherent.** Critical for both parked producers, and it's earned: nothing retries a parked runtime, no app's handlers run while the app runtime is parked, and the only way back is a person. That matches what the store's severity contract says critical is for. Warning for an auto-disabled app is the same reasoning the task-failure path already used ("a dead job stays dead, but nothing else degrades while it waits") — one app off, everything else still runs. The frontend keys the ambient surface on critical only; warning gets the panel row. So the strip and red bell will not start crying wolf for a single disabled app.

**"Make all three arrive" is the real fix, and the commit message is honest about it.** Two of the three producers (app-runtime parked, auto-disable) wrote to the DB without publishing `notification.created`, so an open app saw them only when some unrelated fact re-pushed the feed. Verified the mechanism end to end: any `notification.*` fact routes through the bus projection to `projectNotificationsUpdated`, which broadcasts unread count, critical count, and critical title to every client — that's the strip's only input. With the publish added, the arrival is immediate, not incidental.

**Seam complete.** All four `AddNotification` callers in the daemon now publish the fact — plugin parked and task-failure were already doing it, this PR closes the last two. No other producers, no plugin-SDK write path that bypasses it.

**Witnesses showed up.** All three touched tests pass, full store package green. The daemon package has two failures — both process-spawn tests that fail identically at the base commit in this sandbox; CI's Daemon job is green on the head. Changelog and Changes gates pass, diff-check clean. The changelog note earns its place: "nothing retries a parked runtime, so the notification announcing it is the last word" is exactly the reasoning that justifies critical.

One observation, not a nit: the parked-plugin body still says "reinstall the plugin, or restart attn" — there's no `attn plugin restart` verb, unlike the app runtime's `attn app runtime restart`. The text is honest, so it's fine to ship; just noting that this PR raised the volume on that notification, and the recovery-path asymmetry between the two parked states is now more visible.

PR #841, opened 2026-08-10
docs(plans): the prose density gate — two layers, findings never scores, no Python

victorarias:
Agent-written prose drifts dense even under explicit style guidance — #818 measured comments growing from 6% to 17% of the diff before a manual sweep. This plan adds the verification half: a detector that reads what an agent wrote and reports span-anchored simplification findings back to it, in the writing loop.

Two layers, grounded in a cited state-of-the-art survey (in the doc):

- **Deterministic, pure Go**: Vale-shaped rules that survive technical prose (nominalizations, noun strings, actor-hiding passive, a user-curated reject vocabulary), a CPIDR-style idea-density port (the metric that says *overloaded* rather than *long*, and chop-resistant by construction), and paired anti-gaming tripwires — staccato-run and referent-overlap checks that fire exactly when the length rule is being gamed into fragments.
- **Model judge**: Haiku-class, one narrow criterion per call per paragraph — the granularity where small-model judging is reliable — emitting capped `{quoted span, objection, rewrite}` findings. Never scores: scores drift lenient and collapse prose when optimized against.

Ground rules the plan commits to: no Python dependency (Go ports or cuts); no threshold ships without calibration on a labeled corpus; advisory to the writing agent before it is ever a blocking gate anywhere.

Docs-only: one new plan doc plus its changelog fragment. No behavior change; live verification not applicable.

### DIFF
--- a/changelog.d/prose-density-gate-plan.yaml
+++ b/changelog.d/prose-density-gate-plan.yaml
@@ -1,0 +1,9 @@
+kind: internal
+area: docs
+change: >
+  New plan: the prose density gate — a two-layer detector (pure-Go
+  deterministic tripwires plus a criterion-scoped small-model judge) that
+  reports span-anchored simplification findings back to the agent that
+  wrote the text, grounded in a state-of-the-art survey; findings never
+  scores, thresholds calibrated on a labeled corpus, advisory before
+  blocking, no Python dependency.
--- a/docs/plans/2026-08-10-prose-density-gate.md
+++ b/docs/plans/2026-08-10-prose-density-gate.md
@@ -10,9 +10,11 @@ in the writing loop, what to simplify and where.
 The deliverable is always a **finding** — a quoted span, the objection,
 and when possible a suggested rewrite. Never a score: an agent can fix
 "this sentence does three jobs, split it"; it cannot fix "readability 32",
-and score-shaped targets are documented to collapse prose into fragments
-([RL against a readability score](https://arxiv.org/pdf/2410.17088);
-Hemingway is the interactive version of the same failure).
+and optimizing against score-shaped readability targets is documented to
+degrade the prose itself — RL against a readability formula collapsed
+outputs into degenerate one-sentence compressions
+([receipt](https://arxiv.org/pdf/2410.17088)), and Hemingway-style length
+rules are the interactive push toward staccato.
 
 ## Shape
 
@@ -66,8 +68,13 @@ from the research gets a Go port; anything needing a real parser is cut.
 ### Model judge
 
 One narrow question per call, per paragraph — the granularity at which
-Haiku-class judges hold up (cross-family agreement κ≈0.72 on
-criterion-scoped judging; reliability collapses on "grade this document").
+Haiku-class judges hold up: cross-family agreement between small judges
+reaches κ=0.716 on structured, criterion-scoped judging and degrades as
+criterion complexity rises
+([receipt](https://arxiv.org/pdf/2606.16890)); small models judge
+adequately on narrow rubrics but not on open-ended quality
+([SLMJury](https://arxiv.org/html/2606.07810)). Slice 2 re-measures this
+on our own corpus before trusting it.
 The reconcile classifier is the in-repo precedent: tool-less, cheap,
 deterministic harness around a small model.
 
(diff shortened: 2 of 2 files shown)


figgyster (approved):
Docs-only plan, and I read the whole file before signing. Receipt check first, since the doc's thesis is receipts:

- #818's 6%→17%: the merged PR's own body says "tripled since May (6% → 17% of lines)". Real.
- Trott & Rivière 2024: abstract reports r=0.76/0.74 zero-shot against human judgments, beating formulas and psycholinguistic indices. "r≈0.75" is a fair rounding.
- The CPIDR DOI resolves to the actual idea-density paper (BRM 40(2), 2008), and the chop-resistant argument survives contact: fragmenting moves words, deletes no propositions.
- Garden-plan link resolves, `attn prose` is a free namespace, the changelog fragment matches sibling shape, no whitespace dirt in the diff.

The design is right where it matters. Findings over scores is the correct feedback contract for an optimizing agent: "split this sentence" is actionable, "readability 32" is a dare. CPIDR was the right metric to pick because it survives the obvious evasion. The anti-gaming pair means the adversary got thought about before the tripwire was needed, which is the whole ballgame for a gate aimed at agents. Advisory-first is the right sequencing, and "a legitimate sentence hitting the gate means the threshold is wrong; remeasure" is the correct discipline. Also, the doc practices what it preaches. A prose-density plan that reads this well is the strongest argument for the gate.

**The loop-back is the under-specified half.** The detector gets rules, metrics, tripwires, calibration. The delivery gets "boundary moments, advisory" through embedded-skill guidance, which is the same guidance channel #818 measured drifting. Sequencing slice 3 after the garden exists is honest, but the gate's activation then leans on the exact behavior the plan exists to correct. If the answer is "the garden makes the wiring cheap," write that down; if it's "agents will run it because it's cheap and the findings are actionable," say that too. Right now the doc's strongest section ends where its riskiest question begins.

**κ≈0.72 is the one number without a receipt.** Every other claim in this doc carries a citation or points at an artifact. "Cross-family agreement κ≈0.72 on criterion-scoped judging" appears bare and asserted, and it's load-bearing: it justifies the one-criterion-per-call design. If it came out of the reconcile classifier's eval, point at it; if it's literature, cite it; if it's a working assumption, label it, because slice 2's spot-check can't re-measure a number the plan already presents as fact. Same disease, milder: the RL citation documents readability scores as weak rewards, while the collapse-into-fragments outcome is the plan's inference from the literature family, not the abstract. Directionally fair, one clause ahead of its source.

figgyster (changes requested):
Second lap on the receipts, and credit where it's due first: the loop-back fix is exactly the answer I asked for, and better than the ones I suggested. Wiring activation into the ticketing surface attn already owns — `attn delegate` and `attn ticket new` run the deterministic layer on briefs and descriptions, advisory in the result — makes the gate mechanical where the prose actually flows, and reserving guidance for surfaces attn never touches turns my question into a design rule. Verified both verbs exist on main. That half is closed, properly.

But the κ fix traded a bare number for a receipt that doesn't exist, and the RL rewording still runs ahead of its source. Both are in the changed region, both are load-bearing, and both check in two minutes:

1. **arXiv 2606.16890 is not a judging paper.** It's "Compositional Reasoning Depth Predicts Clinical AI Failure" — EHR question-answering, hop-count taxonomy, accuracy decline. No κ, no inter-rater agreement, no criterion-scoped judging anywhere in it. The only 0.72 in the abstract is an odds ratio per hop ("OR per hop 0.72"), which looks like where the number came from: proximity, not source. κ=0.716 still has no valid receipt. The paragraph already carries the honest fallback ("Slice 2 re-measures this on our own corpus before trusting it") — either cite the real source, or drop the broken receipt and let the re-measurement language carry the number as a labeled working assumption.

2. **"Degenerate one-sentence compressions" is not in 2410.17088.** I read the full HTML. The paper documents the failure differently: "disproportionate weights for sentence-level rewards contribute to unstable training and lower simplification quality," plus truncation of redundant content in the balanced variant. That's a real, quotable finding and it's on-theme — sentence-level readability rewards degrade output. Reword to what the source shows and the sentence gets stronger, not weaker.

SLMJury (2606.07810) checks out — the closed-ended vs open-ended capability split is exactly what its abstract reports. That receipt is good.

The loop-back design rule stays as is. Fix the two receipts and I'll re-approve on the commit.

figgyster (approved):
Loop closed. Verified both fixes on the new head against their sources:

- The RL sentence now says what 2410.17088 actually shows — "disproportionate weights for sentence-level rewards contribute to unstable training and lower simplification quality" is the paper's own finding (present verbatim in the full text).
- The κ=0.716 figure and its receipt are gone. SLMJury stays as the cited ground, the narrow-vs-open-ended split is a fair gloss of its abstract, and the sufficiency claim is now honestly labeled: working assumption, measured by slice 2, per criterion. That is exactly the fallback I offered, and dropping the number beats relocating it.

The loop-back design rule and the ticketing wiring carry over untouched. Docs-only, changelog fragment in place, CI green. Approving on the new head.

PR #842, opened 2026-08-10
fix(apps): a parked runtime stays parked through delivery

victorarias:
Found by the A4 exit proof: parking the app runtime did not hold.

`awaitAppRuntime` — the path every single event delivery takes — called the same
`supervise.Ensure` that `attn app runtime restart` calls, and `Ensure` un-parks.
The bus retries a failing delivery forever, so a runtime attn had given up on was
handed a fresh restart budget every couple of minutes, burned it on ten more
crashes, and parked again. Parked was unreachable in practice, and each parking
on the way wrote its own notification — which, now that those are `critical`,
would have become a permanent ambient alarm.

Measured on the exit-proof profile with a host binary that exits immediately,
before this change: seven parkings and seven critical notifications in fourteen
minutes, one every ~2m10s. Each delivery also burned the full 10s connect
timeout while the doomed child was starting.

## The change

`supervise` grows `EnsureUnlessParked`, for a caller that runs per unit of
traffic rather than per deliberate act: it starts a stopped child, and answers
`ErrParked` for one the supervisor has given up on rather than reviving it.
`Ensure` is unchanged, so the deliberate ways in — `attn app runtime restart`,
daemon startup — still revive.

Dispatch uses it and turns `ErrParked` into a runtime failure that names the
state and the way back. It stays a runtime failure, not an app failure: the app
is not charged for the outage on its auto-disable clock.

## Live verification

Throwaway profile `a4exit`, daemon installed from this branch, real app
(`ticketwatch`) subscribed to `ticket.*`, host binary swapped for one that exits
3.

Parked once, at generation 13, after the supervisor's ten attempts. Then eight
minutes of traffic — bus retries plus three new tickets:

```
00:24:50 parked notifications since restart: 1 | state: parked, generation 13
...
00:31:51 parked notifications since restart: 1 | state: parked, generation 13
```

No restart, no second notification, where the pre-fix rate was one every ~2m10s.
The retries themselves now fail in 0ms instead of waiting out the connect
timeout, and say why:

```
STARTED               SEQ  EVENT           STATUS         MS  ERROR
2026-08-10T22:31:51Z  603  ticket.created  runtime_error  0   the app runtime is parked after repeated
                                                              crashes and is not being restarted; `attn app
                                                              runtime status` shows why it exited and `attn
                                                              app runtime restart` tries again
```

Restoring the host and running `attn app runtime restart` revived it
("revived the app runtime: it was parked after crash-looping"), and the app
drained its backlog to `cursor 679, 0 event(s) behind`.

## Tests

- `internal/supervise`: `TestEnsureUnlessParkedLeavesAParkedChildAlone` — parks a
  child, calls `EnsureUnlessParked` five times for `ErrParked` with no new start
  and one give-up announcement, then proves `Ensure` still revives.
- `internal/daemon`: `TestDispatchLeavesAParkedRuntimeParked` — three dispatches
  into a parked runtime leave the generation where it was, write one
  `app_runtime_parked` notification, record three `runtime_error` invocations,
  keep the app off the auto-disable clock, and return an error naming
  `attn app runtime restart`.

Both go red on the one-word mutation (`startAppRuntime(false)` → `true`): the
daemon test on the error text first, and on the generation assertion alone once
that one is removed.

### DIFF
--- a/changelog.d/parked-app-runtime-stays-parked.yaml
+++ b/changelog.d/parked-app-runtime-stays-parked.yaml
@@ -1,0 +1,13 @@
+kind: fixed
+area: app
+change: >
+  A parked app runtime now stays parked. Delivering an event to an app no longer
+  restarts a runtime that attn has given up on: the delivery fails with an error
+  naming the state and `attn app runtime restart`, which is still the way back.
+notes: >
+  Every delivery attempt went through the same call that `attn app runtime
+  restart` uses, and the bus retries a failing delivery forever, so a broken
+  runtime was handed a fresh restart budget every couple of minutes and parking
+  never held. Measured on a host binary that exits immediately: three parkings
+  and three critical notifications in five and a half minutes, which the ambient
+  surface would have turned into a permanent alarm.
--- a/internal/daemon/app_consumers.go
+++ b/internal/daemon/app_consumers.go
@@ -14,6 +14,7 @@ import (
 	"github.com/victorarias/attn/internal/docstore"
 	"github.com/victorarias/attn/internal/jobs"
 	"github.com/victorarias/attn/internal/store"
+	"github.com/victorarias/attn/internal/supervise"
 )
 
 // Apps as bus consumers: registration, delivery, the invocation log, and the
@@ -472,7 +473,11 @@ func (d *Daemon) awaitAppRuntime(ctx context.Context) (*appRuntimeConnection, er
 	if runtime != nil {
 		return runtime, nil
 	}
-	if err := d.ensureAppRuntime(); err != nil {
+	if err := d.startAppRuntimeForDispatch(); err != nil {
+		if errors.Is(err, supervise.ErrParked) {
+			return nil, runtimeFailure(
+				"the app runtime is parked after repeated crashes and is not being restarted; `attn app runtime status` shows why it exited and `attn app runtime restart` tries again")
+		}
 		return nil, runtimeFailure("%v", err)
 	}
 	wait := d.appConnectWait()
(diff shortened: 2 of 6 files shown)


figgyster (approved):
Good catch, and it's the kind that only shows up when the exit proof actually runs. #840 made the parked runtime critical, and the bus retries forever, so `Ensure` on the dispatch path was handing a dead host a fresh ten-attempt budget every couple of minutes: parked was a state the runtime passed through, not rested in, and each pass wrote another critical notification. The ambient alarm this would have become is the whole reason #840 raised the volume, so this is the fix that makes that severity honest.

Checked the seam end to end:

- The split is minimal and the deliberate ways in are untouched: `Ensure` still revives (only the restart RPC calls it), `EnsureUnlessParked` answers `ErrParked` unwrapped so `errors.Is` holds, and a never-started child still lazy-starts through the same call. No state where a child has neither process nor timer except stopped and parked, and both behave.
- `ErrParked` maps to a runtime failure that names the state and the one command out of it, and the stall clock is cleared, not merely not-advanced, so the app isn't charged for an outage that isn't its fault. The test pins all of it: error text, generation stability across three dispatches, exactly one notification, no stall, and restart still revives.
- Mutation claim holds. Flipped `EnsureUnlessParked` back to `Ensure` locally and the supervise test goes red on the exact `ErrParked` assertion. The daemon half I can't run here - this sandbox's app-store reads come back empty on base too, so it's environmental - but CI's Daemon check is green, and the generation assertion is exactly the tripwire that mutation would cross.
- Full `internal/supervise` package green locally; CI Daemon, react-doctor, Changes, Changelog all green; only Bench A/B still turning.
- The two pre-fix numbers in the body and the changelog (seven in fourteen minutes, three in five and a half) looked like a discrepancy until I did the division: same ~2m10s period, different windows. They corroborate. No nit there.

One observation, not a blocker: "daemon startup still revives" is really "a fresh supervisor has no memory of the park." There's no boot-time Ensure; a daemon restart forgets the park and the first dispatch lazy-starts the host again, re-arming the crash loop and eventually writing a second critical notification while the host is still broken. Right behavior for a deliberate act, but worth knowing that a daemon restart, not just the restart verb, resets the clock on a broken host.

Approve. The exit proof earned its keep.

PR #843, opened 2026-08-10
feat(apps): stage A4 — the app registry and the shared runtime

victorarias:
Stage A4 of the extension platform: attn grows **apps** — automation a user (or
their agent) writes, applies live, watches run, breaks, and rolls back, without
touching attn's source or restarting anything.

Plan: `docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md`.

## What an app is, and why it is not a plugin

A **plugin** integrates the outside world into attn — an agent driver, a
worktree provider. Its own supervised process, installed rarely, effectively
part of the platform. A device driver.

An **app** is automation built on attn's platform: it consumes domain facts from
the durable bus, keeps state in its own document-store namespace, and runs
inside one shared supervised runtime with content-addressed versions, an
invocation log, a kill switch, and auto-disable. Written by agents, applied
live, cheap and numerous.

A failing plugin takes an integration down. A failing app is auto-disabled while
everything else keeps running. Different trust, rate, and blast radius, so
different mechanism, on purpose.

An app is a directory:

```toml
name = "ticketwatch"
attn_app_api = 1
entrypoint = "src/index.ts"

[[subscribe]]
events = ["ticket.*"]

[[collections]]
name = "tickets"
fields = ["status"]
```

`attn app apply <path>` parses the manifest, regenerates the `Handlers` type the
entrypoint must satisfy, typechecks, bundles, hashes, and records the result as
an immutable version. **The apply pipeline never evaluates app code**: a broken
app fails before the version flips, and the version you were running is still
the version you are running.

## The six slices

| | |
|---|---|
| #794 | `internal/supervise` — the process supervisor extracted from the plugin runtime: restart backoff, generation fencing, stability window, disconnect grace, give-up parking, per-child log capture. Two consumers now. |
| #793 | Durable bus consumers that register and unregister while the daemon runs — an app installed today starts at head, and an app removed today stops holding the retention floor down. |
| #797 | The registry tables and the lifecycle CLI: `attn app list/status/enable/disable/remove`, the `app:<name>` consumer and `app/<name>` namespace both derived from the one name. |
| #817 | The apply pipeline, scaffold and codegen: `attn app new`, `apply`, `rollback`, `dev`. |
| #826 | The runtime sidecar: dispatch, the invocation log, the 15-minute auto-disable clock, and `attn app runtime status/restart`. |
| #840, #842 | Severity on the notifications the runtime raises, and the parked-runtime fix the exit proof found. |

## As-built decisions worth reviewing

**One shared runtime, started lazily.** Not one process per app: a daemon with
no apps due a fact never starts it, and the first delivery pays the cold start
(41ms measured here, 77ms in slice 5). The sidecar is a compiled bun binary
shipped in the bundle, cross-compiled for Linux for remote daemons.

**Stall, never skip.** A failing delivery is retried with the bus's backoff and
the cursor does not move. That is what makes an app trustworthy and also what
makes it dangerous — a permanently broken app would pin the log's retention
floor for every consumer — which is why the auto-disable clock exists: fifteen
minutes stuck on the same event and attn disables the app, publishes the fact,
and writes a notification naming `attn app enable <name>`.

**Two failure classes, told apart.** A handler that throws is the app's fault
and moves its clock. A runtime that dies is not, and *clears* it — an app must
not be charged for the minutes the sidecar was down.

**Content-addressed versions.** Applying the same source twice records no new
row. Rollback moves a pointer and builds nothing; every version stays on disk.

**The migration renumber.** The registry migration was numbered 96, then 97,
then 98, then 101, and is now **102** — each main sync found the number taken.
This is not cosmetic: the runner keeps one scalar schema version and skips
anything at or below it, so a database that already ran main's 99 and 100 would
never have created these tables from 98. Proved both ways in the exit proof.

## The exit proof

The roadmap's exit for A4 is the whole loop run live, not a test suite. Run on a
throwaway profile installed from this branch, plus the OrbStack aarch64 VM.
Receipts are in the plan doc under "Slice 6 receipts"; the short version:

- **Write and apply.** Scaffolded, applied as version 1 (558 bytes), consumer
  registered **at head** — an app installed today is not handed a backlog — and
  the runtime not started yet.
- **Invocations.** The first `attn ticket new` started the sidecar and ran the
  handler: cold 41ms, warm 0-1ms, the document written, `attn app logs` carrying
  the handler's own output.
- **Break it.** A version that dereferences nothing stalled the consumer on one
  event; `app status` named the event, the attempt count, the stack, and the
  instant it would be disabled. The clock was waited out in **real time** — the
  injectable clock is test-only. It disabled itself at 16m0s across 15 attempts
  (the window is measured at the next delivery, and backoff had stretched to a
  minute), published `app.enabled.changed`, and wrote a warning notification
  naming the way back.
- **Recover and roll back.** Fixed, applied as version 3, `attn app enable` with
  the stall clock cleared, backlog drained rather than skipped. Rollback to an
  explicit version moved the pointer both directions.
- **Park the runtime.** The host binary was replaced with one that exits 3. Ten
  restarts, then parked: `attn app runtime status` says why and how to leave,
  every app's status says PARKED in the app's own words, the notification is
  critical so the sidebar bell goes red, and `attn app runtime restart` revived
  it and the app drained its backlog.
- **Linux.** The daemon and the sidecar cross-compiled for linux/arm64 and
  dispatched on the VM, `Linux aarch64`, document written, cursor at head. The
  missing-sidecar error was witnessed there too, and it names the binary, where
  it looked, how to build it and how to override it.

### What the proof found

**Parking did not hold (#842).** Dispatch — the path every delivery takes —
called the same `supervise.Ensure` that `attn app runtime restart` calls, and
`Ensure` un-parks. Since the bus retries forever, a runtime attn had given up on
got a fresh ten-attempt budget every couple of minutes: **seven parkings and
seven critical notifications in fourteen minutes**, one every ~2m10s. Fixed with
`EnsureUnlessParked`, which leaves a parked child alone and answers `ErrParked`;
dispatch turns that into a runtime failure naming the state and the command out
of it. Re-verified live: one parking, then eight minutes of traffic with the
generation unchanged and no second notification.

**Two notification producers never published `notification.created` (#840).**
The app-runtime-parked and auto-disable notifications were written to the
database without publishing the fact, so they only reached an open app whenever
something else happened to re-push the feed. Fixed in the same PR that gave a
parked runtime the `critical` severity it needs to earn the ambient surface.

## Open items — recorded, not fixed

Each is a decision for Victor rather than an oversight:

1. **A hub-managed remote endpoint never gets a sidecar.**
   `internal/hub/bootstrap.go:435` (`installRemoteBinary`) streams exactly one
   file — the `attn` binary — and there is no reference to `attn-app-runtime`
   anywhere under `internal/hub/`. A remote daemon installed by the hub hits
   "the app runtime binary attn-app-runtime is not installed" forever. Shipping
   the second file is small, but it is new behavior in the bootstrap path, which
   is why it is not in this epic. **Epic blocker or follow-up is Victor's
   call.**
2. **Bare `attn app rollback <name>` picks the numerically previous version id**,
   not the previously-running one. With history 1 good / 2 broken / 3 fixed it
   rolls to the broken one. Naming the version explicitly is always correct.
   "The previous version" has two honest readings; picking one is a product
   decision.
3. **No parked-only retention clock.** While the runtime is down, deliveries
   record `runtime_error` and clear the app's stall clock, so nothing disables
   the app and its consumer holds the retention floor for as long as the outage
   lasts. `attn bus status` names it — `app:ticketwatch (retention floor)` with
   an ERROR line saying how far behind and for how long — so it is visible, not
   silent. Whether the floor should also have a limit is the question A4 leaves
   open; the roadmap gate did not ask for one.
4. **A restarted daemon forgets a park.** A fresh supervisor has no memory of
   it, so the first dispatch lazy-starts a still-broken host and the crash loop
   re-arms. Right behavior for a deliberate act; worth knowing that it is a
   second way back, not just the restart verb. (figgyster's observation on
   #842.)

## Surfaces

- **CLI.** `attn app new/apply/dev/list/status/logs/enable/disable/remove/rollback`,
  `attn app runtime status/restart`, `attn bus status/disable/enable`.
- **Daemon and app.** Registry state, invocation log, runtime phase and the
  notifications all broadcast through `wireProjections`.
- **Protocol.** 228, both lockstep spots, `generated.go`/`generated.ts`
  regenerated.
- **Linux.** Daemon and sidecar cross-compile and dispatch on aarch64, witnessed
  on the VM.
- **Docs.** `docs/glossary.md` carries app vs plugin; the plan doc carries the
  gate answers and the slice 5/6 receipts; every PR shipped a changelog
  fragment.

Verified live on a profile installed from this branch after each main sync. On
the latest one: preflight PASS at protocol 228 on both lockstep checks, and the
app registry intact across the merge — that profile's database went to 102 and
still lists its app. On the sync before it, a fresh `attn ticket new` ran the
handler in 27ms.


figgyster (approved):
**Approve.** The staging merge delivers what the plan promised: main takes one fully-CI'd, fully-reviewed merge, and the whole diff is what I stood on this time, not the slices.

What I verified on the assembled tree:

- Build green, `go test` green across supervise, apps, appbuild, protocol, store, and the daemon's app suite. One daemon failure — `TestDaemon_StartInstalledPlugins_SpawnsProviderPlugin` — fails identically on main's HEAD in this sandbox (a process-spawn environment flake, not a regression; CI Daemon is green on both runs).
- Protocol 227 holds in both live lockstep spots (`constants.go` and `useDaemonSocket.ts`), the version-match test passes, `tsc --noEmit` clean over the regenerated types. The native client is archived, so "both lockstep spots" is the honest count.
- Migration 101 sits above 100 in the ladder, and the runner's at-or-below skip semantics are documented in the code, not just the plan. The "proved both ways" exit-proof claim matches what the migration runner actually does.
- CI green wall-to-wall on both runs: Daemon, Frontend, E2E 1–3, Generated types, Tauri, Linux daemon amd64 + arm64, Plugins, Bench A/B, react-doctor 80/100 with zero errors. Merge state CLEAN.

The exit proof survives a suspicious read. I re-checked each recorded open item against the code, not the prose: the hub bootstrap does stream exactly one file with no `attn-app-runtime` anywhere under `internal/hub/`; bare rollback does pick the numerically previous version (the code comment shows "newest that isn't current" was considered and refused, because it would roll *forward*); runtime failures do clear the stall clock so a down sidecar never charges an app; and daemon-restart-forgets-park is the observation I made on #842, promoted from review comment to named open item. That's the loop working: a hole in a slice PR becomes a decision with an owner in the staging PR.

Three observations, none blocking:

1. **The hub-sidecar gap is the one with teeth.** A hub-installed remote daemon hits "the app runtime binary is not installed" forever, with nothing telling the user the fix lives on the hub side. Deferring it is the right call, but the follow-up should make the error say so, or someone spends an evening rebuilding a sidecar a remote can never have.
2. **Restart-revives-park has a volume knob now.** #840 made each parking critical, and #842 kept a parked runtime parked through dispatch. But a daemon restart re-arms the crash loop, and each cycle writes another critical notification. Recorded as a decision, and it's the right one — worth keeping on the follow-up list, since it's a question of how loud the alarm is, not whether it fires.
3. **The head commit is a docs fix, and it's right.** "Four things observed," not three — the list has four bullets. Counting the observations right, on the head commit, after the proof ran. That's the receipts culture with the receipts on display.

The arc reads coherent end to end: supervise extraction → runtime bus lifecycle → registry → apply pipeline → sidecar → severity and parking fixes, staged onto main in one merge, with the whole loop proven live on a throwaway profile and witnessed on the Linux VM. The extension platform has a real automation surface now. Merge it.

figgyster (approved):
Re-approving on the merged head (880b714d) — signal belongs on the commit that merges, and this is the one.

What the sync was: main's #844/#845/#847 riding in, plus the renumber. A4's own tree is untouched — the only A4-file deltas are the migration comment (101→102) and the protocol bump (227→228). Verified on the assembled head: build green; supervise/apps/appbuild/store/protocol and the daemon app suite green; the protocol lockstep test green; the migration ladder monotonic 1–97, 99, 100, 101 (agent messages), 102 (app registry), with 98 still burned; generated types regenerated from the merged tsp (Generated types passes on both runs). The renumber commit tells the history straight: 96→97→98→101→102, each main sync finding the number taken.

CI on the new head: one run is fully green — every job, all ten, on this exact SHA. The second run (same commit, pushed twice) collected three failures, and I ran each one down:

1. **Daemon** — `TestAgentMsgSizeRefusalsSurviveTheSocket`: fails ~1 in 5 on this tree, survived 20 straight runs on main and failed inside 40. It's a race between #844's socket test and #845's startup recovery reaping no-evidence sessions: the test adds its sessions after the socket appears, while the recovery goroutine is still deciding who gets reaped. Pre-existing on main, not created by this merge — this tree just fires it more often. It will keep flaking main's CI too; the fix is the test waiting for recovery to settle before it adds sessions.
2. **Frontend** — `PresentRoot` "j leaving a skipped file": passes 53/53 locally and passed in the green run. A presentation-tour UI test nothing in this diff goes near.
3. **Frontend E2E (2/3)** — the two CI runs on the same commit collided on profile-default WebSocket port 19849. Infra, not code: the other run's E2E 2/3 passed.

None of the three implicates the epic. Same verdict as before, now sitting on the SHA that will actually merge.

PR #845, opened 2026-08-11
fix(sessions): recover resumable sessions after machine crashes

victorarias:
## Intent / Why

A machine crash could delete sessions that attn still had enough durable state to reopen. Startup recovery made that decision from a static agent policy and skipped some activity states before examining resumability, so survival depended on the agent name and what the session happened to be doing when the machine went down.

This change gives startup recovery one invariant: a missing-runtime session survives only when it was not intentionally closed, has a persisted launch intent, and has a restoration target that the revive path actually consumes. That target may be an agent-native resume ID, conversation-host history or its intent seed, or a persisted per-session plugin handle. Activity state remains user context; it no longer decides whether the session is kept or reaped.

## Summary

- Route both startup pruning and worker reconciliation through the same durable-resumability decision.
- Remove the legacy SessionRecoveryPolicyProvider and Claude override, the daemon recovery-policy helper, the idle/recoverable reconciliation bypass, and the plugin capability-only fallback.
- Cover the invariant with evidence-driven daemon tests and an isolated real-app machine-crash recovery scenario, and document recoverability in the glossary and changelog.
- Register the socket size-limit test synthetic sessions after daemon startup. Those fixtures intentionally have no launch intent, so the new invariant correctly reaps them during startup; creating them afterward keeps that unrelated test focused on socket behavior.

## Verification

- `make test` — passed.
- `pnpm --dir app test` — 2,709 frontend tests passed.
- `pnpm --dir app exec playwright test e2e/session-restore-reconnect.spec.ts` — focused reconnect E2E passed.
- `"$profile_app/Contents/MacOS/attn" preflight` — bundled preflight passed.
- `pnpm --dir app run real-app:scenario-crash-recovery` — isolated real machine-crash recovery scenario passed.

### DIFF
--- a/app/e2e/session-restore-reconnect.spec.ts
+++ b/app/e2e/session-restore-reconnect.spec.ts
@@ -54,7 +54,7 @@ test.describe('Session restore and reconnect harness', () => {
     await expect(page.locator('[data-testid="sidebar-session-restore-s2"][data-state="waiting_input"]')).not.toHaveClass(/selected/);
   });
 
-  test('reconnects after daemon restart and marks stale Claude sessions recoverable when worker is missing', async ({ page, daemon }) => {
+  test('reconnects after daemon restart and drops a stale session it cannot bring back', async ({ page, daemon }) => {
     await daemon.start();
 
     await daemon.injectSession({
@@ -75,11 +75,10 @@ test.describe('Session restore and reconnect harness', () => {
 
     await daemon.restart();
 
-    // Session remains tracked and is marked recoverable (claude sessions with a
-    // missing worker can be revived from their stored launch intent).
-    await expect(page.locator('[data-testid="session-reconnect-s1"]')).toHaveCount(1, { timeout: 15000 });
-    await expect(page.locator('[data-testid="session-reconnect-s1"][data-state="recoverable"]')).toBeVisible({ timeout: 15000 });
-    await expect(page.locator('.warning-banner')).toContainText(/can be recovered/i, { timeout: 10000 });
+    // The injected row was never launched, so nothing on disk can restore its
+    // conversation: startup recovery reaps it along with its pane.
+    await expect(page.locator('[data-testid="session-reconnect-s1"]')).toHaveCount(0, { timeout: 15000 });
+    await expect(page.locator('[data-testid="sidebar-session-reconnect-s1"]')).toHaveCount(0);
 
     await daemon.injectSession({
       id: 'reconnect-s2',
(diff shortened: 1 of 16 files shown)


figgyster (approved):
Approve. The invariant is real, and it's the right one: a missing-runtime session survives only when it was not closed on purpose, has a launch intent, and has a restoration target the revive path will actually consume. Activity state and agent name are out of the decision entirely, and both the startup prune and worker reconciliation now ask the same question.

Checked the coherence hard, since this whole change stands on "the revive path actually consumes it":
- `reviveSessionForAttach` refuses without a `LaunchIntent`, and `canReviveSession` requires exactly that before anything else — the mark and the gate agree.
- The resume-id resolution in `canReviveSession` is the same call the spawn pipeline and reload path use (`ResolveSpawnResumeSessionID` + `ResumeAvailable`), so a session marked recoverable is one the spawn path will genuinely hand a live transcript/rollout to. Claude's lazy-transcript case is the proof: a zero-turn session has a persisted self-id but no transcript, so `ResumeAvailable` says no and the row is reaped instead of resurrected into the `claude --resume` non-zero exit the old code happily marked recoverable. That's the defect this kills.
- The `InitialPrompt` seed branch can only fire for conversation-capability plugin hosts — `spawn_pipeline.go` guards `intent.InitialPrompt` behind that capability ("no other runtime gets this") — so the seed is always the brief the replacement host would reopen, never a half-conversation masquerading as a full one.
- Plugin sessions are judged on the persisted handle (open run row or metadata), read from the store because the plugin hasn't reconnected at startup — and a cleanly-exited run plus history under `hosts/state` still survives via `hostSessionStateDirHoldsConversation`. The old capability-only fallback promised recovery from a driver's advertisement; the new code verifies the conversation is where a replacement looks. Strictly more honest.

The tests are written in the right currency: real transcript/rollout files in temp tool homes, all four activity states crossed with claude/codex, re-decision of already-parked recoverable sessions, intentional-close respect, pane-follows-verdict. The real-app scenario is the strongest witness — an actual SIGKILL of every worker plus the daemon via the profile's own registry, the token verified inside the codex rollout file on disk (not the pane echo), the zero-turn claude asserted transcriptless before the crash, and the revived pane reading the pre-crash token back. Ran the Go side myself: daemon/agent/store green (the two plugin-spawn failures reproduce identically on main — environmental), `diff --check` clean, CI green on the head including all three E2E shards.

Three observations, none blocking:

1. The glossary's `recoverable` entry lists three restoration targets but leaves out the intent seed — which is a headline case in this PR (the zero-file conversation host that "never got to say" its brief). The code and PR body both include it; the glossary's "or" list reads as exhaustive. Worth a sentence.
2. Plugin contract tightening worth saying out loud for plugin authors: `resume`/`conversation` capability alone no longer rescues a session. The per-session evidence has to exist — history under `hosts/state`, an open run handle, or metadata. For the reference host this changes nothing (its history was already claimed to live under attn's data dir); for any plugin that kept history elsewhere, the old code was promising a Reload that reopened nothing. Now it reaps instead, which is the honest version — but it's a behavior change for those sessions.
3. Residual window, fail-closed and tiny: a codex session that crashes between the rollout landing on disk and the daemon persisting the observed resume id gets reaped even though the conversation exists — the scenario explicitly waits for the persisted id before crashing, so the window is known. Could be recovered via the cwd+start-time transcript lookup if it ever bites; flagging it so it's on the record.

PR #846, opened 2026-08-11
fix(annotations): refresh live transcript messages

victorarias:
## Intent / Why

Completed assistant commentary should be annotatable as soon as it appears, including while the agent is still running a tool. The previous path refreshed around settled turns and independently rediscovered and rescanned transcripts, so valid Codex prose could be rejected as non-agent text.

## Summary

- Make the daemon's per-session transcript watcher the single authority for exact transcript identity, incremental decoding, and the bounded assistant-message window.
- Publish session-scoped invalidations for completed assistant messages and refresh mounted terminals on invalidation and reconnect.
- Replace the ambiguous pending flag with explicit `discovering`, `ready`, and `unavailable` states in protocol 228.
- Remove broad transcript guessing, the duplicate full-file recent-message reader, content-hash identities, state-gated refresh, and optional invalidation delivery.
- Disambiguate Codex assistant prose when the user prompt contains the requested response verbatim.
- Cover exact same-directory isolation, partial and replaced transcripts, reconnect behavior, and the packaged mid-turn Codex annotation flow.

## Verification

- `go test ./... -count=1`
- `pnpm --dir app test`
- `pnpm --dir app run e2e`
- `make check-types`
- `make build-linux-amd64`
- `make build-linux-arm64`
- Packaged Codex terminal-annotation scenario with the tool held open until the annotation was filed.


figgyster (approved):
Approve. This is the guess-path deletion the arc has been building toward: #845 made resumability a checked fact about disk, and now the annotatable window is a checked fact about the live watcher. The duplicate full-file reader, the content-hash identities, the cwd/newest fallback, the state-gated refresh — all gone, and the code is smaller for it. `decodeEventRecord` gives the paged reader and the follower one normalization path, so provider parsing and echo-dedup can't drift apart again.

Verified what a sandbox can verify, and then some. `git diff --check` clean; `internal/transcript`, `internal/protocol` (incl. the version lockstep test), and `internal/store` all green; the daemon suite green except the two plugin-spawn tests that fail identically on base — environmental, known pair. `make check-types` regenerates byte-identical files, so schema→Go→TS are in real lockstep at 228, not just hand-matched. Full frontend run: 243 files, 2711 passed, 20 skipped — beats the PR's own claim. tsc clean, daemon builds. Pulled the prebuilt libghostty-vt from the pinned release (checksum vs the lock) so the watcher tests ran against the real archive: the lifecycle, exact-identity, paired-record single-invalidation, and same-directory isolation tests all pass. The packaged macOS scenario with the held tool I can't run from here — the plan doc's word is the witness on that one.

One: after a daemon restart the restored watcher bootstraps from the tail — 256KiB for codex/claude, 512KiB for copilot — and that window is the session's annotatable horizon for the daemon's whole lifetime. Messages older than the tail never re-enter it (no backfill; `OmittedPrefix` only logs). Existing annotations persist fine, keyed to cursors. But the old code re-scanned the full file after restart, so "annotate a mid-session message after a restart" genuinely narrowed, and the user has no way to see it — `truncated` never renders, the daemon log is the only trace. The plan doc explains the mechanism; the changelog doesn't name the consequence. Worth a line so it reads as a decision, not a silent regression.

Two: the scenario's old first act — drag before the agent said anything, refusal names its reason, retires itself after 6s, daemon logs it — had a live packaged witness. The rewritten scenario skips it entirely in favor of the mid-turn flow, which is the right headline, but the notice self-retirement path is now only unit-tested. The code didn't change there; the witness did. If that refusal behavior matters (it was the "feature broken vs. working" tell), it deserves a cheaper scenario step than it had.

The Codex prompt-quote disambiguation is the sleeper hit — that's a real bug the old alignment would have fought forever, and the `assistantLead` tie-break plus the harness's bullet-row requirement cover both halves.

PR #847, opened 2026-08-11
feat(testing): record PR evidence and publish it to the evidence repo

victorarias:
## What

A process for putting screen recordings in PR descriptions as evidence of testing, plus the script that runs it: `scripts/pr-evidence.sh`.

- `record` captures a single app window (`screencapture -x -v -l <windowid>`, window id resolved via CoreGraphics) to mp4 — retina, no cropping, no other windows.
- `publish` converts each mp4 to a GIF (10fps, ≤960px, per-clip palette), pushes both to [attn-pr-evidence](https://github.com/victorarias/attn-pr-evidence) under a per-branch directory, and prints sha-pinned markdown ready to paste into the PR description.

AGENTS.md documents the process next to live verification, so agents record the verification run they already perform.

## Why mp4 + GIF

GitHub only inline-plays videos uploaded through its own web UI (user-attachments), which has no API. Every repo-hosted mp4 form — raw URL, blob `?raw=true`, release asset, `<video>` tag — renders as a plain link or is stripped. A GIF referenced from a public repo's raw URL does render inline. Measured, one section per embed form: [attn-pr-evidence#1](https://github.com/victorarias/attn-pr-evidence/issues/1). So the GIF is the inline preview and the mp4 beside it is the full-quality master.

Numbers with receipts: busy terminal content converts at ~320KB/s of GIF at the publish defaults (6s live attn window → 1.9MB), so the 20s record default stays well under GitHub's 10MB image render limit; the script warns past it.

## Evidence

The process, run on itself — recorded and published by this branch's `pr-evidence.sh`:

dogfood

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/40633cff8e09601a0c76f66e4226228e5fd85463/salty-pigeon/dogfood.mp4)

https://claude.ai/code/session_01GcYDxzNUutctekU9xV2ZDA

### DIFF
--- a/scripts/pr-evidence.sh
+++ b/scripts/pr-evidence.sh
@@ -68,7 +68,7 @@ cmd_record() {
   # 20s default: busy terminal content converts at ~320KB/s of gif at the
   # publish defaults (receipt: 6s live attn window -> 1.9MB), so 20s stays
   # well under GitHub's 10MB inline-render limit; 30s flirts with it.
-  local app="attn-dev" seconds=20 out=""
+  local app="" seconds=20 out=""
   while [[ $# -gt 0 ]]; do
     case "$1" in
       --app) app="$2"; shift 2 ;;
@@ -77,6 +77,12 @@ cmd_record() {
       *) die "record: unknown argument $1" ;;
     esac
   done
+  # Record the profile the caller is already verifying on (attn-<profile>.app
+  # owns a window named after itself); never guess between running profiles.
+  if [[ -z "$app" ]]; then
+    [[ -n "${ATTN_PROFILE:-}" ]] || die "no --app and no ATTN_PROFILE; pass --app <owner> (e.g. attn-<profile>) or select a profile first"
+    app="attn-$ATTN_PROFILE"
+  fi
   [[ -n "$out" ]] || out="evidence-$(date +%Y%m%d-%H%M%S).mp4"
 
   local id_and_width
--- a/scripts/pr-evidence.sh
+++ b/scripts/pr-evidence.sh
@@ -1,0 +1,163 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# PR evidence recordings: capture an app window to mp4, publish mp4+gif to the
+# evidence repo, and emit the markdown to paste into a PR description.
+#
+#   pr-evidence.sh record  [--app <owner>] [--seconds N] [--out FILE]
+#   pr-evidence.sh publish [--dir <name>] [--fps N] [--width N] FILE.mp4 ...
+#
+# GitHub inline-renders a GIF referenced from a public repo's raw URL, but a
+# repo-hosted mp4 (raw, blob, or release asset) is only ever a click-through
+# link, and <video> tags are stripped. Receipts, one section per embed form:
+# https://github.com/victorarias/attn-pr-evidence/issues/1. Hence the pair:
+# GIF for the inline preview, mp4 as the full-quality master.
+
+EVIDENCE_REPO="${ATTN_PR_EVIDENCE_REPO:-victorarias/attn-pr-evidence}"
+
+usage() {
+  sed -n '4,9p' "$0" | sed 's/^# \{0,1\}//'
+  exit 2
+}
+
+die() {
+  echo "pr-evidence: $*" >&2
+  exit 1
+}
+
+# Prints "<window-id> <width>" for the largest on-screen layer-0 window owned
+# by $1; with WINID_LIST=1, prints every attn* owner instead (for the error
+# path). CGWindowID is not reachable from AppleScript, so this goes through a
+# throwaway swift script.
+window_id_for_app() {
+  local owner="$1"
+  local swift_src
+  swift_src="$(mktemp -t pr-evidence-winid).swift"
+  cat > "$swift_src" <<'EOF'
+import CoreGraphics
+import Foundation
+let opts: CGWindowListOption = [.optionOnScreenOnly, .excludeDesktopElements]
+guard let list = CGWindowListCopyWindowInfo(opts, kCGNullWindowID) as? [[String: Any]] else { exit(1) }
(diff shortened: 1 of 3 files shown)


scripts/pr-evidence.sh
```diff
@@ -0,0 +1,163 @@
...
+    if (( gif_bytes > 10485760 )); then
+      echo "warning: $gif is $((gif_bytes / 1048576))MB; GitHub inline-renders images only under 10MB — shorten the clip or lower --fps/--width" >&2
+    fi
+    names+=("$base")
+  done
+
+  git -C "$clone" add "$dir"
+  git -C "$clone" commit --quiet -m "evidence: $dir"
+  # Another agent may publish concurrently; rebase once and retry.
+  git -C "$clone" push --quiet || { git -C "$clone" pull --rebase --quiet && git -C "$clone" push --quiet; }
```
figgyster: The retry chain has a silent 404 path. When the rebase leg itself fails — auth, network, shallow-clone hiccup — set -e does not fire: the failing command is the first of the inner `&&` list, so it's exempt, and the script sails past the failed push to `rev-parse HEAD` and prints "published" with a sha that never left the machine. The raw URLs then 404 in the PR description. Verified empirically with this exact control-flow shape (push fails, rebase fails -> script continues, exit 1); also verified the intended path works: concurrent push, rebase in a real --depth 1 clone, second push, correct rebased sha printed.

Fix is one token: `... || die "push to $EVIDENCE_REPO failed"` on the end of the chain.

figgyster (approved):
The evidence process, with receipts I went and checked. The evidence repo is public, pushed at 16:19:20Z about a minute before the PR opened, and both salty-pigeon files resolve at the pinned sha. The dogfood is real, not a screenshot of a screenshot. The mp4/GIF split is the kind of thing someone measures before believing, and attn-pr-evidence#1 does that: one section per embed form, so the decision is a fact list, not a guess. CI agrees with the blast radius: E2E, react-doctor, Changelog, Changes green; the code suites skipped is path filtering doing its job for a scripts/docs-only change. Changelog fragment is valid per the gate. Script is clean, usage is honest, the 10MB warning is a warning not a trap.

Two observations, neither blocking:

1. Public by default, and the subject is attn. The scheme only works because the evidence repo is public, and what gets recorded is the app whose whole job is showing agent sessions, ids, transcripts. The AGENTS.md flow reads "record the run, publish" with no pause in between. For a human that's a glance; for an agent following the doc it's a fire-and-forget push of whatever's on screen to the open internet. One line in the doc — check the clip before it ships — would close the loop cheaply.

2. The throwaway swift file from mktemp never gets cleaned up. cmd_publish already owns the trap pattern; window_id_for_app could borrow it.

Also worth saying: the recorded evidence lives in a separate repo pinned by sha, which means PR descriptions stay readable and the media never rots when a branch force-pushes. Good shape. Ship it.

figgyster (approved):
Loop closed, and faster than I got the first review out. All three items answered in one commit:

1. The retry chain now ends in `|| die` — re-ran the control flow with the new shape, all legs: total failure dies loudly, push #2 failure dies loudly, and the intended concurrent-publish path (push rejected, rebase, push again) still sails through. The silent 404 path is gone.

2. The swift file gets cleaned — and the cleanup is careful, status captured before the `rm` so the failure signal still reaches the caller, including the `.swift`-less mktemp original.

3. The AGENTS.md privacy note is sharper than what I asked for: names the repo public, names what the clip shows (session names, transcripts, tickets), and gives the agent a rule — watch before publishing, re-record rather than push.

CI matches the blast radius again: E2E, react-doctor, Changelog green; code suites skipped by path filter; the only running check is the same one that passed on the previous head. Nothing open on the file. Re-approving on the merge candidate.

figgyster (approved):
Second lap, and this one isn't answering me — it's hardening the tool on its own terms. The hardcoded `attn-dev` default is gone, replaced by the profile you're actually verifying on: `record` now targets `attn-$ATTN_PROFILE`, and when neither `--app` nor the profile is there, it dies with a message that teaches the fix instead of guessing. That's the profile-isolation invariant leaking into tooling, and it fits the arc — the evidence recorder should record the exact thing under test, not an assumed dev build.

Ran the matrix with the macOS bits stubbed: no `--app` and no `ATTN_PROFILE` dies (exit 1, correct message), `ATTN_PROFILE=dev` targets `attn-dev` and proceeds, `--app` overrides the profile, an empty `ATTN_PROFILE` dies the same as unset. The empty-vs-unset distinction is handled — ${ATTN_PROFILE:-} covers both, so recording prod means passing `--app attn`, which the die message points at.

CI green for what runs: E2E, react-doctor, Changes, Changelog all pass; code suites skipped by path filter again. AGENTS.md matches the new behavior, changelog fragment still accurate. Nothing open on the file. Approving on the merge candidate.

figgyster (approved):
Lap four, and the interface found its final shape. `--profile <name>` is now the primary flag, and it closes the gap I flagged last lap: bare prod is `attn.app`, so `--profile prod` maps to the bare `attn` owner while every other profile maps to `attn-<name>`. ATTN_PROFILE got demoted to fallback with an honest reason in the comment — it only exists in the shell that eval'd profile-env, so it can't be the contract; the named profile is.

Verified the matrix with macOS bits stubbed, all six legs: `--profile dev` targets attn-dev, `--profile prod` targets bare attn, `--app` still overrides, no flags dies with "which window?" and both remedies, ATTN_PROFILE fallback works, and an explicit `--profile` beats the env var. Precedence is clean: flag, then env, then a loud death.

CI green for what runs again — E2E, react-doctor, Changes, Changelog; code suites skipped by path filter. AGENTS.md matches the new shape and ties the profile name to the one you'll `profile clean` later, which is the right lifecycle hook. One nit, barely worth the ink: the window-lookup failure message still says only "Pass --app <owner>" — accurate as a remedy, just doesn't mention `--profile`. Approving on the merge candidate.

PR #848, opened 2026-08-11
fix(testing): wait on daemon startup signals

victorarias:
## Summary

- replace the shared E2E daemon startup poll and fixed deadline with Unix-socket creation and child-exit signals
- include isolated stdout, stderr, and daemon-log evidence when startup fails
- expose a recovery-cycle completion signal so daemon startup tests assert post-recovery state without sleeping or polling
- document the active issue #796 root causes and preserve the existing product assertions unchanged

## Root cause

Several E2E failures attributed to terminal redraw and LocationPicker scenarios happened before those scenarios ran: a healthy daemon child had not created its socket before the fixture's fixed deadline. Separately, `TestDaemon_PrunesSessionsWithoutLivePTYOnStart` queried the store while asynchronous startup PTY recovery could still be pruning the seeded session.

CI found the same ordering bug in `TestAgentMsgSizeRefusalsSurviveTheSocket`: it seeded its sender and target after the socket appeared but before recovery settled, so recovery could prune them before the message was handled. It now waits on the same recovery completion edge before seeding.

The dominant split-paint ledger entry is already fixed on main by `487adce1`; its later occurrences ran older branch code.

## Validation

- `go test ./internal/daemon`
- targeted startup-prune and recovery-signal tests, 25 consecutive runs
- socket message size-refusal test, 25 consecutive runs
- targeted tests under `go test -race`
- `pnpm --dir app test` — 243 files, 2,714 passed, 15 skipped
- Playwright readiness success and early-exit diagnostics
- exact LocationPicker hover/Enter and chunked split-redraw scenarios against the branch binary
- `go run ./cmd/changelog-check`

No protocol or user-visible behavior changes.

### DIFF
--- a/app/e2e/daemon-readiness.spec.ts
+++ b/app/e2e/daemon-readiness.spec.ts
@@ -1,0 +1,49 @@
+import { test, expect } from '@playwright/test';
+import { spawn, type ChildProcess } from 'child_process';
+import * as fs from 'fs';
+import * as net from 'net';
+import * as os from 'os';
+import * as path from 'path';
+import { waitForDaemonSocket } from './daemonReadiness';
+
+async function stopChild(proc: ChildProcess): Promise<void> {
+  if (proc.exitCode !== null || proc.signalCode !== null) return;
+  await new Promise<void>((resolve) => {
+    proc.once('exit', () => resolve());
+    proc.kill('SIGTERM');
+  });
+}
+
+test('daemon readiness follows socket creation without a wall-clock poll', async () => {
+  const tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'attn-ready-'));
+  const socketPath = path.join(tempDir, 'attn.sock');
+  const proc = spawn(process.execPath, ['-e', 'process.stdin.resume()'], { stdio: 'pipe' });
+  const server = net.createServer();
+
+  try {
+    const ready = waitForDaemonSocket(proc, socketPath);
+    await new Promise<void>((resolve, reject) => {
+      server.once('error', reject);
+      server.listen(socketPath, resolve);
+    });
+    await expect(ready).resolves.toBeUndefined();
+  } finally {
+    await new Promise<void>((resolve) => server.close(() => resolve()));
+    await stopChild(proc);
+    fs.rmSync(tempDir, { recursive: true, force: true });
+  }
+});
+
+test('daemon readiness reports an early child exit with its diagnostics', async () => {
+  const tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'attn-ready-'));
+  const socketPath = path.join(tempDir, 'attn.sock');
+  const proc = spawn(process.execPath, ['-e', 'process.exit(23)']);
+
+  try {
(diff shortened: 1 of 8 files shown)


figgyster (approved):
**Approve** — this is the main-side fix for the startup race I flagged during the #843 staging review, and it's built the right way: readiness became a real lifecycle edge instead of a deadline with a guilty-until-proven scenario.

**What I verified**

- Ordering invariants hold. `setRecovering(true)` (daemon.go:991) precedes both the socket bind (~1005) and `signalStarted()` (1146), so every wait path — `waitForSocket`-then-`waitForRecovery` or `startedCh`-then-`waitForRecovery` — is guaranteed to observe recovery either in flight or complete. The query-before-recovery window is structurally closed, not just narrowed.
- Signal semantics are sound: the channel is created only on a fresh recovery cycle, closed exactly once under the lock (the `recovering` flag guards double-close), and `recoverySettledSignal` returns a shared always-closed channel outside recovery. The shared helper upgrade carries ~10 call sites (daemon, ws_eviction, toxiproxy), not just the prune test, and none regress against the old poll's behavior.
- Ran the two targeted tests 15× clean, 5× clean under `-race`. Full `./internal/daemon` suite: green except the known environmental plugin-spawn pair, which I reproduced identically on `origin/main` — not this PR.
- Both new Playwright seam specs pass for real (socket-creation and child-exit-with-diagnostics). `tsc --noEmit` clean, `changelog-check` exit 0, `git diff --check` clean. Playwright's default testMatch won't collect `daemonReadiness.ts` as a spec.
- The diagnostic chain is real, not aspirational: the daemon opens `config.LogPath()` = `attnDir()/daemon.log`, and `ATTN_DATA_DIR` is the highest-precedence data-dir override — so adding it to both launchers makes `daemonStartDebugInfo` read the actual isolated log. Verified through config.go's chokepoint.
- The cutoff protection in `pruneSessionsWithoutPTY` (post-start sessions are skipped) makes the new wait ordering safe for the prune test's seeded session.

**Non-blockers**

1. `waitForRecovery` is now an unbounded `<-signal`. The old helper had a 5-second named tripwire; if recovery ever wedges, the failure mode degrades to Go's 10-minute package timeout with no diagnostic. The E2E side documents the identical trade (runner budget as outer tripwire); the Go side has no runner-level equivalent. Theoretical — `recoverPTYBackend` is internally bounded at 10s — and the deadline was precisely the false-failure source this PR removes, so I'm not asking for a change. Just putting the arithmetic on record. A `select` with a generous (30s+) deadline would keep the diagnostic without reintroducing the race, if it ever bites.

2. The ledger plan names three findings, but the `agent_msg` socket-test race from my #843 notes isn't among them, and `agent_msg_test.go` doesn't use the new signal. I couldn't reproduce it on either head in 20 runs each — it may well have been a misattribution to the same root-cause family. But if it's still alive in CI, the recovery signal is now available to close it the same way; worth a look the next time the ledger gets an entry.

Good plan doc discipline, by the way — "failures on branches that predate an existing fix are historical evidence, not a reason to change current code again" is exactly the right read of the split-paint entries. The ledger closes with receipts, not vibes.

PR #850, opened 2026-08-11
feat(harness): record scenario runs to per-launch mp4 segments

victorarias:
## What

`ATTN_HARNESS_RECORD=1` makes every `createScenarioRunner` scenario record its app window to `<runDir>/recording-NN.mp4` — one segment per app launch, listed in `recording.json` — with zero per-scenario wiring. The env var flows through the serial matrix and soak runner, so any leg can be recorded. Default off: a matrix run should not pay recording CPU/disk unasked. Recording failures trace and never fail a run.

Recordings serve two purposes: PR evidence (publish a segment with `scripts/pr-evidence.sh publish`, from #847) and failure diagnosis — a failing run's artifacts now can include the actual footage next to pane text and screenshots.

## How

Capture is a new `WindowRecorder.swift` (compiled on demand and codesigned like `InputDriver.swift`) built on ScreenCaptureKit's desktop-independent window filter, driven by a window-id poller in `windowRecording.mjs` that starts, rotates, and finalizes segments as the scenario's window appears, relaunches, or dies. `scenarioRunner.mjs` starts it when enabled and stops it in `finalizeRunner`, covering success, failure, and signal exits.

Three measured facts shaped this (receipts in the swift/mjs headers):

- The harness parks bridge-driven windows almost fully off-screen (`uiAutomationClient` → `window_park`, ~20px visible). `screencapture -v -l` composites only the on-screen region — it recorded black frames with a 20px sliver. ScreenCaptureKit's `SCContentFilter(desktopIndependentWindow:)` records the window's full content wherever it is: verified with a 97%-off-screen window returning complete retina frames.
- A recorder whose window closes never exits on its own; segments must be stopped explicitly, and an app relaunch means a new segment against the new window id — hence the poller.
- SCK delivers frames only when the window repaints, so quiet stretches cost nothing; a final re-append at stop time gives each file its true wall-clock duration.

## Verification

- 12 unit tests over the poller/segment lifecycle (rotation, stop races, SIGKILL tripwire, build-failure disable); all 204 harness tests pass.
- Live on a profile installed from this branch: `scenario-tr204-local-relaunch-formatting` with recording on passed and produced two segments — rotation observed at the relaunch, both playable, frames showing the full parked window (below). A control run without the env var produced zero recording artifacts and no recorder work.

## Evidence

Segment 1 of the tr204 run (launch → session → relaunch), recorded by this branch's harness recorder from a window parked 97% off-screen, published with `pr-evidence.sh`:

tr204 harness recording

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/c9bf6cbef3066f4b08014e9121d311aae42f82f0/candid-heron/tr204-recording.mp4)

https://claude.ai/code/session_01GcYDxzNUutctekU9xV2ZDA

### DIFF
--- a/app/scripts/real-app-harness/scenarioRunner.mjs
+++ b/app/scripts/real-app-harness/scenarioRunner.mjs
@@ -3,6 +3,8 @@ import os from 'node:os';
 import path from 'node:path';
 import { assertPackagedAppBuildMatchesCurrentSource } from './buildPreflight.mjs';
 import { createRunContext, emitVerdict, FIRST_FAILURE_MAX_LENGTH } from './common.mjs';
+import { MacOSDriver } from './macosDriver.mjs';
+import { createScenarioRecorder, recordingEnabled } from './windowRecording.mjs';
 
 // Collapse to a single line and cap length so the verdict's firstFailure field
 // can never break the one-line ATTN_VERDICT contract regardless of what the
@@ -186,6 +188,20 @@ export function createScenarioRunner(options, {
     process.stdout.write(line);
   };
 
+  // ATTN_HARNESS_RECORD=1 records the app window to runDir/recording-NN.mp4,
+  // one segment per app launch. The recorder follows the window by polling, so
+  // no scenario has to wire anything; when disabled nothing runs at all.
+  let recorder = null;
+  if (recordingEnabled()) {
+    const recordingDriver = new MacOSDriver({ appPath: options.appPath });
+    recorder = createScenarioRecorder({
+      runDir,
+      resolveWindowId: () => recordingDriver.mainWindowId(),
+      log: appendTrace,
+    });
+    recorder.start();
+  }
+
   const runRegisteredCleanup = async (reason) => {
     if (cleanupPromise) {
       return cleanupPromise;
@@ -218,6 +234,10 @@ export function createScenarioRunner(options, {
       return;
     }
     finalized = true;
+    // Not awaited: the recorder's screencapture child keeps the event loop
+    // alive until it has finalized its file, and an orphaned recorder (signal
+    // path exits immediately) finalizes on its own after the SIGINT.
+    void recorder?.stop();
     releaseScenarioLock?.();
     process.removeListener('exit', exitHandler);
     for (const [signal, handler] of signalHandlers.entries()) {
--- a/app/scripts/real-app-harness/windowRecording.mjs
+++ b/app/scripts/real-app-harness/windowRecording.mjs
@@ -102,7 +102,9 @@ export function startWindowRecording({ windowId, outputPath, command, spawnFn = 
           ? `recorder ignored SIGINT for ${FINALIZE_TRIPWIRE_MS}ms and was SIGKILLed; the file is likely unplayable`
           : bytes === 0
             ? `recorder exited ${code} with no output${stderr ? `: ${stderr.trim()}` : ''}`
-            : null;
+            : code !== 0
+              ? `recorder exited ${code} leaving a possibly unplayable file${stderr ? `: ${stderr.trim()}` : ''}`
+              : null;
       return { windowId, outputPath, bytes, exitCode: code, failure };
     },
   };
--- a/app/scripts/real-app-harness/windowRecording.mjs
+++ b/app/scripts/real-app-harness/windowRecording.mjs
@@ -1,0 +1,214 @@
+import { execFile, spawn } from 'node:child_process';
+import { createHash } from 'node:crypto';
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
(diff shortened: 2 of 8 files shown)


app/scripts/real-app-harness/scenarioRunner.mjs
```diff
@@ -218,6 +234,10 @@ export function createScenarioRunner(options, {
       return;
     }
     finalized = true;
+    // Not awaited: the recorder's screencapture child keeps the event loop
+    // alive until it has finalized its file, and an orphaned recorder (signal
+    // path exits immediately) finalizes on its own after the SIGINT.
+    void recorder?.stop();
```
figgyster, line 240: Fire-and-forget is the right call for the mp4 — the child self-finalizes, and cleanup quits the app before this runs, so a signal-path orphan stops at window close anyway. One thing that only lands on the normal completion path: recording.json, since `process.exit()` races the not-awaited `stop()` continuation. Footage is the evidence, so this is fine for failure diagnosis — just noting the manifest is best-effort on signal exits.

app/scripts/real-app-harness/windowRecording.mjs
```diff
@@ -0,0 +1,214 @@
...
+      }, FINALIZE_TRIPWIRE_MS);
+      tripwire.unref();
+      const { code, spawnError } = await exited;
+      clearTimeout(tripwire);
+
+      let bytes = 0;
+      try {
+        bytes = fs.statSync(outputPath).size;
+      } catch {}
+      const failure = spawnError
```
figgyster, line 99: A writer failure mid-stream (finishWriting fails → exit 4) with a partial file on disk doesn't count as a segment failure — only spawn errors, SIGKILL-forced stops, and zero-byte outputs do. The manifest carries exitCode so it's diagnosable, but a dead-finalize mp4 (no moov atom) marked usable is the one case where "recording failures trace" goes quiet. Consider non-zero exit with bytes > 0 a failure too.

figgyster (approved):
Went through this with the whole file open. The recorder's a serious piece of machinery — desktop-independent SCK capture because `screencapture -v` films black where the harness parks its windows, a poller that rotates segments on relaunch, content-hash rebuild plus codesign so the TCC grant survives rebuilds, and every failure path tracing instead of failing the run. The three receipts in the swift header are the right shape of evidence, and the published tr204 footage is real: checked the mp4's a valid mp42 and both assets sit in the evidence repo.

Verified the seams against the tree, not just the diff: `mainWindowId()` returns null when the window's gone, so the poller's truthiness check is safe; matrix and soak both spawn legs with `env: process.env`, so the var passes through untouched; the codesign pattern matches InputDriver's; scenarios register `quit_app` cleanup, so a signal-path orphan finalizes itself when the window dies. Ran the suite in the box: 12/12 new tests, 199 passed / 5 skipped across the harness — the 204 claim holds up. CI's green for what ran, changelog fragment's the right shape, diff-check clean.

One real gap, and it's not in this diff: the frontend job's path filter doesn't cover `app/scripts/**`, so these 12 new tests — and every future harness change — never run in CI. Frontend shows "skipping" on this head. The tests run fine on ubuntu; I just ran them. The filter just never sends them there. Adding `app/scripts/real-app-harness/**` (or `app/scripts/**`) to the frontend filters list closes it, and a harness regression starts failing loudly instead of shipping green with a local-only witness.

Two smaller notes inline. Neither blocks — approving on the merge candidate.

app/scripts/real-app-harness/scenarioRunner.mjs
victorarias: Agreed on both counts — the footage self-finalizes and the manifest is best-effort on signal exits by design; left as is.

app/scripts/real-app-harness/windowRecording.mjs
victorarias: Fixed: a non-zero exit with bytes on disk is now a failure ("recorder exited N leaving a possibly unplayable file" + stderr), with a unit test. The signal-kill case (code null) already fell under the SIGKILL tripwire branch.

figgyster (approved):
Loop-closer inside ten minutes — the filter line, the partial-file flag, and an eyewitness test to pin it. Verified against the new head:

- `app/scripts/**` lands in the frontend filter: the Frontend job is running on this PR for the first time as of this commit, so harness tests now ship with CI coverage instead of a local-only witness.
- A non-zero exit with bytes on disk is now a segment failure, stderr included, and the new unit test pins the exit-4-with-partial-bytes case exactly. Ran the suite in the box on the new head: 200 pass / 5 skip, all green.
- The manifest note stays as designed — agreed, the footage is the evidence and the process exit tells the rest.

Both threads resolved. Re-approving on the merge candidate.

PR #851, opened 2026-08-11
fix(daemon): wait for startup recovery before seeding agent-msg socket test sessions

victorarias:
## What

`TestAgentMsgSizeRefusalsSurviveTheSocket` failed ~1 in 5 under CI load with
`daemon error: sender_session_not_found`. The test now waits for startup
recovery to finish (`waitForRecovery`) before seeding its sessions. Test-only
change; no daemon behavior is touched.

## Root cause

The test runs `go d.Start()` and waits only for the socket file, but the
socket appears before startup recovery runs. It then seeds sessions via
`addCharacterizationSession`, which deliberately backdates
`LastSeen`/`StateUpdatedAt` to `2000-01-01`. Startup recovery's
`pruneSessionsWithoutPTY` reaps sessions with no live PTY, and its own-run
guard only protects sessions updated after recovery started — a backdated row
is exactly what it cannot protect. Under load the reaper lands after test
setup and removes the sender, so the daemon answers the send with
`sender_session_not_found`.

The fix is the package's existing pattern: `TestDaemon_RegisterAndQuery` and
friends already call `waitForRecovery(t, d)` right after `waitForSocket`.

## Every instance, not just the one that flaked

Audited every `go d.Start()` site in `internal/daemon` tests: this is the
only one that seeds backdated sessions after starting the daemon. Every
other post-start site registers sessions through the socket with fresh
timestamps (protected by the prune's cutoff guard) or spawns real PTYs, so
no other test carries this race.

## Verification

Reproduced first, then fixed, on the same harness: 14 `timeout`-bounded CPU
spinners on a 10-core machine, one test-binary run per iteration.

- Unfixed (origin/main): **1 failure in 40 runs**, exactly
  `sender_session_not_found`.
- Fixed: **120/120 pass** under the same load.
- Full `go test ./internal/daemon` green on an unloaded machine.
- `gofmt`, `go vet ./internal/daemon`, and `go run ./cmd/changelog-check`
  clean.

This is a test-only Go change with no app-observable behavior, so unit-level
verification is the right tier per the repo guide; no live app install.

https://claude.ai/code/session_01JFyBEwqep7dhgd3cWFjbFE

### DIFF
--- a/changelog.d/flaky-agent-msg-startup-reaper.yaml
+++ b/changelog.d/flaky-agent-msg-startup-reaper.yaml
@@ -1,0 +1,11 @@
+kind: internal
+area: daemon tests
+change: >
+  The agent-msg socket test now waits for startup recovery to finish before
+  seeding its backdated sessions, so the startup reaper can no longer remove
+  the sender mid-test.
+notes: >
+  TestAgentMsgSizeRefusalsSurviveTheSocket failed under CI load with
+  "sender_session_not_found": the socket appears before recovery runs, and
+  the deliberately backdated characterization sessions are outside the
+  reaper's own-run guard. Test-only change; no daemon behavior touched.
--- a/internal/daemon/agent_msg_test.go
+++ b/internal/daemon/agent_msg_test.go
@@ -266,6 +266,10 @@ func TestAgentMsgSizeRefusalsSurviveTheSocket(t *testing.T) {
 	go d.Start()
 	defer d.Stop()
 	waitForSocket(t, sockPath, 5*time.Second)
+	// The socket appears before startup recovery finishes, and recovery reaps
+	// sessions with no PTY — these seeded rows are deliberately backdated, so
+	// the reaper's own-run guard cannot protect them. Seed only after it ran.
+	waitForRecovery(t, d)
 	addCharacterizationSession(t, d, "sender-session-id", protocol.SessionAgentClaude, protocol.SessionStateIdle)
 	addCharacterizationSession(t, d, "target-session-id", protocol.SessionAgentClaude, protocol.SessionStateIdle)
 


figgyster (approved):
**Approve.** This is the case I left open in #848 — the agent_msg socket test racing #845's startup reaper — and the fix is exactly right. Verified every leg of the story against the code, then put both versions through the same gauntlet.

**Root cause checks out, line by line.** `characterizationOldTimestamp` is literally `2000-01-01T00:00:00Z`; `pruneSessionsWithoutPTY`'s own-run guard (`sessionUpdatedAfter(session, cutoff)`) protects only rows moved after recovery started; a backdated row is exactly what it can't protect. Reaped sender → `sender_session_not_found`.

**The two-wait pattern is race-free, and here's why:** `setRecovering(true)` (daemon.go:1009) precedes the synchronous socket bind (1047), and `setRecovering(false)` (1114) fires only after `performStartupPTYRecovery` — which contains the reaper — completes. So `waitForSocket` returning means recovery has either started (flag is up, `waitForRecovery` blocks on it) or already finished (reaper done, seeding safe). No window where the reaper lands after the seed. The deferred reconciliation pass can't bite either: it's only scheduled when the initial reconcile reports skips, and a fresh test daemon's store is empty during recovery, so the report is all zeros.

**A/B on the same harness, my box:** unfixed, under `GOMAXPROCS=2` + 12 CPU spinners, **14/150 fail** — `sender_session_not_found`, plus `session_not_found` when the reaper takes the target too, same root cause. Fixed, identical conditions: **150/150 pass**. Your 1-in-40 was the gentler harness; under harder skew the race is ~1-in-11. Fix holds.

**Audit claim verified:** I checked every `go d.Start()` site in `internal/daemon` — this is the only one seeding backdated rows post-start. daemon_test.go and stop_terminality_test.go both use fresh `now`/`nowStr` timestamps (cutoff-guard protected); auto_settle never starts the daemon at all.

**Local suite:** full `go test ./internal/daemon` green except the known environmental plugin pair (`TestPluginDriverEndToEnd_InstalledProcessLaunchReportAndResumeThroughWorkerPTY`, `TestDaemon_StartInstalledPlugins_SpawnsProviderPlugin`) — reproduced identically on base; CI's Daemon job runs them green. `go vet` and `changelog-check` clean. CI on this SHA: Daemon, Changelog, Bench A/B, both Linux daemons, Frontend E2E all green.

Case closed on the ledger question I flagged in #848 — with better receipts than I asked for.

PR #852, opened 2026-08-11
feat(enrollment): record every daemon's home, and fence home-level state

victorarias:
Stage 2 of the home–garden–crew arc: the vocabulary in code, the enrollment
record, and the fence. Plan:
[docs/plans/2026-08-10-home-garden-crew-arc.md](docs/plans/2026-08-10-home-garden-crew-arc.md)
(its stage-2 checklist is ticked here).

Every daemon has exactly one home. A **home daemon** is standalone and complete —
it owns its garden and its crew, and every fresh install is one. An **outpost** is
a daemon enrolled to a home: it keeps owning its sessions, PTY, and PR flows, and
owns none of the home-level state. This PR records that relationship and fences
what depends on it, so the garden (stage 3) and the crew (stage 4) can be built
home-only without owing a migration when the uplink (stage 5) lands.

Nothing about a single machine changes: one daemon on its own is a home, and
everything outposts do today is untouched.

## What lands

**The record.** `internal/enrollment` owns two files in the data dir — `daemon-id`
(who this daemon is, moved here from `internal/daemon/instance_id.go` unchanged)
and `enrollment.json` (whose it is) — under the same flock discipline. A fresh
install writes a record naming itself, so "am I a home?" is answered by a file
rather than by the absence of one.

**The fence.** `Status.RequireHome` is the single place anything asks whether
home-level state may live here; the daemon reaches it through
`Daemon.requireHome`, and nothing checks the record ad hoc. It fails closed on a
record it cannot read — a daemon must not claim to be a home it cannot prove. Its
refusal names what refused, this daemon, its home, both ways forward, and the
plan:

```
refused the garden and the crew on this daemon: it is an outpost, and home-level state lives at its home.
  this daemon: d-d1d8528f1876c31d336286281fc5c327 (outpost)
  its home:    d-0123456789abcdef0123456789abcdef
The garden and the crew have exactly one owner — the home daemon — and the uplink that would
carry this ask home is not built yet.
Do this on the home daemon (d-0123456789abcdef0123456789abcdef), or make this daemon its own home again with `attn enrollment leave`.
Why: docs/plans/2026-08-10-home-garden-crew-arc.md
```

It already has a production caller: enrolling a remote says "your garden and crew
live here", so a daemon that is itself an outpost passes the fence first and
enrolls nobody.

**Reporting.** Health says `"enrollment": "home"` or `"outpost of <id>"`;
`initial_state` carries `home_daemon_id` (protocol 229). The record is re-read on
every ask rather than cached at startup, because a sync and `attn enrollment
leave` both rewrite it under a live daemon — a cached answer would keep refusing
until the next restart.

**Enrolling.** A home enrolls a remote by running `attn enrollment enroll --home
<id>` there over ssh, right after installing the binary and before starting the
daemon. The record's rules live in one place instead of a shell script that would
drift from them, and the wording the outpost produces is what the home shows.

**Re-home is a decision.** A remote already enrolled to a different home refuses
with exit 3, and that refusal stops the sync with the remote's own words. The ask
is answered on that machine: `attn enrollment leave` makes it a home again, and
the second home's sync then enrolls it. Anything else a remote cannot answer — an
older binary with no `enrollment` command — is logged and does not block the
sync, because enrollment is a record, not a precondition for sessions.

**The operator surface.** `attn enrollment` (status, and the refusal it would
give), `attn enrollment enroll --home <id>`, `attn enrollment leave`. The way in
has a way out and a way to see it.

## Not in scope

The uplink (stage 5), any garden or crew state, any server work. The fence exists
precisely so those wait without debt.

## Live verification

Throwaway profile `fence3q` (installed from this branch, preflight PASS, protocol
229 end to end, cleaned up after). No recording: nothing in the app window
changes — this is daemon, CLI, and wire.

- Fresh install wrote its own home record: `enrollment.json` naming
  `d-d1d8528f1876c31d336286281fc5c327`, beside `daemon-id`, both 0600.
- `attn enrollment` → `home` / "this daemon owns them"; health →
  `"enrollment": "home"`; `initial_state` over the WebSocket →
  `home_daemon_id == daemon_instance_id`.
- Enrolled the live daemon to another home id: health, `initial_state`, and the
  fence all moved with no restart; the full refusal above is what `attn
  enrollment` printed.
- A second home asked to take it: exit 3, record untouched, refusal naming both
  homes and `attn enrollment leave`. After `leave`, the second home's enroll
  succeeded.
- **The real ssh leg** against the OrbStack remote (`attn-remote@orb`, Linux
  arm64): enrolled a machine that had never run a daemon there, re-ran it
  (unchanged, quiet), then dialed it from a second home and got the refusal back
  across ssh verbatim. That run found a bug a shim could not: enrolling into a
  data dir that does not exist yet failed on the lock file. Fixed, with a test
  that walks the whole first-contact sequence. The remote's profile dir and
  binary were removed afterwards.

`go test ./internal/... ./cmd/...`, `go vet ./...`, `make check-types`,
`pnpm --dir app test` (2714 passing), and `make build-linux-{amd64,arm64}` are
green.

https://claude.ai/code/session_01XbVniVF2PqnuJRunFxdPKa


figgyster (changes requested):
Stage 2 of the spine, in code. Read the whole case file — the record, the fence, the ssh leg, the CLI, the protocol bump, the plan tick-list, and every test that pins a decision. Then ran the box.

What I verified:

- **The record holds up.** Fresh install writes a self-named record, so "am I a home?" is answered by a file, not by an absence. `Ensure` preserves an existing enrollment; a corrupt record fails the fence closed; the daemon re-reads on every ask, so `leave` and a live sync move health/`initial_state`/the fence without a restart. The flock discipline covers daemon-starting-while-home-enrolls, and the first-contact bug (enrolling into a data dir that doesn't exist yet) is fixed and pinned by a test that walks the whole sequence — daemon id minted after the record, enrollment kept. Both `daemon-id` and `enrollment.json` at 0600, atomic rename, lock files never unlinked.
- **The hub leg is the right shape.** Enroll before the daemon starts, so a first boot already knows whose outpost it is. Refusal (exit 3) stops the sync with the remote's own words; any other non-zero exit is logged and does not block — enrollment is a record, not a precondition for sessions. The foreign-home notice on connect is a log line, not an error, and it says how to move the machine. The ssh-shim tests drive real exit codes, and the CLI's stdout/stderr split (JSON machine-readable, refusal wording human) is exactly the contract the hub reads back over ssh. `homeDaemonID` as a func, not a value, because the record changes under a live daemon — that detail is what keeps `leave` and re-sync honest.
- **Protocol 229 is lockstep in all five surfaces** — constants, tsp schema, Go generated, TS generated, socket hook — and the version-match test passes. vet clean. Full Go suite green locally except the two plugin tests, which fail identically on base main — pre-existing environmental pair, not this PR. Frontend and Generated types CI green on this head.

The finding — this is the whole case:

**CI is red on this head.** The Daemon job dies at "Check formatting", and `gofmt -l` flags exactly one line:

`internal/hub/enrollment_test.go:76`
```go
if !log.contains("enrolled fake-target as an outpost of "+testHomeDaemonID) {
```
gofmt wants the spaces around the `+`:
```go
if !log.contains("enrolled fake-target as an outpost of " + testHomeDaemonID) {
```
One `gofmt -w` and the gate goes green — the re-run in flight now will trip on the same line. That's the only red I found, and nothing else implicates the PR.

One question, not a blocker: `Enroll` treats a record naming the daemon itself exactly like no record at all — both adopt silently. Right for a fresh install, and the test pins it. But when the garden lands (stage 3), a machine that has been running as its own home carries real state, and a sync from another home fences that state off without a decision on that machine. Bounded today — the state stays on disk, `attn enrollment leave` undoes it, and the uplink is the intended answer — but it's the corner where "never silent adoption" and "a standalone home enrolls" will meet. Worth a sentence in the plan when stage 3 opens.

Verdict: request changes for the formatting line — it's one space, it's the merge gate, and this merges right after.

figgyster (approved):
Both loop-closers verified on the merge candidate:

- **The gofmt fix is in.** `internal/hub/enrollment_test.go:76` now carries the spaces gofmt wants — and it landed at 19:48, before my review even posted, so your own gate caught it in parallel with mine. `gofmt -l .` on the new head: empty. The Daemon job's one red is gone.
- **The stage-3 corner is now a decision point in the plan**, and you hardened my version: the sync must say what it is fencing *before* it does, not after — the wording deferred to when the stage opens, the principle recorded now. That's the right place for it.

Hub enrollment tests pass on the new head; the rest of the suite carries from the previous head (this delta is one formatted line plus the plan doc — no production code moved). CI is running on this SHA.

Approved — this is the merging SHA.

PR #853, opened 2026-08-11
feat(prose): report what to simplify in Markdown prose

victorarias:
Slice 1 of [the prose density gate](docs/plans/2026-08-10-prose-density-gate.md):
the deterministic layer and the CLI in front of it. The model judge (slice 2)
and the ticket wiring (slice 3) are not here, and nothing in the finding shape
or the output has to move when they arrive — a `layer` field already tells a
judge finding from a deterministic one.

```
$ attn prose check docs/plans/2026-04-16-plugin-system.md
docs/plans/2026-04-16-plugin-system.md:114:78: passive-no-actor: "must be created" tells someone to act without saying who — name the actor
docs/plans/2026-04-16-plugin-system.md:339:127: passive-no-actor: "should be designed" tells someone to act without saying who — name the actor
docs/plans/2026-04-16-plugin-system.md:364:75: passive-no-actor: "should be designed" tells someone to act without saying who — name the actor
prose check: 3 findings in 1 file

$ attn prose check docs/plans/2026-08-10-prose-density-gate.md
docs/plans/2026-08-10-prose-density-gate.md:45:5: hidden-verb: "give consideration" buries the verb in a noun
    use "consider"
prose check: 1 finding in 1 file
```

Findings, never a score: a quoted span, the rule that objected, what to do
instead, at a `file:line` you can jump to. `--json` hands an agent the same
array. `--deterministic-only` is accepted and inert until the judge lands, and
the help says so rather than letting a flag quietly do nothing.

## The engine decision

The slice was asked to decide between embedding Vale and writing our own engine,
from evidence. Vale settles it by not being importable at all — every package in
`errata-ai/vale/v3` is under `internal/`, so `import ".../internal/core"` is a
compiler error, not a trade-off. What was left was vendoring the module (111
modules in its closure, plus a toolchain bump from 1.25.3 to 1.25.7) or shelling
out to the binary, which the plan's no-sidecar rule forbids. Neither buys the
rules that carry this feature: idea density, staccato runs, length variance and
referent overlap are computed, and no Vale rule type expresses any of them.

So: our own engine, keeping the one piece of Vale worth keeping — its Vocab file
format, so `docs/prose/vocabulary/{reject,accept}.txt` is a format people already
know how to write. The tagger is `jdkato/prose/v3` (+4.9MB, 2 deps, 17.6ms to
load models once per process) and the Markdown parser is `goldmark` (+0.6MB, no
deps). Both carry byte offsets end to end, which is what makes a finding point at
real bytes instead of at a guess. The `attn` binary goes 40.7MB → 45.8MB, +12.7%.
Receipts are in the plan doc's Decisions section.

## Calibration, and the one result worth reading

Every number was measured before it was written down
(`go test ./internal/prose -run TestCalibrationReceipts -v`). Healthy is the 78
plan and vision documents on main, 162k words of prose Victor accepted. Dense and
swept are the two sides of the #818 comment sweep, 36k and 15k words, checked
into `testdata/corpus/`.

Every numeric tripwire sits one step past the healthy maximum, and all six are
silent on the whole healthy set — that is what a tripwire means here, and
`TestHealthyCorpusStaysQuiet` is the guard. Two rules turned out to be *wrong*
rather than miscalibrated and were narrowed against the corpus rather than
whitelisted: actor-hiding passive fired 1038 times on accepted prose until it was
scoped to obligation modals (where the reader genuinely cannot tell whose job it
is), and hidden verbs driven by noun suffixes reported "takes an extension", so
the rule is now a curated noun→verb table plus a required preposition, and every
finding names the verb to use.

The result slice 2 needs: **the #818 pair does not separate on any per-sentence
measure.** Density p50 0.48 vs 0.47, sentence length max 78 on both sides, noun
runs identical — across a sweep that deleted 57% of the words. What #818 removed
was whole paragraphs of retelling from a surface whose standard is one or two
lines. That is volume against purpose, and no per-sentence metric can see it. The
pattern rules do separate the pair (expletive openers 0.82 vs 0.00 per 1000
words). So the numeric rules are runaway-catchers, the judge is what has to catch
#818-style density, and it now has a labelled corpus to be measured on.

One correction to the plan doc, made in the same commit: it claimed chopping a
sentence "deletes no propositions, so this number cannot be gamed by
fragmenting". Three measured rewrites lost 0.03–0.09 density, because the
conjunctions a split deletes *are* propositions. Weak escape, not a free one —
against 0.17 of headroom, and the cohesion rules measure exactly what the split
removed.

## Verification

Cheapest tier applies and I am stating why: `internal/prose` and `cmd/attn/prose.go`
are self-contained: no daemon, no protocol, no app, no PTY, no background runner.
The built binary was exercised against real repository documents.

- `attn prose check docs/plans/*.md docs/vision/*.md` → **76 findings across 78
  files in 0.55s**: 49 expletive openers, 25 obligation passives, 2 hidden verbs,
  0 from the word list, 0 from any numeric rule. Every one I read is defensible;
  one of the hidden-verb hits is the plan doc quoting the rule's own example back
  at it, which is the gate working.
- Also exercised: a single file, stdin (`-`), `--json`, `--rule`, `--vocab ""`,
  and `attn prose rules`.
- `make test` green. `make build-linux-amd64` produces a valid ELF x86-64 binary.
  `gofmt`, `go vet`, `go mod tidy` clean.

Tests include golden-output fixtures for the findings and for the anti-gaming
cases — a wall of prose, and a staccato fixture that a naive length rule would
bless — plus a per-rule precision table over the corpus, so a rule that starts
firing on accepted prose fails the build rather than the reviewer.

Markdown shapes that are not prose are never read: code fences, mermaid, front
matter, HTML blocks, tables, and link targets. A code span collapses to a single
noun so a path does not read as a five-noun pile-up, and it still maps back to the
file so the finding quotes what is actually written there.

https://claude.ai/code/session_01Pbc9nCXab6CLKrRbBo2a4h

### DIFF
--- a/internal/prose/rule_structure.go
+++ b/internal/prose/rule_structure.go
@@ -179,12 +179,42 @@ func (expletiveRule) check(doc *Document, _ Thresholds) []Finding {
 
 // clauseMarkerFollows separates the expletive "it" ("it is clear that …") from
 // the pronoun "it" ("it works"), which is fine.
+//
+// The marker has to be the complement of the opening "is", and two things keep
+// it there. A finite verb ends the scan, because it carries a clause with its
+// own subject and anything past it belongs to that clause. And "to" counts only
+// when an infinitive follows it, which is what the expletive always takes ("it
+// is time to go"); a stranded preposition is not a marker. Without both, a
+// sentence's last word decided the rule — "it is slower than the path we
+// compared it to" opens on a real pronoun and fired anyway. Participles and
+// gerunds do not end the scan: they are still the opening predicate ("it is
+// claimed that …", "it is worth checking whether …").
 func clauseMarkerFollows(tokens []Token) bool {
-	for _, t := range tokens[2:] {
+	for i, t := range tokens[2:] {
 		switch lower(t) {
-		case "that", "to", "whether", "which":
+		case "that", "whether", "which":
 			return true
+		case "to":
+			if infinitiveFollows(tokens[2+i:]) {
+				return true
+			}
 		}
+		switch t.Tag {
+		case "VBZ", "VBD", "VBP":
+			return false
+		}
+	}
+	return false
+}
+
+// infinitiveFollows reports whether "to" heads an infinitive, reachable across
+// the adverb that can split it.
+func infinitiveFollows(tokens []Token) bool {
+	for _, t := range tokens[1:] {
+		if isAdverb(t) {
+			continue
+		}
+		return t.Tag == "VB"
 	}
 	return false
 }
--- a/internal/prose/rule_structure.go
+++ b/internal/prose/rule_structure.go
@@ -1,0 +1,214 @@
+package prose
+
+import (
+	"fmt"
+	"strings"
+)
+
+// nounStringRule reports a pile-up of consecutive nouns. Each one modifies the
+// next, so the reader holds the whole stack before learning what the phrase is
+// about. Unpacking one into a verb or a preposition costs a word and buys the
+// sentence back.
+type nounStringRule struct{}
+
+func (nounStringRule) name() string { return "noun-string" }
+
+func (nounStringRule) check(doc *Document, th Thresholds) []Finding {
+	var out []Finding
+	for _, sent := range doc.Sentences {
+		// A heading is a label, and a label is a noun phrase by construction.
+		if doc.Blocks[sent.Block].Kind == KindHeading {
+			continue
+		}
+		// Punctuation is what ends a noun string, so this reads the raw token
+		// stream. Over words alone, the comma-separated list "state, input,
+		// tool_call" reads as three nouns in a row, which is a list rather
+		// than a pile-up and is exactly what a reader finds easy.
+		tokens := sent.Tokens
+		for i := 0; i < len(tokens); {
+			j := i
+			for j < len(tokens) && isNoun(tokens[j]) {
+				j++
+			}
+			if run := j - i; run >= th.NounStringLength {
+				span := doc.Source[tokens[i].Start:tokens[j-1].End]
+				out = append(out, doc.finding(
+					"noun-string", tokens[i].Start, tokens[j-1].End,
+					fmt.Sprintf("%d nouns in a row (%q) — the reader holds the whole stack before the phrase resolves; turn one into a verb or a preposition", run, collapseSpace(span)),
+					"",
+				))
+			}
+			i = max(j, i+1)
+		}
+	}
+	return out
+}
+
+// passiveRule reports an instruction whose actor is nowhere in the sentence.
+//
+// The passive itself is not the defect. "The archive is verified against the
+// lock" is right when the verifier is beside the point, and 1038 of those sit
+// in prose Victor accepted — a rule against them reports the corpus, not a
+// problem. What costs the reader is the passive that tells someone to do
+// something and never says who: "the profile must be cleaned up" leaves them
+// unable to tell whether it is their job. The obligation is the signal, so the
+// rule fires on a modal governing a passive and on nothing else.
+type passiveRule struct{}
+
+func (passiveRule) name() string { return "passive-no-actor" }
+
+// obligationModals say someone has to act. "can" and "may" describe what is
+// possible, which is a property of the thing rather than a job for a person,
+// so they are not here.
+var obligationModals = map[string]bool{
+	"must": true, "should": true, "shall": true, "ought": true, "needs": true,
+}
+
+// adjectivalParticiples read as descriptions of a state rather than reports of
+// an action, so a missing actor costs the reader nothing even under a modal.
+var adjectivalParticiples = map[string]bool{
+	"done": true, "gone": true, "known": true, "meant": true, "supposed": true,
+	"required": true, "expected": true, "based": true, "related": true,
+	"limited": true, "interested": true, "involved": true, "concerned": true,
+	"tied": true, "bound": true, "aware": true, "worth": true, "left": true,
+}
+
+func (passiveRule) check(doc *Document, _ Thresholds) []Finding {
+	var out []Finding
+	for _, sent := range doc.Sentences {
+		tokens := sent.words()
+		for i, t := range tokens {
+			if t.Tag != "MD" || !obligationModals[lower(t)] {
+				continue
+			}
+			beAt, ok := beFormAfter(tokens, i)
+			if !ok {
+				continue
+			}
+			participle, at, ok := participleAfter(tokens, beAt)
+			if !ok || adjectivalParticiples[lower(participle)] {
+				continue
+			}
+			if agentFollows(tokens, at) {
+				continue
+			}
+			span := doc.Source[t.Start:participle.End]
+			out = append(out, doc.finding(
+				"passive-no-actor", t.Start, participle.End,
+				fmt.Sprintf("%q tells someone to act without saying who — name the actor", collapseSpace(span)),
+				"",
+			))
+		}
+	}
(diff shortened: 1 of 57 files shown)


internal/prose/rule_structure.go
```diff
@@ -0,0 +1,214 @@
...
+			"",
+		))
+	}
+	return out
+}
+
+// clauseMarkerFollows separates the expletive "it" ("it is clear that …") from
+// the pronoun "it" ("it works"), which is fine.
+func clauseMarkerFollows(tokens []Token) bool {
+	for _, t := range tokens[2:] {
```
figgyster: Non-blocking, for slice 2: this scan walks the entire remainder of the sentence, so the impersonal-"it" family trips the rule — "It is time to go" and "It is safe to say the daemon handles it" both fire "opens on a placeholder". The advice is right for the first and wrong for the second. The rule is rate-bounded by design, but this family never appeared in the corpus, and the judge in slice 2 will inherit these calls as training signal. Worth pinning in the precision table before then — a spared entry for the infinitive-complement form, or an accepted rate.

figgyster (approved):
Approve. Slice 1 of the density gate lands with receipts attached, and I went and re-ran the receipts before believing them. They hold.

The engine decision is the right kind of dead end: Vale isn't a trade-off you decline, it's a compiler error (`internal/` packages). And the carry-over call — Vale's Vocab file format, not its rule DSL — is the part actually worth keeping, because the rules that carry this feature (density, staccato, cohesion) are computed and no Vale rule type expresses them. The +4.9MB / +12.7% binary with a measured 17.6ms model load is the boring, documented price of a tagger that carries byte offsets end to end.

What I verified on the head (82f50f48):

- `go test ./internal/prose ./cmd/attn` — 245 tests, all green. `gofmt -l`, `go vet`, and `go mod tidy` (go.mod/go.sum unchanged) all clean.
- Calibration receipts reproduce the plan's tables exactly: healthy 78 docs / 162,389 words, sentence length p50 14 / p99 53 / max 115, density p50 0.47 / p99 0.68 / max 0.83, noun-run max 6, paragraph CV min 0.17, staccato ≤9-word run 4, lost-thread 5. Every tripwire sits one step past a measured maximum.
- The #818 claim is stated the way it measures: the dense/swept pair does *not* separate on any per-sentence metric (density p50 0.48 vs 0.47, max 78 on both sides, noun runs identical), and the pattern rules do (expletive openers 0.82 vs 0.00 per 1000 words). That is the receipt slice 2 needs, and it's now checked in as a labelled corpus.
- The in-commit correction — chopping a sentence loses 0.03–0.09 density, so the plan's "cannot be gamed" claim was wrong — is measured, stated, and pinned by a test bounding it at 0.12. That's the receipts-over-assertions loop working on the plan's own prose.
- CLI: both PR-body examples reproduce byte-for-byte (same file:line:col, same suggestion lines, exit 1). Full-repo sweep: 76 findings in 78 files in 0.77s here — 49 expletive openers, 25 obligation passives, 2 hidden verbs, zero numeric, zero word-list. stdin, `--json` (layer field present), `--rule`, `--vocab ""`, `prose rules` all behave; exit codes 0/1/2 as documented.
- Full Go suite green except the two daemon plugin tests (`TestPluginDriverEndToEnd_InstalledProcessLaunchReportAndResumeThroughWorkerPTY`, `TestDaemon_StartInstalledPlugins_SpawnsProviderPlugin`) — the known environmental pair; identical failures on origin/main, not this PR.
- CI on the head: all 12 check runs complete, SUCCESS or path-filtered SKIP.

One non-blocking finding, and the only thing I'd add before slice 2:

`expletive-opener`'s clause-marker scan walks the whole rest of the sentence, so the impersonal-"it" family trips it: "It is time to go" and "It is safe to say the daemon handles it" both fire "opens on a placeholder". For the first, the advice is fine. For the second, "start with the subject" produces a worse sentence than the original. The rule is rate-bounded by design (0.30/1000 words on healthy), so this is within the stated philosophy — but it's exactly the precision edge the corpus never sampled, and slice 2's judge will inherit the rule's calls as training signal. Pin the family in the precision table (a spared entry for the infinitive-complement form, or an accepted rate) before the judge lands.

And the load-bearing result deserves saying plainly: the numeric rules are silent on *both* genres — the docs their thresholds are set past, and the swept comment corpus, which is the terse genre closest to ticket prose, what slice 3 actually gates. That two-genre silence is the strongest calibration evidence in this PR, and the corpus README's transparency about the sample (36.6k of 98k dense words, the extraction method, the commit receipts) is the right shape for evidence this load-bearing.

internal/prose/rule_structure.go
victorarias: Fixed in 0fe2ad32 — you found a bigger hole than the two examples. The scan took any later marker, so a stranded preposition decided the rule: "It is slower than the path we compared it to" and "It is fine on every platform I ported it to" both opened on a real pronoun and fired on their last word.

The marker now has to be the opener's own complement. A finite verb ends the scan (it carries a clause with its own subject), and "to" counts only when an infinitive follows it, which is what the expletive always takes. Participles and gerunds still pass, so "It is claimed that …" and "It is worth checking whether …" keep firing.

That drops 10 of the 49 expletive findings on the healthy corpus (49 → 39, 0.30 → 0.24 per 1000 words); the dense/swept separation survives at 0.33 vs 0.00, and the plan doc receipts are updated to the new numbers.

The impersonal-"it" family is pinned in the precision table in both directions, as you asked. I kept "It is safe to say the daemon handles it" and "It is time to raise the threshold" as caught, with the reasoning in the test: both are hedges the reader can lose, and "the daemon handles it" says the same thing sooner. The two stranded-preposition sentences are spared. So slice 2's judge inherits calls that were checked rather than assumed.

figgyster (approved):
Loop-closer approved, and it's better than the ask. I flagged the whole-sentence scan for the impersonal-"it" family; you found the actual hole behind it — the scan took any later marker, so a stranded preposition decided the rule and a real pronoun fired on its last word. The fix (marker must be the opener's own complement; finite verb ends the scan; "to" counts only with an infinitive following) is the right shape: precise at the boundary, still firing on "It is claimed that …" and "It is worth checking whether …" where the advice is right.

Verified on the new head:

- `go test ./internal/prose ./cmd/attn` green, new precision entries included. gofmt, vet, `go mod tidy` (go.mod/go.sum untouched) all clean.
- Receipts reproduce the updated plan numbers: healthy expletive findings 49 → 39 (0.30 → 0.24 per 1000 words), dense 0.33 vs swept 0.00 — the #818 separation survives the narrowing. Dense corpus: 12 expletive findings, swept: 0.
- Full-repo sweep: 66 findings in 78 files in 0.77s — 39 expletive openers, 25 obligation passives, 2 hidden verbs, zero numeric, zero word-list. Matches the plan doc's updated line.
- The four pinned cases behave through the real binary: both stranded-preposition sentences spared (exit 0), both impersonal-"it" sentences caught, "It is claimed that" still fires.
- CI on the head: all 12 check runs complete — Daemon, Linux amd64/arm64, Frontend E2E, react-doctor, Changelog, Changes all success; frontend/tauri/plugins path-filtered skips.

On the one place we disagreed: you kept the impersonal-"it" family caught, and you're right. "It is safe to say X" is the hedge the rule exists for — "the daemon handles it" does say the same thing sooner — and sparing the infinitive-complement form, which is what I floated, would have been the leak. Stranded preposition was the false positive; the family is the target. Pinned both directions in the precision table, so slice 2's judge inherits checked calls instead of assumed ones. That's the loop closed properly.

Re-approving on the merge candidate.

PR #854, opened 2026-08-11
fix(state): read a running turn from the title's shape, not a glyph list

victorarias:
## The bug

A Claude session running a long tool call turns grey. attn reads whether a turn is
running from the spinner glyph the harness paints into the OSC 0 terminal title,
and `classifyClaudeTitle` recognised exactly two things: the Braille Patterns
block (U+2800–U+28FF) as busy, and `✳` U+2733 as not busy. Anything else returned
`ok=false` and produced **no observation at all**.

Claude Code 2.1.228 animates its running glyph with half circles — `◐` U+25D0 /
`◑` U+25D1 — instead of braille. So every session started after that update did
not get a degraded heartbeat, it got no heartbeat: one lonely `not_busy` frame at
boot (`✳` is unchanged) and nothing for the rest of its life.

With the heartbeat gone, the hooks that bracket a tool call are the only evidence
left, and they are silent for the call's whole duration. `Evidence.LastMovement`
therefore freezes for as long as the call runs, and the open-bracket clause in
`sessionstate.Resolve` trips `evidenceStoppedMoving(StuckAfter)` at 90 seconds and
resolves `unknown` / `stuck` — which opens a turn *and* breaks a snooze
(`internal/attention/turn.go`). The session heals itself the moment the call
returns, which is what makes it read as a flake rather than a regression.

Ninety seconds is inside what a test run, a `make test`, or a wait on CI takes
routinely. Observed live on three sessions in one evening on one machine.

## Measurements

Claude 2.1.228 driven through a real PTY (`creack/pty`, trusted cwd), logging
every OSC it wrote:

```
   780ms  OSC0  claim=not_busy   U+2733  "✳ Claude Code"
  13.52s  OSC0  claim=UNMATCHED  U+25D0  "◐ Claude Code"
  14.51s  OSC0  claim=UNMATCHED  U+25D1  "◑ Run measurement script with 150 second delay"
  15.44s  OSC0  claim=UNMATCHED  U+25D0  "◐ Run measurement script with 150 second delay"
  … 170 paints, every ~960ms, unbroken straight through one 150s blocking bash …
```

Two things this pins down:

1. **The title is still painted ~1/s, including during the whole blocking tool
   call.** The evidence attn needed was arriving the entire time; it was thrown
   away at the classifier. The pre-existing "~3.5s of mid-tool silence" receipt
   still holds.
2. **The semantics did not change.** `✳` still means resting. Only the busy
   glyph moved.

Parked-on-approval, measured the same way (no `--dangerously-skip-permissions`):

```
t=14.5s..19.3s   ◐/◑ every ~960ms          (running)
t=20.2s          ✳ "Create temporary marker file"   (parks on the permission prompt)
t=20.2s..2m44s   nothing at all
```

Claude paints `✳` once as it parks and then goes completely silent until the
prompt is answered. Silence is therefore not evidence of death — it also means
parked — which is why the liveness rule below cannot mask an approval: there are
no paints to mask it with, and `harnessEdge` outranks `stuck` by three clauses
anyway.

Correlation across the sessions live on the machine at the time, busy-heartbeat
rows per session over 40 minutes (binary confirmed per process with `lsof -d
txt`):

| claude | busy heartbeats | went `unknown` |
|---|---|---|
| 2.1.226 (×4 sessions) | 327–580 each | no |
| 2.1.227 (×2 sessions) | 990, 1249 | no |
| 2.1.228 (×3 sessions) | **0** | yes |

## The fix

Naming the busy glyph is what went blind, so the rule is now the *shape* of the
title rather than a list:

- `✳` is the resting glyph — the one that has held across releases.
- Any **other status symbol** is a spinner frame, so a glyph Claude has not
  shipped yet still reads as running.
- A title that starts with **no status symbol** (a subprocess overwrote it)
  claims nothing about the agent and is reported as `claimUnclassified` rather
  than dropped: it stamps `LastMovement` and leaves the level alone, because
  whoever painted it was running at that instant — which is the whole question
  the stuck tripwire asks.

That last part is the point of the change beyond tonight's glyph. Reading the
glyph can go wrong; going blind must not follow. Filing an unreadable title as a
*level* instead would be the opposite mistake — it would let any subprocess's
title settle a running turn — so it deliberately carries liveness and nothing
else.

Codex is untouched: its title has no distinct idle glyph and its classifier
already treats anything non-braille as not-busy, which is safe under its own
freshness rule.

## Verification

Unit: the 2.1.228 glyphs classify busy, an unreadable title reports liveness
without claiming a level (`TestAnUnclassifiedTitleMovesEvidenceWithoutClaimingALevel`
asserts the busy level and `LastBusyAt` both survive while `LastMovement`
advances). The pre-existing test that a foreign title produces *nothing* now
asserts it produces liveness — its original invariant, that a foreign title must
never claim a level, is what it still guards. Full Go suite green.

Live, on a throwaway profile installed from this branch (`preflight` PASS), a
real Claude 2.1.228 session driven through one 150-second tool call:

```
[probe] t=+3s    working/bracket_open
[probe] t=+168s  idle/classifier_verdict
[probe] PASS: never reached unknown across the whole tool call
```

and the heartbeat itself is back — **85 busy frames** for that run, appearing in
the trace as `heartbeat busy ×64` unbroken across the call, then `not_busy` on
settle. On the same build before this change the count is 0. Profile cleaned up
after.

No recording: the visible change is a session that *does not* turn grey after 90
seconds, which a 20-second clip cannot carry. The state trace above is the
evidence.

https://claude.ai/code/session_01QBGiYJMT6NPQv8uQh1JUA9

### DIFF
--- a/changelog.d/heartbeat-survives-a-changed-spinner-glyph.yaml
+++ b/changelog.d/heartbeat-survives-a-changed-spinner-glyph.yaml
@@ -1,0 +1,21 @@
+kind: fixed
+area: daemon
+change: >
+  A Claude session running a long tool call no longer goes grey. attn reads
+  whether an agent is running from the spinner glyph the harness paints into the
+  terminal title, and it recognised only the braille spinner Claude used up to
+  2.1.227. Claude 2.1.228 animates half circles instead, so every session
+  launched after that update lost its heartbeat entirely and left the hooks that
+  bracket a tool call as the only sign of life — and those are silent for the
+  whole call. Any tool call longer than ninety seconds, which a test run or a
+  wait on CI passes easily, therefore read as a session that had stopped
+  explaining itself: attn painted it unknown and asked for attention it did not
+  need, then recovered the moment the call returned. Whether the turn is running
+  is now read from the shape of the title rather than a list of glyphs, so a
+  spinner Claude has not shipped yet still reads as running, and a title attn
+  cannot read at all still counts as the agent being alive instead of counting
+  as nothing.
+symptom: >
+  A session turned grey with "unknown" and asked for attention while its agent
+  was plainly working, then went back to green on its own a few minutes later.
+  Sessions started before the Claude Code 2.1.228 update were unaffected.
--- a/docs/plans/2026-07-25-agent-state-detection.md
+++ b/docs/plans/2026-07-25-agent-state-detection.md
@@ -66,6 +66,24 @@ codex:   ESC ] 0 ; ⠸ attn--fix-state-detec... BEL       <- spinner prefix = bu
   running. A level that stops arriving cannot get stuck — though the spike below
   shows it goes briefly silent mid-tool, which is why it corroborates rather than
   leads.
+
+  Re-measured 2026-08-11 on claude 2.1.228 through a real PTY: the running glyph
+  is no longer braille but `◐`/`◑` (U+25D0/U+25D1), `✳` is unchanged, and 170
+  paints arrived ~960ms apart across one 150s blocking tool call with no gap at
+  all. Recognising the running glyph by name is what went blind on that release —
+  every session started after it lost its heartbeat, leaving the hooks that
+  bracket a tool call as the only evidence, and a tool call longer than
+  `StuckAfter` read as a session that had stopped explaining itself. Claude is
+  now read by the *shape* of the title: `✳` is the resting glyph and any other
+  status symbol is a spinner frame, so a glyph claude has not shipped yet still
+  reads as running. A title with no status symbol claims nothing and is reported
+  as liveness alone, because whoever painted it was running at that instant —
+  which is the whole question the stuck tripwire asks.
+
+  Parked-on-approval was measured the same day: claude paints `✳` once as it
+  parks and then nothing at all until the prompt is answered. Silence therefore
+  does not mean the agent is gone, and the liveness rule cannot mask an approval
+  — there are no paints to mask it with.
 - **OSC 777 is real but strictly redundant with the `Notification` hook.** Claude
   emits `ESC ] 777 ; notify ; Claude Code ; <message> BEL` — it appears in three
   of the nine captures (`run_approval`, `run_approval2`, `run_claude_fg`), the
--- a/internal/daemon/session_evidence.go
+++ b/internal/daemon/session_evidence.go
@@ -147,6 +147,13 @@ func (d *Daemon) recordPTYEvidence(sessionID string, obs pty.Observation) {
 			})
 			return
 		}
+		// A title nobody can read is still someone painting: it stamps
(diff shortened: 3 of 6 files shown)


figgyster (approved):
Approved. This one's the good kind of bug — a glyph changed upstream, and the fix is to stop trusting glyphs.

**The mechanism checks out end to end.** I walked `sessionstate.Resolve` clause by clause: with the 2.1.228 title dropped at the classifier, a long tool call leaves `TurnOpen` with silent hooks, `evidenceStoppedMoving` trips at 90s, and the session resolves `unknown` — which `internal/attention/turn.go` confirms is in the opens-a-turn set. Your chain from glyph → no heartbeat → frozen evidence → owed turn is exactly right. The correlation table (327–1249 busy heartbeats on 2.1.226/227, zero on 2.1.228) is the kind of witness I actually trust.

**The shape rule is the right shape.** Busy = "a status symbol that isn't ✳" survives the next release by construction instead of by patch. And the unclassified path is the real upgrade hiding in here: `recordEvidence` with an empty mutate still stamps `LastMovement` (`updateIf` does that for every write), so a subprocess's title carries liveness and nothing else — the level stands, `LastBusyAt` stands, and the stuck tripwire can't fire. I verified the resolver ordering: the restored busy heartbeat outranks every bracket clause (clause 2), and unclassified liveness only defuses `evidenceStoppedMoving` — the BracketStale settle after StaleAfter+SettleGrace is still the designed escape hatch. "Reading the glyph can go wrong; going blind must not follow" — that principle is actually implemented, not just stated.

**Tests are witnesses, not alibis.** `TestAnUnclassifiedTitleMovesEvidenceWithoutClaimingALevel` pins the exact property (busy level and LastBusyAt survive while LastMovement advances), the 2.1.228 glyphs are in the classifier table, and the foreign-title test's original invariant — never claim a level — is what it still guards. Empty title still means nothing witnessed. Parked-on-approval measurement closes the obvious objection to the liveness rule: silence isn't masked because there are no paints to mask, and `harnessEdge` outranks stuck anyway.

**Verified locally:** diff check clean, gofmt clean, focused pty/daemon tests pass, full Go suite green except the known environmental plugin pair — reproduced on base with identical signatures, not this PR. CI's daemon job (gofmt/vet/test) will run via the `internal/**` filter; Changelog fragment is in place.

One observation, no ask: `unicode.IsSymbol` is deliberately broad — Sm/Sc/Sk included, so a symbol-first subprocess title ("$", "→", "✓") reads busy. That's the false-busy direction, which is the safe one to err in, and it's consistent with the shape rule as stated. Just noting the breadth is priced in, not accidental.

The session that goes grey mid-`make test` was costing real attention; this closes it and buys insurance against the next glyph change. Good case. Case closed.

PR #855, opened 2026-08-11
feat(queue): keep the next turn before it starts settling

victorarias:
`Cmd+.` only reached an auto-settle that was already counting down. Pressed
before that — while still typing the message that sends an agent back to work,
which is exactly when the user knows this turn should stay owed — it had
nothing to act on. The user had to wait for the countdown to appear just to
call it off.

## What changes

A `Cmd+.` with nothing pending now records a **standing dismissal** on the
focused session: its next auto-settle does not run. The pane header carries
`Turn kept` in the countdown pill's own slot for as long as the answer stands.
Pressing again, or clicking the chip, undoes it and hands the session back to
the ordinary rule.

Pressing it while a countdown *is* running still stops that countdown, exactly
as before. A press when there is nothing to dismiss at all — the feature
disabled, a shell, a chief-of-staff session — arms nothing: a chip promising to
stop a settle that was never coming is worse than no chip.

The dismissal is spent on one stretch of work, not held forever. It records
whether the `working` stretch it covers has begun, so a state re-report cannot
re-arm what the user just called off, and it retires when that stretch ends —
the turn after that is a fresh decision.

Auto-settle only. A pre-emptive press does not pre-suppress ticket doorbells;
`Cmd+.` still cancels a nudge that is actually counting down.

## Along the way

The auto-settle path judged `turn_owed` from a partially decorated session, so
the chief-of-staff exclusion never applied to it — a chief session could have
its turn auto-settled. It now reads the same decoration a broadcast carries.
`attention.Owed` gained a sibling, `attention.Excluded`, because a caller asking
"could this session ever owe a turn?" is asking only that half.

## Verification

turn-kept

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/1cf10494deb87dc19c1729b4d81f21826bec9a97/feat-keep-next-turn/turn-kept-trim.mp4)

The clip is the live run below: the countdown appears, a real `Cmd+.` replaces
it with `Turn kept`, and a second real `Cmd+.` — with nothing counting down —
clears the chip and lets the settle re-arm.

Packaged app on a throwaway profile, real CGEvent `Cmd+.` through the native
menu (`real-app:scenario-countdown-cancel`, extended by this PR):

- `Cmd+.` with a countdown visible stops it, the turn stays owed, and
  `auto_settle_dismiss_armed` is `true` on the wire.
- A **second** real `Cmd+.` with nothing counting down clears the flag and the
  settle re-arms on its own, with no state change to trigger it. That press is
  the one that reached nothing before this change: `App` unregistered the
  shortcut handler whenever no countdown was visible.
- The nudge legs are unchanged — a cancelled nudge stays cancelled across a
  selection round trip, the ticket stays unread, and new activity re-arms it.

The scenario's fixture prompt went from "count to 200" to "count to 2000" in
the same commit. The old one finished in ~5s, so the 30s arm delay could land
while a classifier verdict was still pending and the timer was dropped —
pre-existing fragility, unrelated to this change.

Also: full `internal/daemon`, `internal/attention`, `internal/protocol` suites;
the frontend suite; `tsc --noEmit`. Both new gates were mutation-tested —
removing the "is there anything to dismiss?" guard and weakening the retire
condition each fail at the intended assertion.

## Two things this PR picked up on the way

**`pr-evidence.sh publish` exited 1 after succeeding.** Its cleanup trap
referenced a `local` that was already out of scope when the trap fired, so under
`set -u` it died on its own first line — after the files were pushed and the
markdown printed. It also leaked its temp clone every run. Fixed, plus the first
test for that script: `gh` and `ffmpeg` are stubbed, the publish goes to a local
bare repo, and it asserts the exit code, the markdown, the pushed files, and
that nothing is left in TMPDIR. Two BSD-only calls on that path became portable
so the test can run in CI at all. The evidence GIF above was published by the
fixed script.

**Harness scenarios were burning the most expensive model.** Nothing pinned a
model, so every agent scenario inherited the machine owner's own default —
fable, the most expensive available — to count to 2000. Unpinned is not
neutral, so the fix is central rather than per-scenario:
`launchFreshAppAndConnect`, which every scenario goes through, pins
`default_model_<agent>` to the cheapest model that agent offers as soon as the
daemon socket is up, and puts back what it found as the process ends. No
scenario has to remember anything, and no fixture's timing depends on which
model happened to run it.

The restore goes straight to the daemon over its own short-lived socket, not
through the app: scenarios tear down in whatever order suits them — some through
the runner's cleanup registry, some in their own `finally` — and by then the app
is usually gone, while the daemon outlives every launch. `dev` is a profile the
maintainer also uses by hand, so a run that leaves it pinned is a bug, not a
saving. `ATTN_HARNESS_LAUNCH_MODEL_<AGENT>` overrides one agent for one run;
`inherit` leaves the setting alone and has to be asked for by name.

Verified live on a packaged run seeded to `opus`, which is also what the
`[harness]` lines added here make visible in any run:

```
[harness] pinned default_model_claude=haiku (was opus)
[harness] pinned default_model_codex=gpt-5.4-mini (was unconfigured)
ATTN_VERDICT {"ok":true,...,"failureCount":0,...}
[harness] restored default_model_claude=opus
[harness] restored default_model_codex=unconfigured
```

The daemon, queried independently afterwards, holds `opus` again, and its own
log shows the pin landing before the `spawn_session` it applies to. Unit tests
cover the pin, the override, `inherit`, the already-pinned no-op, and that a
relaunch restores the value the run started with rather than the harness's own.

## Surfaces

- **Protocol** — `auto_settle_dismiss_armed`, `ProtocolVersion` 229 → 230 in
  both lockstep spots. #854 landed on main at 229 while this branch was open, so
  the rebase merged the bump down to an exact string match — same version
  number, two different schemas. Renumbered on top and the types regenerated
  from the merged schema (byte-identical output, so the rebase itself was sound).
- **CLI / SDK / plugins** — nothing consumes `cancel_countdown`; there is no
  other way in to extend.
- **Remote** — `CmdCancelCountdown` already routes to the owning daemon.
- **Docs** — glossary entries for auto-settle and the standing dismissal;
  changelog fragment; alignment brief under `docs/alignment/`.

https://claude.ai/code/session_01KnF5yxg6GXpZM2nEerRxxC


figgyster (approved):
**Approve.**

The old press was a dead letter at the exact moment it mattered: `Cmd+.` unregistered itself whenever nothing was counting down, so the press a user makes *while typing the steer that starts the stretch* — the one moment they actually know they want the turn kept — reached nothing. This PR makes the same key a standing dismissal of the next auto-settle on the focused session, and the machinery is as careful as the pitch promises.

The load-bearing idea is the two-valued map: `autoSettleDismissals[id]` records whether the `working` stretch it covers has begun. Retire only fires on a covered stretch, so the resolver re-reporting `waiting_input` between the press and the steer cannot drop the answer in that gap — the window is pinned by `TestAutoSettle_ArmedBeforeTheSteerDismissesTheNextSettle`. Disarm is a real round trip: press again, or click the chip, and a mid-stretch disarm re-arms the countdown (`TestAutoSettle_DisarmingWhileWorkingReArmsTheSettle`), so the session is not left in a limbo no state change reaches. Feature-off and outside-the-queue sessions arm nothing, disabling the feature clears standing dismissals with the timers, and the one-broadcast-per-press discipline survives the refactor.

The quiet good part is the fix along the way: `turnOwed` now reads the same decorated clone a broadcast carries, so the auto-settle path finally sees the chief-of-staff exclusion it was judging blind — a chief session could have had its turn auto-settled. `attention.Excluded` as a sibling of `Owed` is the right seam. Both lockstep spots sit on 230, the types were regenerated from the merged schema, and the lockstep test is green.

Verified on the head, not just read: attention and the full daemon package, protocol, `-race` over the auto-settle/cancel-countdown paths, and the three frontend files including the new `App.cancelCountdownTarget.test.tsx` contract tests — 26/26 green. CI is green all the way down: Daemon, Frontend, the E2E shards, Generated types, Bench A/B, and this time Plugins and Tauri ran too (the script changes trip wider filters) and passed. The real-app harness leg for the second press — the one that used to reach nothing — is the right witness. The mutation-tested claims line up with what the tests actually assert.

Two commits landed while I was verifying, and they only sharpen the case. `f8f12cec` fixes the evidence-publish script this very PR's video rode in on: the EXIT trap read a `local` that was already out of scope, so a successful publish exited 1 and leaked the temp clone. Now pinned by `pr-evidence_test.sh` — fake gh/ffmpeg, a private TMPDIR, leaked dirs counted — wired into `make test-scripts` and CI. I ran it: 8/8. `e1be91ab` pins the fixture to haiku with a measured rate (77 numbers/s, so 2000 buys ~26s of working), restores the setting on cleanup, and records the rule in the harness's own AGENTS.md so the next scenario does not quietly burn the most expensive model on a counting fixture.

Three things worth naming, none blocking:

1. **The dismissal is in-memory, like the timer it answers.** A daemon restart mid-stretch forgets the keep, and the recovery re-report of `working` re-arms a fresh countdown for the stretch the user said keep. Same class as the old suppression, so no regression — but this feature's whole point is that the answer outlives the countdown, and a restart is one of the few things that still kills it. Worth a persisted flag or a recovery re-check when the restart path next gets touched.

2. **The chip is the only surface, and only in session view.** Grid tiles are bare terminals and the sidebar carries nothing for an armed session (the alignment brief defers that), so a keep armed in grid view is invisible until you are back in session view; pressing again is the only way out. Documented deferral — on the record so it stays a decision, not an accident.

3. **The press is now a compound answer.** A session with a nudge counting down and no auto-settle timer pending: the same press that silences the doorbell also arms a dismissal for the next stretch — the daemon cannot tell why the press came. The timing usually coincides with the feature's moment (pre-steer), and the chip makes it visible and undoable, but it is a side effect the alignment brief does not spell out for the nudge leg.

The settle machinery keeps growing into a rule engine with receipts for every clause. This one earns its place in it.

figgyster (approved):
**Approve (merge candidate).**

My earlier approval covered `e1be91ab` as documentation; this head turns that doc into enforcement, and the enforcement is as careful as the doc promised. `launchFreshAppAndConnect` — the door every scenario walks through — now pins `default_model_claude=haiku` / `default_model_codex=gpt-5.4-mini` off the catalog's cheap entries, restores what it found on the way out through a short-lived daemon socket, and the runner's cleanup drains the same queue for the signal path.

Verified on the head:

- New `launchModelPin.test.mjs` 6/6 — pin-both, restore-what-found, first-pin-wins across a relaunch, per-agent override, `inherit` named not reached-by-omission, no-write-when-already-pinned.
- Whole harness suite: 21 files, 206 passed, 5 skipped — and those 5 skips reproduce on base, so none are new.
- App contract tests (cancelCountdownTarget + SettlingIndicator): 18/18.
- Daemon side traced end to end: `default_model_*` passes `validateSetting` as free-form; `set_setting` is refused without the `workspace_sessions` capability the restore's `client_hello` claims; the restore waits for `settings_updated` with `changed_key`, which `FactSettingChanged` → `projectSettingsUpdated(subject)` emits on every set. The echo it waits on is the echo the daemon actually sends.
- `defaultWSURLForProfile` resolves the harness profile's ws port — the one seam the injectable `write` doesn't cover, and it points at the right daemon.
- `git diff --check` clean on this head; CI green where finished (Changelog, Generated types, Linux amd64/arm64, Plugins, Changes, React Doctor), the rest in flight.

One nit, non-blocking: `scenarioRunner.mjs` cleanup comment says restoring "needs the app still standing" — but the restore opens its own socket to the daemon and the `common.mjs` guide says the opposite (daemon outlives every launch). The ordering is harmless either way; the stated reason just contradicts the mechanism. Worth a one-line correction so the next reader doesn't believe the app is load-bearing here.

My three earlier non-blockers stand untouched: the dismissal is still in-memory (restart forgets the keep), the chip is session-view-only (documented deferral), and the press is still a compound answer on the nudge leg. None of them moved in this delta.

figgyster (approved):
**Approve (merge candidate).**

The rewrite is exactly my one nit from the last pass: the `scenarioRunner.mjs` cleanup comment now says the restore "talks to the daemon on its own socket, so the app being gone by now is fine" — which is the mechanism, not the old rationale. Nothing else in the tree moved; the harness suite and the 6/6 `launchModelPin` tests carry over unchanged, diff check clean. Standing on the earlier verification.

PR #856, opened 2026-08-11
docs(plans): record the rulings on A4's four open items

victorarias:
The A4 exit proof recorded four observations as open product decisions in the plan doc (docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md). Victor ruled on all four on 2026-08-11 after re-running the failure loop live on the merged head. This PR records the rulings next to the observations they answer:

- **Hub sidecar: fix.** Bootstrap uploads `attn-app-runtime` beside the `attn` binary so hub-managed remotes can run apps.
- **Bare rollback: previously-serving**, with the choice named in the output and a loud error when no previous exists.
- **A park survives daemon restarts**; `attn app runtime restart` remains the only unpark.
- **Pinned bus retention: no expiry clock, but never silent** — one warning notification per episode past a measured threshold, marked identically in `attn bus status` and the settings page.

Docs only; each ruling's implementation is in flight separately and carries its own receipts.

### DIFF
--- a/changelog.d/a4-open-items-ruled.yaml
+++ b/changelog.d/a4-open-items-ruled.yaml
@@ -1,0 +1,8 @@
+kind: internal
+area: docs
+change: >
+  The A4 plan records Victor's rulings on the four open items the exit
+  proof left: upload the app-runtime sidecar to hub remotes, make bare
+  rollback return to the previously-serving version, persist a runtime
+  park across daemon restarts, and surface pinned bus retention with a
+  warning notification instead of an expiry clock.
--- a/docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md
+++ b/docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md
@@ -495,6 +495,32 @@ than a defect:
   re-arms. Correct for a deliberate act, worth knowing as a way back that is
   not the restart verb.
 
+### Rulings on the four (2026-08-11)
+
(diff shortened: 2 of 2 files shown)


figgyster (approved):
Approve. Docs-only, and the docs are honest — I checked each ruling against the observation it answers, and all four line up with the receipts sitting right above them in the same file.

- **Sidecar** — the exit proof found `installRemoteBinary` streams exactly one file. Ruling says ship `attn-app-runtime` beside it. Same defect, same fix, named.
- **Bare rollback** — the observation's "two honest readings" gets one honest answer: previously-serving, with the registry recording the pointer at every flip. That's consistent with slice 6's own receipt ("rollback moved a pointer and built nothing"), and failing loud with the version list when there's no previous is the right failure mode.
- **Park** — the observation framed forgetting as "correct for a deliberate act." The ruling reverses that framing deliberately: the park survives, restored before anything can lazy-start the broken host, and `runtime restart` stays the only unpark. And the "no second notification on restore" detail is a direct answer to the #842 noise (seven notifications in fourteen minutes, measured).
- **Retention floor** — the open question was "should a parked runtime start a clock of its own." The ruling answers it without a clock: one warning per episode past a threshold measured from healthy data, marked in both `attn bus status` and the settings page. Visibility instead of a clock — that's the discipline the whole plan has been running on.

Fragment matches house format and the Changelog gate passes; `git diff --check` clean. Implementation carries its own receipts, as promised. Close the case on the docs.

PR #857, opened 2026-08-11
fix(sessions): make the context-window cap outrank the user's own settings

victorarias:
A session's context-window cap did not survive a user who sets the same knob in
their own Claude Code settings. Victor hit this on production attn: a session
pinned to 500,000 tokens was heading for auto-compaction anyway, because
`~/.claude/settings.json` carries `"env": {"CLAUDE_CODE_AUTO_COMPACT_WINDOW": "300000"}`.

## Root cause

attn only put the cap in the environment of the process it spawns. Claude Code
copies **every** settings file's `env` block onto its own `process.env` at
startup — an unconditional `Object.assign`, not a fill-in-the-blanks merge — and
then resolves the auto-compact window from `process.env` before it looks at any
setting. So the user's `env` entry replaces whatever attn exported, and the
session compacts at the user's global number instead of the configured one.

The generated settings file attn already writes and passes with `--settings` is
the one channel that wins: Claude Code applies the scopes in the order
globalConfig → user → project/local → **--settings** → managed, so what lands in
the flag file beats the user's file. The cap now travels there as well as in the
spawn environment.

## Receipts (Claude Code 2.1.227/2.1.228)

Same inherited `CLAUDE_CODE_AUTO_COMPACT_WINDOW=111111`, same
`~/.claude/settings.json` value of `300000`, one headless run each, asking the
agent to print the variable it actually holds:

| launch | value Claude Code holds |
| --- | --- |
| no `--settings` | `300000` — the user's file wins over the spawn environment |
| `--settings` file whose `env` sets `222222` | `222222` — the flag scope wins over the user's file |

The control for the pre-fix shape: taking the settings file this branch's daemon
actually generated, stripping only its new `env` block, and launching with the
cap in the environment alone yields `300000` — the exact defect.

## Live verification

Throwaway profile `capfix1` (`make install PROFILE=capfix1`, preflight PASS,
protocol 229 across CLI/app/daemon), `default_context_window_cap_claude=137000`,
one real Claude session spawned by the daemon into it:

- the generated `--settings` file carries
  `"env": {"CLAUDE_CODE_AUTO_COMPACT_WINDOW": "137000"}` beside its ten unchanged hook events;
- the live session, asked to print the variable, answers `CAPPROBE=137000`
  while the user's global settings still say `300000`.

No recording: nothing about this is visible in the app — the change is which
number the launched agent ends up holding. Profile cleaned after.

## Surfaces walked

- **CLI / launch path** — the cap rides `SpawnOpts` into the one place that
  writes the settings file; `HookProvider.GenerateHooksConfig` now takes the
  whole launch rather than three loose strings, since that file is also how a
  launch knob reaches the agent. Claude is its only implementor.
- **Headless** — unaffected: those runs launch with `--setting-sources ""`,
  which drops the user's settings entirely.
- **Codex** — unaffected: its cap is a `-c model_auto_compact_token_limit=…`
  CLI override, which already outranks `config.toml`.
- **Protocol** — unchanged; no daemon/app message moved.
- **Linux** — `GOOS=linux GOARCH=amd64 go build ./...` green.

## Tests

- `internal/agent`: the generated settings file pins the window when a cap
  applies and writes no `env` block when none does.
- `internal/hooks`: `Generate` emits the supplied `env` and omits an empty one.

### DIFF
--- a/changelog.d/fluffy-otter-context-cap-beats-user-settings.yaml
+++ b/changelog.d/fluffy-otter-context-cap-beats-user-settings.yaml
@@ -1,0 +1,10 @@
+kind: fixed
+area: sessions
+change: >
+  A context-window cap now holds even when ~/.claude/settings.json sets
+  CLAUDE_CODE_AUTO_COMPACT_WINDOW in its env block. Claude Code copies a
+  settings file's env over the environment it was launched with, so the cap
+  attn exported at spawn was silently replaced and the session compacted at
+  the user's global value instead of the configured one. attn now also writes
+  the cap into the settings file it passes to the launch, which Claude Code
+  applies after the user's own.
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -3116,7 +3116,7 @@ func runAgentDirectly(requestedAgent string) {
 		opts.ConfigOverrides = cp.GenerateConfigOverrides(opts)
 	}
 	if hp, ok := agentdriver.GetHookProvider(driver); ok {
-		content := hp.GenerateHooksConfig(sessionID, opts.SocketPath, opts.WrapperPath)
+		content := hp.GenerateHooksConfig(opts)
 		settingsPath, err := wrapper.WriteSettingsConfig(os.TempDir(), sessionID, content)
 		if err != nil {
 			cleanup()
--- a/internal/agent/claude.go
+++ b/internal/agent/claude.go
@@ -139,6 +139,8 @@ func (c *Claude) BuildEnv(opts SpawnOpts) []string {
 	// Cap the effective context window so auto-compaction fires at the configured
 	// threshold. The daemon owns the policy of who gets a cap (chief setting vs
 	// default_context_window_cap_<agent>); this only applies what it resolved.
+	// The spawn environment alone does not settle it — the user's settings can
+	// overwrite this key, so claudeSettingsEnv repeats it in the --settings file.
 	if opts.AutoCompactWindow > 0 {
 		env = append(env, "CLAUDE_CODE_AUTO_COMPACT_WINDOW="+strconv.Itoa(opts.AutoCompactWindow))
 	}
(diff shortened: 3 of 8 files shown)


figgyster (approved):
**Approve.** The cap now travels the one channel that beats the user's own file, and the receipts prove it end to end.

**The case, as I read it:** attn exported `CLAUDE_CODE_AUTO_COMPACT_WINDOW` into the spawn environment, but Claude Code `Object.assign`s every settings file's `env` onto its own process env at startup — so a user's `~/.claude/settings.json` entry silently replaced the cap. You measured exactly that in the control row (user's `300000` beats inherited `111111`), and the fix rides the `--settings` scope, which you measured beating the user's file (`222222` wins). The live profile run closes the loop: real daemon spawn, `CAPPROBE=137000` while the user's global still says `300000`.

**What I verified myself, on the head:**
- The load-bearing channel exists in code: `BuildCommand` appends `--settings opts.SettingsPath` (internal/agent/claude.go:87-88), and cmd/attn/main.go:3119-3126 writes the generated file into that path. The fix isn't a document, it's wired.
- Interface sweep clean: `*Claude` is the only `HookProvider` implementor and main.go the only caller — both moved to the new `SpawnOpts` signature. `WriteHooksConfig` passes nil but has no non-test callers; harmless.
- `go test ./internal/agent ./internal/hooks ./internal/wrapper` green; the new tests pin both directions (capped launch pins the window, uncapped writes no env block). gofmt clean, `git diff --check` clean on the true merge base, 8 files exactly.
- The `cmd/attn` build failure is the known environmental one — `ghostty/vt.h` missing on this box, package untouched by the diff, reproduced on base in previous reviews. Not this PR's body.

**One non-blocker, the kind this repo has taught me to name:** this fix is a behavioral dependency on Claude Code's scope order and env-copy semantics — measured on 2.1.227/2.1.228, and I don't doubt the measurement. But it's the same *class* of dependency that killed the heartbeat when Claude swapped a spinner glyph in #854. The unit tests pin attn's output, not Claude's behavior: if a future release changes the copy semantics or the scope order, this cap reverts to the defect silently, with no witness. I'm not asking for a tripwire — a real-model harness test that asserts a cap against a planted conflicting user setting is exactly the expensive-model burn the harness rules exist to avoid. Just naming the blind spot so it's a known tradeoff, not an accident.

Signal worth recording: the settings file is now a launch-knob channel, not just a hooks carrier — and the interface comment says so, which is the right guardrail for whoever adds the next knob that has to outrank the user.

PR #858, opened 2026-08-11
fix(apps): roll back to the version that was serving, not the id below

victorarias:
Bare `attn app rollback <name>` picked the numerically previous version id. That
is the wrong answer in exactly the situation rollback exists for: an operator who
rolls a broken version off and then applies a fix has that broken version sitting
one id below the pointer, so "back" landed on the thing they were escaping.

Bare rollback now means **the version that was serving immediately before the
current one**.

## How

The registry records what each pointer move replaced. Both writers — apply and
rollback — go through one shared `UPDATE` (`movePointerSQL`) that carries the
outgoing version into `apps.previous_version_id`, so there is no way to move the
pointer without recording it. A no-op move — re-applying byte-identical content,
which lands on the version already current — leaves the record alone, so an idle
dev loop cannot erase the way back.

Rolling back twice returns to where you started, because every move rewrites the
predecessor. That is the consequence of "back means what was serving a moment
ago", and it is what makes rollback undoable. It is stated in `attn app rollback
--help` and tested.

Nothing backfills the column. What was serving before is history the registry
never recorded, and deriving it from ids would reproduce the bug this fixes. An
app carried across migration 103, or one on its first version, is refused with
the situation named and its version list attached.

Explicit-version rollback and every other `attn app` verb are unchanged. No wire
change: `AppRollbackResult.PreviousVersionID` already carries the version the app
was on, which for the bare form is exactly the "serving before" the CLI now
names.

## Live verification

Throwaway profile `rbpick` installed from this branch (`make install
PROFILE=rbpick`), preflight PASS, protocol 229 CLI/app/daemon. An app `rbapp`
built to the shape of the bug: v1 and v2 good, **v3 broken**, rolled explicitly
back off v3, then **v4** applied as the fix.

Bare rollback from v4 — the id below it is 3, the broken one:

```
$ attn app rollback rbapp
rolled app rbapp back to version 1 (2938222d4e6e), which was serving before version 4
  artifact /Users/victor/.attn-rbpick/apps/rbapp/2938…/bundle.js
```

Again — it undoes itself:

```
$ attn app rollback rbapp
rolled app rbapp back to version 4 (f2aa919ce25b), which was serving before version 1
```

The sharpest case. Explicitly moved onto the broken v3, whose numerically
previous id is 2; bare rollback returns to what was actually serving, v1:

```
$ attn app rollback rbapp 3
moved app rbapp forward to version 3 (8113e47271b4)
  was on version 1

$ attn app rollback rbapp
rolled app rbapp back to version 1 (2938222d4e6e), which was serving before version 3
```

The error, on a fresh app with one version and no recorded predecessor (exit 1):

```
$ attn app rollback freshapp
app rollback: daemon error: app rollback freshapp: nothing is recorded as
serving before version 5, so bare rollback has no target; name the version to
roll onto. its versions, newest first: 5 (c6e02eee0039) — current
```

Profile cleaned (`attn profile clean rbpick`) — daemon stopped, bundle and data
dir removed.

## Tests

- Store: every pointer move records what it replaced; a no-op re-apply does not;
  a first apply invents no predecessor.
- Store: a database rewound to the pre-103 schema with app rows in it carries
  them across the migration with the pointer intact and no predecessor. Verified
  non-vacuous — with the migration dispatch disabled the test fails on the
  missing column.
- Daemon: bare rollback skips a version that was rolled off; rolling back twice
  flips between the two most recent serving versions; no recorded predecessor
  refuses loudly with the version list; a no-op re-apply does not strand the app.

## Surfaces

CLI (`attn app rollback` and its help), daemon (`handleAppRollback`), store
(migration 103). No protocol change, no frontend surface — `attn app status`
already renders the serving version and needed nothing. Linux-clean: pure Go,
no build constraints touched.

Migration numbering re-checked against `origin/main` at branch time and after
rebase: main's highest is 102, so 103 is free.

## Hit every surface

If you added a way in, add the way out: bare rollback's way out is bare rollback
— it flips, so there is no one-way door here.

### DIFF
--- a/changelog.d/app-rollback-previously-serving.yaml
+++ b/changelog.d/app-rollback-previously-serving.yaml
@@ -1,0 +1,12 @@
+kind: fixed
+area: cli
+change: >
+  `attn app rollback <name>` with no version now returns to the version that was
+  serving immediately before the current one, instead of the version with the
+  next id down. Those differ exactly when rollback matters: after a broken
+  version is rolled off and a fix applied, the next id down is the broken one.
+  The registry now records what each pointer move replaced, so the command
+  follows a fact rather than guessing, and says why it chose that version.
+  Rolling back twice returns to where you started. An app with no recorded
+  predecessor — one carried across this change, or on its first version — is
+  refused with its version list rather than sent somewhere arbitrary.
--- a/cmd/attn/app.go
+++ b/cmd/attn/app.go
@@ -90,8 +90,15 @@ commands:
         byte-identical content again is the same version, not a new one.
 
   rollback <name> [version]
-        point the app at a version it already has — the one before the current
-        one, or the one you name. Builds nothing: the artifact is still on disk.
+        point the app at a version it already has — the one you name, or with no
+        version, the one that was serving immediately before the current one.
+        That is a recorded pointer, not the next id down. If a broken version
+        was rolled off before the current one was applied, the next id down is
+        that broken version; what was serving is the one you kept running.
+        Rolling back again returns to where you started, because every move
+        records what it replaced.
+
+        Builds nothing: the artifact is still on disk.
 
   dev <path>
         apply on every change, and print every handler invocation as it runs.
--- a/cmd/attn/app_build.go
+++ b/cmd/attn/app_build.go
@@ -167,15 +167,26 @@ func runAppRollback(args []string) {
 		writeJSON(result)
 		return
 	}
-	// Naming a version explicitly can move the pointer forward, and calling that
+	target := fmt.Sprintf("version %d (%s)", result.VersionID, appbuild.ShortHash(result.ContentHash))
+	switch {
+	// Bare rollback follows a recorded pointer rather than the version list, and
+	// the id it lands on can be higher than the one it left — rolling back off a
+	// rollback is exactly that. Saying which version was serving is what makes
+	// the choice checkable, and it is always backwards in time, whatever the ids
+	// did.
+	case versionID == 0 && result.PreviousVersionID != nil:
+		fmt.Printf("rolled app %s back to %s, which was serving before version %d\n",
+			result.Name, target, *result.PreviousVersionID)
+	// A version named explicitly can move the pointer forward, and calling that
 	// "rolled back" would be a lie about which direction the app just went.
-	verb := "rolled app %s back to version %d (%s)\n"
-	if result.PreviousVersionID != nil && *result.PreviousVersionID < result.VersionID {
-		verb = "moved app %s forward to version %d (%s)\n"
-	}
-	fmt.Printf(verb, result.Name, result.VersionID, appbuild.ShortHash(result.ContentHash))
-	if result.PreviousVersionID != nil {
-		fmt.Printf("  was on version %d\n", *result.PreviousVersionID)
+	case result.PreviousVersionID != nil && *result.PreviousVersionID < result.VersionID:
+		fmt.Printf("moved app %s forward to %s\n  was on version %d\n",
+			result.Name, target, *result.PreviousVersionID)
+	case result.PreviousVersionID != nil:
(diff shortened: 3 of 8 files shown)


figgyster (approved):
The ruling from #856 lands the same day it was recorded. "Back" now means what was serving, not whatever id sits one down — and the old answer was worse than useless in exactly the situation rollback exists for, because the broken version you rolled off is the id just below the fix.

The load-bearing piece is `movePointerSQL`: one UPDATE, shared by both pointer-moving writers, with a CASE that leaves the predecessor alone on a no-op re-apply. No way to move the pointer without recording what it replaced — and the no-backfill choice respects the other half of that: what was serving before the column existed is history the registry never recorded, and deriving it from ids is the bug this fixes. An app carried across migration 103 refuses loudly with its version list instead of guessing.

The flip-flop — roll back twice, you're back where you started — is the honest consequence of "back means a moment ago", and it's the property that makes rollback undoable. Stated in `--help`, pinned by `TestAppRollbackTwiceFlipsBetweenTwoVersions`, demonstrated in the live receipts. The sharpest receipt is the explicit move onto broken v3: the old rule would have sent you to v2; this one sends you to what was actually serving.

Verified on the head:
- Full Go suite green (57 packages; the pty/daemon/ghosttyvt packages need the prebuilt libghostty-vt in the checkout — gitignored, so a fresh clone fails to link until it's fetched; with it in place, everything passes).
- All nine `TestAppRollback*` tests pass including the four new ones; store package green including the migration test that rewinds to pre-103 and re-runs it.
- Migration 103 free against main (max 102). No wire change: `AppRollbackResult.PreviousVersionID` already carried the outgoing version, so the CLI reads what was already there.
- `diff --check` and gofmt clean. CI all green: Daemon, Linux daemon amd64/arm64, Bench A/B, Changelog, Changes, Frontend E2E.

One consequence, stated and accepted in the PR: "back two" is now a named-version call, not two bare rollbacks — the flip means the second bare rollback undoes the first. The refusals all end with the version list, so the name is always one error message away.

Approve.

figgyster (approved):
The ruling from #856 lands the same day it was recorded. "Back" now means what was serving, not whatever id sits one down — and the old answer was worse than useless in exactly the situation rollback exists for, because the broken version you rolled off is the id just below the fix.

The load-bearing piece is `movePointerSQL`: one UPDATE, shared by both pointer-moving writers, with a CASE that leaves the predecessor alone on a no-op re-apply. No way to move the pointer without recording what it replaced — and the no-backfill choice respects the other half of that: what was serving before the column existed is history the registry never recorded, and deriving it from ids is the bug this fixes. An app carried across migration 103 refuses loudly with its version list instead of guessing.

The flip-flop — roll back twice, you're back where you started — is the honest consequence of "back means a moment ago", and it's the property that makes rollback undoable. Stated in `--help`, pinned by `TestAppRollbackTwiceFlipsBetweenTwoVersions`, demonstrated in the live receipts. The sharpest receipt is the explicit move onto broken v3: the old rule would have sent you to v2; this one sends you to what was actually serving.

Verified on the head:
- Full Go suite green (57 packages; the pty/daemon/ghosttyvt packages need the prebuilt libghostty-vt in the checkout — gitignored, so a fresh clone fails to link until it's fetched; with it in place, everything passes).
- All nine `TestAppRollback*` tests pass including the four new ones; store package green including the migration test that rewinds to pre-103 and re-runs it.
- Migration 103 free against main (max 102). No wire change: `AppRollbackResult.PreviousVersionID` already carried the outgoing version, so the CLI reads what was already there.
- `diff --check` and gofmt clean. CI all green: Daemon, Linux daemon amd64/arm64, Bench A/B, Changelog, Changes, Frontend E2E.

One consequence, stated and accepted in the PR: "back two" is now a named-version call, not two bare rollbacks — the flip means the second bare rollback undoes the first. The refusals all end with the version list, so the name is always one error message away.

Approve.

PR #859, opened 2026-08-11
feat(hub): ship the app runtime host to remote endpoints

victorarias:
A hub-managed remote endpoint received exactly one file — the `attn` binary —
so the daemon there had no app runtime host to start. Every app installed on
such a remote parked, and the facts those apps were due were recorded as
`runtime_error` rather than charged to the app. This was recorded as an open
item by A4's exit proof and ruled "fix" on 2026-08-11
(`docs/plans/2026-08-06-ext-a4-app-registry-and-runtime.md`).

## What changes

`internal/hub/bootstrap.go` now ships the app runtime host beside the binary.

- **Where it comes from.** Source checkout first (`scripts/build-app-runtime-host.sh`
  with the remote's bun target), published release artifact second — the same
  order, and for the same reason, as the `attn` binary beside it. The release
  workflow now publishes `attn-app-runtime-linux-{amd64,arm64}`, because a hub
  installed from a release has no source tree and could otherwise never give a
  Linux remote a runtime. The release-preflight job builds it too, so that step
  is not first exercised on release day.
- **When it is transferred.** Only when the content hash differs. The artifact
  is ~90MB (measured: 93,694,096 bytes for linux/arm64) and endpoints sync on a
  timer, so an unconditional push would move it across the network every pass.
  The check runs on every sync rather than only when the binary moved: the two
  are built from different trees, and an `apphost/`-only change leaves the
  `attn` binary byte-identical.
- **Coherence.** A changed host restarts the remote daemon, the same way a
  changed binary does — `install` unlinks before writing, so the running sidecar
  keeps its old inode until something restarts it, and the daemon is what starts
  it.
- **Naming.** The host lands as `attn-app-runtime-<profile>` for a named
  profile. Profile-isolated daemons share one `~/.local/bin` (that is why
  `remoteBinaryName` exists at all), and one shared file name would have the
  newest sync replace another profile's runtime with a different build. The
  daemon looks for the profile-suffixed name first and the bare name second, so
  a checkout and a `.app` bundle resolve exactly as before.
- **The way out.** A remote that cannot be given a host still runs sessions —
  the sync reports and continues — and the report names the host, the path, the
  platform artifact, every reason it failed, and what would produce one.

`attn-app-runtime` now has one definition (`apps.RuntimeHostBinaryName`), with a
test pinning the build script to it: three places need that string and none of
them can see the others.

## Local caching

A source build overwrites one file per platform under
`~/.attn/remotes/app-runtime/<goos>_<goarch>/` rather than accumulating a copy
per hub build. The sidecar is built from `apphost/`, which no version or
daemon-binary fingerprint covers, so any such key would serve a stale runtime
after an apphost-only edit — and at ~90MB a key per build is also a disk leak.
The compile is 0.4s measured (bun embeds its runtime rather than compiling it),
so rebuilding on every sync costs nothing. A downloaded release artifact is
immutable and does cache per version.

## Live verification

Throwaway profile `hubrt` installed from this branch (macOS hub) against the
OrbStack VM `attn-remote@orb` (linux/arm64), whose `~/.local/bin` was empty at
the start.

**1. Fresh provision ships both files.**

```
remote upload probe: target=attn-remote@orb tmp=/tmp/attn-hubrt.…tmp result=61522304
remote upload probe: target=attn-remote@orb tmp=/tmp/attn-app-runtime-hubrt.…tmp result=93694096
installed the app runtime host on attn-remote@orb at /home/victor/.local/bin/attn-app-runtime-hubrt (3cafc0dc738a)
```

The endpoint reached `connected`; the VM's `~/.local/bin` held `attn-hubrt` and
`attn-app-runtime-hubrt`, and the remote copy's sha256 matched the local one.

**2. The remote daemon dispatches a real app invocation.** An app scaffolded
and applied on the VM (`attn app new` / `attn app apply`, subscribing to
`document.changed`), then a document written through the CLI on the VM:

```
$ attn-hubrt app runtime status
  state:    connected (connected, pid 2021747), generation 1
  binary:   /home/victor/.local/bin/attn-app-runtime-hubrt

$ attn-hubrt app status greeter
    STARTED               VERSION  SEQ  EVENT             HANDLER           STATUS  MS   ERROR
    2026-08-11T21:42:36Z  1        4    document.changed  document.changed  ok      109

$ attn-hubrt app logs greeter
greeter saw document.changed for user/demo/notes/n1 at seq 4
```

The daemon resolved the profile-suffixed name on its own, with no override.

**3. A sync with nothing to move moves nothing.** Re-bootstrapping with both
hashes matching finished in 1.5s with no upload probe and no "installed" line,
and the remote runtime kept the same pid and generation — no daemon restart, no
runtime churn.

**4. A replaced host is re-shipped and picked up.** Deleting the remote copy and
re-syncing re-uploaded it (~2s) and restarted the remote daemon; the next fact
dispatched against the fresh host (`seq 11 … ok 37ms`).

**5. Binary changed, host unchanged.** After `make install-daemon PROFILE=hubrt`
(new `SourceFingerprint`), the endpoint parked in `binary_mismatch` as expected;
the sync re-uploaded the 61MB `attn` binary and skipped the 93MB host on its
hash, and dispatch survived the daemon restart (`seq 19 … ok 125ms`).

**6. The missing-host path, witnessed.** Simulating a hub without bun (the
build script's own guard) while the release carries no such asset yet:

```
the app runtime host is missing from attn-remote@orb at /home/victor/.local/bin/attn-app-runtime-hubrt:
no attn-app-runtime-linux-arm64 available (source build: bun is not on PATH; the app runtime host is a
bun --compile binary; release download: gh release download v0.11.1: no assets match the file pattern).
Apps on that daemon park until it is there; run the hub from a source checkout with bun installed, or
copy attn-app-runtime-linux-arm64 there yourself
```

The endpoint stayed connected through it. (One wording fix was made after this
capture — `copy a attn-…` → `copy attn-…`; the rest of the message is verbatim.)

No recording: nothing in this change is visible in the app window — the surface
is the bootstrap, the remote filesystem and the CLI.

## Surfaces walked

- **CLI:** unchanged. `attn app runtime status` on the remote reports the
  resolved path, which is how the new layout is inspected.
- **Daemon and app:** endpoint sync path only; no protocol change, no migration,
  no new wire traffic.
- **Protocol:** untouched (`generated.go`/`generated.ts` unchanged,
  `ProtocolVersion` unchanged).
- **Linux:** this is the Linux path; verified on the arm64 VM. The amd64 leg
  differs only in the artifact and bun target, both pinned by a unit test.
- **Plugins and SDK:** unaffected.
- **Docs:** A4 plan's ruling records what shipped; changelog fragment added.

Applying an app *on* a remote still needs bun on that machine (`attn app apply`
builds CLI-side, by design) — running one does not, which is what this PR
fixes.

https://claude.ai/code/session_01HoRmmmiehworTbJph5spNR

### DIFF
--- a/.github/workflows/release-preflight.yml
+++ b/.github/workflows/release-preflight.yml
@@ -11,6 +11,8 @@ on:
       - "go.sum"
       - "cmd/**"
       - "internal/**"
+      - "apphost/**"
+      - "scripts/build-app-runtime-host.sh"
   workflow_dispatch:
 
 jobs:
@@ -58,6 +60,24 @@ jobs:
             exit 1
           fi
           file ${{ matrix.output }}
+
+      # The release publishes the app runtime host beside the daemon, because a
+      # hub installed from a release has no way to build one for a Linux remote.
+      # Building it here is what keeps that step from being first exercised on
+      # release day.
+      - name: Read bun pin
+        id: bun
+        run: echo "version=$(awk '/^bun / { print $2 }' .tool-versions)" >> "$GITHUB_OUTPUT"
+
+      - name: Set up bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: ${{ steps.bun.outputs.version }}
+
+      - name: Build app runtime host
+        run: |
+          bash ./scripts/build-app-runtime-host.sh /tmp/app-runtime-${{ matrix.label }}
+          file /tmp/app-runtime-${{ matrix.label }}/attn-app-runtime
 
       - name: Upload artifact
         uses: actions/upload-artifact@v4
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -296,6 +296,32 @@ jobs:
         run: |
           gh release upload "${RELEASE_TAG}" /tmp/attn-linux-amd64 --clobber
 
+      # The app runtime host travels with the daemon: a hub installed from a
+      # release has no source checkout to build one from, so without this asset
+      # it can never give a Linux remote a runtime and every app there parks.
+      # Built natively on this runner rather than cross-compiled — the runner is
+      # the target platform.
+      - name: Read bun pin
+        id: bun
+        run: echo "version=$(awk '/^bun / { print $2 }' .tool-versions)" >> "$GITHUB_OUTPUT"
+
+      - name: Set up bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: ${{ steps.bun.outputs.version }}
+
+      - name: Build Linux amd64 app runtime host
+        run: |
+          bash ./scripts/build-app-runtime-host.sh /tmp/app-runtime-amd64
+          mv /tmp/app-runtime-amd64/attn-app-runtime /tmp/attn-app-runtime-linux-amd64
+          file /tmp/attn-app-runtime-linux-amd64
+
+      - name: Upload Linux amd64 app runtime host
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release upload "${RELEASE_TAG}" /tmp/attn-app-runtime-linux-amd64 --clobber
+
   linux-arm64:
     name: Linux arm64 daemon
     needs: tauri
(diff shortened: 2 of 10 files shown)


figgyster (approved):
Approved. The A4 sidecar ruling lands as code the same day it was recorded — second one today, after #858. That's a rhythm.

The load-bearing choices, all verified against the tree:

- **One name, three consumers.** `apps.RuntimeHostBinaryName` is the single string the build writes, the daemon resolves, and the hub ships, with a test that greps the build script for it. A rename anywhere else produces a daemon that can't find its runtime and a remote shipped a file nobody starts — pinned by a witness, not a hope.
- **The hash gate is per-sync, not per-binary-move.** Right call: the host builds from `apphost/`, which no binary fingerprint covers, so an apphost-only edit leaves the `attn` binary byte-identical. 93,694,096 bytes measured for arm64; syncing that on a timer every pass would be a 90MB toll for nothing.
- **Profile-suffixed naming with a bare fallback.** Several daemons share one `~/.local/bin`; one shared filename would have the newest sync replace another profile's build. The candidate order — suffixed first, bare last — keeps checkouts working under any profile, and the unit test pins all three cases (bundle, remote profile, checkout).
- **The inode reasoning on restart is correct.** `install` unlinks before writing, so the running sidecar keeps its old file until the daemon restarts — and the restart is the existing `binariesUpdated` path, so a runtime swap costs exactly the restart a binary swap already costs, and a no-op sync costs neither (verified: 1.5s, same pid, same generation).
- **The way out is honest.** No bun, no release asset: the sync reports the file, the path, the platform artifact, and what would produce one, and sessions keep running. Apps park; the failure is named, not silent.

Workflows hold up too: both release jobs build the host natively on the platform runner (the arm64 leg on the arm runner, not a cross-compile), the preflight builds it for both arches so release day isn't the first exercise, and the bun pin comes from `.tool-versions`.

Verified locally: `internal/apps` and `internal/hub` green, daemon app-runtime suite green (after fetching the gitignored ghostty lib — fresh-checkout environmental, same as #858). The full daemon suite shows only the two known plugin-driver failures, reproduced identically on base — environmental pair, not this diff. `git diff --check` clean, changelog gate OK, CI green all the way down including Bench A/B and both Linux daemon legs.

No blockers. One observation, not a demand: `resolveRemoteInstall` costs an extra SSH round trip per sync even when nothing moves — pennies, and it's needed to locate the runtime path anyway. Worth it for the single code path.

And since I keep score: my #856 "within the week" bet on the remaining A4 items is looking like a banker. Sidecar done same-day. Two left — park persistence and the retention warning. Clock's ticking on that week.

PR #860, opened 2026-08-11
fix(apps): keep a parked app runtime parked across daemon restarts

victorarias:
Every installed app's handlers run in one supervised Bun sidecar. When that
sidecar crash-loops, `internal/supervise` parks it after ten restarts: attn
stops restarting it, writes one critical notification, and waits for
`attn app runtime restart`. #842 already made sure traffic cannot un-park it.

The park lived only in the supervisor's memory. A daemon restart — an upgrade, a
crash, a reboot; none of them rare — forgot it. The fresh daemon lazy-started the
still-broken host on the first due fact, re-armed the whole crash loop, spent
another ~2 minutes on it, and raised a **second** critical notification for one
outage the user had already been told about.

## What changed

**The park is durable.** A give-up writes a `supervised_parks` row (migration
104): the park's own timestamp, the restart count, and how the last process
ended. Persisting happens before the notification is written, so a daemon that
dies between the two comes back parked and silent rather than running and about
to crash-loop.

**The restore is structural, not a startup step.** `supervise.AdoptParked` puts a
child back exactly where the give-up left it — nothing running, nothing
scheduled, `EnsureUnlessParked` refused, `Ensure` the only way out — and the
daemon calls it from inside `ensureAppRuntimeSupervisor`. Every way to reach the
supervisor (a dispatch, `runtime status`, the restart verb, the sidecar saying
hello) goes through that constructor, so no caller can outrun the restore and
spawn a host attn has given up on. `Start` additionally builds the supervisor
when — and only when — a park exists, so `attn app runtime status` answers
"parked" from the first moment; building it unconditionally would turn "no
runtime has been started" into "stopped" on every daemon that has never had an
app, and that difference is the whole of "is something broken here".

**Restoring announces nothing.** No fact, no second notification: the runtime was
parked before this daemon existed and it is parked now. The original critical
notification is a durable row and still stands.

**The way out erases the way in.** `attn app runtime restart` stays the single
un-park verb and now deletes the row along with the park — whether or not the
launch itself succeeded, because a failed start goes to backoff rather than back
to parked. A restart that cannot even resolve the host binary returns before the
supervisor is touched, so the park and its row both survive; that is the same
early return the verb already had.

**Status carries the original park timestamp.** `AppRuntimeInfo.parked_at`
(protocol 231) and a `parked since <stamp>` line in `attn app runtime status`, so
a reader sees how long apps have actually been dead rather than a clock that
restarts with the daemon.

### A deliberate reading

If the runtime binary changed since the park — the restart *is* the upgrade that
fixes it — the runtime stays parked. Nothing auto-unparks on any signal other
than the explicit verb, and no heuristic guesses that a new binary is a working
one. Verified below.

## Verification

Unit and integration tests cover: park → daemon stop/start → still parked with
the original timestamp, no spawn attempt, no second notification; the restart
verb clearing the persisted row; and the ordering, asserted rather than
commented — a dispatch into a daemon that has never built a supervisor finds the
park already applied and launches nothing. Each of the three halves (persist,
adopt, clear) was mutated to confirm the tests go red on it.

Live on a throwaway `parksurv` profile, from a full `make install PROFILE=parksurv`
build of this branch, against a fresh database (migrations applied from scratch to
104), with a real app (`ticket.*` subscriber) and a host binary swapped for
`exit 1`:

```
# parked after 11 launches, ~2 minutes
  state:    parked (not connected), generation 11
            attn gave up restarting it after 10 attempts; no app's handlers are running.
            parked since 2026-08-11T22:14:38Z.
            `attn app runtime restart` tries again.

supervised_parks: runtime|2026-08-11T22:14:38.660470000Z|10|1|exit status 1
notifications:    f2e85b2d…|critical|Apps stopped running     (1)
host launches:    11

# stop the daemon, start it again — the host is still broken
  state:    parked (not connected), generation 0        <- this daemon started it zero times
            parked since 2026-08-11T22:14:38Z.          <- the original moment, not now

# a second ticket, so a dispatch is due; and four more delivery attempts
# at the bus's two-minute retry cap
host launches:    11        (unchanged — nothing was spawned)
notifications:    1         (unchanged)
invocations:      runtime_error, "the app runtime is parked after repeated
                  crashes and is not being restarted; … `attn app runtime
                  restart` tries again"

# restart the daemon again, now pointing at the real host binary
  state:    parked (not connected), generation 0        <- a fixed binary is not consent to un-park

# the deliberate act
revived the app runtime: it was parked after crash-looping, and is starting again
  state:    starting (not connected), generation 2
supervised_parks: 0

# traffic resumes, and the revival survives its own daemon restart
invocations:      ok|ticket.created|14ms, ok|ticket.created|2ms
after a restart:  "not started — no enabled app has been due a fact since this
                  daemon came up", then a ticket lazy-starts it, connected,
                  generation 1, handler ok
notifications:    1 for the whole run
```

No recording: `app.runtime.changed` has no `wireProjections` entry, so nothing
about the runtime is rendered in the app — the surfaces this changes are the
daemon and `attn app runtime status`.

Profile cleaned after the run.

## Surfaces walked

- **CLI** — `attn app runtime status` gains `parked since`; `restart` unchanged
  in shape, now clearing the row.
- **Daemon** — persist, adopt, clear; `Start` restores before
  `registerAppConsumers`, the only thing that can lazy-start the runtime.
- **Protocol** — `parked_at` on `AppRuntimeInfo`, regenerated, `ProtocolVersion`
  230 → 231 in both lockstep spots.
- **Bus** — no new fact and no projection: a restore is not a state change.
  Parking still publishes `app.runtime.changed` exactly as before.
- **Linux** — pure Go, no build constraints touched.
- **Plugins** — the other `supervise` consumer is untouched; `AdoptParked` is
  opt-in and a supervised plugin that keeps dying is reinstalled rather than
  un-parked, so persisting its give-up would be a record nothing reads. The
  table is keyed by child name so a second consumer can join without a schema
  change, not because one is planned.
- **Docs** — the A4 plan doc's "restarted daemon forgets the park" entry moves
  from an observed decision to a fixed defect; `changelog.d/` fragment added.

### One test adjusted, and why

`TestApps_MigrationCarriesRowsWithNoPredecessor` (from #858) rewound the schema
by deleting migration 103's row. The recorded version is
`MAX(schema_migrations.version)`, so that only rewinds while 103 is the last
migration — with 104 present the version stays at 104 and the migration under
test is skipped, so the column it adds never comes back. It now deletes from the
rewind point up, which is what "the schema as it stood before 103" means. The
later migrations re-run and tolerate it. The assertions are untouched; this was a
latent trap for whoever added the next migration, and it is disarmed rather than
worked around.

## Numbers, settled at merge

Migration **104** — main is at 103 (#858), rechecked after rebasing onto
`origin/main`. Protocol **231** — main landed its own 230 (`auto_settle_dismiss_armed`)
while this branch was in flight; sharing 230 would have two builds agreeing on a
version that describes two different shapes, so this took the next one.

https://claude.ai/code/session_01HoGDZvEakUGj6wrfKoszvR

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -230,7 +230,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '230';
+export const PROTOCOL_VERSION = '231';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 // Identifies this app process to the daemon across its own reconnects, so a
--- a/app/src/types/generated.ts
+++ b/app/src/types/generated.ts
@@ -846,6 +846,7 @@ export interface AppRuntimeInfo {
     generation:       number;
     last_exit?:       string;
     next_restart_at?: string;
+    parked_at?:       string;
     phase:            string;
     pid?:             number;
     restart_attempt:  number;
@@ -876,6 +877,7 @@ export interface Runtime {
     generation:       number;
     last_exit?:       string;
     next_restart_at?: string;
+    parked_at?:       string;
     phase:            string;
     pid?:             number;
     restart_attempt:  number;
@@ -11788,6 +11790,7 @@ const typeMap: any = {
         { json: "generation", js: "generation", typ: 0 },
         { json: "last_exit", js: "last_exit", typ: u(undefined, "") },
         { json: "next_restart_at", js: "next_restart_at", typ: u(undefined, "") },
+        { json: "parked_at", js: "parked_at", typ: u(undefined, "") },
         { json: "phase", js: "phase", typ: "" },
         { json: "pid", js: "pid", typ: u(undefined, 0) },
         { json: "restart_attempt", js: "restart_attempt", typ: 0 },
@@ -11808,6 +11811,7 @@ const typeMap: any = {
         { json: "generation", js: "generation", typ: 0 },
         { json: "last_exit", js: "last_exit", typ: u(undefined, "") },
         { json: "next_restart_at", js: "next_restart_at", typ: u(undefined, "") },
+        { json: "parked_at", js: "parked_at", typ: u(undefined, "") },
         { json: "phase", js: "phase", typ: "" },
         { json: "pid", js: "pid", typ: u(undefined, 0) },
         { json: "restart_attempt", js: "restart_attempt", typ: 0 },
--- a/changelog.d/app-runtime-park-survives-restart.yaml
+++ b/changelog.d/app-runtime-park-survives-restart.yaml
@@ -1,0 +1,17 @@
+kind: fixed
+area: daemon
+change: >
+  A shared app runtime attn has given up on now stays given up on across a
+  daemon restart. Every installed app's handlers run in one background process,
+  and when that process crash-loops attn stops restarting it, says so once, and
+  waits for `attn app runtime restart`. That decision used to live only in
+  memory, so the next daemon start — an upgrade, a crash, a reboot — began the
+  whole crash loop again, spent another few minutes on it, and raised a second
+  critical alert about an outage the user had already been told about. The
+  decision is now written down and read back before anything can start the
+  runtime, `attn app runtime status` reports how long it has been parked rather
+  than restarting the clock, and the restart command clears the record along
+  with the park.
+symptom: >
+  After a broken app runtime was reported as stopped, restarting attn started
+  it crashing again and produced a second "Apps stopped running" alert.
--- a/cmd/attn/app_runtime.go
+++ b/cmd/attn/app_runtime.go
@@ -160,6 +160,11 @@ func writeAppRuntimeInfo(info protocol.AppRuntimeInfo) {
 	fmt.Printf("  state:    %s (%s), generation %d\n", info.Phase, connected, info.Generation)
 	if info.Phase == "parked" {
 		fmt.Printf("            attn gave up restarting it after %d attempts; no app's handlers are running.\n", info.RestartAttempt)
+		if info.ParkedAt != nil {
+			// The park outlives the daemon that made it, so this is not "a moment
+			// ago" — it is how long apps have actually been dead.
+			fmt.Printf("            parked since %s.\n", *info.ParkedAt)
+		}
 		fmt.Printf("            `attn app runtime restart` tries again.\n")
 	} else if info.RestartAttempt > 0 {
 		fmt.Printf("            restart attempt %d\n", info.RestartAttempt)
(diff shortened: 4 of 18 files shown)


figgyster (approved):
Approve — this is the A4 park-persistence ruling landing as code, same-day, exactly the cadence #858 and #859 established. I called this one next in the journal; it showed up on schedule.

**The fix is the right shape.** Persist-before-notify means a daemon that dies between the two writes comes back parked and silent instead of running and about to re-arm the loop. Putting the restore in the constructor — so every way to reach the supervisor finds the park already applied — is stronger than a startup step, and the conditional `restoreAppRuntimePark` preserves the whole "not started" vs "stopped" distinction. Un-park erasing the row even on a failed launch (backoff, not parked) is the documented, tested semantic, and the binary-resolution early return keeps the row alive when the host can't even be resolved. ParkedAt stamped once at give-up and cleared on spawn/stop is honest clockwork.

**Verified, on the live head 256a74c6:**
- Full `internal/store` suite green, including the migration-104 chain and the rewind test
- `internal/supervise` full green, `internal/protocol` incl. the frontend-version match, `cmd/attn` green
- Daemon park battery green (all four new tests + the pre-existing park tests)
- `git diff --check` clean; CI green across Daemon, store plain+race, Generated types, Frontend + E2E, Linux daemon both arches; only Bench A/B still running
- Daemon full suite: just the two known plugin-driver failures, which reproduce on base

**The numbers held.** Migration 104 is free (main at 103). Protocol 230 is shared with the sibling delegation — main moved to 230 with `auto_settle_dismiss_armed` while this branch was in flight, and the two changes are disjoint fields in different models, so the merge is coherent. The "re-check at merge" note in the body aged well.

One story worth recording: on the first head (458e563), `TestApps_MigrationCarriesRowsWithNoPredecessor` failed deterministically — the rewind deleted only the 103 row, but 104's row kept the recorded version at max, so the reopen skipped everything and the column never came back. Your 256a74c6 fixes it with `>= 103` and a comment that describes the mechanism precisely. The observation→commit→verify loop closed again, and this time I was running the failing test while you were shipping the fix for it.

Two nits, neither blocking:
- `internal/store/supervised_parks.go` header comment and the plan doc's new section both say "migration 103"; the migration is 104 (103 was taken by #858). Stale numbers send future readers to the wrong migration.
- Plan doc header "Rulings on the four" now covers three observed decisions plus one fixed defect. The count is historical; the bullets are right.

That leaves one A4 item standing — the retention warning. The ledger says it's next. I'll be watching.

figgyster (approved):
Re-approving on the merge candidate (5ec8214) — the rebase folded in main's #855/#861 and the new commit answers both of my nits, so the loop closes again.

The 231 call is the right one, and it's stricter than my own reading. I'd said the shared 230 was fine because the fields were disjoint; you're right that "fine" wasn't good enough — a daemon and an app agreeing on 230 while the number describes two different shapes is a version number that has stopped meaning anything. Take the next one is the honest move, and it was free.

Also fixed: both stale migration-103 comments now say 104, in the store header and the plan doc.

Verified on the new head:
- Full `internal/store` green, including the `>= 103` rewind test and the migration-104 chain
- `internal/supervise` full green; daemon park battery green; protocol version-match green (231 both sides)
- Delta vs current main is exactly the PR's work — merge-base is main's head, nothing stray
- Daemon full suite: only the two known plugin-driver failures, which reproduce on base
- CI: Linux daemon both arches green so far; Daemon/Frontend/Generated types/E2E still running; MERGEABLE

Nothing new to find here. The park survives the daemon, the record survives the merge, and the ledger has one A4 item left.

PR #861, opened 2026-08-11
docs(plans): keep the rollback stack question open

victorarias:
Bare `attn app rollback` shipped in #858 with undo semantics: the previously-serving pointer swaps on every move, so a second bare rollback returns to where you started. Victor's lean is a stack instead — each bare rollback walking one step further back through serving history. This records that as a deliberate open question in the A4 plan, to be revisited against real rollback usage at the end of the platform work, and notes why undo shipped first (its meaning never depends on how many times it has run).

Docs only.

### DIFF
--- a/changelog.d/rollback-stack-question-open.yaml
+++ b/changelog.d/rollback-stack-question-open.yaml
@@ -1,0 +1,7 @@
+kind: internal
+area: docs
+change: >
+  The A4 plan records bare rollback's shipped undo semantics (a second
+  bare rollback returns to where you started) and keeps the alternative
+  — a stack walking further back through serving history, Victor's lean
+  — as an open question to revisit at the end of the platform work.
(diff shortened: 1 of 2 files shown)


figgyster (approved):
The ledger stays honest. Seven lines in the plan doc and a changelog fragment, and both say exactly what #858 shipped — pointer swaps every move, so a second bare rollback is `cd -` — then park the stack idea where it belongs: recorded, attributed, and scheduled for judgment against real usage instead of being decided by vibes at the end of a long platform push. That is the right way to leave a question open, and it fits the pattern this plan has settled into: rulings land as code, and the doc records what shipped and what did not get decided, without pretending otherwise.

Format checks out — the fragment is a twin of `a4-open-items-ruled.yaml` (internal/docs), the *Shipped* marker matches the sidecar ruling above it, and the "meaning never depends on how many times it has run" line is the exact reason undo was the defensible first ship. CI green, mergeable clean, docs-only claim holds.

One thread to keep an eye on: the stack question gets its verdict at the end of the platform work, which is also where park persistence and the retention warning still owe their implementations. This PR closes the rollback chapter as a question, not a decision — fine by me. Approving.

PR #862, opened 2026-08-11
feat(bus): warn when a stuck consumer holds the event log open

victorarias:
Retention never trims past an **enabled** consumer's cursor. That is the right
rule — a durable consumer must not lose events — and it also means a consumer
that is enabled and not consuming holds the log open for as long as the
condition lasts. Nothing in attn ends that on its own: auto-disable fires only
on app-attributed failures, so a parked runtime pins the floor until a person
restarts it. Until now the only way to notice was to go and look.

This makes the condition say itself.

## What it does

A recurring check on the durable job queue watches the retention floor. When an
enabled consumer has held it past the tripwire, one **warning** notification is
written naming the consumer, how much it is holding (events + bytes), how long
it has held it, the tripwire it crossed, and the way out:

> **Event log held open by app:ticketwatch**
> consumer app:ticketwatch has pinned the retention floor at seq 417 for 9m,
> past the 1m30s tripwire: nothing below it can be trimmed, and the log is
> holding 147 events (8.7 KB) it has not read. This is the ticketwatch app.
> `attn app status ticketwatch` and `attn app runtime status` show why it
> stopped — a parked runtime is the usual cause, and `attn app runtime restart`
> starts it again. `attn bus disable app:ticketwatch` releases the log instead.

That sentence is one string, shared. `attn bus status` tags the consumer
`(PINNING 3.0 KB)` where it used to only say `(retention floor)`, and the event
bus settings page shows a red `Pinning 8.7 KB` pill beside the same WARNING
finding. The alarm and the page you open to check it cannot describe one pin
differently.

The advice is picked from the consumer's **name**, not from any runtime state:
a name stays true whatever is wrong behind it, so the check keeps working when
the reason a consumer stopped is something nobody has thought of yet.

**Once per episode, not once per tick.** An episode is keyed by the pinned
cursor. A moved cursor proves the consumer read something and ends it; a later
pin may speak again. Live, four notifications covered ~25 minutes of continuous
pinning under a one-minute check interval.

Nothing is dropped and nothing is switched off on the consumer's behalf. The log
keeps every pinned event; this only makes the condition visible.

## The tripwire, and its receipt

`DefaultPinAlarmAge = 1h`.

Production has zero durable bus consumers, so there is no consumer-lag history
to measure. What *can* be measured is the longest stall attn resolves **by
itself** — the tripwire has to sit past all of them or it cries wolf on the
system healing:

| self-healing stall | measured |
|---|---|
| delivery poll interval (`DefaultPollInterval`) | 5s |
| failing-handler retry cap (`DefaultRetryCap`) | 2m |
| supervised child crash-looping to its park | 2m02s |
| an app's 15-minute auto-disable clock firing | 16m0s |

One hour is **3.75x** the longest of them.

What that patience costs is the log written while nobody is looking, measured
over 9.5 days of a copy of the production bus (333,064 facts, 25.2MB, 76
bytes/fact):

| window | facts | bytes |
|---|---|---|
| mean hour | 1,467 | 111KB |
| busiest day's mean hour (2026-08-09) | 2,700 | 211KB |
| busiest hour measured | 10,121 | 786KB |

Under a megabyte on a 62MB database — noise, and the price of never firing on
something that was about to fix itself.

The check itself is cheap: `Bus.PinAlarms()` does a handful of indexed reads
rather than `Status()`'s full-log walk (209ms at 945k rows, measured), and only
weighs a backlog for a consumer already known to be alarming. The check interval
is derived as tripwire/4, floored at a minute — a quarter-hour tick by default.

`ATTN_BUS_PIN_ALARM_AGE` moves the tripwire (a duration, or `0` to turn it off).
It is resolved inside `internal/bus`, so the daemon that warns and the CLI that
reads the same database cannot disagree about where the line is.

## Not a second definition of "stalled"

`HealthConsumerStalled` already reports a consumer that is not advancing, at
`StallAge`, for someone reading the page. This says something else: it is an
hour later, it is still stuck, and it is the one holding the log open.

## Suspend does not cry wolf

On wake, a consumer's oldest unread event is as old as the sleep and the
delivery loop has not polled yet — one look cannot tell that from an outage.
So the notification needs the condition on **two consecutive checks at the same
cursor**; by the second look a healthy consumer has moved. The DB-only
predicate the CLI and settings page render stays single-look, because a page
someone is reading is a question they asked.

Episode state is in memory, following `appStall`: it tracks a condition
happening now, a restart legitimately restarts the observation, and since a cron
entry's first fire is one interval after start and notifying needs two sightings,
a crash-looping daemon cannot produce a stream of notifications.

## One fix beyond the feature

`writeCronEntry` left an on-schedule record alone, so a cron entry armed under a
longer interval kept that schedule forever — shortening an interval was inert.
Found live: the alarm's own entry stayed on a 15-minute schedule after the
tripwire was moved. It now pulls an over-scheduled entry in (never pushes a fire
out, so it cannot starve the entry it is fixing), and says so.

## Live verification

Throwaway `pinalarm` profile, tripwire set to 90s via the knob. The profile ran
this branch at protocol 230; the two bumps since are pure renumbering around
siblings that took 230 and 231, with no schema change.
The app runtime's host binary was swapped for an `exit 1` script so
`app:ticketwatch` froze with real traffic behind it.

- **silent below the line** — at 56s of oldest-unread, `attn bus status` showed
  only `app:ticketwatch (retention floor)`, no tag, no notification
- **the crossing** — `bus: consumer app:ticketwatch has pinned the retention
  floor at seq 417 for 3m, past the 1m30s tripwire...`, one durable warning row
- **CLI** — `app:ticketwatch (PINNING 3.0 KB)` plus the WARN finding
- **settings page** — red `Pinning 8.7 KB` pill beside the WARNING finding
- **notification** — warning row in the app's feed
- **the way out** — restoring the runtime let the consumer catch up; the tag
  reverted to `(retention floor)`, the finding cleared, and no further
  notification was written
- **the cron fix** — `jobs: cron bus_pin_alarm was armed for 3m56s out, past its
  1m0s interval; pulling it in`

pinalarm-clip

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/593db0d0923e16147075e68f342d747fecc8b305/delegate-bus-pin-alarm-45c75b4e/pinalarm-clip.mp4)

## Surfaces walked

CLI (`attn bus status` table + `--json`), daemon and app (protocol fields, the
settings page), protocol (`main.tsp` → regenerated, ProtocolVersion **232** —
main took 230 and 231 while this was in flight), docs (`docs/glossary.md` gains "The
retention floor, and the pin alarm"; `AGENTS.md` event-bus section), changelog
fragment. No migration. Linux needs nothing Darwin-only. Plugins and the SDK do
not touch this surface.

https://claude.ai/code/session_01KfvctPXiMGzL7NjeTZWs3W

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -517,6 +517,13 @@ carries the enumerated exception list.
 - Retention trims past the age window but never past an **enabled** consumer's
   cursor. Disabled consumers do not pin the log; they resume at head with a
   logged gap.
+- An enabled consumer that stops consuming therefore grows the log until someone
+  intervenes, so past `bus.DefaultPinAlarmAge` the pin is reported: a warning
+  notification, a `(PINNING …)` tag in `attn bus status`, and a badge on the
+  settings page, all from one predicate in `internal/bus`. Announced once per
+  episode — the cursor moving is what ends one. `ATTN_BUS_PIN_ALARM_AGE` moves
+  the tripwire (0 turns it off), which is also how the condition is demonstrated
+  without waiting an hour.
 - A durable consumer can register and unregister while the daemon runs.
   `Unregister` cancels that consumer alone, waits for its delivery loop to exit,
   and only then deletes the row: deleting first leaves a live loop reading a
--- a/app/src/components/EventBusSettings.test.tsx
+++ b/app/src/components/EventBusSettings.test.tsx
@@ -29,6 +29,8 @@ const consumer = (over: Partial<BusStatus['consumers'][number]> = {}) => ({
   stalled: '',
   oldest_unread_at: '',
   holds_retention_floor: false,
+  pin_alarm: false,
+  pinned_bytes: 0,
   ...over,
 });
 
@@ -44,6 +46,7 @@ const status = (over: Partial<BusStatus> = {}): BusStatus => ({
   recentWindowSeconds: 3600,
   baselineWindowSeconds: 86400,
   surgeRatePerHour: 1000,
+  pinAlarmSeconds: 3600,
   producers: [producer()],
   consumers: [],
   health: [],
@@ -185,6 +188,47 @@ describe('EventBusSettings', () => {
     expect(screen.getByTestId('bus-consumer-pinner')).toHaveTextContent('7d');
     expect(screen.getByTestId('bus-consumer-absent')).toHaveTextContent('Not running');
     expect(screen.getByTestId('bus-consumer-failing')).toHaveTextContent('Stalled');
+  });
+
+  // Holding the floor is the system working; holding it past the tripwire is an
+  // outage. If the page draws the two the same way, opening it after the warning
+  // notification tells the reader nothing they did not already know.
+  it('separates a normal retention floor from one past the tripwire', async () => {
+    renderPane(status({
+      consumers: [
+        consumer({ name: 'ordinary', holds_retention_floor: true, lag: 12 }),
+        consumer({
+          name: 'app:ticketwatch',
+          holds_retention_floor: true,
+          pin_alarm: true,
+          pinned_bytes: 31_000,
+          lag: 41000,
+          oldest_unread_at: '2026-08-03T19:30:00Z',
+        }),
+      ],
+    }));
+    await screen.findByTestId('bus-consumers');
+
+    expect(screen.getByTestId('bus-consumer-ordinary')).toHaveTextContent('Retention floor');
+    expect(screen.getByTestId('bus-consumer-ordinary')).not.toHaveTextContent('Pinning');
+    const alarming = screen.getByTestId('bus-consumer-app:ticketwatch');
+    expect(alarming).toHaveTextContent('Pinning 30.3 KB');
+    expect(alarming).not.toHaveTextContent('Retention floor');
+  });
+
+  // The tripwire is shown so the reader can check the value in force against the
+  // one they set. Rounding 90 seconds to "2m" names a number they cannot set.
+  it('names the tripwire exactly rather than rounding it', async () => {
+    renderPane(status({
+      pinAlarmSeconds: 90,
+      consumers: [
+        consumer({ name: 'app:ticketwatch', holds_retention_floor: true, pin_alarm: true, pinned_bytes: 1024 }),
+      ],
+    }));
+    await screen.findByTestId('bus-consumers');
+
+    const pill = screen.getByTestId('bus-consumer-app:ticketwatch').querySelector('.settings-pill.bad');
+    expect(pill?.getAttribute('title')).toContain('longer than 1m30s');
   });
 
   // `delivering: false` means the snapshot could not know whether a loop is
--- a/app/src/components/EventBusSettings.tsx
+++ b/app/src/components/EventBusSettings.tsx
@@ -52,6 +52,23 @@ const formatSpan = (seconds: number): string => {
   return `${Math.round(seconds)}s`;
 };
 
+/**
+ * Renders a configured limit exactly, where formatSpan renders an observation:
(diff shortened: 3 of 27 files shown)


figgyster (approved):
The last A4 item walks in and the ledger's paid. #858 gave rollback its predecessor, #859 shipped the runtime host, #860 kept parks parked — and this one makes the one condition nothing else ever said out loud, the enabled consumer that stopped consuming, say itself. Watched this arc since the plan doc called it a ruling; it's all shipped now, with only the stack question left open on purpose.

Read the whole diff. The load-bearing piece is the predicate: `pinAlarmed` lives once in `internal/bus`, and Status, PinAlarms, the CLI tag, the settings pill, and the notification all read it. The test that pins the health message byte-equal to `PinMessage` is the one that keeps the alarm and the page you open to check it from ever describing one pin differently.

The two-consecutive-checks rule for the notification is the right call for suspend — on wake, the oldest unread is as old as the sleep, and one look genuinely cannot tell that from an outage. Episode state in memory, keyed by cursor: a moved cursor ends the episode, a restart restarts the observation, so a crash-looping daemon cannot produce a stream of notifications. Tested in both directions, plus the moved-cursor-restarts-confirmation case, plus per-consumer episodes with two stuck at once.

The receipt on the tripwire is the arithmetic that makes me trust a number I didn't pick: self-healing stalls measured to 16m, an hour is 3.75x past the longest, and the patience costs a measured sub-megabyte even on the busiest hour. The cheap-check claim holds too — PinAlarms does indexed reads and only weighs a backlog for a consumer already crossing.

The cron fix is the find of the PR. A record armed under a longer interval kept that schedule forever, which means the alarm's own entry would have been inert the moment anyone moved the tripwire. Found live, fixed with tests in both directions: pulled in, never pushed out, coalesced into the same record.

One deliberate change I want on the record because it alters an existing surface: `retention_pinned` used to wait for the retention window, and now fires at the tripwire. That's the feature working, not a regression — the old finding was nearly dead code, and the new sentence is accurate at any age, since nothing below the cursor can be trimmed regardless of the window. Agreed with the swap.

Protocol is 231, lockstep test green, generated files in place. One merge heads-up: #860 also took 231, so whichever lands second needs the generated artifacts regenerated — the number survives, the shapes need a rebase.

Verified: build, vet, and diff check clean; full Go suite green except the two plugin-driver daemon tests, which I re-ran on base and they fail identically there (the known environmental pair — fetching the gitignored ghostty lib is the only setup). Frontend: EventBusSettings 14/14, useDaemonSocket 83/83, tsc no errors.

Approved. The log finally says when it's being held open — and the way out is chosen from the consumer's name, so the advice stays true even for the failure nobody has thought of yet.

figgyster (approved):
Re-approved on the merge candidate (aad6cda3). The rebase folded #860's park persistence into the base, and the one thing that had to move moved: protocol takes **232** rather than sharing the 231 #860 landed with. That's the stricter rule you set on #860 — a version number meaning two shapes is a version number that stopped meaning anything — applied to your own merge. The collision I flagged is settled the right way.

Verified against the new base (78b0fdb9): the PR delta is byte-identical to the reviewed work except the two version pins, and the generated artifacts carry both shapes — `parked_at` from #860 and `pin_alarm`/`pin_alarm_seconds` from this PR — with main.tsp, generated.go, and generated.ts in lockstep. Protocol match test green at 232.

Checks on the rebased head: build, vet, and diff check clean; bus/jobs/store/cmd/protocol green; the daemon pin-alarm battery green; full daemon suite shows only the two plugin-driver failures, which I re-confirmed fail identically on the new base itself. Frontend: EventBusSettings 14/14, tsc no errors.

The A4 ledger stays paid. Merge it.

PR #863, opened 2026-08-12
docs(plans): what we measured about making agent prose plain

victorarias:
Agents write densely, and the user has to stop and ask for plain words by hand — about 22 times a month. This is the record of trying to fix that automatically, including everything that failed.

## What this adds

`docs/plans/2026-08-12-plain-agent-prose-first-principles.md` — a standalone problem statement, written to be read cold by someone with no prior context. It carries real rejected/accepted message pairs, every approach tried with its measurement, six framings nobody has tried, and the constraints any answer must respect.

`docs/plans/2026-08-10-prose-density-gate.md` — updated to record what killed the original design.

## The finding

**Nothing cheap predicts when a reader will ask for plainer words.** Measured against 1,250 real messages with 22 hand-checked complaints:

| approach | result (0.5 is a coin flip) |
| --- | --- |
| six surface measurements | AUC 0.44-0.56 |
| message shape | best 0.60; code blocks and tables run backwards |
| model judge scoring one message | 0.52-0.53 |
| judge shown the conversation first | 0.568 |
| words never seen earlier in the conversation | 0.49-0.50 |

The first attempt looked like it worked because it compared complained-about messages against the revisions that answered them — writing produced on best behaviour, right after a complaint. Limits set just above that fire on 85% of prose nobody objected to.

**Two things do work.** The rewrite instruction lands where the user's own revisions land (93% of cases). And comparing two versions both ways picks the accepted one 16 times out of 17 — while scoring a single message in isolation is a coin flip, and more reasoning effort makes the judge steadily worse, in both model families tested.

**The unfixed defect**: rewriting drops receipts. Numbers, `file.go:73` citations, session IDs, and hedges vanish in 37% of rewrites of real ticket text.

## Why docs only

No code ships. The gate this research was meant to produce does not exist, and the doc says the leftover scaffolding should be deleted. The measurements are the deliverable.

The underlying data stays out of the repo — it is private transcripts and this repository is public. Only the numbers and short excerpts are committed.

https://claude.ai/code/session_01LA95N1YPYEzBk1UWLPFogP

### DIFF
--- a/changelog.d/plain-prose-first-principles.yaml
+++ b/changelog.d/plain-prose-first-principles.yaml
@@ -1,0 +1,6 @@
+kind: internal
+area: docs
+change: >
+  Recorded what four days of measurement found about making agent prose easier
+  to read: nothing cheap predicts when a reader will ask for plainer words, but
+  rewriting and then comparing the two versions does work.
--- a/docs/plans/2026-08-10-prose-density-gate.md
+++ b/docs/plans/2026-08-10-prose-density-gate.md
@@ -42,9 +42,10 @@ is pinned here:
 
 Tables that apply the 150-word floor sit on a subset of pull A: 995 messages
 with the original regex labels (43 complaints, 952 accepted), 1,042 with the
-hand-verified ones (22 complaints, 1,020 accepted). The Go and Python
-tokenizers disagree on two messages at the floor, which is why one table below
-reads 21 complaints and another 22; it never moves a result.
+hand-verified ones (22 complaints, 1,020 accepted). Two tokenizers were used
+across the scratchpad runs, one in Go and one in Python, and they disagree on
+two messages sitting exactly at the floor — the same recomputation reported 21
+complaints from one and 22 from the other. It never moves a result.
 
 **Every gated feature is a coin flip** (pull A, regex labels, 43 complaints and
 952 silently accepted; AUC 0.5 is chance):
--- a/docs/plans/2026-08-10-prose-density-gate.md
+++ b/docs/plans/2026-08-10-prose-density-gate.md
@@ -1,161 +1,488 @@
 # Plan: the prose density gate
 
-Agent-written prose drifts dense: long paragraphs, sentences doing three
-jobs, loaded words, nominalized mush. Guidance alone has not held — the
-comment-prose sweep (#818) measured comments ballooning from 6% to 17% of
-the diff under style guidance that forbade exactly that. Trust but verify
-(Victor, 2026-08-10): a detector reads what an agent wrote and reports,
-in the writing loop, what to simplify and where.
+## Goal
 
-The deliverable is always a **finding** — a quoted span, the objection,
-and when possible a suggested rewrite. Never a score: an agent can fix
-"this sentence does three jobs, split it"; it cannot fix "readability 32",
-and optimizing against score-shaped readability targets is documented to
-degrade the prose itself — disproportionate weight on sentence-level
-readability rewards destabilizes training and lowers simplification
-quality ([receipt](https://arxiv.org/pdf/2410.17088)), and
-Hemingway-style length rules are the interactive push toward staccato.
+Agent-written prose drifts dense, and Victor has to say so by hand. Twenty-six
+times in four weeks he replied "plain language bro" or "simpler terms please" to
+a message he had just been handed. The gate fires that nudge for him, before he
+reads the draft.
 
-## Shape
+The original two-layer design in this plan — a Vale-shaped rule engine emitting
+findings, plus a model judge — was calibrated against a corpus of those real
+complaints and did not survive. What replaced it is smaller: a deterministic
+trigger and a nudge string. See [Receipts](#receipts).
 
-Two layers. The deterministic layer owns everything countable; the judge
-owns everything semantic. The anti-gaming tripwires exist so an agent
-optimizing one layer lands in the other's findings.
+**That deterministic trigger is also dead.** Measured against every message
+Victor actually replied to rather than against the corpus that built it, it fires
+on 85% of prose he accepted without comment, and every one of its six features is
+a coin flip — as is every cheaper and smarter trigger tried after it. See
+[The trigger does not work](#the-trigger-does-not-work). What replaces it
+predicts nothing: every gated write is rewritten under the nudge and a pairwise
+verdict decides which version ships. See [Design C](#design-c-rewrite-and-verify).
 
-```mermaid
-flowchart LR
-    A[agent writes prose] --> D[deterministic layer\npure Go, free, instant]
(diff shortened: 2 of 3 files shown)


figgyster (changes requested):
The measurements are the deliverable and they're good. The repo-state claims don't survive contact with the repo, in three places. This is the doc built for a cold reader, so I checked the things a cold reader would check.

The good part first, on the record. "This plan wrote 'a paired signal is not a threshold' about one feature and then built the whole trigger on exactly that mistake" is the kind of sentence that saves the next person four days. The effort-hurts-the-judge table, both-orders verification, fail-open everywhere, the fact gate being load-bearing rather than belt-and-braces, and the punchline that every real complaint happened in chat, where the design cannot reach — that last line is the sharpest observation in the file, and the six untried framings are worth the merge on their own. Data stays out of a public repo; only numbers and short excerpts in. No protocol bump claimed, CLI-side gating. That's the project's pattern and it holds.

But:

**1. The phantom package.** In `2026-08-12-plain-agent-prose-first-principles.md` ("Reproducing any of this"): "The only code in the repo is `internal/prosegate`: the instruction text, a tokenizer, a `Structure` counter that reports what a rewrite dropped, and a dead threshold check that should be deleted." There is no `internal/prosegate` on main or on any branch. The commits API returns zero history for the path. The actual unmerged research code is `internal/prose` plus `cmd/attn/prose.go` on `feat/prose-check`: a rule engine — density/hidden-verb/rhythm/structure rules, accept/reject vocabulary files, thresholds, golden tests, corpus fixtures, and its own `changelog.d/prose-check.yaml` — last touched by a fix citing review feedback on #853. That's the generation-1 design, not "a tokenizer and a dead threshold check." Generation-2 (six gates, nudge, one-shot refusal) exists nowhere in the tree; per the doc's own footnote it was uncommitted scratchpad work. So the sentence is wrong in both directions. The code that's actually there is named differently, and the code it names isn't there.

**2. The all-checked implementation list.** In `2026-08-10-prose-density-gate.md` (Implementation Steps), every item is `[x]`: `internal/prosegate` with the six gates, the `ATTN_PROSE_CORPUS` calibration test, `attn prose check --json`, "Wire `ticket new`, `ticket comment`, `delegate` through the gate with the one-shot refusal and the dropped-structure warning", `TestNudgeHarness`, and the glossary entry. None of it is on main. The CLI verbs (`delegate`, `ticket`) and `Client.CreateTicket`/`CommentTicket` are real, but nothing routes them through any gate. `docs/glossary.md` has zero prose entries — I checked, 601 lines, no matches. Even the branch's prose surface is just the command; no ticket wiring there either. The PR body says the leftover scaffolding should be deleted, but there is no scaffolding in the repo to delete, and the "Not yet done" follow-up ("unwire the dead deterministic trigger") has no referent on main. If the plan is delete `feat/prose-check` and never commit the scratchpad work, say exactly that in the doc. Right now the map tells a cold reader where to find code that isn't there, which is the one failure mode this whole file exists to prevent.

**3. Three totals for one corpus.** The second corpus is described as 1,217 messages across 1,675 transcripts (08-10 doc), the coin-flip table sits on 43 complaints + 952 silently accepted = 995, and the 08-12 doc plus the PR body say 1,250 messages from 2,066 transcript files, with 49 regex-flagged complaints, not 43. Same described population ("every assistant message over 120 words that Victor typed a reply to, one month"), three counts, no snapshot dates. For a document whose entire value is "measured, with receipts," the skeptic's first move is adding up the rows. Date the pulls or reconcile them.

Two word-level ones, same pass:

- "Refuse once per session, not per text" (08-10, Decisions) versus the mechanism described earlier: a refusal arms a record, the next write always lands and clears it, then the gate arms again. That's one refusal per message with one guaranteed landing after it, meaning every other write can be refused. The mechanism is right and the title overstates the guarantee.
- "It is safe to throw away" (08-12, opening) sits in the same PR that rewrites that plan in full. The rewrite is the right call — the file is the full receipts record — but the line reads at odds with it. One clause, "its design is dead; the file now records the autopsy," fixes it.

The measurements need no changes. The fix is a paragraph of truth about what exists, a footnote on the corpus, and two sentences. Fast.

figgyster (approved):
All five points answered, and answered with the project's own currency — receipts.

1. The phantom package is gone from the record. "Reproducing any of this" now says plainly: no prose code on main or anywhere it would be found by looking; generation 1 is named, located (`internal/prose` + `cmd/attn/prose.go` on `feat/prose-check`), and told to be closed; generation 2 is described as never committed and gone. "There is nothing here to delete and nothing to build on" — that's the sentence the cold reader needed, and it matches what I found in the tree.
2. The checklist stopped lying. "Implementation Steps" became "What was built, and where it is not", every `[x]` dropped, with an explicit "nothing below is a task list" guard. "Never built" replaces "Not yet done" and says flat out that no gate exists to unwire and no threshold code to delete. The glossary bullet now tells the truth and even sets policy: no prose entry until something ships.
3. The corpus got its pinning table, and it's a good one. Two pulls, dated, with file counts, message counts, and label sets; the 150-word floor subset explained (995 regex / 1,042 hand-verified); the 43-vs-49 mystery resolved by label-set, not by hand-waving. The companion doc cross-links the same pinning, so the two files no longer disagree.
4. "Refuse once per session" is now "The retry always lands", with the re-arm spelled out: one refusal per message, one guaranteed landing after, gate re-arms. The title now matches the mechanism.
5. "Safe to throw away" became "its design is dead; that file now records the autopsy" — the exact one-clause fix, done properly.

One note, non-blocking: the new tokenizer sentence says "one table below reads 21 complaints and another 22", but no table in either doc reads 21 — the complaint counts live in prose (22/1,020, 22 of 49, 22 times a month). If the 21-vs-22 split lived in the scratchpad runs, say so ("the two tokenizer runs disagreed at the floor, 21 vs 22 complaints; it never moves a result") rather than pointing at a table that isn't there. The pinning is otherwise airtight, and this is the only sentence I could not verify against the document itself.

Measurements untouched, still the honest core they were. This one's ready to merge.

figgyster (approved):
The note's fixed, and the fix has the same shape as the rest of this PR — the claim now lives where the evidence lives. "Two tokenizers were used across the scratchpad runs, one in Go and one in Python, and they disagree on two messages sitting exactly at the floor — the same recomputation reported 21 complaints from one and 22 from the other." No more pointing at a table that isn't there; the referent is the runs themselves, which is the truth of how the numbers came out. Verified the delta: one file, four lines, exactly this sentence. Everything else carries over from 6873a235.

And thanks for the correction on the rewriter table — good to have it on the record that the four "not measured" rows weren't a chosen gap but a CLI failure during scoring. The doc already carries that ("the rest died on a CLI failure while being scored"), so the record reads complete cold.

Delta vs main is still exactly the three docs files. Re-approving on the merge candidate.

PR #865, opened 2026-08-12
docs(plans): point the prose research at the branch that holds it

victorarias:
Follow-up to #863. Two pointers in those docs were about to dangle, and now resolve.

## What was wrong

The docs said the mining and experiment scripts "live in the session scratchpad" — that is `/private/tmp`, which gets swept. And the untried-framings list offered the replay harness as the way to answer "does standing guidance survive a long session?" without saying where that harness was, because it was nowhere.

## What changed

Everything is preserved on the branch `research/prose-gate-generation-2`, under `research/prose/`, with a README naming what each piece answered and the traps worth knowing before rerunning any of it. The docs now point there, including the direct path to the harness.

## Why keep a branch whose design is dead

The corpus cannot live in a public repository — it is private transcripts. But the **miners** can, and pointed at `~/.claude/projects` they rebuild the whole corpus in a couple of minutes. Keeping the tools is what makes the data disposable, and it is what makes every number in those docs repeatable by someone who was not there.

The one piece that outlives the gate is the replay harness: it rebuilds the conversation up to a message the user objected to, answers in their place with whatever intervention is being tested, and measures what comes back. Every real answer in the plan docs came out of it, including the one that killed the design.

The docs also now carry the warning forward: the miners label complaints by pattern match, and those labels are wrong more than half the time (49 flagged, 22 genuine). Every published number came from labels read by hand.

That branch must never merge. This PR is docs only.

https://claude.ai/code/session_01LA95N1YPYEzBk1UWLPFogP

### DIFF
--- a/changelog.d/prose-research-branch-pointer.yaml
+++ b/changelog.d/prose-research-branch-pointer.yaml
@@ -1,0 +1,6 @@
+kind: internal
+area: docs
+change: >
+  Pointed the prose-research plan docs at the branch that now holds the mining
+  and experiment scripts, so the measurements can be repeated without keeping
+  the private transcripts they were taken from.
--- a/docs/plans/2026-08-10-prose-density-gate.md
+++ b/docs/plans/2026-08-10-prose-density-gate.md
@@ -426,19 +426,23 @@ write still goes through.
 
 ## What was built, and where it is not
 
-**None of this is in the repository.** It was written and measured in a session
-scratchpad, and the measurements above are the only thing worth keeping from it.
-Nothing below is a task list; it is a record of what the numbers were produced
-with. Unticked boxes would imply work still owed, and none is.
+**None of this is on `main`.** The measurements above are the only thing worth
+keeping from it. Nothing below is a task list; it is a record of what the
+numbers were produced with. Unticked boxes would imply work still owed, and
+none is.
 
-Two generations of prose code exist, and neither is on `main`:
+Two generations of prose code exist, both on unmerged branches:
 
 - **Generation 1**, a rule engine — density, hidden-verb, rhythm and structure
   rules, accept/reject vocabulary files, golden tests, corpus fixtures — lives
-  unmerged on the branch `feat/prose-check` as `internal/prose` plus
-  `cmd/attn/prose.go`. The corpus measurements killed its design; the branch
-  should be closed.
-- **Generation 2**, described below, was never committed at all. If any of it is
+  on `feat/prose-check` as `internal/prose` plus `cmd/attn/prose.go`. The
+  corpus measurements killed its design before it was reviewed.
+- **Generation 2**, described below, is preserved on
+  `research/prose-gate-generation-2`, together with every mining and experiment
+  script under `research/prose/`. That branch must never merge. It is kept
+  because the miners rebuild the corpus from `~/.claude/projects` on demand,
+  which is what makes the private data disposable, and because the replay
+  harness there outlives whatever gets built next. If any of the gate itself is
   ever worth reviving it is the nudge text and the structural counter, both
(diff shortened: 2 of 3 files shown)


figgyster (changes requested):
Checked the branch claims against the tree before commenting, since this PR is literally about pointers that resolve. Almost all of it holds:

- `research/prose-gate-generation-2` exists, with `research/prose/` + README, the four miners, the Python experiments, and `internal/prosegate/harness_test.go` — which is env-gated (`ATTN_PROSE_CORPUS`/`ATTN_PROSE_HARNESS`) and does exactly what both docs say it does.
- `main` is clean: 404 on both `research/` and `internal/prosegate`. The gen-1 map is accurate (`feat/prose-check` carries `internal/prose` + `cmd/attn/prose.go`). README carries the 49-flagged/22-genuine warning verbatim. Changelog fragment fits convention; `git diff --check` clean.

One claim does not survive contact with the branch — in `docs/plans/2026-08-12-plain-agent-prose-first-principles.md`, "Reproducing any of this":

> What the branch keeps instead are the miners, which rebuild the whole corpus from `~/.claude/projects` in a couple of minutes: the paired rejected/accepted cases, the 1,250 messages with their replies, and the real ticket text (read out of the production database, read-only).

The miners do not rebuild the ticket text, because there is no tickets miner. The README table lists extract, silent, silentctx, features, glossary — nothing that touches ticket text. And the scripts that consume it, `exp4.py` and `exp5.py`, both open `../ticketreg/tickets.jsonl` — a path that exists nowhere on the branch (whole-tree grep for `ticketreg`: nothing). The old text said that file was "copied read-only from the production database" — a manual step, not a miner, and the README gives no way to reproduce it ("run from a directory holding the mined JSONL" does not cover it).

So the reproduction story works for the transcript corpus and breaks on the ticket experiments: exp4 (the 37% receipt-loss number) and exp5 (the unfinished rewriter comparison) would fail at first open with no pointer to where the input comes from. That is the one gap in "every number repeatable by someone who was not there" — the same claim-checking bar this PR exists to satisfy.

Small fix, one or two sentences: either say plainly that the ticket text was a read-only copy out of the production DB and is not regenerated by the miners (and where exp4/exp5 expect it), or drop it from the "miners rebuild the whole corpus" sentence. Everything else verifies.

figgyster (approved):
My request-changes came back answered with receipts, and I re-verified every claim against the tree before approving.

The ticket-text gap is closed three ways:

- The miner now exists: commit `d4c69a2` on `research/prose-gate-generation-2` adds `research/prose/mine_tickets.py`. It queries `tickets.description` and `ticket_events.comment`, enforces the same 100-word floor the experiments use, and writes one `{"kind": ..., "text": ...}` per line.
- The README on that branch documents it: copy-first commands (`cp ~/.attn/attn.db /tmp/attn-copy.db`, then `python3 mine_tickets.py ... > ../ticketreg/tickets.jsonl`), the warning that exp4/exp5 "fail at first open" without it, and the production-path refusal.
- The docs on this PR now say the true thing: the miners rebuild the *transcript* corpus from `~/.claude/projects`; ticket text has its own miner and one manual step — copy `~/.attn/attn.db` first, because production is read-only to agents and a live daemon is writing to it.

I checked the last claim against the code itself: the script refuses any path ending in `.attn/attn.db` outright (`sys.exit("refusing to read the production database; copy it first")`) and opens via `mode=ro` URI. Exactly what both the README and the plan docs say.

Delta vs main is still just the changelog fragment plus the two docs; `git diff --check` clean; head 1ddf91a matches steering. The reproduction story now holds end to end — transcript corpus from the miners, ticket text from a DB copy, both experiments fed, every number repeatable by someone who was not there. Approving.

PR #866, opened 2026-08-12
fix(apps): charge the app that broke the shared runtime, not its neighbours

victorarias:
An adversarial review of the A4 landing arc on `main` produced 19 candidate
findings; 9 were refuted on inspection. This is every surviving one that was
worth fixing, in one PR.

## The shared runtime blamed the wrong app

One supervised Bun process runs every installed app's handlers. Two ways an app
could break it for everybody and never be held responsible.

**A handler that never yields freezes the loop for every app.** The 60s dispatch
timeout then fired for apps whose handler was never reached, each was charged,
and the 15-minute stall clock would auto-disable innocent apps with a
notification telling their author to fix a handler that never ran.
`app.runtime.ping` existed for exactly this and had zero callers in Go.

A dispatch that times out now asks before it charges anyone. Ping answers → the
loop is turning, this app's own handler is what hung, charged as before. Ping is
silent → the loop is frozen, one handler is holding it, and only that app is
charged; everyone else is recorded as a runtime failure with its clock cleared,
and their error names the culprit.

**Which handler is holding it is the host's to say — the first build got this
wrong and live verification caught it.** Charging "the earliest dispatch" read
the daemon's own send order, which is not the order handlers hold the loop. Three
apps were dispatched into a runtime frozen by a 120s synchronous spin; the daemon
charged the app whose handler is `await ctx.collections.seen.put(...)` — the
documented shape — and let the spinner off. A handler that awaits an attn API
yields at once, so it is still unanswered when a spinner dispatched behind it
freezes everything, including that first handler's own reply.

The host now announces each entry on the socket immediately before calling the
handler (`app_runtime.entered`, a notification), and the culprit is the most
recent entry with no answer yet. An entry is dropped when its dispatch is
answered, when a ping answers, or when the process dies. This rests on one
measured fact: a Bun socket write issued immediately before a synchronous spin
reaches the peer at once rather than queueing behind it — with a 5s spin, the
frame arrived ~1ms after the write and ~5s before the spin ended.

**An unhandled rejection killed the sidecar for everyone, charged to nobody.**
The A4 plan ruled that a sidecar-wide crash is never attributed to an app,
reasoning that the runtime cannot know whose code caused it. That premise is
false, and the ruling made the worst thing an app can do the one thing it was
structurally exempt from being disabled for. **Overturned, and recorded as such
in the plan doc.** The host reads the app's name off the stack, reports it, and
exits; `appCrashStrikes = 3` inside `appCrashWindow` disables that app with a
notification.

The stack is the only honest witness and also the right one: Bun's
`async_hooks.createHook` fires no callbacks at all here and
`AsyncLocalStorage.getStore()` is `undefined` inside these handlers, both
measured — and a floating promise rejects long after its dispatch returned, so
"which app is running" routinely names an innocent. A crash naming no app is
charged to nobody.

Three, because `supervise.DefaultGiveUpAfter` (10) parks the whole sidecar after
~2 minutes of crash-looping and the culprit has to go first; above one, because a
single crash can be a machine event whose stack merely passes through a bundle.
`TestCrashStrikesFireBeforeTheSupervisorParksTheRuntime` pins that ordering.

## Rollback named ids nothing would show you

A rollback refusal names a version id, and no surface listed them. `attn app
status` now prints the version ids it takes, capped at the 10 most recent with
the total beside it.

## Receipts

`appRuntimePingWait = 2s` is a tripwire past a localhost round trip measured at
344µs and 416µs on a live daemon — ~5,000× — and is only ever spent on a dispatch
that already burned its whole timeout. The daemon's ping log prints microseconds
rather than milliseconds, because an answered ping rounds to `0s` and tells the
reader nothing about the margin.

`recentVersionLimit = 10` is ~200 bytes per `AppVersionInfo`, ~2KB total.

## Also in here

- `internal/hub`: a remote's daemon is revived before the app runtime host is
  shipped to it, rather than after.
- Four comments that had drifted from their code.
- One finding was deliberately skipped and noted in the plan doc: `apply`
  reports `Ok` when consumer wiring failed.

## Verification

Live on an installed non-production profile (`hubfix`), preflight PASS, protocol
233 in step across CLI, app and daemon. Four apps: a 120s synchronous spin, a
well-behaved app awaiting an attn API, one leaving a floating rejection, and one
awaiting something that never settles while the loop keeps turning.

- Frozen loop: the spinner is the one on the auto-disable clock; all three others
  get `runtime_error` naming it. One victim's dispatch started *before* the
  spinner's and is still correctly not blamed — the inversion that broke the
  first build.
- Turning loop: `apps: dawdler hit the dispatch timeout; the app runtime answered
  a liveness ping after 344µs` — charged to itself, correctly.
- Crash attribution: `apps: leaker crashed the app runtime
  (unhandledRejection), strike 1 of 3` … `strike 2 of 3`, and enabling an app
  clears its streak.

`appRuntimeAPIVersion` goes to 2 — the daemon↔host wire gained a frame and the
two are version-checked at hello.

Go suites (`internal/daemon`, `cmd/attn`, `internal/hub`) and the frontend suite
are green. Both new rules were mutation-tested: reverting the culprit rule to
"earliest entry" and dropping the answered-handler cleanup each fail a test that
names the real defect.

### DIFF
--- a/docs/plans/2026-08-12-a4-review-fixes.md
+++ b/docs/plans/2026-08-12-a4-review-fixes.md
@@ -125,25 +125,31 @@ the earliest dispatch therefore charges the best-behaved app in the common case,
 which is a worse failure than charging everyone.
 
 The daemon cannot work this out; only the host knows which handler it called. So
-the host announces each entry on the socket immediately before invoking the
-handler (`app_runtime.entered`, a notification — nothing to answer), and the
-culprit is the most recent entry with no answer yet. An entry is dropped when its
-dispatch is answered, when a ping answers, or when the process dies.
+the host announces each handler entering and leaving on the socket
+(`app_runtime.entered` / `app_runtime.left`, notifications — nothing to answer),
+entry announced immediately *before* the call, and the culprit is the most recent
+entry that has not left.
 
-An answered ping does not prove no handler is on the loop — a yielded handler
-that never settles is still there. It proves nothing is *holding* it, which is
-what makes the wipe safe: attribution is moot while the loop turns. The cost is
-one corner. If a handler yields, has its entry wiped by a ping, and only then
-resumes and spins, there is no entry naming it and attribution falls back to
-charging whoever timed out, until some new dispatch enters and is announced.
+Both halves are load-bearing. A first pass sent only `entered` and dropped
+entries when a liveness ping answered, which cost two corners: a yielded handler
+that came back and spun was named by nothing, and a stale entry outlived its
+handler. Both came from the same mistake — inferring that a handler left. The
+only thing the daemon can infer that from is the loop still turning, and a
+handler that yielded and never settled is on a turning loop, so the inference is
+wrong exactly for the handler that has to be named. The ledger is now dropped
+only on facts: an entry when the host says that handler left, all of them when
+the process dies. The price is one extra frame per dispatch, against a dispatch
+rate bounded by the bus fact rate.
 
 The receipt this depends on: a Bun socket write issued immediately before a
 synchronous spin reaches the peer at once rather than queueing behind it —
 measured with a 5s spin, the frame arrived ~1ms after the write and ~5s before
-the spin ended. Without that, the announcement would arrive too late to be the
-witness, and the design would not work at all.
+the spin ended. It has to be measured across two processes; a single-process
+probe cannot show it, because one loop cannot read what the spin blocks. Without
+that, the announcement would arrive too late to be the witness, and the design
+would not work at all.
 
-`appRuntimeAPIVersion` goes to 2: the daemon↔host wire gained a frame, and the
+`appRuntimeAPIVersion` goes to 2: the daemon↔host wire gained frames, and the
 two are version-checked at hello because they ship together.
 
 ## 3 — one app's unhandled rejection kills the sidecar for everyone (medium)
--- a/docs/plans/2026-08-12-a4-review-fixes.md
+++ b/docs/plans/2026-08-12-a4-review-fixes.md
@@ -62,8 +62,10 @@ Generosity costs nothing on the broken path: every ssh here carries
 `ConnectTimeout=10`, so an unreachable remote fails in seconds regardless.
 
 `TestRemoteBudgetsCoverTheArtifactsTheyCarry` fails if either budget stops
-covering its artifact, so growing the sidecar again cannot silently re-create
-this bug.
+covering the artifact sizes recorded in it. It does not measure the artifacts, so
+it does not notice one growing by itself — the tripwire is that a sidecar which
+outgrows its budget has to update those constants to pass, and updating them is
+when someone reads the arithmetic.
 
 ### Carried with it
 
@@ -126,8 +128,14 @@ The daemon cannot work this out; only the host knows which handler it called. So
 the host announces each entry on the socket immediately before invoking the
 handler (`app_runtime.entered`, a notification — nothing to answer), and the
 culprit is the most recent entry with no answer yet. An entry is dropped when its
-dispatch is answered, when a ping answers (the loop is turning, so nothing is on
-it), or when the process dies.
+dispatch is answered, when a ping answers, or when the process dies.
+
+An answered ping does not prove no handler is on the loop — a yielded handler
+that never settles is still there. It proves nothing is *holding* it, which is
+what makes the wipe safe: attribution is moot while the loop turns. The cost is
+one corner. If a handler yields, has its entry wiped by a ping, and only then
+resumes and spins, there is no entry naming it and attribution falls back to
+charging whoever timed out, until some new dispatch enters and is announced.
 
 The receipt this depends on: a Bun socket write issued immediately before a
 synchronous spin reaches the peer at once rather than queueing behind it —
--- a/docs/plans/2026-08-12-a4-review-fixes.md
+++ b/docs/plans/2026-08-12-a4-review-fixes.md
@@ -1,0 +1,150 @@
+# A4 landing-arc review fixes
+
+An adversarial review of the A4 app-registry/runtime arc as landed on main
+(#843 `4fbd6f71`, #858 `cc49fd95`, #859 `b873da05`) produced 19 candidate
+findings; 9 died against the code. This is what we do about the rest.
+
+The review report itself — every confirmed finding with its evidence, and the
+refutations, which are the more useful half — is the `a4-arc-review.md` artifact
+on ticket `journey-review`. This doc is only the disposition.
+
+Findings land on `epic/a4-review-fixes` in three PRs, because two of them change
+when an app gets auto-disabled and that is existing behavior worth reviewing as
+one diff.
+
+## 1 — the hub starved the step that revives a remote (high)
+
+`EnsureRemoteReady` shipped the ~93.7MB app runtime host *before*
+`ensureRemoteDaemonRunning`, both out of one 60-second budget the manager
+imposed at both call sites — including the one taken when the remote's daemon is
+down. On a link under ~20 Mbit/s the upload consumed the budget and execution
+never reached the revive, so the endpoint loop retried roughly every 90 seconds
+forever. Each attempt left up to 93.7MB in the remote's `/tmp`: the staging name
+carries pid+UnixNano so retries accumulate rather than overwrite, there was no
+cleanup on any error path, and `/tmp` is tmpfs on many systemd distros.
+
+A regression: before #859 a down remote whose binary already matched went
+straight to starting its daemon.
+
+**Fix.** Two phases, in the order that matters, each with its own budget:
+
+- `makeRemoteReady` — platform, versions, the attn binary, enrollment, daemon
+  start and its readiness wait. This is everything the remote needs to host
+  sessions.
+- `shipRemoteAppRuntime` — the sidecar, content-gated as before. It runs only
+  after the daemon is up, and its failure is reported rather than returned,
+  because a remote without a sidecar still runs sessions.
+
+The manager's blanket 60s is gone; the phases own their deadlines, since one
+number could only ever be right for one of them.
+
+### The receipts behind the two budgets
+
+Sized as tripwires past the slowest link anyone would sync a remote over —
+5 Mbit/s, a bad tether — not around a measured happy path. The previous 60s was
+sized for a ~40MB payload and #859 grew it to ~152.6MB without moving it, which
+is what made a healthy case reach it.
+
+| term | measured | at 5 Mbit/s |
+| --- | --- | --- |
+| attn binary (`attn-linux-arm64`) | 58,934,608 B | 94s |
+| app runtime host (`linux_arm64`) | 93,694,096 B | 150s |
+| daemon readiness wait | ≤ 35s (`remoteDaemonReadyTimeout`) | — |
+| sha256 of the runtime, local | 0.17s | — |
+| sha256 of the runtime, remote | 0.38s | — |
+| `bun build --compile` of the runtime | 0.13s | — |
+| bare ssh round trip (local VM) | 0.03s | — |
+
+`remoteReadyBudget = 180s` (94 + 35 + round trips). `appRuntimeShipBudget =
+300s` (150 + hashing + install + a possible runtime restart, generously).
+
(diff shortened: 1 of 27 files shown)


figgyster (approved):
Cross-examined the whole file, and the receipts hold up under questioning. The bug is real and the arithmetic is the witness: 152.6MB inside one 60s budget needs ≥2.54 MB/s with zero overhead, and the 35s readiness wait lives in the same context — a ~37s transfer and a 35s wait cannot both fit, so the math fails before link speed even testifies. And #859 (b873da05) did grow the payload without moving the budget, so this was a regression, not a design that never worked.

The fix has the right shape: order matters first, budgets second, and the sidecar's failure is reported instead of being allowed to take the daemon down with it. The two phases genuinely own their deadlines — the manager blanket is gone from both call sites, and the tests pin the property that matters, that the ship phase cannot inherit the ready phase's deadline (the remaining-time assertion catches the shared-lineage bug shape even if someone reintroduces it by derivation). The `context.WithoutCancel` cleanup defer is the right call: a dropped transfer cleans up even when the parent context died, and the staging name carries pid+UnixNano, so the leak was real and compounding.

Verified against the tree, not just the prose:
- `runSSHExit` signature matches the bounce's usage; `remoteAttnCommand`, `apps.RuntimeHostBinaryName`, `ConnectTimeout=10` all present.
- The JSON the bounce parses is real: `AppRuntimeStatusResult.Runtime *AppRuntimeInfo`, `Running bool json:"running"` — and a nil runtime (never started) correctly skips the bounce, so no Bun process gets spawned on a remote with no enabled apps.
- All three plan-doc commit receipts resolve: 4fbd6f71 (#843), cc49fd95 (#858), b873da05 (#859).
- Budget arithmetic matches the doc table: 94.3s + 35s = 129.3s < 180s; 149.9s < 300s.
- Diff clean, changelog fragment present, and CI ran gofmt + vet + the full Go suite on this exact head — the five new tests included — all green.

Two non-blocking observations for the record:
1. A parked runtime (crash-looping, daemon stopped retrying) also reads as "not running", so a replaced sidecar is not bounced onto it — it keeps the old inode until `attn app runtime restart` there. Same corner the old daemon-restart path left parked runtimes in, so no regression, just noting the file swap does not reach a parked runtime automatically.
2. The doc's out-of-scope environment figures (2.1GB binaries dir, 11G workers logs) trace to the review artifact on ticket `journey-review`, not to anything the tree can witness. Not doubting them — flagging that they are the one class of claim in here that is machine-measured rather than tree-verifiable, same bar this repo has been holding its own docs to lately.

Approve. The high-severity finding of the A4 review, fixed the way the autopsy said: revive the daemon first, budget the sidecar separately, and tell the truth about what failed instead of letting it kill the endpoint loop.

figgyster (approved):
Reviewed the rewritten head, not the one I approved this morning — the branch came back at 20:04 as the whole A4 review-fixes bundle, base moved to main. The hub half is byte-identical to what I already signed off (internal/hub/ unchanged between the old head and this one), so this review is about the new half. The new half holds up.

Verified against the tree, not the prose:

- The attribution design rests on one measured fact, so I measured it: a Bun socket write issued immediately before a 5s synchronous spin reaches a separate-process peer in the same millisecond, ~5s before the spin ends. Reproduced with two processes — the real apphost/daemon topology. (My first attempts ran server and client in one process, which fakes the failure: a shared loop can't read what the spin blocks. Those are the unreliable witnesses; the two-process run is the one that counts.)
- `app.runtime.ping` had zero Go callers on main; it is now the arbiter, served off the host's read loop without touching app code. The culprit rule — most recent entry with no answer, announced before the handler call — is pinned by tests that fail if you revert to earliest-entry or drop the answered-handler cleanup.
- Crash attribution: 3 strikes inside the stall window, supervisor parks at 10, a crash naming no app charges nobody, enable clears both clocks. The overturned A4 ruling is recorded in the plan doc beside the original, with the measured receipts (AsyncLocalStorage undefined, async_hooks silent) in the text.
- Rollback: the refusals already ended with the version list (`versionsSentence`); status now prints the same ids with the serving one marked. Fixed at the root, not in the copy.
- Protocol 233 untouched (optional field), generated.go/ts/tsp in lockstep, appRuntimeAPIVersion 1→2 on both sides of the host wire and version-checked at hello.
- New wedge/crash/status tests all pass; hub, cmd/attn, protocol suites green; diff --check clean; CI green all the way down (Daemon, Frontend, Changelog, Generated types, React Doctor, Bench A/B). One local daemon failure — TestDaemon_StartInstalledPlugins_SpawnsProviderPlugin — reproduces on main; it's the known plugin-runtime flake, not this PR.

Two non-blocking notes inline: the budget tripwire pins constants rather than the artifact, which makes the "cannot silently re-create this" claim read stronger than the test actually is; and one sentence about what a ping answer proves over-describes, in the plan doc and the code comment both.

internal/hub/bootstrap_phases_test.go
```diff
@@ -0,0 +1,140 @@
...
+		t.Fatalf("EnsureRemoteReady() = %v, want nil — the daemon is up and sessions work", err)
+	}
+	if len(logged) == 0 {
+		t.Fatal("a sidecar that could not be shipped must still be reported")
+	}
+}
+
+// The budgets are tripwires, not fits: a healthy sync must never be near them.
+// The receipts they are sized from live beside their declaration.
+func TestRemoteBudgetsCoverTheArtifactsTheyCarry(t *testing.T) {
```
figgyster, line 130: Receipt nit, same family as #863/#871: this test pins the artifact sizes as constants (58,934,608 / 93,694,096), so it fails if a budget stops covering the sizes recorded *here* — not if the sidecar grows. A host that grows to, say, 130MB sails through (208s at 5 Mbit/s, still inside 300s), and the plan doc's "growing the sidecar again cannot silently re-create this" is only true if whoever grows it also updates these constants. The headroom is real — 300s covers ~187MB at 5 Mbit/s, roughly 2x today's artifact — so this is a wording/receipt fix, not a code one. The honest version: "fails if either budget stops covering the sizes recorded here, so a sidecar that outgrows its budget must update this test to pass" — the update becomes the tripwire moment.

docs/plans/2026-08-12-a4-review-fixes.md
```diff
@@ -0,0 +1,232 @@
...
+Dispatch order is not the order handlers hold the loop. A handler that awaits an
+attn API yields immediately, so it is still unanswered when a spinner dispatched
+after it freezes everything, including that first handler's own reply. Blaming
+the earliest dispatch therefore charges the best-behaved app in the common case,
+which is a worse failure than charging everyone.
+
+The daemon cannot work this out; only the host knows which handler it called. So
+the host announces each entry on the socket immediately before invoking the
+handler (`app_runtime.entered`, a notification — nothing to answer), and the
+culprit is the most recent entry with no answer yet. An entry is dropped when its
```
figgyster: Papercut #4 of this very PR is "comments that describe something the code does not do", so: "when a ping answers (the loop is turning, so nothing is on it)" over-describes. A yielded handler that never settles is still on the loop — the ping proves nothing is *holding* it, and the wipe is still right because attribution is moot while the loop turns. But if that yielded handler later resumes and spins, its entry is already forgotten, and attribution degrades to the old self-charge until a new dispatch enters. The residue note in `attributeWedgedDispatch` covers the two-freezer corner but not this one — and the code comment's "each of which proves that handler is off the loop" makes the same claim a shade stronger than it is. Worth a clause on both.

figgyster (approved):
Reviewed the rewritten head, not the one I approved this morning — the branch came back at 20:04 as the whole A4 review-fixes bundle, base moved to main. The hub half is byte-identical to what I already signed off (internal/hub/ unchanged between the old head and this one), so this review is about the new half. The new half holds up.

Verified against the tree, not the prose:

- The attribution design rests on one measured fact, so I measured it: a Bun socket write issued immediately before a 5s synchronous spin reaches a separate-process peer in the same millisecond, ~5s before the spin ends. Reproduced with two processes — the real apphost/daemon topology. (My first attempts ran server and client in one process, which fakes the failure: a shared loop can't read what the spin blocks. Those are the unreliable witnesses; the two-process run is the one that counts.)
- `app.runtime.ping` had zero Go callers on main; it is now the arbiter, served off the host's read loop without touching app code. The culprit rule — most recent entry with no answer, announced before the handler call — is pinned by tests that fail if you revert to earliest-entry or drop the answered-handler cleanup.
- Crash attribution: 3 strikes inside the stall window, supervisor parks at 10, a crash naming no app charges nobody, enable clears both clocks. The overturned A4 ruling is recorded in the plan doc beside the original, with the measured receipts (AsyncLocalStorage undefined, async_hooks silent) in the text.
- Rollback: the refusals already ended with the version list (`versionsSentence`); status now prints the same ids with the serving one marked. Fixed at the root, not in the copy.
- Protocol 233 untouched (optional field), generated.go/ts/tsp in lockstep, appRuntimeAPIVersion 1→2 on both sides of the host wire and version-checked at hello.
- New wedge/crash/status tests all pass; hub, cmd/attn, protocol suites green; diff --check clean; CI green all the way down (Daemon, Frontend, Changelog, Generated types, React Doctor, Bench A/B). One local daemon failure — TestDaemon_StartInstalledPlugins_SpawnsProviderPlugin — reproduces on main; it's the known plugin-runtime flake, not this PR.

Two non-blocking notes inline: the budget tripwire pins constants rather than the artifact, which makes the "cannot silently re-create this" claim read stronger than the test actually is; and one sentence about what a ping answer proves over-describes, in the plan doc and the code comment both.

figgyster (approved):
Commit 48d173d4 answers both notes, and the answers are better than my wording — they name the corners instead of just dropping the overclaim.

- The budget test now says what it actually is: "fails if either budget stops covering the artifact sizes recorded in it. It does not measure the artifacts, so it does not notice one growing by itself — the tripwire is that a sidecar which outgrows its budget has to update those constants to pass, and updating them is when someone reads the arithmetic." That's the honest receipt, and making the constant-update the tripwire moment is the right mechanism — the person growing the artifact is the person who must re-read the arithmetic.
- The ping-answer sentence now proves only what it proves: "an answered ping does not prove no handler is on the loop — a yielded handler that never settles is still there. It proves nothing is *holding* it." And the residue I flagged is now documented as "Known residue, both cheap" in the code — the two-freezer case and the wiped-yielded-handler case, with the fallback behavior (self-charge until a new dispatch enters) stated outright. The inline wipe comment defers to the residue note instead of re-claiming.

Verified on the merge candidate: comment/doc-only delta, diff --check clean, hub suite green, all eleven wedge/crash/status/budget tests green. CI's one red job is the libghostty-vt asset-download flake (connection died five times, zig fallback with no zig) — the same infra flake that's been biting the Linux daemon arm64 gate all day, unrelated to this diff. Approving.

figgyster (approved):
Commit a8211917 doesn't document the residue I flagged — it removes the source of it. The host now announces `app_runtime.left` in a `finally` after the handler settles, and the ledger drops entries on facts only: the host says the handler left, or the process dies. The ping-wipe is gone.

That kills both corners at once, which is the part I like. The one I named — a yielded handler whose entry a ping wiped, coming back and spinning unnamed — is gone because a ping no longer wipes anything; `TestAHandlerThatYieldedIsStillNamedWhenItComesBackAndSpins` pins exactly that sequence, and it fails under the previous design. The one I didn't name, which the plan doc now does — a stale entry outliving a handler whose dispatch the daemon already abandoned — is gone because `left` arrives when the abandoned handler finally settles, where before nothing ever dropped that entry. The daemon stopped inferring, and both corners were inference bugs. The wire ordering holds up: `finally` runs before the async function's promise resolves, so `left` precedes the answer frame on the same FIFO socket, and for the spinner `left` never runs at all — the entry stays, which is the point.

Verified on the merge candidate: the process-death wipe survives (clearAppRuntimeConnection → forgetEnteredHandlers), the fake sidecar mirrors the real frame order (left before answer), all twelve focused tests plus the hub suite green, diff --check clean, and CI is green where it counts — both Linux daemon gates pass this round, the asset-download flake cleared. Also noted with appreciation: the plan doc now records that the load-bearing receipt has to be measured across two processes, because a single-process probe cannot show it — the exact trap I fell into this morning, written down so nobody else has to. Approving.

PR #867, opened 2026-08-12
feat(garden): plant a seed and see it appear

victorarias:
Slice 1 of the garden — "plant and see". An agent plants a seed in one line and
it shows up in the app, without anyone touching the app.

A **seed** is the unit of work: one document, one short id, a title, a markdown
body, a state. Anything worth handing off, parking, or attributing is a seed;
in-session scratch is not. The **garden** is all of them, and it belongs to a
home daemon. Plan:
[docs/plans/2026-08-06-the-garden-vertical-slices.md](docs/plans/2026-08-06-the-garden-vertical-slices.md).

## What ships

- **`internal/garden`** — what a seed *is*: the `core/garden` namespace, its two
  collections, the statuses, the id shape, the slug rule, the limits, and the
  markdown export. No daemon, no store, no wire: the CLI and a future in-app
  renderer cannot disagree about what a crown looks like.
- **`attn seed plant | ls | show | export`**, `--json` on all of them. Planting
  costs one line and returns the id, because the daemon already knows who is
  asking: the session id stamps the workspace, so an agent never names its own
  context. `export` renders a seed to markdown stamped as generated — the bridge
  that lets today's file-addressed review and annotation surfaces work on a seed
  until slice 6 renders the seed itself.
- **The full designed seed schema, declared now** — status, workspace, planter,
  tender, step slug, edges, and the still-inert `template`/`vars`/`gate` fields.
  Only a slice of it is behaviorally live; declaring it now is what keeps later
  slices out of a migration.
- **`garden.planted`** with the seed as its subject, projected to a
  `garden_seeds_updated` push. A bulk planting coalesces to one wire message.
- **A garden panel** listing the seeds of the workspace being looked at, with a
  toggle to the whole garden. It holds no fetch of its own: it renders what the
  daemon pushed.
- **The enrollment fence** on every seed command. On an outpost each one refuses
  by name, saying which home to run it on and how to leave the enrollment.
- **Glossary**: the garden, seed, plant, plot, crown, packet.

## Decisions this slice was asked to make

**Seed id shape**: `s-` plus six Crockford base32 characters — `s-7k3f9m`. The
alphabet drops `i`, `l`, `o`, `u`, so no two characters are confusable when read
aloud or copied by hand; six of them is 2^30 ids, which nothing a single garden
approaches. Minted from `crypto/rand` rather than derived from the title,
because a derived id renames itself when the title is corrected. No `/` anywhere
in the alphabet, which is what keeps qualifying an id at a federation boundary a
prefixing job (`<daemon-id>/<local-id>`) rather than a re-identification — the
arc's pre-commitment, pinned by a test that asserts a minted id is both a legal
document id and free of `/`.

**App transport**: the classic snapshot projection, not the docstore's live-query
surface. That surface is IPC-only (`handleDocSubscribe` speaks over `net.Conn`)
and the app speaks WebSocket, so riding it meant building the app-side
subscription seam first — A3.4 Stage 2's whole scope — to deliver a list of at
most a few hundred small documents. The whole garden is pushed newest-first and
the panel scopes client-side, so switching workspaces costs no round trip.
Revisit when the garden outgrows one push. Both rulings are recorded in the plan.

## Numbers and their receipts

Measured against the production `~/.attn` and this repo on 2026-08-12: 59
tickets, longest title 81 chars, mean 18; longest description 14,920 chars;
largest plan doc in the repo 75,843 bytes. So: title cap 400 chars, body cap
1 MiB, slug cap 100 — tripwires nothing healthy touches. Every refusal names the
limit, the value, and the ask.

## Live verification

Throwaway profile `gardenslice1`, installed from this branch; bundled preflight
PASS (0 failed, 0 warnings; CLI, app and daemon all on protocol 233).

A real Claude session in a real workspace ran

```
./attn seed plant 'the garden' -m - < docs/plans/2026-08-06-the-garden-vertical-slices.md
```

and the seed appears in the panel while the agent is still printing the id.
That, plus one seed per remaining slice, is the garden's first planting: the
plan for the garden now lives in a garden.

garden-slice1

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/74a8ad8a1e2e9236214d0b1e9086172d7e16533e/feat-garden-slice-1/garden-slice1.mp4)

Also exercised live: `plant → ls → show` round-tripping in JSON, workspace
scoping vs `--all`, `export` to a file and to stdout, and the refusals — an
id with a letter the alphabet excludes says which letter and why, an unknown id
says how to list the garden, and `ls` with nothing to scope to names both ways
out instead of silently answering for the whole garden.

**The bug live verification caught.** The first planting stamped every seed with
an empty workspace, so the panel filtered all of them away. A session's
workspace is decorated onto the record at broadcast time from the live registry;
the persisted column only leads while startup rebuilds it, and the planting read
the stored record. The test fixture had hidden it by writing the workspace
straight onto the stored session, which no live path does — it now associates
through the registry the way the app does, and the round-trip test fails when
the stamp reads the wrong side.

## Surfaces walked

- **CLI** — `attn seed` is the agent surface; every subcommand takes `--json`.
- **Daemon and app** — three commands, one event, the fence, the projection, and
  the panel that renders it.
- **Protocol** — `main.tsp` → `make generate-types` → `ProtocolVersion` 233, and
  `useDaemonSocket.ts` with it.
- **Linux** — `make build-linux-amd64` green; nothing here is Darwin-only.
- **Docs** — glossary entries, the plan's slice-1 ledger and its two rulings, and
  a changelog fragment.
- **Plugins/SDK** — untouched; whether extensions may reach `core/garden` is a
  named open question in the plan, deliberately still open.

## Known and deliberate

- A seed removed through the generic `attn doc delete` does not refresh the
  panel: the projection listens for `garden.*`, and the docstore's own
  `document.changed` is not a garden fact. Nothing in slice 1 removes a seed
  through a garden verb; slice 2's lifecycle verbs each publish their own fact.
- No `Qualify()` helper for the `<daemon-id>/<local-id>` form. Slice 1 has no
  boundary to cross, so the pre-commitment is pinned by a test rather than by
  code with no caller.


figgyster (approved):
**Approved.** Slice 1 plants, and the seed shows up where the app is looking.

Verified against the tree, not the description:

- **Protocol 233 holds end to end.** `constants.go` and `useDaemonSocket.ts` both at 233; `generated.go`/`generated.ts` in lockstep with the new `main.tsp` models (Seed, SeedEdge, SeedVar, the three messages, GardenSeedsUpdatedMessage, seeds on `initial_state` and WebSocketEvent); `TestFrontendProtocolVersionMatchesDaemon` passes. The old native-ui `lib.rs` bump is moot — the native client is archived on main, so there's no Rust handshake left to drift.
- **The id shape is what the body says.** `s-` + six Crockford characters, unbiased modulo (256 is a multiple of 32, no rejection loop), no `/`, docstore-legal — pinned by `TestNewIDIsStorableAndQualifiable`, including the 500-mint distinctness check.
- **The limits carry their receipts.** 400 ≈ 5× the measured 81-char longest title, 1 MiB ≈ 14× the 75,843-byte plan doc, slug cap past anything measured. Refusals name the limit, the value, and the ask — `TestValidatePlantNamesEveryRefusal` checks the numbers are actually in the messages.
- **The workspace-stamp fix is the real thing.** `d02ca45a` swaps `store.Get` for `decoratedSession`, and the fixture now associates through the registry with the persisted column left empty, so the round trip fails when the stamp reads the wrong side. The bug live verification caught now has a witness that can't be faked by the fixture.
- **One planting = one push; bulk coalesces to one.** `TestGarden_PlantingPushesTheGardenOnce` and `TestGarden_BulkPlantingCoalescesToOnePush` pin both sides of the panel's whole contract. The fence refuses all three verbs on an outpost, naming the surface, the home id, `attn enrollment leave`, and the plan — and an outpost's `initial_state` carries zero seeds.
- **The panel holds no fetch.** It renders what the daemon pushed, scopes client-side, shows the whole garden when there's no workspace to scope to, and its empty state names `attn seed plant`. All four scope tests pin it.
- **Docs.** Glossary carries the six terms; the plan ledger ticks all seven slice-1 items with both rulings recorded; the dead prose-gate repoint matches the #863–#865 sweep. Changelog fragment conforms.

Ran the box: `internal/garden` ok; all nine daemon Garden tests pass; the protocol version test passes; garden frontend tests 7/7; the six touched App test files 28/28; tsc clean; vet clean. CI green all the way down — 14 jobs including Generated types, E2E ×3, and Linux daemon on both arches; Plugins/Tauri path-skipped, correctly.

Two local notes, neither this PR's doing: the two plugin-runtime daemon tests fail on this box identically on `main` (pre-existing environment thing — CI's Daemon job is green), and the committed rapid `.fail` file is the tool's own format, trailing space and all, doing its job as the regression pin for the "ǅ" slug case.

One non-blocking nit: the id-comment's collision arithmetic (`internal/garden/garden.go:147-149`) says ten thousand seeds give "about 5e-5" chance that any two collide. Birthday math says ~4.5% (n(n−1)/2N with N = 2^30), and the per-planting reading against 10k existing seeds is ~1e-5 — 5e-5 is neither. The design is safe either way: planting is create-only, the store refuses the second write, and the planter mints again — `TestGarden_PlantingIsCreateOnly` pins that. Comment fix only, but this repo's bar is arithmetic with receipts, and that sentence misses by three orders of magnitude.

The garden's first planting is the plan for the garden, planted from a real session. Good dogfood. Looking forward to slice 2 giving the statuses something to do.

figgyster (approved):
**Re-approved on the merge candidate** (50bf6c0e) — both new commits verified against the tree.

**50bf6c0e — mint again.** My arithmetic nit came back as a behavior fix plus the corrected numbers, and the correction is right this time: 1e4/2^30 ≈ 1e-5 per mint at ten thousand seeds, n(n−1)/2N ≈ 4.7% for some pair over the garden's life, and 32^6 = 1,073,741,824 exactly. And the code now matches what the old comment only promised: `mintAndPlant` retries on `docstore.IsConflict` (real — internal/docstore/docstore.go:165), three tries as a tripwire, and a mint source that only repeats itself gets a refusal naming the cause instead of the planter eating a coin-flip refusal. `TestGarden_PlantingMintsAgainWhenAnIDIsTaken` pins both paths — collision recovers on a fresh mint, stuck source names itself. All eleven daemon Garden tests pass locally, `internal/garden` ok.

**5eb5298b — sidebar honesty.** `activeRemoteSession` was read in the actions memo but never listed as a dependency — a stale-answer bug of the exact kind the memo exists to prevent, and the one-line dep addition is the honest fix. Dropping the `export` from `formatPlantedAt` is safe: nothing outside GardenPanel.tsx imports it (grep confirms), and Fast Refresh gets its state preservation back. tsc clean; the new head's React Doctor job is 0 errors / 0 warnings.

CI on 50bf6c0e: 14/14 green — Daemon, Frontend, E2E ×3, Generated types, Changelog, Changes, Bench A/B, Linux daemon on both arches, React Doctor. Plugins/Tauri path-skipped.

One observation, not a finding: the mint-again retry is the right call, but it means a collision is now *silently* retried inside the daemon — three mints max, so the worst case is a refused planting with an honest cause. If you ever want the sync engine (or a plot planting a hundred seeds) to know a collision happened, the fact log is where that would show. Slice 2 material at most.

Garden's planted, the receipts check out, and the loop closed faster than the review cooled. On to slice 2.

PR #868, opened 2026-08-12
fix(annotations): align across rendered links

victorarias:
## Intent

Long Codex answers can contain Markdown links whose absolute targets are rendered in the terminal as shorter paths. The annotation projection previously used one greedy token alignment; the first rewritten link could stop alignment, making later agent-written prose fail with the generic ownership refusal.

## Summary

- Parse Markdown link labels and destinations into source-offset-preserving tokens, then match them to visible path suffixes or OSC 8 URI metadata.
- Build a monotone chain from unique anchors and fill each local segment, so one renderer substitution leaves a sparse hole instead of truncating the rest of the answer.
- Preserve the reverse-selection and paint-containment gates, and explicitly exclude user-prompt rows.
- Add Codex-style multi-link regressions, a design note, and a changelog fragment.

## Verification

- `pnpm --dir app test` — 2,721 passed, 15 skipped.
- `pnpm --dir app exec tsc --noEmit`.
- Pre-commit frontend build.
- `pnpm --dir app run e2e` — 196 passed, 1 skipped.
- Packaged isolated app, bundled preflight, and the real Codex terminal-annotation scenario — passed; the annotation survived an app relaunch and a second turn.
- Production-snapshot replay — the visible assistant tail aligned from source offsets 5,876–9,638 instead of only 8,069–9,638; mean alignment time was 0.79 ms over 200 runs.

### DIFF
--- a/app/src/components/GhosttyTerminal.tsx
+++ b/app/src/components/GhosttyTerminal.tsx
@@ -1167,6 +1167,12 @@ export const GhosttyTerminal = forwardRef<GhosttyTerminalHandle, GhosttyTerminal
         totalRows: () => terminal.getScrollbackLength() + terminal.rows,
         rowText: (row) => selectionLineAtBufferRow(row, 0, terminal.cols),
         rowTextRange: (row, startCol, endCol) => selectionLineAtBufferRow(row, startCol, endCol),
+        hyperlinkUri: (row, col) => {
+          const history = terminal.getScrollbackLength();
+          return row < history
+            ? scrollbackHyperlinkUri(terminal, row, col)
+            : hyperlinkUriAt(terminal, row - history, col);
+        },
       };
     }, [selectionLineAtBufferRow]);
 
--- a/app/src/utils/terminalAnnotations.gate.test.ts
+++ b/app/src/utils/terminalAnnotations.gate.test.ts
@@ -367,4 +367,46 @@ describe('anchorForSelection', () => {
     expect(anchor!.messageKey).toBe('turn-1');
     expect(older.slice(anchor!.start, anchor!.end)).toContain('retry wrapper');
   });
+
+  it('accepts agent prose after a Markdown link was rendered as a shortened path', () => {
+    const markdown = 'Evidence: [source](/Users/tester/src/services-pilot/a/b/Source.java:19). '
+      + 'The conclusion after the source remains annotatable.';
+    const rows = [
+      '• Evidence: a/b/Source.java:19.',
+      '  The conclusion after the source remains annotatable.',
+    ];
+    const store = new TerminalAnnotationStore();
+    store.setMessages([{ key: 'turn-1', markdown }]);
+    const grid = new FakeGrid(rows);
+    const startCol = rows[1].indexOf('conclusion');
+
+    const anchor = store.anchorForSelection(grid, {
+      startRow: 1,
+      startCol,
+      endRow: 1,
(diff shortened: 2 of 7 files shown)


figgyster (approved):
Approve.

The case: the greedy walk dies on a run of rendered links. I checked the old arithmetic by hand — four consecutive `[label](/abs/path.java:NN)` substitutions each burn the RESYNC_WINDOW=4 lookahead without a match, so the walk stalls at the first rewritten link and every word after it reads as TUI chrome. The symptom in the changelog is the failure mode in the code.

The fix replaces the single diagonal with a monotone chain of unique anchors. I walked the DP like an alibi:

- Keys are namespaced (`target:`/`path:`/`path2:`), so a shortened path's basename can't collide with a plain word's norm. The Fenwick max is batched by source index, so one source token can't double-anchor through two of its own aliases. The exclusive grid query is what makes a visible label beat its own OSC 8 target — the label-precedence test is testing that ordering, and the ordering is real, not incidental.
- The chain is monotone in both indices. Segments are bounded by anchors on both ends and can't overlap them; the per-island gap budget means a long message no longer spends its whole resync allowance in one place.
- `explicitUser` matches the repo's own capture corpus (`• Working\n\n› Explain this codebase`) — `›`/`❯` rows are the user's, and hard-excluding them closes the loop the changelog names: a selection landing on the user's own prompt now refuses at selection time instead of minting an annotation that never paints.
- The gate got stricter in the right direction: `quotesAnchor` moved from string-substring to token-sequence containment, so a mid-word slice the old `includes` would have waved through now refuses. Token-boundary containment still holds in both directions — I poked that too.

Verified against the tree: `diff --check` clean; the two touched suites pass 43/43; the full frontend run is 2,716 passed / 20 skipped across 243 files (your 2,721/15 is a five-test environment-skip delta, zero failures); `tsc --noEmit` clean; CI green on the exact head — frontend plus all three E2E shards, react-doctor and changelog checks; daemon/tauri/plugins correctly skipped for a frontend-only change. I also built a case the suite doesn't cover — prose *between* two rendered links — and the middle stays annotatable, with the full destination mapping back to markdown offsets.

Two non-blocking notes:

- `normalizedWords` is dead in production now — only the projection tests still call it. Harmless, but it's a test helper living in the source file.
- The scary case I checked: an alias-only (strength-1) anchor can split a segment on a pairing that isn't the true one. Uniqueness on both sides is what keeps that rare — two links to files sharing a basename+line make the key non-unique and it drops out of the anchor set — and what leaks through is still bounded by the 0.6 row-confidence and the containment gate. "Candidates, not proof" is doing real work in that comment, not just apologizing.

figgyster (approved):
Re-approving on 54d57066.

The new commit is the weak-alias case from my first pass, fixed with receipts: basename-only identity is gone, so `wrong/Shared.java:42` no longer anchors a link to `/repo/right/Shared.java:42` — the two new tests pin both the distinct-by-directory mapping and the wrong-directory refusal, and `quotesAnchor` refuses the basename-only containment. The trade is documented in the plan doc (final directory + filename must survive), which is the honest version of the old comment's "candidates, not proof". And `normalizedWords` moved into the test file where it belonged — the dead-export nit, closed.

Verified: diff --check clean; touched suites 45/45; full frontend run 2,718 passed / 20 skipped on this head; tsc clean; CI green across frontend, all three E2E shards, react-doctor, and changelog checks.

PR #869, opened 2026-08-12
fix(delegate): validate models and checkout placement

victorarias:
## Intent

Delegation could pass preflight with a model identifier that the selected agent did not resolve as intended, then silently start on a fallback model. Repository placement was also coupled to workspace placement, so `--workspace --no-worktree` could reuse a checkout from an unrelated repository.

This makes both decisions explicit and rejects ambiguous input before launching an agent.

## Summary

- add one built-in model resolver shared by `preflight` and `delegate`
- resolve supported Claude aliases to canonical model IDs and reject unknown built-in IDs with supported choices and suggestions
- keep plugin model identifiers driver-owned
- make the source checkout authoritative for `--no-worktree`, including when the pane is placed with `--workspace`
- reject conflicting `--cwd`/`--repo` and existing `--worktree-path` repository inputs with both resolved locations
- clarify workspace placement versus repository/worktree placement in CLI and bundled delegation guidance
- add regression coverage for model aliases, invalid identifiers, mixed workspaces, repository conflicts, and valid worktree combinations

## Verification

- `go test ./cmd/attn ./internal/client ./internal/agent ./internal/preflight`
- focused `internal/daemon` delegation tests
- freshly built CLI checks for canonical alias resolution, invalid model rejection, and updated help output
- pre-commit Go build

The full Go harness and packaged-app install could not complete because Spotify Artifactory returned `401` while fetching TypeScript/frontend dependencies. All unaffected runnable packages and daemon shards passed.

### DIFF
--- a/changelog.d/delegate-admission-and-placement.yaml
+++ b/changelog.d/delegate-admission-and-placement.yaml
@@ -1,0 +1,12 @@
+kind: fixed
+area: cli
+change: >
+  Delegation now rejects unknown built-in model identifiers before launch and
+  resolves supported aliases through the same catalog as preflight. Workspace
+  placement no longer silently chooses a different repository when
+  --no-worktree is used: the source checkout remains authoritative unless
+  --cwd explicitly selects another checkout, and conflicting Git locations
+  report both resolved repositories before any agent starts.
+symptom: >
+  A shorthand model could pass preflight but select an unintended fallback, or
+  --workspace --no-worktree could start an agent in another repository.
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -742,18 +742,18 @@ task source:
   --ticket <id>               adopt an existing ticket and use its description
   --confirm                   take over a ticket with a non-orphan assignee
 
-placement:
-  (no flags)                 add a pane to the source workspace; Git repositories
-                             get a new worktree automatically
-  --new-workspace            create a workspace using the source directory
-  --workspace <id>           add a pane to an existing workspace
-  --cwd <path>               create a workspace at an existing directory
-  --worktree <branch>        choose the new worktree's branch
-  --no-worktree              reuse the resolved checkout instead
+workspace placement (where the pane appears):
+  (no flags)                 add a pane to the source workspace
+  --new-workspace            create a workspace for the delegated pane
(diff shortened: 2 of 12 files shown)


figgyster (approved):
Approve. This closes two holes I've been watching: preflight blessing a model the delegate then silently downgrades, and `--workspace --no-worktree` inheriting a checkout from whatever repo the workspace's registration artifact happened to point at.

Verified against the tree and the running binary:

- `ResolveLaunchModel` is genuinely shared — preflight and the daemon resolve through the same catalog, and the preflight report carries the canonical ID back (`opus` → `claude-opus-5` in the live report, no model check failure).
- Rejections are legible: `opus-5` → "did you mean `claude-opus-5`?", `gpt-5.6` → "did you mean one of luna/terra/sol?", and the old documented `claude-opus-4-8` gets pointed at `claude-opus-5` instead of dumped.
- The placement fix is exactly as scoped: the client always sends a worktree request unless `--no-worktree`, so plain `--workspace` still creates a worktree (behavior preserved); only the nil-request path now means "stay in the source checkout". The mixed-workspace test pins it, and I traced the spawn cwd to confirm the session actually runs in the source checkout, not just the record.
- Conflict errors fire before any side effect — the worktree-path check runs before git touches anything, the `--cwd`/`--repo` check before worktree creation — and both surface through the durable operation as a terminal Failed with the error attached; nothing strands in Preparing.
- Ran the claimed suites: `cmd/attn`, `client`, `agent`, `preflight`, the full daemon delegation set (the three new tests pass), `go vet`, `git diff --check`, and `go test ./...` — 57 packages green. The only two failures are the plugin-runtime daemon tests, which fail identically on current main; pre-existing env flake, not this PR. Built the CLI and exercised the help text plus all three resolution/rejection paths live. CI green across the board on this head.

One note, non-blocking: the catalog is now the admission authority for built-in models, and nothing in the tree ties it to what the driver actually accepts. A provider shipping a new model means hard rejection until someone updates the catalog; a retired model left in the list sails through and the silent-fallback risk this PR kills reappears one layer down, at the driver. A test pinning the catalog against the driver's accepted set — or at least an update-contract comment on the catalog — would close that last gap. The intent is right; just make sure the catalog can't quietly go stale.

PR #870, opened 2026-08-12
docs(agents): a comment claims only what this commit does

victorarias:
The agent guide's comment-discipline bullet said what a comment is for — a constraint, an invariant, a measured receipt — but never that the claim must be true of the commit that carries it. That gap has a measured incentive, not just a hypothetical one: an agent writes the comment for the design it intends, ships the part that was in scope, and the comment silently becomes a claim about code that does not exist. Two instances of that class shipped in garden slice 1 (#867) — an id comment promising a mint-again retry nothing implemented (caught in review, became a real fix) and a log line promising a panel behavior no consumer rendered (caught by a post-merge adversarial review).

One addition to the bullet: a comment claims only what this commit does; an unbuilt intention belongs in the plan doc or as a named gap in the PR body, never in a comment.

Docs and changelog fragment only; no behavior changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -377,8 +377,9 @@ forgets that position and `--since <RFC3339>` replays from an instant.
   comment claims only what this commit does: describing the design you
   intend, then shipping the part in scope, quietly turns the comment into a
   lie. An unbuilt intention belongs in the plan doc or as a named gap in the
-  PR body, never in a comment. Godoc on an exported symbol is one line. A package header is a few lines plus a
-  link to the design doc, never a retelling of it. Never narrate the next line
+  PR body, never in a comment. Godoc on an exported symbol is one line. A
+  package header is a few lines plus a link to the design doc, never a
+  retelling of it. Never narrate the next line
   or argue that the change is correct: that talk belongs in the PR, and a
   comment addressed to the reviewer is a defect.
 - Conventional commit titles with a scope, in plain language:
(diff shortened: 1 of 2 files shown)


figgyster (approved):
Both receipts in the body check out against the tree. The #867 seed-id comment that promised a mint-again retry nothing implemented — that one became 50bf6c0e after review. And the `garden: %d seeds, pushing the newest %d (limit %d); the panel reports the rest` log line, where the panel is fetch-free and no consumer renders "the rest." Two real instances of the same class, now a rule. That is how a norm should get written: after the second body, not before the first.

The commit practices what it preaches — adds the rule and the changelog fragment, nothing else, and the unbuilt intentions stay in the PR body where the new rule says they belong. Changelog gate passes, CI green on the head.

One non-blocking nit: the grafted sentence leaves a 110-char line where the file wraps at ~80 — "Godoc on an exported symbol is one line. A package header is a few lines plus a" all ran together. Cosmetic, and the file already carries longer lines elsewhere, but it is a one-line rewrap in a PR whose whole subject is doc discipline.

figgyster (approved):
The rewrap lands on the new head: bullet region now maxes at 77 chars, wrapped to the file width, and the 110-char graft is gone. Changelog gate and CI green on 3cedab3. That closes the one nit — and it is the third time today the nit-fix loop paid out in a real commit, not prose. Still good.

PR #871, opened 2026-08-12
test(bus): fail when a projection stops writing to the wire

victorarias:
## The hole

`TestWireTrafficComesFromProjections` closes one direction of the event-bus
boundary: a function that reaches the hub without a fact behind it fails.
Nothing failed a projection whose push disappeared.

Measured with `go test -overlay`, deleting only the send:

| mutation | `go test ./internal/daemon -count=1` |
| --- | --- |
| `broadcastMessage` deleted from `projectGardenSeeds` | **ok**, 146.9s |
| `broadcastMessage` deleted from `projectTicketsUpdated` | **ok**, 129.5s |

The shape is what hid it, not the feature. A projection whose message is its own
top-level type (not a `WebSocketEvent`) is invisible to the hub's
`broadcastListener`, so garden, the ticket board and the workflow run each grew a
test-only broadcast hook that fires immediately before the real send — and every
assertion in those features reads the hook. Delete the send, keep the hook, and
nothing notices. Each new feature copies the shape.

## The mirror

`TestEveryProjectedFactReachesTheWire` (`internal/daemon/bus_wire_join_test.go`).
For every fact a `wireProjections` entry matches, it builds a daemon with a
seeded session, workspace, layout and presentation, taps `wsHub.wireTap` — which
sees the marshalled bytes of all five send paths, and no hook — publishes that
one fact, and asserts the **exact set** of `event` names that reached the wire.

Three things make it hold without maintenance:

- The fact vocabulary is parsed out of `bus.go` (`Fact…` constants) and the
  projection table is read at runtime, so a new fact — including one falling under
  an existing wildcard like `ticket.*` — is **seen** without anyone remembering
  this test exists.
- Being seen is what makes it loud: a projected fact with no fixture fails **by
  name**, saying what to add, and a fixture whose fact no longer exists fails as
  stale. That fixture is a single line — no subject, no payload — for **41 of the
  79** facts, which is **14 of the 41** projection entries covered whole. A
  subject or payload appears only where the projection re-reads an entity or
  decodes a body.

  (Corrected after review: an earlier version said "26 of the 41 projections push
  against a bare daemon and a bare fact". That came from the probe used to size
  the work — an empty world, counting anything that reached the wire, including a
  push whose event name was empty — so it did not reproduce by reading the table.
  The figures above are counted from the fixtures as written.)
- A declared fact matched by *no* projection must be listed in `factsWithoutWire`
  with the consumer that does read it — the same enumerated-exception discipline
  `wireSenderExceptions` already uses. Without it, a projection deleted by
  accident is indistinguishable from a fact that never had one.

`TestWireTrafficComesFromProjections` is untouched.

### Where the line sits: the event name, and deliberately nothing deeper

It has to be at least that deep. Publishing `workspace.context.changed` with no
payload decodes to a zero value and pushes a message with an **empty** event
field — "pushed something" accepts that. And the event name is precisely what
copy-pasting a projection gets wrong, which is the failure mode this exists to
close.

It should not go deeper here. Payload bytes already belong to the wire-trace
goldens and to each feature's own tests. One table owning every feature's payload
would rot on the first field added and would be regenerated rather than read.

Asserting the *complete* set rather than "at least one" also catches the opposite
defect: a second projection entry matching the same fact and double-pushing.

## Receipts

All `go test ./internal/daemon -count=1`:

| run | result |
| --- | --- |
| garden push deleted, this test absent | ok, 146.9s |
| garden push deleted, this test present | **FAIL** — `garden.planted` put nothing on the wire |
| ticket board push deleted, this test absent | ok, 129.5s |
| ticket board push deleted, this test present | **FAIL** — every `ticket.*` fact put nothing on the wire |
| unmutated | ok, 185.0s |

The "absent" runs replace the new test file with a bare `package daemon` through
the same overlay, so before and after are measured in one tree.

```
--- FAIL: TestEveryProjectedFactReachesTheWire/garden.planted
    publishing "garden.planted" put nothing on the wire; its projection is
    contracted to send [garden_seeds_updated].
    Either the projection stopped sending — which is the defect this test exists
    for — or it needs a subject or payload the fixture does not give it.
```

**The guard is not vacuous.** Four mutations of the mechanism itself, each
failing with a message naming what to fix: a removed fixture (`"garden.planted"
is projected but has no wireFixture`), a fixture for a fact that does not exist
(`not a declared fact constant in bus.go`), a fixture declaring the wrong event
(`put [garden_seeds_updated] on the wire, want [tickets_updated]`), and a deleted
`wireProjections` entry (`has no projection and no entry in factsWithoutWire`).

**One correction to the report this came from.** The third hook-shaped feature,
`workflow_broadcast.go`, was inferred to have the same hole and does not: deleting
its send fails `TestWireTraceProducerGolden`, because a workflow run is one of the
producers that golden drives. That is exactly the difference between a
hand-maintained producer list and a table-complete one — garden was merged without
an entry in it, and nothing noticed.

`go vet ./internal/daemon` clean; the new test passes under `-race` (3.7s).

## What a future projection author has to do

Add a one-line fixture naming the events the projection sends — and nothing else,
in the common case, because 41 of the 79 existing fixtures are exactly that one
line. There is nothing to remember: the test fails by name until the fixture is
there, and says what to supply. It is never silent.

(Corrected after review: the first version of this section claimed "nothing",
which was true of being *noticed* and not of being *covered*.)

## Live verification

Exempt: test-harness only. No production file changed — the diff is one new test
file, a changelog fragment, and two documentation edits. `wireTap` already existed
in `wsHub` for the wire-trace goldens; nothing was added to production wiring, so
there is no app-observable behavior to verify.

https://claude.ai/code/session_01S99Ao2Ckrnzw976ghoUhtX

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -495,7 +495,12 @@ consumer — runs the matching entry in `wireProjections`
 (`internal/daemon/bus.go`) to produce the wire traffic, often a snapshot
 re-push. Every state-change broadcast goes through it;
 `TestWireTrafficComesFromProjections` fails on a new one that does not, and
-carries the enumerated exception list.
+carries the enumerated exception list. Its mirror,
+`TestEveryProjectedFactReachesTheWire`, publishes every fact that has a
+projection and reads the bytes the hub sent, so a projection that stops sending
+— or sends the wrong event — fails too. Both are driven by the live table: a new
+projection is covered without touching either test, and a fact reaching no
+projection has to name the consumer that does read it.
 
 - A fact without a subject is a snapshot invalidation, not a fact. If the
   producer does not know the entity id, that is the bug to fix — change the
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -498,9 +498,10 @@ re-push. Every state-change broadcast goes through it;
 carries the enumerated exception list. Its mirror,
 `TestEveryProjectedFactReachesTheWire`, publishes every fact that has a
 projection and reads the bytes the hub sent, so a projection that stops sending
-— or sends the wrong event — fails too. Both are driven by the live table: a new
-projection is covered without touching either test, and a fact reaching no
-projection has to name the consumer that does read it.
+— or sends the wrong event — fails too. Both are driven by the live
+table, so a new projection cannot go unnoticed: it needs a one-line fixture
+naming the events it sends, the test fails by name until it has one, and a fact
+reaching no projection has to name the consumer that does read it.
 
 - A fact without a subject is a snapshot invalidation, not a fact. If the
   producer does not know the entity id, that is the bug to fix — change the
--- a/changelog.d/wire-join-every-projection-proved.yaml
+++ b/changelog.d/wire-join-every-projection-proved.yaml
@@ -1,0 +1,15 @@
+kind: internal
+area: testing
+change: >
+  The daemon's fact-to-wire join is now proved for every projection at once.
+  A projection could stop sending and nothing failed: three features assert
+  against a broadcast hook that fires just before the real send, so deleting
+  the send left the whole package green — measured, by deleting the garden
+  snapshot push and again the ticket board push. TestEveryProjectedFactReachesTheWire
+  publishes each fact that has a projection and reads the bytes the hub
+  actually put on the wire, asserting the exact set of events it produced. It
+  is driven by the live projection table and the fact constants themselves, so
+  a projection added later is covered without anyone editing the test; a fact
+  that reaches no projection has to say why. This is the missing half of
+  TestWireTrafficComesFromProjections, which only ever caught traffic that
+  skipped the bus.
--- a/docs/plans/2026-08-01-ext-a1-event-bus.md
+++ b/docs/plans/2026-08-01-ext-a1-event-bus.md
@@ -339,3 +339,70 @@ CLI, app and daemon.
   `setting.changed` fact and one `settings_updated` carrying
   `changed_key=auto_settle_enabled`, while the startup tailscale and backup
   facts pushed settings with no changed key, as before.
+
+## Proving the join, both directions (2026-08-12)
+
+`TestWireTrafficComesFromProjections` only ever closed one direction: it fails a
+function that reaches the hub *without* a fact behind it. Nothing failed a
+projection whose push disappeared. Measured with `go test -overlay`: delete the
+`broadcastMessage` call from the garden snapshot projection and
+`go test ./internal/daemon` is green in 147s; do the same to the ticket board's
+and it is green in 130s.
+
+What made it invisible is a shape, not a feature. A projection whose message is
+its own top-level type (not a `WebSocketEvent`) is unobservable through the
+hub's `broadcastListener`, so garden, the ticket board and the workflow run all
+grew a test-only broadcast hook that fires immediately before the real send.
+Every assertion in those features reads the hook. Delete the send, keep the
+hook, and nothing notices — and each new feature copies the shape.
+
+`TestEveryProjectedFactReachesTheWire` (`internal/daemon/bus_wire_join_test.go`)
+is the mirror. For each fact that a `wireProjections` entry matches, it builds a
+daemon with a seeded session, workspace, layout and presentation, taps
+`wsHub.wireTap` — which sees the marshalled bytes of all five send paths, and no
+hook — publishes that one fact, and asserts the **exact set** of `event` names
+that reached the wire.
+
+Three properties make it hold without maintenance:
+
+- The fact vocabulary is parsed out of `bus.go` (`Fact…` constants) and the
+  projection table is read at runtime, so a new fact under an existing wildcard
+  (`ticket.*`, `pr.*`, `garden.*`) is covered with no edit at all.
+- A projected fact with no fixture fails by name, and a fixture whose fact no
+  longer exists fails as stale. Most fixtures are one line — 26 of the 41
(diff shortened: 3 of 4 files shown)


figgyster (approved):
Approve. Ran the case myself, mutations and all — the receipts check out.

**The shape of the hole, confirmed.** garden.go:259, ticket_board.go:121, workflow_broadcast.go:82 all fire a feature-test hook right before the real send. Delete the send, keep the hook, the whole package stays green. That's a shape hiding the defect, not any one feature's fault.

**The observation point, checked rather than trusted.** wireTap predates this PR (testharness.go:309 wires it unconditionally) and sits on every broadcast path (websocket.go:532, 612); the two direct per-client sends are request/response traffic, not projections, so the tap covers all projection traffic. `NewForTesting` leaves the event bus nil, so `publishFact` runs projections synchronously — no timing, no races, which is why `-race` passes in seconds.

**The guard is not vacuous — I mutated it myself.** Deleted the `broadcastMessage` call in `projectGardenSeeds`: FAIL with the exact message quoted in the PR body ("garden.planted put nothing on the wire"). Same for `projectTicketsUpdated`: every `ticket.*` fact fails. Both mutations reverted after.

**The workflow correction is honest.** Deleting workflow_broadcast.go's send would fail `TestWireTraceProducerGolden`, because a workflow run is a golden step driven through the projection (wire_trace_producer_test.go:81). The report's self-correction holds, and it names the real difference: a hand-maintained producer list vs a table-complete one.

**The empty-event justification checks out one level down.** `bus.Event.Decode` returns nil on an empty payload (bus.go:52), so a payload-less `workspace.context.changed` decodes to a zero value and pushes a message with an empty event field — a "pushed something" assertion would swallow that. `EventNames()` maps empty to "?" rather than dropping it, so the exact-name assertion catches it. The depth argument lands: the event name, deliberately nothing deeper.

**Coverage arithmetic.** 86 declared facts: 79 with fixtures, 7 in `factsWithoutWire` with named consumers — consistent with the comments in bus.go itself. 40 of the 79 fixtures are bare one-liners. The exact-set assertion also catches the opposite defect: a redundant projection entry matching the same fact double-pushes, and the intentional `plugin.disconnected` dual-push is encoded as a two-event contract. Local: vet clean, full daemon suite fails only the two known plugin-runtime env flakes (fail on main too, pre-existing). CI green: Daemon, Changelog gate, Linux daemon amd64/arm64, Frontend E2E.

One non-blocking nit, and it's the class this repo just codified in #870: the docs overclaim by exactly one line. The PR body and the committed plan say a new fact under an existing wildcard is "covered with no edit at all", and the future-author section says "Nothing, in the common case" — but the loop requires a `wireFixture` entry for every projected fact. A new `ticket.*` fact fails by name ("is projected but has no wireFixture") until someone adds the one-line fixture. That's never silent and the message says exactly what to add — which is better than "no edit", not worse. Wording fix in the plan doc: "covered by a one-line fixture; the test fails by name and says what to add."

figgyster (approved):
Re-approve. Commit 5875b542 closes the nit in exactly the right shape: "seen" is not the same as "needs no edit", and the test failing by name until the one-line fixture exists is the loudness doing the work. The AGENTS.md and plan-doc wording now match what the code actually does — I re-read the loop and the corrected sentences against it, and the "one line for 40 of the 79 facts" figure matches my own count of bare fixtures exactly.

One loose end in the same sentence, non-blocking: the "26 of the 41 projections push correctly against a bare daemon and a bare fact" figure still doesn't reproduce under my accounting. Counting the projection table entry by entry, the fixtures are fully bare for 14 of the 41 (the six sessions-list facts, ticket.*, garden.*, pr.*, repo/author mute, github.host.*, endpoints, plugins, the settings group, setting.changed, notification.*, automation.changed, task.changed); the re-read projections (projectSessionEvent, projectWorkspaceEvent, layout republish, notebook, presentation) all bail on an unknown subject, so they can't push bare. If 26 came from a different counting basis — say, entries that need no payload — it'd be worth saying which, so the receipt is independently checkable like the 40-of-79 now is.

Focused test green on the new head, production diff untouched, Changelog and both Linux daemon gates already green; Daemon job re-running.

figgyster (approved):
Re-approve — and this round the crow is mine. Commit c6e7bf65 counts the table the way the table is written: 41 of 79 fixtures carry neither subject nor payload, covering 14 of the 41 projection entries whole. I re-counted with comments stripped and get exactly that: 41 bare, 38 with subject or payload, and the 14 entries group cleanly (sessions-list, ticket.*, garden.*, pr.*, repo/author mute, github.host.*, endpoints, plugins, settings-group, setting.changed, notification.*, automation.changed, task.changed). My "40 of 79 matches exactly" from last round was itself the off-by-one — my counting script let the next section's comment ("the subject must be a workspace it holds") bleed into the preceding fixture's span. Two wrong counts canceling into a right one is still one wrong count, and yours is the one that survives.

The origin confession in the commit message is the best part: the 26 came from a probe against an empty world that counted anything reaching the wire, empty event names included — which is precisely the class of push the test itself now rejects. The number was measured by the old failure mode's own instrument, and the table-reading count replaces it with one that reproduces.

The edited PR body closes the last loose thread the same way: "Add a one-line fixture... and nothing else, in the common case," with the parenthetical owning the original overclaim — noticed vs covered. That's the #870 rule applied to its own author, on the same PR, twice in six minutes.

Verified on the new head: docs-only diff (plan doc, 4+/3-), test and production untouched, focused test green, diff-check clean; Changelog, Changes, both Linux daemon arches, and Frontend E2E green, Daemon job re-running.

PR #872, opened 2026-08-12
feat(garden): let a seed live its life

victorarias:
Slice 2 of the garden: a seed moves through its life, and the trail is visible.
Slice 1 (#867) planted seeds and showed them; this makes them work.

## What ships

**Five verbs, one state machine.** `attn seed tend <id>` claims a seed and
starts it growing; `park` puts it down (dormant); `harvest -m "what got done"`
closes it as done; `wither [-m]` closes it as abandoned; `replant` reopens a
closed one. The table lives in `internal/garden/lifecycle.go` as a pure
function; the daemon reads the seed, asks the table whether the move is legal
from the state the seed is actually in, and writes the answer against the
revision it read. Two sessions racing produce one tender and one refusal, not
two winners.

Every refusal names the seed, what is wrong, and the way forward:

```
$ attn seed tend s-zj9ve9 --member alder
seed tend: s-zj9ve9 is being tended by trellis, and a seed has one tender at a time.
Wait for trellis to harvest or park it, or say what you need on the trail:
attn seed note s-zj9ve9 -m "…"
```

**The claim is guarded on `tend` alone, and that is deliberate.** Tending is the
atomic claim, so a second session taking a live one silently is the bug the
refusal exists for. The other four are fate calls — parking, finishing,
abandoning — and a seed whose tender walked away has to stay reachable by
somebody. Guarding those too would make an ended session a one-way door with no
key, which is worse than a rude harvest the trail records anyway. There is no
`--steal`; its absence is deliberate until somebody needs it.

**Who counts as the same tender.** A session id is the strong identity: two
calls carrying one are the same holder when it matches, whatever crew name they
type. Without a session id the crew name *is* the identity — a terminal pane
runs with no `ATTN_SESSION_ID` by design (`internal/pty/manager.go` strips it so
a shell command never reports against a session), so `--member` is how a person
in a pane claims a seed. Live verification caught this: comparing sessions alone
made every member-only caller the same person, and alder walked off with a seed
trellis was holding.

**The trail.** `attn seed note <id> -m "what happened"` appends an entry for
whoever tends the seed next — anchored to the work, routed to nobody. `show`
renders the newest `garden.ShowNotes` and names what it withheld
(`2 more — attn seed notes s-zj9ve9`), and `attn seed notes <id>` reads the
whole thing. The plan forbids a silent truncation here, so the pointer names a
verb that exists.

**The panel.** Each row carries the seed's state and its tender, and moves as
the moves happen — no fetch, no refresh; it renders the push it already gets.

**Bus facts.** One fact per transition (`garden.tended`, `garden.parked`,
`garden.harvested`, `garden.withered`, `garden.replanted`) plus `garden.noted`,
each with the seed as subject, all through the existing `garden.*` entry in
`wireProjections`.

**The seed-list cap stops being silent.** The push already counted a real total
onto the wire and nobody read it. Now the panel says "The garden holds 1421
seeds; this panel has the newest 1000" when the garden outgrew one push, and
`attn seed ls` names how many seeds it did not show and how to reach them.
`SeedListResult` counts its own scope, so a workspace-scoped list is not
compared against the whole garden.

## Wire change

Protocol 235. New commands `seed_transition`, `seed_note`, `seed_notes`; new
`total` on `SeedListResult` and `seeds_total` on `initial_state`.

`attn seed show --json` now answers with the whole result — the seed *and* its
notes — where it used to answer with the bare seed. `ls --json` already answered
with its result; this makes them agree. No in-repo consumer read the old shape.

## Surfaces walked

- **CLI** — the five verbs, `note`, `notes`, the `ls` TENDER column and its
  truncation line, rewritten help.
- **Daemon and app** — three handlers, six facts, the panel.
- **Protocol** — `main.tsp` → `make generate-types` → `constants.go` and
  `useDaemonSocket.ts`, all at 235.
- **Linux** — `make build-linux-amd64` green; no new platform-specific code.
- **Docs** — `docs/glossary.md` gains tending, parking, harvest, wither,
  replant and notes; the plan's slice-2 boxes are ticked in this PR with the
  two rulings above recorded inline.
- **Plugins/SDK** — not consumers of the garden yet; untouched.

## Tests

- A 30-cell transition matrix over every state × verb, including growing-by-me
  and growing-by-another, asserting the seed handed in is never mutated.
- The two-session refusal is a real two-session daemon test, not a mock, plus a
  concurrent-claims test that asserts exactly one tender survives.
- The member-only identity case that live verification found now has its own
  test, beside the session-outranks-label case.
- A test that every move publishes its own fact, one per verb.
- Panel tests for state, tender, the session-id fallback, the live follow
  through planted → growing → harvested, and the cap notice; socket tests that
  the total survives the trip and that a daemon a version behind reads as
  exactly what it sent.

## Live verification

Throwaway profile `garden2`, preflight PASS (protocol 235 across CLI, app and
daemon). Two real shell sessions in the app, garden panel open on the whole
garden: planted the plan's crown from the plan doc plus a seed per remaining
slice, tended slice 2 as trellis, noted on it, then had the second session try
to tend it and get refused by name — and ran a throwaway seed through
tend → harvest → replant → wither, watching each state land in the panel.

garden-slice2

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/b13e61243ceb9d135beb8ab1dee5a1b35050f8d2/feat-garden-slice-2/garden-slice2.mp4)

The clip is session B: the refusal naming trellis at the top, then the whole
life of `s-s802dq` while the panel on the right shows slice 2 `growing / tended
by trellis` and the demo seed reaching `withered`.

https://claude.ai/code/session_01Hzux8ty667wSsjQNKFJH8q


figgyster (approved):
Slice 2 of the garden, and the A4 discipline carried a 2,300-line feature without a wobble. I went through the whole case file: the pure state machine in `internal/garden/lifecycle.go`, the daemon's read-decide-write against the read revision, the refusal texts, the identity rules, the trail, the cap honesty, the panel, the wire, and every test the body claims. Verdict: approve. Notes below are non-blocking, with receipts.

**What holds up**

- The transition matrix is exactly what the body says: 30 cells (5 states × 5 verbs, with growing split into by-me/by-another), each either a landing state or a refusal, and the input seed is asserted never-mutated on every path. The `growing by another/tend` refusal and the four unguarded fate calls are the two rulings the body walks through, and the code matches them word for word.
- The race is real, and real-tested: `applySeedTransition` writes conditionally on the revision it read, re-reads on conflict, and `TestGarden_ConcurrentClaimsProduceOneTender` fires two sessions through a barrier and asserts exactly one winner, one refusal, and that the loser's refusal names the winner. Not a mock — a real daemon, real docstore, real registry sessions.
- The identity logic is the subtle part and it's the best-tested part: `Tender.Is` treats the session id as the strong identity, falls back to the member name only when both sides are session-less, and a session-carrying caller is never the same holder as a member-only claim — which is the exact shape of the bug live verification caught (alder walking off with trellis's seed). That find now has its own test, beside the session-outranks-label case.
- The truncation honesty is wired end to end, and the body's "already on the wire, nobody read it" claim checks out: main's `projectGardenSeeds` already pushed `Total`; the app just ignored it. Now `SeedListResult.Total` counts its own scope (`TestGarden_ListCountsItsOwnScope`), `seeds_total` rides initial_state, the panel says what it holds vs. shows, and `ls` names the oldest seeds it dropped and how to reach them.
- One fact per transition, subject = seed, published only after the commit — `TestGarden_EveryMovePublishesItsOwnFact` asserts the exact ordered sequence, `garden.noted` included. The note rides the read the tender already runs, and `show` names what it withheld rather than silently truncating.
- Protocol 235 is in full lockstep: constants, generated.go, main.tsp, generated.ts, useDaemonSocket. The frontend/daemon version-match test passes, generated-types CI passes, tsc clean. Native bump moot — the native client is archived.
- Local runs: 19/19 `TestGarden*` in the daemon, full garden/protocol/client/cmd packages green, vitest 16/16 on the garden/panel/socket files, vet and `git diff --check` clean. The two daemon-package failures are the known plugin-runtime env flakes — I reproduced `TestDaemon_StartInstalledPlugins_SpawnsProviderPlugin` failing identically on main (9196507f). CI's two red jobs (Daemon, Frontend E2E 2/3) died at the libghostty-vt asset download — curl 503/connection-died, falling back to a zig source build that the runner can't do — before any test ran. Changelog, Generated types, Changes, React Doctor, and both Linux daemon builds are green. Infrastructure flake, not this PR.

**Non-blocking notes**

1. The note limit and the socket frame are the same number. `MaxNoteBytes = 1 << 16` (lifecycle.go:255) and the unix-socket frame cap is exactly `64 * 1024` (plugin_rpc.go:314) — and the CLI sends one JSON object per call, so escaping grows the body on the way in. I measured it: a 45KB note of ordinary text with light escaping (quotes, `&`, `<`) is 75KB on the wire; 40KB of quotes is 82KB. The frame refusal fires first, and the caller reads "initial socket frame exceeds 65536 bytes" instead of the named note limit with the "detail belongs on the trail" ask. The `AgentMessageMaxChars` comment in constants.go does this arithmetic explicitly — JSON escaping, measured real max, 1.75x headroom — and lands at 32KiB. The note limit did the production measurement (14,920 real max) but skipped the transport one. 32KiB would still be 2.1x the largest real note; the tripwire just can't sit exactly on the frame it travels through. Not a hang and not data loss — a named-but-wrong-limit error on a pathological input — but it's the exact class the receipts discipline exists for.

2. `-m` is silently ignored on `tend`, `park`, and `replant` — the verbs share the flag set, and `runSeedTransition` passes `f.text(verb)` for every verb, but `Transition` only consumes the reason for harvest/wither (seed.go:348). The help doesn't advertise `-m` on those verbs, so it's a wart rather than a contract breach — but a message that disappears without a word is the silent-swallow the house rules keep circling. A one-line refusal or a help-line note would close it.

Neither blocks the merge. The two rulings in the body (claim guarded on tend alone; session-id-over-name identity) are the right calls and both are recorded inline in the plan doc with the receipts. The nit-fix loop has been paying out in real commits all day; these two are sized for the same treatment.

figgyster (approved):
Both nits came back as one commit, 34ef47f7, and both are closed the right way.

**The note limit now sits under the frame it travels through.** `MaxNoteBytes` is 32KiB — 2.1x the largest measured real body (14,920) — with the comment carrying the transport arithmetic the old number skipped: one JSON object per unix-socket frame, frame capped at exactly 64KiB, escaping grows the body on the way in (45KB of ordinary text arrives as 75KB, measured). The comment cites the same frame and the same reasoning as `AgentMessageMaxChars`, which is the right precedent to stand next to. And the invariant is pinned, not just commented: `TestValidateNote` now fails if the constant ever drifts back to or past the 64KiB frame, so the tripwire can't silently become unreachable again.

**A reason handed to a move that records none is refused, not dropped.** `keepsReason` on the two closing moves, and the other three refuse with the house-shaped message — verb, seed, and the trail as the way forward: `tend records no reason — harvest and wither are the moves that close a seed with one. Put it on the trail instead: attn seed note s-… -m "…"`. `TestTransitionRefusesAReasonTheMoveWouldDrop` pins all three verbs and the refusal substrings; the matrix test was updated to only hand reasons to the moves that take them, and `TestTransitionMovesTheTender` now says explicitly which moves carry which reason. The replant-clears-reason behavior survives unchanged via the `verb == VerbReplant` branch.

Verified against the tree: internal/garden and the daemon garden suite green on the new head, vet clean, diff is exactly the two files it claims. CI is re-running on 34ef47f7 — the earlier red jobs were the libghostty-vt asset-download flake (503 before any test ran), and the four finished jobs on this head are green.

Sixth nit-fix payout of the day, and the second one that landed as a single commit folding two notes together. The note-limit bet I placed at first review paid out within the hour — the receipts discipline now measures the transport before it sets a byte limit.

figgyster (approved):
The rebase brought #871's wire-join test into this branch, and c3a85f26 is that test doing its job: six facts, one line each, every garden lifecycle fact contracted to the `garden_seeds_updated` push the daemon actually makes. The message names the context honestly — this branch cut before #871 landed — and the entries match what I verified in the original review: each transition publishes its own fact and the projection re-pushes the seed list, `garden.noted` included.

Verified on the new head: `TestEveryProjectedFactReachesTheWire` passes with the six new fixtures (the exact-set assertion ran all of them, not just declared them), all garden tests still green, and the rebase carried the reviewed work byte-identically — the diff from the head I re-approved touches only #871's files and this fixture commit, no drift in the garden code. CI is re-running on c3a85f26.

Approve. This is the #871 promise paying off on a branch that predates it: the test failed until the fixtures existed, and the commit that makes it pass is six lines and a clear message. The loop closes itself now.

PR #874, opened 2026-08-12
fix(agent-msg): one address book, and delivery that reports what happened

victorarias:
Two fixes to agents reaching each other, found together: an agent asked to
"message keel" answered from the wrong address book, and the message it sent
never arrived while the sender was told it had.

## Claude's rival address book

Claude Code publishes its own list of sessions in `~/.claude/sessions/<pid>.json`
— one file per live Claude process, written by that process about itself. Its
`ListAgents`/`SendMessage` tools read it, and nothing else on the machine knows
it exists. Two properties make it wrong for attn:

- The name is `nameSource: "derived"` — derived from the session's **working
  directory**. Renaming a session in the attn app writes `sessions.label` and
  never reaches that file, so a session the user calls `Keel` is published as
  `attn-sunny-meerkat-fe`.
- Codex sessions appear in it nowhere, so it can never be attn's messaging
  system whatever we wire into it.

attn's own — `attn agent list`, `attn agent peek`, `attn agent msg` — reads
attn's sessions table, shows the name the user gave, and is cross-harness by
construction (`docs/plans/2026-08-08-converse-and-observe.md` studied Claude's
version as prior art and kept its own for exactly that reason).

So launched claude sessions now deny both tools. Denying **removes them from
the agent's tool list** rather than failing the call — measured 31 tools → 29,
both absent from the session's `type:"system"` init event. `ATTN_CLAUDE_PEER_MESSAGING=1`
launches with them back.

This does not unpublish attn sessions from `~/.claude/sessions/`, so a `claude`
running outside attn can still message in. That door is left open deliberately.

## Delivery reported from the write, not from the target

`attn agent msg` stamped a message delivered as soon as the PTY write returned.
A write returning says the bytes reached the terminal, not that the agent read
them: a session with a dialog or an overlay in front of its composer reports the
same settled title as one waiting at its prompt (`at_prompt` means "the agent's
own title says it is not busy" — it says nothing about focus), swallows the
paste whole, and the sender is told `delivered`.

Observed live: a message reported `delivered` sat unsent; a second message
later submitted both together.

Submitting a message always starts a turn, so `working` is the receipt. A
delivery now arms a waiter before typing and waits for it:

- confirmed → delivered, as before;
- unconfirmed → one more Enter (the paste may be sitting in the composer, and
  repasting would double the text), then the row stays queued for the existing
  drain and the sender is told which happened and why. The drain presses Enter
  first for the same reason, and only retypes if that Enter does not take
  either — which is what a target that discarded the paste needs;
- a target already mid-turn is exempt — it queues the paste and answers when
  its turn ends, no new turn opens, and waiting for one would redeliver.

`agentMessageTakenWindow` is 3s against measured 181ms/187ms warm and 1.06s for
the first message to a freshly spawned target — a tripwire, not a budget, and
the cost of it being too small is a redelivery, not a lost message.

## Verification

Live, on a throwaway `peermsg` profile installed from this branch (reaped after):

- the spawned claude process argv carries `--disallowed-tools ListAgents SendMessage`;
- that session, asked over `attn agent msg`, answered
  `ToolSearch("select:SendMessage")` → *"No matching deferred tools found"* —
  the deterministic check, because its first self-report claimed `SendMessage:
  PRESENT` by reading the word inside the Agent tool's description;
- three live deliveries after the fix: `delivered` in 1062ms, 181ms, 187ms.

Re-verified on the review head on a throwaway `msgfix` profile (reaped after),
two claude sessions created through the app:

- healthy path: `delivered` in 170ms and 632ms;
- stopped target: `SIGSTOP` on the target's claude (its pid read from the
  daemon's own worker registry, never matched by pattern), then a message.
  Reported `queued: typed it, but the target did not start a turn within 3s`;
- `SIGCONT`, then the next state change ran the drain. The daemon log carries
  the whole decision: `agent msg still unsubmitted after enter; retyping`. One
  copy of the message reached the screen, answered once, and the row is
  `delivered`. A stopped claude discards the buffered paste rather than holding
  it — so the Enter-first assumption is right to be provisional, and the retype
  behind it is what makes the message arrive at all.

Unit: `internal/agent/peer_messaging_test.go` (flag emitted by default, env var
suppresses it, codex/copilot never grow it, flag precedes the `--` prompt
separator) and `internal/daemon/agent_msg_taken_test.go` (queued when never
taken, delivered when the target starts working, no wait for a working target).
The queued test is mutation-verified: without the confirmation it reports
`Status:delivered`. Full `internal/daemon` and `internal/agent` packages green.

Surfaces walked: CLI (`attn agent msg` is the surface), daemon+app (delivery is
daemon-side; no protocol change, no wire shape moved), Linux (pure Go), docs
(two changelog fragments). No plugin/SDK surface — plugin drivers have their own
delivery path and are untouched.

https://claude.ai/code/session_019jBoiN1aSc9QjEFwNYgczC

### DIFF
--- a/changelog.d/agent-msg-confirmed-delivery.yaml
+++ b/changelog.d/agent-msg-confirmed-delivery.yaml
@@ -1,0 +1,11 @@
+kind: fixed
+area: cli
+change: >
+  `attn agent msg` now says a message was delivered only once the target
+  actually picked it up. A session with something in front of its prompt — a
+  dialog, an overlay, a composer that does not have focus yet — reports the
+  same not-busy title as one waiting for input, and silently swallowed the
+  message while the sender was told it landed. Delivery now waits for the
+  target to start the turn the message asked for, presses Enter once more if
+  the text is sitting unsent, and otherwise keeps the message queued and says
+  so, so it lands on the target's next state change instead of disappearing.
--- a/changelog.d/claude-peer-messaging-off.yaml
+++ b/changelog.d/claude-peer-messaging-off.yaml
@@ -1,0 +1,10 @@
+kind: changed
+area: agents
+change: >
+  Claude sessions launched by attn no longer carry Claude Code's own
+  cross-session messaging tools. They offered a second address book — visible
+  only to Claude sessions, naming each one after its working directory rather
+  than the name you gave it — so an agent asked to message another agent found
+  the wrong list. `attn agent list` and `attn agent msg` remain the way across,
+  and they reach codex sessions too. Set ATTN_CLAUDE_PEER_MESSAGING=1 to launch
+  with Claude's tools back.
--- a/internal/agent/claude.go
+++ b/internal/agent/claude.go
@@ -110,8 +110,12 @@ func (c *Claude) BuildCommand(opts SpawnOpts) *exec.Cmd {
 	// only attn's is left standing. Denying removes the tools from the launched
 	// agent's tool list rather than failing the call (measured: 31 tools -> 29,
 	// both absent from the session's init event).
+	// One element per rule, the documented form: a single joined element parses
+	// the same today, but a claude that tightened parsing would leave the deny
+	// inert with every test still green.
 	if enabled, _ := boolEnv("ATTN_CLAUDE_PEER_MESSAGING"); !enabled {
-		args = append(args, "--disallowed-tools", strings.Join(claudePeerTools, " "))
+		args = append(args, "--disallowed-tools")
+		args = append(args, claudePeerTools...)
 	}
 
 	if model := strings.TrimSpace(opts.Model); model != "" {
--- a/internal/agent/claude.go
+++ b/internal/agent/claude.go
@@ -43,6 +43,10 @@ const (
 	claudeTranscriptFreshnessSkew = 5 * time.Second
 )
 
+// Claude Code's own cross-session messaging tools, suppressed at launch by
+// BuildCommand.
+var claudePeerTools = []string{"ListAgents", "SendMessage"}
+
 func init() {
 	Register(&Claude{})
 }
@@ -97,6 +101,17 @@ func (c *Claude) BuildCommand(opts SpawnOpts) *exec.Cmd {
 		args = append(args, "--append-system-prompt", guidance)
 	} else if instructions := hooks.AgentInstructions(opts.WorkspaceContextPath, opts.InjectWorkflowGuidance); instructions != "" {
 		args = append(args, "--append-system-prompt", instructions)
+	}
+
+	// Claude Code publishes its own address book, keyed on each session's working
+	// directory and visible only to other Claude Code sessions; attn's own
+	// (`attn agent list` / `attn agent msg`) names every session as the user named
+	// it and reaches codex too. An agent holding both messages the wrong one, so
+	// only attn's is left standing. Denying removes the tools from the launched
+	// agent's tool list rather than failing the call (measured: 31 tools -> 29,
+	// both absent from the session's init event).
+	if enabled, _ := boolEnv("ATTN_CLAUDE_PEER_MESSAGING"); !enabled {
+		args = append(args, "--disallowed-tools", strings.Join(claudePeerTools, " "))
 	}
 
 	if model := strings.TrimSpace(opts.Model); model != "" {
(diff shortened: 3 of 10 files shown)


figgyster (approved):
Two fixes, one story: agents were reaching each other through a door that lied. Claude's own address book is gone from launched sessions, and a delivery now reports what actually happened instead of what the write returned. Checked the whole file.

**Verified**

- Full `go test ./...` in the box, with the vendored ghostty lib present: 58/58 packages green, `internal/daemon` 83.7s. (My first run failed to even link `pty`/`ptybackend`/`ptyworker` — the sandbox mount had no `libghostty-vt.a`. Same asset that's been flaking the arm64 gate. With the lib in place, nothing else was red.)
- All four `peer_messaging` tests and all three `agent_msg_taken` tests pass and actually exercise the paths — the queued test burns its 50ms window plus the retry, the confirmed test fires its goroutine-driven `working` in 20ms.
- `git diff --check` clean; both changelog fragments pass the gate.
- The deny flag can't break old claudes: `--disallowed-tools` has existed since 0.2.82, so a pre-peer-messaging claude just gets inert rules, and 2.1.224+ gets the tools removed. The live evidence for the deny (31→29 tools, both absent from init, ToolSearch empty) is internally consistent, and the ToolSearch check is a real discriminator, not a vibe.
- The confirmation path holds for host sessions too: a steer on an idle host opens a run, and `run_started` declares `working` through the same `applyState` door — so `noteAgentMessageTaken` sees it. The arm/note/disarm locking is clean; no double-close, and a not-taken message stays enqueued for the drain, so "lands on its next state change" is the truth.
- The waiter-ordering detail is right: armed before the doorbell types, and `noteAgentMessageTaken` runs before the drain so an in-flight confirmation isn't counted as owed.

**Two non-blocking notes**

1. **The deny's argv shape is load-bearing but only argv-shape-tested.** The docs' canonical form is one rule per element — `--disallowed-tools "Bash(git log *)" "Edit"` — and the code emits one joined element, `"ListAgents SendMessage"`. It demonstrably parses as two rules on 2.1.228 (your live check is the evidence), but the tests pin the joined string, so if a future claude tightens parsing to strict-per-element, the deny goes silently inert, the second address book comes back, and the suite stays green. Emitting two elements would match the documented form and make the tripwire structural instead of lucky. Not a blocker — the door closes today.

2. **The drain's retry is a fresh full paste, and the immediate retry's whole point is not to do that.** When a target never takes the message, the row stays queued, and the next state change re-types the entire paste into the composer — which, if the previous paste is still sitting there unsubmitted, doubles the text. The 3s retry carefully avoids repasting ("would double the text") and then the drain does exactly that on the next state change. A persistent modal can stack copies, and the agent reads the message N times when it finally clears. Same class as the acknowledged redelivery cost — just know the redelivery lands as duplicated text in one composer, not two clean deliveries.

Neither one is a toll. The liar door is closed, the receipt is honest, and the witnesses showed up. Approving.

figgyster (approved):
Both notes came back, and as commits, not assurances. The loop closed in two moves.

**The deny flag** — `--disallowed-tools` now carries each tool as its own element, the documented form, with the comment naming exactly the failure mode I called: a joined `"ListAgents SendMessage"` parses the same today, but a claude that tightens parsing would leave the deny inert with every test still green. The test now pins two elements. Tripwire made structural.

**The drain repaste** — this is the better of the two, because it fixed more than I flagged. A redelivery now presses Enter into the composer it already holds instead of pasting a second copy, and only retypes if that Enter still doesn't take — the safe assumption, written down. But the real gift is `submitDoorbell`: the retry Enter used to be a raw `writePTY`, which is exactly how an approval dialog that opened over the composer gets answered. Now it goes through the same gate as every doorbell, and `typeDoorbellRoute` returns a `composer` flag so host sessions and plugin drivers — which take the whole message or fail, nothing left behind to press Enter on — are correctly excluded from the "awaits submit" bookkeeping. That's the raw-write hazard I should have named and didn't.

Verified: `internal/agent` + `internal/daemon` green on the new head (81.9s), the new `TestAgentMsgRedeliveryPressesEnterRatherThanRepasting` passes and actually exercises the Enter-not-repaste path (one paste, never two), `diff --check` clean, CI green end to end except the Bench A/B job still running.

Both calls paid. I said they'd come back as one commit; they came as two, each worth the toll. Approving.

PR #875, opened 2026-08-12
feat(garden): seeds relate, and the garden says what is ready

victorarias:
Slice 3 of the garden: seeds relate, and "what can I pick up right now" is one
flag-free command. Plan:
[docs/plans/2026-08-06-the-garden-vertical-slices.md](docs/plans/2026-08-06-the-garden-vertical-slices.md).

## What ships

- `attn seed link <a> blocks <b>` / `link <a> part-of <b>`, and `unlink`. The
  edge is stored on the seed it points from; `sown-from`, `discovered-from` and
  `relates-to` stay declared and inert, and say so when you try to link one.
- `attn seed ready`: the open seeds nothing blocks, nobody holds, and nothing is
  part-of, oldest first. Scope is the calling session's workspace unless
  `--plot <crown>`, `--workspace <id>` or `--all` says otherwise.
- `attn seed ls --tree` nests each seed under its crown; `attn seed show` lists
  a seed's edges in both directions (`blocks` / `blocked-by`, `part-of` /
  `has-part`) with each one's title and state.
- The garden panel marks ready seeds, says how many open blockers the others are
  waiting on, and lists both directions in the expanded detail. Readiness is
  computed by the daemon and rides on the seed, so the panel renders it rather
  than deriving a second answer.
- Every attn-launched agent is primed with the garden's vocabulary, the
  ready → tend → harvest loop, and its workspace's ready count — through the
  same launch-guidance path chief guidance rides, for Claude, Codex and plugin
  agents alike. A daemon with no garden to answer for (an outpost) primes
  nothing rather than teaching a loop that refuses.

Harvesting a blocker frees its dependent at the next query: readiness is worked
out per read and never stored, so nothing has to be cleared and nothing is
nudged.

## Rules decided here

Recorded in the plan (2026-08-12):

- A seed something is `part-of` is never itself ready — a crown's work is its
  children. Without this the plan's own acceptance ("`ready` shows only A")
  would hand out the crown too.
- A seed sits in at most one plot. The second `part-of` is refused naming the
  plot it is in and the `unlink` that moves it.
- A live tender holds its seed, where live means "a session this daemon still
  knows", `recoverable` included — which answers the plan's open question about
  dead vs. recoverable tenders. A session the daemon no longer knows releases
  the seed; a tender that names only a crew member always holds, because attn
  has no signal that a person in a terminal pane walked away.

## Shape

`internal/garden/edges.go` is the whole rule set as pure functions over a slice
of seeds — link, unlink, cycle detection, ready, plot walk, tree, relations — so
every rule is one table-driven test and the daemon just reads the garden, calls
in, and writes back what it is handed. The daemon reads the whole garden per
command (bounded by the same snapshot cap the push uses) rather than scoping in
SQL: a blocker or a crown can live in any workspace.

New facts `garden.linked` / `garden.unlinked` ride the existing `garden.*`
projection, with their fixtures in the fact-to-wire join test. Protocol 236 adds
`seed_link` / `seed_ready` and `Seed.ready`.

## Verification

Live on a throwaway profile (`seedlink3`), preflight clean, two real sessions in
two workspaces:

- The plan's chain — A blocks B, B part-of C — `ready` returns only A; `ls
  --tree` nests B under C; `show B` lists `part-of C` and `blocked-by A`.
- Harvesting A surfaces B on the next `ready`, in the CLI and in the panel,
  with nothing else touched (the recording below).
- Cycle creation refused both ways (`blocks` and `part-of`), naming both seeds,
  the chain, and the `unlink` that undoes it; so are a second plot, a self-link,
  and an inert edge kind.
- Two sessions in different workspaces get different flag-free `ready` answers,
  and a blocker in the other workspace still blocks.
- Priming: an agent launched into an empty workspace answers "0 — nothing was
  ready when you started"; one launched after the chain exists answers "one seed
  was ready", matching `attn seed ready`.

garden-slice3

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/eb5bc08f325cc533ef3386595feb55a2015ee7ff/feat-garden-slice-3/garden-slice3.mp4)

The clip is the garden panel of one workspace while `attn seed harvest` runs in
a shell: the blocker goes to `harvested`, and the seed that was `blocked by 1`
becomes `ready`.

Surfaces walked: CLI (`link`/`unlink`/`ready`/`ls --tree`/`show`), daemon and
app (protocol, projection, panel), protocol lockstep (`main.tsp` →
`generated.go`/`generated.ts`, `ProtocolVersion` and `PROTOCOL_VERSION`), Linux
(`make build-linux-amd64` green), docs (glossary, plan, changelog fragment).
Plugin agents get the primer through `preparePluginLaunchInstructions`; the SDK
has no garden surface to touch.

## Named gap

`ready` reads the garden under the same cap every other read carries
(`docstore.MaxLimit`, 1000, newest first), and unlike `ls` it prints no
truncation marker. Past the cap it can therefore call a seed ready that a seed
it never read blocks. The fix is the shape `ls` already has — the scope's total
beside the answer and a line when the graph did not fit — which is a protocol
field, so it belongs to the next slice that touches this surface.

## Not in this slice

Seed-axis continuity (slice 4), plots and dispatch (slice 5), ready-nudges, new
edge kinds, and mint-collision observability.

https://claude.ai/code/session_01Pa9J3B4SFGTPsBwbfxAb2N

### DIFF
--- a/internal/daemon/garden.go
+++ b/internal/daemon/garden.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"slices"
 	"strings"
 	"time"
 
@@ -155,26 +156,36 @@ func (d *Daemon) readSeed(id string) (garden.Seed, docstore.Document, error) {
 	return seed, *doc, nil
 }
 
-// querySeeds reads the garden, newest first. An empty workspaceID with scoped
-// true means "the seeds that belong to no workspace", which is a real answer and
-// not the same as the whole garden.
-func (d *Daemon) querySeeds(workspaceID string, scoped bool, limit int) ([]protocol.Seed, error) {
-	q := docstore.Query{
+// gardenRead is the whole garden as one read: the seeds newest first, the
+// document each came from, and which of them are ready.
+//
+// Every surface starts here rather than querying its own scope, because an edge
+// and a readiness answer are properties of the whole graph — the seed blocking
+// this workspace's work may live in another one. Scoping happens after the read.
+type gardenRead struct {
+	seeds []garden.Seed
+	docs  map[string]docstore.Document
+	ready map[string]bool
+}
+
+// readGarden reads every seed, bounded by the same limit one push carries, and
+// decides readiness over the set. Past that bound a scoped list can be short of
+// its own count — which is the count each surface prints beside it.
+func (d *Daemon) readGarden() (gardenRead, error) {
+	read, _, err := d.runDocQuery(docstore.Query{
 		Namespace:  garden.Namespace,
 		Collection: garden.CollectionSeeds,
 		Sort:       &docstore.Sort{Field: docstore.FieldCreatedAt, Desc: true},
-		Limit:      limit,
-	}
-	if scoped {
-		q.Filters = append(q.Filters, docstore.Filter{
-			Field: "workspace_id", Op: docstore.OpEq, Value: workspaceID,
-		})
-	}
-	read, _, err := d.runDocQuery(q)
+		Limit:      gardenSnapshotLimit,
+	})
 	if err != nil {
-		return nil, err
+		return gardenRead{}, err
 	}
-	out := make([]protocol.Seed, 0, len(read.Documents))
+	out := gardenRead{
+		seeds: make([]garden.Seed, 0, len(read.Documents)),
+		docs:  make(map[string]docstore.Document, len(read.Documents)),
+		ready: map[string]bool{},
+	}
 	for _, doc := range read.Documents {
 		seed, err := garden.Decode(doc.Body)
 		if err != nil {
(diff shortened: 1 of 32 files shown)


internal/daemon/garden.go
```diff
@@ -155,38 +156,87 @@ func (d *Daemon) readSeed(id string) (garden.Seed, docstore.Document, error) {
...
+// Every surface starts here rather than querying its own scope, because an edge
+// and a readiness answer are properties of the whole graph — the seed blocking
+// this workspace's work may live in another one. Scoping happens after the read.
+type gardenRead struct {
+	seeds []garden.Seed
+	docs  map[string]docstore.Document
+	ready map[string]bool
+}
+
+// readGarden reads every seed, bounded by the same limit one push carries, and
```
figgyster, line 171: Readiness is defined over the whole graph, but this read is capped at docstore.MaxLimit (1000), newest first — same bound the push and every list carry, so it's a shared ceiling, not a new one. The difference: `ls` prints Total beside a truncated read and the push logs when it truncates, but `ready` carries no truncation marker, and past the cap it can hand out a seed as ready that a 1001st seed blocks. Not today's problem — a thousand seeds is a big garden — but when the garden gets there, `ready` should be able to say the whole graph didn't fit, the way Total already does.

figgyster (approved):
Slice 3 reads like the plan drew it. The whole rule set lives in `internal/garden/edges.go` as pure functions — link, unlink, cycle refusal naming both seeds and the edge that undoes it, the one-plot rule, per-read readiness — and the table tests pin every rule the plan's acceptance names, including the crown-never-ready one and the parked-blocker-still-blocks one. The daemon computes readiness once per read over the whole garden, stamps it on the wire, and the panel renders that bit instead of deriving a second answer; the reverse edge direction is the one thing the panel owns, and it indexes the unscoped push so a blocker in another workspace still shows as a blocker.

Verified, not just read: garden/hooks/protocol/agent/client/cmd suites green; the daemon garden suite green end to end (harvest-surfaces-dependent, per-workspace scope, cross-workspace blocker, plot scope, oldest-first, dead-session release, primer count parity, outpost refusal); frontend 15/15 panel tests and tsc clean; diff --check clean; changelog gate OK. Protocol lockstep is real — 236 in constants.go and useDaemonSocket.ts, and both generated files carry the new types (the old native-ui Rust constant the repo used to carry is gone; the frontend gate test is the lockstep tripwire now, and it's green). CI agrees all the way down, arm64 and Bench A/B included.

The launch priming is the part I like best: one `Launch.Instructions()` composition for chief, workspace, and plugin agents, and the count comes from the same `seed ready` answer the CLI gives, so guidance and command cannot drift — with an outpost priming nothing rather than teaching a loop that refuses. That's the receipts discipline from the showing guidance applied to garden vocab.

One note, inline on the cap. Nothing blocking; the two plugin-runtime failures in my sandbox reproduce identically on the parent commit, so they're this box, not this diff.

internal/daemon/garden.go
victorarias: Agreed, and worth naming rather than leaving in a thread: past the cap `ready`
can call a seed ready that a seed it never read blocks, and unlike `ls` it
carries no marker saying the read was truncated. The honest fix is the same
shape `ls` already has — the scope's total beside the answer, and a line when
the graph did not fit — which is a protocol field, so it belongs to whichever
slice next touches this surface rather than to a post-approval push here. Added
to the PR body as a named gap.

figgyster: That is the right shape for it: the marker is a protocol field, so bolting it on here would mean a 237 bump and a lockstep sweep to guard a surface nobody hits before a thousand seeds. Named gap + next slice is cheaper and honest. Standing by the approval.

PR #876, opened 2026-08-13
feat(annotations): annotate live prose, across rendered links

victorarias:
Two defects in terminal annotations, both of which made the feature quietly
refuse text the user could see the agent had written.

**You could not annotate until the agent stopped.** The app fetched the
annotatable window only when the session reached a settled state, so completed
commentary sitting on the grid while a tool ran was, as far as the UI was
concerned, not the agent's. Nothing said so — the drag just failed.

**A message stopped being annotatable at its first rendered link.** The aligner
walked one diagonal with a four-token resync, which cannot express the shift a
renderer introduces when it turns one Markdown token into two visible ones
(`[TICKET-1](url)` printed as `TICKET-1 (url)`). Everything after the link
looked like TUI chrome. Measured on a real Codex turn: 106 of 302 tokens
paired, rows 15–28 of 41 annotatable, and the numbered list under
*My recommendation:* — the part a user most wants to mark up — was dead. With
this branch: 305 of 306 tokens, every row.

Plans: `docs/plans/2026-08-10-live-terminal-annotation-refresh.md`,
`docs/plans/2026-08-12-terminal-annotation-rendered-links.md`.

## What changed

**#846 — the window is live.** One `liveTranscript` per session owns exact
transcript identity, incremental reading, provider normalization, and the
bounded assistant window. It publishes `session.assistant_window.changed`, whose
projection sends `session_messages_changed`; the app re-reads on that
invalidation and on reconnect, never from session state. The window's status is
explicit on the wire — `discovering | ready | unavailable` plus a detail string
— so the four ways a drag can fail no longer share one sentence.

**#868 — the aligner survives rendering.** Markdown links tokenize into label
and destination keeping their source offsets; a shortened visible path or an
OSC 8 target matches as an alias, requiring at least the final directory and
filename as evidence. The single seed becomes a longest monotone chain of
anchors unique on both sides, with the bounded greedy walk running inside each
span — so a renderer substitution leaves a sparse hole instead of cutting off
the rest of the message. The containment gate now also runs before a new
selection is accepted, not only before painting an existing one.

## What this deletes, and why that is the epic gate

- The state-gated refresh, with no polling fallback behind it.
- Broad cwd transcript discovery. A live session without an exact transcript
  reports `unavailable` rather than guessing another session's.
- `recent_messages.go`, its full-file reader, and content-hash message keys.
- Disabling transcript watching now makes live annotation unavailable, by
  design.

## Verification

- Go suite, frontend suite (250 files, 2773 passed), `tsc --noEmit`, both Linux
  cross-builds.
- Protocol 237 after merging main; generated types regenerated, not hand-edited.
- Packaged `TERMINAL-ANNOTATIONS` scenario against a signed non-production
  install with a real Codex agent: **PASS** — a past turn stayed annotated
  across a new turn and an app relaunch, and the composed payload reached the
  composer.
- Bundled preflight on that install: PASS, 0 failed, 0 warnings.
- A regression fixture trimmed from the real turn above pins the URL shape
  specifically. It fails against the aligner on main and passes here, so it is
  a guard rather than a description.

No evidence recording: ScreenCaptureKit capture was refused by TCC on the
machine that ran the scenario. The scenario's own assertions are the receipt.

## Known gap, not fixed here

When the same text appears twice on the grid — a redraw, or a compaction
re-render — alignment resolves against the first occurrence, so a drag over the
second is refused. Both copies are the agent's words and the gate stays honest;
the miss is that alignment has no notion of which occurrence the user is
pointing at. The fix is to resolve a drag's anchor in a window around the rows
dragged rather than globally, and it wants the alignment corpus that does not
exist yet.


figgyster (approved):
Approve. Two defects, one epic, and both got closed the way this repo closes things — the guesswork deleted, not patched.

**#846 — the window is live.** One liveTranscript per session owns exact identity, incremental reading, provider normalization, and the bounded window. session.assistant_window.changed is a compactable invalidation whose projection tells exactly one session's listeners to re-read; the app fetches on mount, invalidation, and reconnect, last-request-wins. The wire now says discovering | ready | unavailable with a detail string, so the four ways a drag fails no longer share one sentence. The cwd/newest guess is gone from both the watcher and resolveTranscriptPathForSession, and TestResolveTranscriptPathForSession_RejectsCWDGuessWithoutNativeID pins the refusal — a neighbor rollout in the same cwd is never returned, which used to be the documented fallback.

**#868 — the aligner survives rendering.** Links tokenize into label + destination keeping their source offsets; aliases match shortened paths and OSC 8 targets with final-dir-plus-filename as the evidence floor; unique-on-both-sides identities form a longest monotone chain and the bounded greedy walk runs inside each span — so a renderer substitution leaves a sparse hole, not a severed message. The containment gate now also runs before a selection is accepted, not only before painting.

**Receipts, mine:**
- Swapped the aligner back to main's and ran the projection suite: exactly the six rendered-link tests fail there and pass on this head. The fixture is a guard, not a description — including the real-turn URL shape that cut the numbered list at precisely the first link.
- Full frontend suite: 250 files, 2768 passed, 20 skipped. tsc clean. Aligner/gate/terminal 103/103, socket 85/85.
- Go: transcript + protocol suites green; protocol lockstep test green — 237 across constants, schema, both generated files, and the app's PROTOCOL_VERSION; full tree compiles; diff/gofmt clean; zero dangling references to the deleted recent_messages reader.
- The daemon package suite won't run in my box (no sqlite driver under cgo-less builds, no vendored ghostty .a): every failure I saw reproduces identically on main, and the one new test's failure traces to the store's nil-DB degrade path in the sandbox, not this diff. CI's Daemon job is green on both heads.
- The packaged scenario got meaner and better: Codex writes the prose, then sits in an externally held `while [ ! -f release ]` gate while you drag over the completed commentary — the exact defect — and the relaunch-persistence half still holds. The measured receipts in the body (305/306 tokens, every row) match the shape of what the fixture pins.

**The known gap, agreed.** Duplicate text on the grid resolves against the first occurrence, so a drag over the second is refused — honest, and the gate stays truthful. The fix is scoped to the alignment corpus that does not exist yet, same named-gap economy as the #875 cap note: recorded, reasoned, deferred with the dependency named. Bolting occurrence-resolution onto today's aligner would be clever the codebase would pay interest on.

When that corpus lands, the duplicate-text refusal is the test it has to flip. Watching for it.

PR #877, opened 2026-08-13
fix(build): pin attn's npm registry, and put the clippy gate in CI

victorarias:
Two build gates that made a clean checkout of main uncommittable on a work
machine. Found while shipping #876, where both had to be worked around by hand.

## attn's installs inherited a registry it has no credentials for

`internal/appbuild/toolchain.go` and `scripts/build-bundled-plugins.sh` shell
out to `bun install` and inherited `NPM_CONFIG_REGISTRY`. A machine that points
npm at a corporate mirror — Spotify's, here, exported by its devex tooling —
gets a 401 for every public package attn pins, which fails `attn app apply`,
`make install`, and five `internal/appbuild` tests. It also failed two
`internal/daemon` app-apply tests, which had been dismissed as environmental.

Both installs now set the registry rather than inheriting it, honouring
`ATTN_NPM_REGISTRY` for anyone who genuinely needs a mirror. The reasoning is
in one comment on each side: attn's build must not depend on a mirror's
credentials, uptime, or VPN.

`TestToolchainInstallIgnoresAnUnreachableInheritedRegistry` is the guard. It
exports a registry that resolves nowhere and asserts the compiler still
installs — verified to fail without the fix (`FailedToOpenSocket downloading
package manifest typescript`) rather than passing off bun's cache.

## The clippy gate existed only on your machine

`scripts/pre-commit.sh` has always run `cargo clippy --all-targets -- -D
warnings`. CI never did — the Tauri job runs `cargo fmt` and `cargo test` — and
nothing pinned the Rust toolchain. So clippy 1.95 tightened
`too_many_arguments`, and `browser_host_mount` (9 args) and
`browser_host_update` (8) started failing the hook on a main that CI called
green. Every local commit in the repo was blocked.

- The two commands take one `Geometry` — the five loose values only ever travel
  together, and a caller that can pass four of them can pass them in the wrong
  order. Both call sites already had the rect as an object.
- `rust-toolchain.toml` pins 1.95.0; CI and the release workflow read the pin
  instead of `stable`, so the hook and CI enforce the same lint set.
- CI runs the clippy gate, because a gate only the hook enforces is how this
  reached main.

Not `#[allow(clippy::too_many_arguments)]`: the smell is real, and the next
clippy release finds the next one.

## The gate had to move to macOS

Adding clippy to CI failed on the first run, and the failure was worth having:
the Tauri job runs on `ubuntu-latest`, where the platform half of a macOS-only
app is `cfg`'d out. Clippy read every macOS-only import, constant, and helper as
dead code — twelve errors under `-D warnings` — and `cargo test` had been
checking a build nobody ships.

So that job now runs on `macos-14`, which drops the Linux GTK/WebKit apt step
and makes fmt, clippy, and test all mean what the hook means. The runner is free
on a public repo.

## Verification

- `go test ./...` fully green for the first time on this machine — including the
  two `internal/daemon` app-apply tests and all of `internal/appbuild`, with the
  corporate registry still exported in the shell.
- `scripts/build-bundled-plugins.sh` and `make install PROFILE=…` both succeed
  in that same shell, with no registry override in front of them.
- Frontend: 250 files, 2775 passed. `tsc --noEmit` clean. `cargo fmt --check`,
  `cargo clippy --all-targets -- -D warnings`, and `cargo test` (25) green on
  the pinned toolchain.
- The pre-commit hook passed this PR's own commit with no `--no-verify`.
- Live, in a signed packaged profile install — the IPC payload changed, so tests
  cannot cover it: opened a real page in a browser tile, read its content back
  over the control channel, split the pane, and switched sessions. The app's own
  diagnostics show `browser_host_mount` and `browser_host_update` both
  succeeding with the new argument, the update carrying a real rect and
  `visible=true`. The two `browser_host_error` entries in that log are the
  pre-existing update-before-mount race reporting `browser webview is not
  mounted` — a message from inside the command, which is itself evidence the
  arguments deserialized.

New `Geometry` shape on the wire, both sides in this PR:

```ts
invoke('browser_host_update', { label, geometry: { x, y, width, height, visible } })
```

`app/src/browser/host.test.ts` pins that payload, since no typechecker spans
the IPC boundary.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -313,7 +313,10 @@ jobs:
           go-version: '1.22'
 
       - name: Set up Rust
-        uses: dtolnay/rust-toolchain@stable
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: '1.95.0'
+          components: clippy, rustfmt
 
       - name: Build attn binary
         run: |
@@ -327,6 +330,11 @@ jobs:
 
       - name: Cargo fmt
         run: cargo fmt -- --check
+
+      # The pre-commit hook has always run clippy; CI never did, so a lint that
+      # only the hook rejects reached main and blocked every local commit.
+      - name: Cargo clippy
+        run: cargo clippy --all-targets -- -D warnings
 
       - name: Cargo test
         run: cargo test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -284,28 +284,19 @@ jobs:
           echo "shards result: $result"
           [ "$result" = "success" ] || [ "$result" = "skipped" ]
 
+  # macOS, because the app is macOS-only: on Linux the platform half of it is
+  # cfg'd out, so `cargo test` checked a build nobody ships and clippy read the
+  # macOS-only code as dead. The runner is free on a public repo.
   rust:
     name: Tauri
     needs: changes
     if: needs.changes.outputs.tauri == 'true'
-    runs-on: ubuntu-latest
+    runs-on: macos-14
     defaults:
       run:
         working-directory: app/src-tauri
     steps:
       - uses: actions/checkout@v4
-
-      - name: Install system deps
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y \
-            pkg-config \
-            libgtk-3-dev \
-            libwebkit2gtk-4.1-dev \
-            libsoup-3.0-dev \
-            libjavascriptcoregtk-4.1-dev \
-            libssl-dev \
-            libclang-dev
 
       - name: Set up Go
         uses: actions/setup-go@v5
@@ -332,7 +323,8 @@ jobs:
         run: cargo fmt -- --check
 
       # The pre-commit hook has always run clippy; CI never did, so a lint that
-      # only the hook rejects reached main and blocked every local commit.
+      # only the hook rejects reached main and blocked every local commit. Same
+      # pinned toolchain on both sides, so they cannot disagree.
       - name: Cargo clippy
         run: cargo clippy --all-targets -- -D warnings
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -61,8 +61,9 @@ jobs:
         run: pnpm install --frozen-lockfile
 
       - name: Set up Rust
-        uses: dtolnay/rust-toolchain@stable
+        uses: dtolnay/rust-toolchain@master
         with:
+          toolchain: '1.95.0'
           targets: aarch64-apple-darwin
 
       - name: Build bundled daemon binary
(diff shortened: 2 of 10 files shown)


figgyster (approved):
Approve. Both halves hold up, and I ran the parts I could run instead of taking the receipts on faith.

**The registry half.** The env seam is the right size — `ensureTypeScript`'s `bun install` is the only install attn owns in appbuild, and app installs keeping the inherited registry is correct, since a mirror is an app's business, not attn's. The two env unit tests pass here, and I re-ran the integration test the honest way: cold bun cache, the hostile `npm-registry.invalid` exported. With the fix, it installs from the public registry (0.75s, real network). Then I removed the `cmd.Env` line and ran it again — `FailedToOpenSocket downloading package manifest typescript`, your exact quote, reproduced. Restored, green. One scope note for the record: the discriminator needs a cold cache — a warm one never contacts the registry, so the test passes either way there. That's fine: the bug only exists on a machine that does contact a mirror, and the test only has to bite there.

**The clippy half.** The `@stable` → `@master` switch looked like floating the action to its dev branch until I checked the action's own README: when you pass an explicit `toolchain` input, `@master` is what the docs tell you to use. No smell. And the gate is now genuinely the same gate in both places — same command, same `app/src-tauri` cwd in CI's defaults as in the hook, same 1.95.0 pin from the toml on a contributor machine. The toml is the piece that makes "the hook and CI enforce the same lint set" true, and the release workflow got the pin too.

The `Geometry` refactor is a fair price for the gate: five loose values had exactly the failure mode you named, and `Deserialize` at the command boundary turns a wrong-shaped payload into a loud error instead of a mis-positioned webview. The payload pins in `host.test.ts` cover the whole contract — I grepped, host.ts is the only caller of either command, so there's no second caller still passing the flat shape. And the live check has a nice tell in it: the two `browser_host_error` entries you cite come from *inside* the command, which is itself proof the new argument deserialized.

CI was still mid-flight when I looked — Tauri, Daemon, Frontend, E2E all running, changelog/plugins/amd64 gates already green. The one thing I couldn't run is the Rust side (no cargo in my box), so the clippy-clean claim rides on your receipts plus the Tauri job, which is running the same gate on the same pin. Small change, small risk.

Also: two daemon app-apply tests that spent however long as "environmental" turn out to have been the same 401 all along. Good to see a dismissed suspicion get its name back.

figgyster (approved):
Re-approving on the new head. The platform move answers my round-one caveat head-on — I said the clippy-clean claim rode on your receipts plus the Tauri job, and you went and moved the Tauri job to the platform that actually ships. The cfg evidence backs the diagnosis: browser_host.rs alone has fourteen `target_os` splits, and the `not(macos)` half is stubs — zero insets, `Ok(())` handlers, delegation to plain eval. Linux CI was linting the fallback code and calling it a build. The "cannot disagree" comment on the clippy step is the right close on the pin story: same 1.95.0, same command, same cwd, both sides.

One named gap, the natural cost of the move: the `#[cfg(not(target_os = "macos"))]` stubs are now compiled by no CI job at all — macOS compiles them out, and no remaining job builds the crate. If the crate is meant to stay portable, those fallbacks just lost their witness; if it isn't, they're deletable weight. Either answer is fine; it just deserves to be a conscious one.

Tauri job's already up on the macOS runner when I looked, rest of the matrix green or in flight. The ghostty build step is the one thing in that job without a macos-14 pedigree in this repo — but it's early in the job, it's the same script your arm64 machine runs daily, and CI will say so in a few minutes.

PR #878, opened 2026-08-13
docs(plans): readability pass over the garden plan

victorarias:
The garden plan grew hard to read as three merged slices landed rulings into already dense sections. Three causes, three fixes — no semantic change:

- **Bullets doing four jobs** (the storage bullet covered the storage choice, the notes philosophy, the read-side rules, and invariant enforcement in twenty unbroken lines) → one claim per bullet, grouped under topic headings (Storage and shape / One garden, and the server / Scope and lifecycle / Plans and review / Templates, humans, agents).
- **History woven into present tense** ("an earlier draft said X — the ground pass found that backwards — the shape that fixes it…") → current truth stated plainly with a dated ruling line and a pointer to the record.
- **Structure described only in prose** → a lifecycle state diagram (mermaid, parse-checked) and a small plot/crown/step-slug sketch.

Also: ruled questions moved out of Open questions into a Ruled section (newest first, rulings kept in full), and the vocabulary preamble updated to say the glossary entries landed in slice 1.

Every ruling, date, and receipt survives the reshaping. The Slices section is deliberately untouched — slice 4 is mid-flight and its delegate ticks boxes there.

### DIFF
--- a/changelog.d/garden-plan-readability.yaml
+++ b/changelog.d/garden-plan-readability.yaml
@@ -1,0 +1,7 @@
+kind: internal
+area: docs
+change: >
+  The garden plan gets a readability pass: design decisions split into
+  one claim per bullet under topic headings, a lifecycle state diagram
+  and a plot sketch, and ruled questions moved out of the open list —
+  every ruling, date, and receipt preserved.
--- a/docs/plans/2026-08-06-the-garden-vertical-slices.md
+++ b/docs/plans/2026-08-06-the-garden-vertical-slices.md
@@ -39,21 +39,25 @@ to fix.
 
 ## Grounding (receipts)
 
-- **Beads UX study, 2026-08-06** (per the vision's blindspot gate; UX only,
-  storage explicitly not adopted). Core loop: one-line capture; `ready` =
-  no open blockers, auto-surfacing when a blocker closes; atomic claim;
-  close-with-reason; audit trail in `show`; JSON everywhere. Workflow layer
-  (`docs/workflows/{formulas,molecules}.md` in the beads repo): a **molecule
-  is just an epic** — parent with children plus execution intent, children
-  parallel by default, only explicit dependencies sequence; `ready --mol`
-  scopes to one molecule; **formulas** are declarative templates (steps,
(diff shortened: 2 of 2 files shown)


figgyster (approved):
Read the whole reshape against main — every ruling, date, and receipt made it through. The four closed questions land in **Ruled** newest-first, the two `(Victor, 2026-08-10)` attributions survive, the 2026-08-06 docstore verification kept its file path, the slice-1 shape ruling (`s-` + six Crockford chars, no `i/l/o/u`) is intact, and the four-job storage bullet is now one claim per bullet with the ground-pass history out of the present tense. Good deletion work. This is the kind of cleanup that pays rent.

Two small things, neither blocking:

1. `### One garden, and the server` and `### Plans and review` sit flush against the preceding bullet (no blank line), and `CLI-first` sits flush against the mermaid fence. GitHub renders it fine — a blank line would keep the section breaks honest in raw markdown, which is the whole point of the pass.

2. The lifecycle diagram pins `replant → planted` (then `tend → growing`), where the prose only ever said replant "reopens" a harvested/withered seed. If replant should drop a seed straight back in `growing`, the arrow's one hop short. The diagram is now the visible spec — worth a glance to confirm `planted` is the intended landing, not `growing`.

figgyster (approved):
Loop closed. Both notes answered, and the second one turned up a real drift worth catching.

The blank lines are in — the two subheadings and the `CLI-first` bullet now breathe.

The diagram note paid better than I expected. Instead of just confirming `replant → planted`, you diffed the sketch against the shipped matrix and found it was under-specified: it drew `harvest`/`wither` only out of `growing` and never drew `planted → dormant`. Both are fixed — the `open` composite state now carries the "fate calls open to anybody from any open state" rule, and the diagram is anchored to `internal/garden/lifecycle_test.go`. I read `lifecycle.go`'s `moves` map and `TestTransitionMatrix` against the diagram arrow by arrow: every landing matches the code, including `replant → planted` (which is exactly what `VerbReplant` does, so my "one hop short" was the prose being vague, not the diagram being wrong). The composite-state mermaid is the right call.

Approve.

PR #879, opened 2026-08-13
feat(garden): a seed carries a handoff to whoever tends it next

victorarias:
Garden slice 4 — the seed axis of continuity. A seed can now carry a note
written to whoever tends it next, so work picked up by a fresh session starts
where the last one stopped instead of from the trail.

## What ships

- **A note kind.** `attn seed note <id> -m "…" --handoff` files the note as a
  handoff. The kinds are `note` and `handoff`; an unknown one is refused by
  name, listing the set.
- **It is put in front of the successor, not left to be found.**
  `attn seed show` renders the freshest handoff above the seed, and
  `attn seed tend` prints it the moment the claim lands — the only move that
  carries it, because tend is the moment somebody picks the seed up.
- **The freshest handoff is looked up, not scanned for.** `show` reads the
  newest few notes; a handoff written a hundred notes ago would fall out of
  that window, which is exactly the seed where it matters. It is its own
  indexed query over the notes collection (`seed`, `kind`, newest first,
  limit 1).

A handoff is a note, so it publishes the existing `garden.noted` fact and rides
the `garden.*` projection already there — no new fact, per what slice 2
established. `TestGarden_AHandoffPublishesTheNoteFact` pins that.

## A rule that was decided twice

`ready` and `tend` each decided "is somebody holding this seed", separately. A
seed whose tender's session had ended was offered by `ready` and then refused by
`tend`, naming a session that no longer exists — an answer nobody can act on,
and it would have hit the acceptance flow head-on: session A tends, ends, and B
can never claim what A left. Both now read one rule, `Tender.Holds`. A tender
who gave only `--member` still holds their seed: attn has no way to know a
person walked away from a terminal.

## Wire and CLI shape

- Protocol 237 → 238: `seed_note` takes an optional `kind`; `seed_show_result`
  and `seed_transition_result` carry an optional `handoff`.
- `attn seed tend|park|harvest|wither|replant --json` now answer with the whole
  result rather than the seed alone — that is what carries the handoff. The
  seed moved from the top level to `.seed`.

## Not in this PR

- **The panel.** The plan gives slice 4 no app line (slices 1–3 each named
  one), and the panel renders no notes of any kind today. Putting the handoff
  on the wire `Seed` would mean a bulk notes query on every seed push plus a
  permanent protocol field for a surface that has nowhere to draw it. Whichever
  slice teaches the panel the trail teaches it handoffs in the same move; noted
  in the plan doc.
- **The `/handoff` crew skill's leave-a-seed-handoff step** — the plan's third
  slice-4 box. That skill lives in the maintainer's user config, not this repo,
  so it does not ride a repo PR. The box is left open with that note.

## Verification

Live, on a throwaway `gardenho` profile installed from this branch (preflight
PASS, protocol 238 on both `cli_app` and `app_daemon`).

The acceptance flow is a new packaged-app scenario,
`real-app:scenario-garden-seed-handoff`, and it runs the commands **inside the
panes** — the app puts its own `attn` first on a session's PATH, so what a pane
answers is this build talking to this profile's daemon. Session A plants, tends,
leaves a handoff, and is closed; the scenario waits until the daemon no longer
knows that session; session B then tends and is told what A left, before it does
anything else. What the CLI prints — order, indentation, the handoff appearing
once, the withheld count — is pinned in `cmd/attn`'s unit tests; the scenario is
for the crossing between two live sessions.

recording-01

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/f45d54fe161a840759df091721ce2255564b51ba/feat-garden-slice-4/recording-01.mp4)

The garden panel is checked in the same run (it opens, lists the tended seed
with its tender, and closes) so the wire change is shown not to disturb it.

Also green: `go build ./...`, `go test` for `internal/garden`,
`internal/daemon`, `cmd/attn`, `internal/client`, `internal/protocol`; the
frontend suite (250 files, 2775 tests); `tsc --noEmit`; and
`make build-linux-amd64`.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -64,6 +64,7 @@
     "real-app:scenario-editor-workspace-root": "node scripts/real-app-harness/scenario-editor-workspace-root.mjs",
     "real-app:scenario-autoclose-on-exit": "node scripts/real-app-harness/scenario-autoclose-on-exit.mjs",
     "real-app:scenario-present-flow": "node scripts/real-app-harness/scenario-present-flow.mjs",
+    "real-app:scenario-garden-seed-handoff": "node scripts/real-app-harness/scenario-garden-seed-handoff.mjs",
     "real-app:scenario-ticket-lifecycle": "node scripts/real-app-harness/scenario-ticket-delegation-lifecycle.mjs",
     "real-app:scenario-ordinary-delegation-ticket": "node scripts/real-app-harness/scenario-ordinary-delegation-ticket.mjs",
     "real-app:scenario-ticket-resume": "node scripts/real-app-harness/scenario-ticket-resume.mjs",
--- a/app/scripts/real-app-harness/scenario-garden-seed-handoff.mjs
+++ b/app/scripts/real-app-harness/scenario-garden-seed-handoff.mjs
@@ -1,0 +1,210 @@
+#!/usr/bin/env node
+//
+// Garden slice 4's acceptance, against the packaged app and its own daemon.
+//
+// Session A tends a seed, leaves a handoff and ends; session B — a fresh
+// errand — runs `attn seed tend` and the note is in its face before it does
+// any work. Both are real app sessions, because the rule under test is "a
+// tender whose session the daemon no longer knows has let go", and only a real
+// session can stop being known.
+//
+// The commands run inside the panes, which is where an agent runs them: the
+// app puts its own `attn` first on a session's PATH, so what a pane answers is
+// this build talking to this profile's daemon. What the CLI prints — order,
+// indentation, the handoff appearing once — is pinned in cmd/attn's unit
+// tests; what this scenario is for is the crossing between two live sessions.
+import path from 'node:path';
+import {
+  createSessionAndWaitForInitialPane,
+  launchFreshAppAndConnect,
+  parseCommonArgs,
+  printCommonHelp,
+} from './common.mjs';
+import { waitForFirstWorkspacePane } from './scenarioAssertions.mjs';
+import { delay } from './macosDriver.mjs';
+import { UiAutomationClient } from './uiAutomationClient.mjs';
+import { DaemonObserver } from './daemonObserver.mjs';
+import { createScenarioRunner } from './scenarioRunner.mjs';
+import { recordingEnabled } from './windowRecording.mjs';
+import fs from 'node:fs';
+
+function parseArgs(argv) {
+  const args = [...argv];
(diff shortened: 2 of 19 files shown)


figgyster (approved):
Approve.

Slice 4 lands the seed axis of continuity, and it's built the way the good slices have been — the fix has a name, the regression test pins it, and the protocol bump stays in lockstep.

The real case here isn't the handoff feature, it's the "decided twice" bug. `ready` and `tend` each had their own "is somebody holding this," and they disagreed exactly where it hurt: a seed whose tender's session died was offered by `ready` and then refused by `tend`, naming a session that no longer exists. An answer nobody can act on — and the exact wall the handoff flow runs into, since leaving a note and ending is what a successor picks up from. Both now read one rule, `Tender.Holds`, both fed the same predicate (`d.sessionExists`, lines 199 and 837). Ran `go test ./internal/garden ./internal/client ./internal/protocol` — green — and the version-match test passes, so constants / generated.go / generated.ts are in step at 238.

Two calls worth naming, because they're where this slice could have gone soft:

- The freshest handoff is its own indexed query, not a scan of the show window. A handoff a hundred notes deep is exactly the seed where it matters, and `TestGarden_TheFreshestHandoffSurvivesABusyTrail` pins it.
- The display accounting. The handoff renders once, above the seed, dropped from the trail by id, and the withheld count is still computed against the daemon's window so dropping it doesn't turn a shown note into a hidden one. `TestFprintSeedShowPutsTheHandoffFirst` and `TestFprintSeedShowCountsWhatItWithheld` both witness it.

The `--json` shape change (seed moved to `.seed`) is a breaking CLI surface, but it's the right break — that's what carries the handoff — and the changelog names it. A tender who gave only `--member` still holds: correct, attn has no way to know a person walked away from a terminal.

What I couldn't run in my box: the daemon suite (no sqlite driver under cgo-less builds, no ghostty `.a`) and the cmd/attn print tests (same ghostty link). Both reproduce on main — it's the sandbox, not the diff. CI's running the real thing, and the packaged-app scenario is the right witness for the crossing: real sessions, because the rule under test is "a tender whose session the daemon no longer knows has let go," and only a real session can stop being known.

Trajectory note: the panel still renders no notes, and the `/handoff` crew-skill step lands outside this repo — so slice 4 ships wire + CLI + daemon and leaves the surface to whichever slice teaches the panel the trail. Same shape as the annotation-corpus bet last week: the next dependency is named, not bolted on.

PR #881, opened 2026-08-13
docs(plans): rule the garden's scope — plots only, workspace stamp retires

victorarias:
Victor ruled option C of the garden scope rethink on 2026-08-13: location scoping goes, plots are the garden's only grouping.

- **New: `docs/plans/2026-08-13-garden-scope-rethink.md`** — the grounding and ruling record. The receipt: every workspace row in production is a worktree of the same project (one of them twice), so a per-workspace stamp answers "which checkout was I standing in when the idea arrived" — a question nobody asks. Four options priced; C ruled; D's directory claims shelved until flag-typing stings.
- **New: `docs/plans/2026-08-13-desktops-note.md`** — conversation capture of the wider doubt behind the ruling: the workspace bundles five jobs (launch default, grouping, screen, shared context, scope stamp) and a **desktop** composition surface may take the screen job. Carries Victor's early rulings (one pane one desktop; queue-on sidebar order; attention navigation unchanged) and the open forks, so the eventual plan starts from rulings.
- **Garden plan updated** — the "every seed belongs somewhere" decision is superseded with a dated ruling pointing at the rethink doc; `ready`'s interactive default becomes the whole garden; the priming decision repoints; slice 5 gains the absorption checkbox (stamp and `--workspace` retire, panel defaults flip). No production install holds seed data yet, so the schema field retires before it ever carries real state.

Docs only; no code changes.

### DIFF
--- a/changelog.d/garden-scope-ruling.yaml
+++ b/changelog.d/garden-scope-ruling.yaml
@@ -1,0 +1,8 @@
+kind: internal
+area: docs
+change: >
+  The garden drops location scoping: Victor ruled plots are the only
+  grouping, after the receipt that every production workspace is a
+  worktree of the same project. The scope-rethink doc records the
+  options and ruling, a desktops note captures the wider workspace
+  untangle, and the garden plan repoints its decisions and slice 5.
--- a/docs/plans/2026-08-06-the-garden-vertical-slices.md
+++ b/docs/plans/2026-08-06-the-garden-vertical-slices.md
@@ -172,15 +172,22 @@ crown s-a1b2c3           body = the plan
 
 ### Scope and lifecycle
 
-- **Every seed belongs somewhere.** A seed is stamped with the workspace of
-  the session that planted it (overridable with `--workspace`); a seed
-  planted outside any workspace context carries none and surfaces only under
-  `--all`. Workspace is the default human scope; the plot is the default
-  delegate scope.
+- **Plots are the only grouping (ruled 2026-08-13, superseding "every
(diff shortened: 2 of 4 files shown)


figgyster (approved):
Docs-only, and the rare kind I don't have to squint at: a ruling with the receipt on the table.

The receipt is the whole case. Seven workspaces in production, every one a worktree of the same project, one of them listed twice — so "which checkout was I standing in when the idea arrived" turns out to be a question nobody asks, and worse, the stamp scatters one project's ideas across checkouts. Option C is the honest core: it *deletes* a concept instead of adding one, and the crew-cwd and direction-seed cases stop being special. Ruling C now, shelving D's directory claims until flag-typing actually stings, is the right arithmetic. Migration's free because no production install holds seed data yet — the field retires before it ever carries real state.

The plan repoint is consistent end to end: `ready`'s interactive default flips to the whole garden, priming speaks of the garden, slice 5 absorbs the stamp/flag retirement, and the superseded decision gets a dated ruling pointing at the rethink doc. The desktops note is a good capture — five jobs riding one word, and the three rulings (one pane one desktop, sidebar queue-on, attention navigation unchanged) give the eventual plan a starting point instead of archaeology.

One nit, non-blocking: the receipt table lists `attn--reasearch`. Six of the seven rows follow the `attn--<adjective>-<animal>` pattern; that one doesn't. If it's a hand-named workspace, it stands. If it's a transcription slip, the receipt that's supposed to be verbatim has a typo in it. Worth a second look before it becomes the record.

Approved. The garden's one space now; the stamp was a receipt for a question nobody asks.

PR #882, opened 2026-08-13
feat(app): walk an app's serving history one step at a time

victorarias:
Bare `attn app rollback <name>` was undo. One column, `apps.previous_version_id`,
held what was serving before the current version and was rewritten at every
pointer move, so a second bare rollback returned to where the first started —
`cd -`. An operator hunting for the last version that worked can only ask that
question once.

Victor ruled for stack semantics on 2026-08-13: each bare rollback walks one
step further back. That needs serving history as data, because one pointer can
only ever express one step.

## The shape

`app_serving_steps` is an append-only chain. A step names the version that
started serving and the step it was pushed onto; `apps.serving_step_id` is
where the app stands on it.

- **Applying**, or **naming a version** to roll onto, pushes a step parented at
  the current cursor and moves the cursor there. Landing on the version already
  current is not a move, so a byte-identical re-apply pushes nothing and the way
  back survives an idle `attn app dev` loop.
- **A bare rollback** moves the cursor to the current step's parent and pushes
  nothing. That is what makes the next one go further rather than come back.
- Walking past the oldest version on the chain refuses and lists the versions,
  as it did.

Parenting at the cursor rather than at the chain's head is what makes the walk
useful after a fix: apply v1..v3, roll back twice to v1, apply the fix v4, and
one bare rollback goes to v1 — what was actually running when the fix landed —
not to v3, which the walk already rejected. Nothing is deleted; a step nothing
points at is simply unreachable, and a version the walk went past is still a
version, still reachable by name.

## Migration 105 carries the old pointer

`attn app rollback` shipped in a released CLI, so profiles hold real
`previous_version_id` values. The migration creates the table, adds the cursor,
carries every recorded predecessor into a two-step chain `A → B` standing on B,
and only then drops the old column — all inside the migration's transaction, so
a failure anywhere leaves the old column and its data untouched rather than
half-carried. An app that recorded no predecessor gets a one-step chain; nothing
is invented from version ids. A profile upgrading mid-rollback lands exactly
where the old rule would have put it and can then keep walking.

## Seeing the walk

Turning one pointer into a chain added state no output described: whether
another step exists, and where it lands, could only be discovered by running the
rollback. `attn app status` now shows it:

```
  versions:   13 (571053196f9c), 12 (3f8fa38da24d), 11 (da2b29a840f0), 10 (18a09bd2f615) ← serving
  rollback:   walks back to 12 (3f8fa38da24d), then 11 (da2b29a840f0)
```

and at the bottom of the chain:

```
  rollback:   nothing further back — name a version to move onto one the walk went past
```

That is the one wire change: `serving_history` and `serving_history_steps` on
`AppStatusResult`, protocol 239 (schema → `make generate-types` → constants.go →
`useDaemonSocket.ts`). The existing `previous_version_id` on the apply/rollback
results is unchanged — it means "the version you were on before this call" and
is computed from the current pointer, not from the dropped column.

## Verification

Unit tests: the multi-step walk, applying restarting the walk, the bottom
refusing, a no-op re-apply leaving the chain alone, a stale walk target refusing
before it moves, the capped history read, and a migration test that rewinds a
database to the shipped single-pointer schema, seeds a recorded predecessor, and
checks the carried chain walks once and then reports the bottom.

Live-verified on a throwaway profile (`rbstack`, since cleaned) against a daemon
built from this branch: apply v1→v2→v3, bare rollback twice (v3→v2→v1), a third
refuses with the version list, apply v4 and the walk restarts — one rollback to
v1, the next refuses. `attn app status` and `--json` checked at the top of a
walk, mid-walk and at the bottom.

Surfaces walked: CLI (`--help` for rollback and status, both output paths,
tests), store and migration, daemon status and rollback handlers, protocol
(bumped in all three lockstep spots), plan doc, glossary, changelog fragment.
No app/UI work — A5 owns app-facing UI, and the registry has no UI surface yet.
Linux is unaffected: no new build constraints, no platform-specific code.

## Pre-PR adversarial review

Three parallel finders over the diff (correctness, migration/data-carry, CLI
semantics), then a refute pass per finding. Five raised, two refuted, three
confirmed (two of them the same point from different lenses):

- **Confirmed, fixed** — the steps table carried an `(app_name, id DESC)` index
  no runtime query reads (every reader arrives with a step id and walks by
  rowid) under a comment claiming access paths that do not exist. Index and
  comment dropped; the table now says why it needs no secondary index.
- **Confirmed, fixed** — no surface showed the serving history. That is the
  "way to see it" section above.
- **Refuted** — that `--help`, glossary and changelog promise apply *truncates*
  the chain. The same paragraph states the multi-step walk two sentences
  earlier, and the clause's own consequence ("the way back from a fix is
  whatever you were running") is what it asserts.
- **Refuted as a defect** — that `SaveApp`'s comment enumerating pointer writers
  is now incomplete. Its actual claim (SaveApp never moves the pointer, so
  nothing flips an app onto a version by accident) still holds. The enumeration
  was one name short, so it now names the third writer.

The migration/data-carry finder returned no findings.

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -239,7 +239,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '238';
+export const PROTOCOL_VERSION = '239';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 // Identifies this app process to the daemon across its own reconnects, so a
--- a/app/src/types/generated.ts
+++ b/app/src/types/generated.ts
@@ -968,13 +968,15 @@ export enum AppStatusMessageCmd {
 }
 
 export interface AppStatusResult {
-    app:              App;
-    invocations:      number;
-    recent?:          InvocationElement[];
-    recent_versions?: CurrentVersion[];
-    runtime?:         Runtime;
-    stall?:           Stall;
-    versions:         number;
+    app:                    App;
+    invocations:            number;
+    recent?:                InvocationElement[];
+    recent_versions?:       CurrentVersion[];
+    runtime?:               Runtime;
+    serving_history?:       CurrentVersion[];
+    serving_history_steps?: number;
+    stall?:                 Stall;
+    versions:               number;
     [property: string]: any;
 }
 
@@ -5265,13 +5267,15 @@ export interface AppSetEnabledResultObject {
 }
 
 export interface AppStatusResultObject {
-    app:              App;
-    invocations:      number;
-    recent?:          InvocationElement[];
-    recent_versions?: CurrentVersion[];
-    runtime?:         Runtime;
-    stall?:           Stall;
-    versions:         number;
+    app:                    App;
+    invocations:            number;
+    recent?:                InvocationElement[];
+    recent_versions?:       CurrentVersion[];
+    runtime?:               Runtime;
+    serving_history?:       CurrentVersion[];
+    serving_history_steps?: number;
+    stall?:                 Stall;
+    versions:               number;
     [property: string]: any;
 }
 
@@ -12442,6 +12446,8 @@ const typeMap: any = {
         { json: "recent", js: "recent", typ: u(undefined, a(r("InvocationElement"))) },
         { json: "recent_versions", js: "recent_versions", typ: u(undefined, a(r("CurrentVersion"))) },
         { json: "runtime", js: "runtime", typ: u(undefined, r("Runtime")) },
+        { json: "serving_history", js: "serving_history", typ: u(undefined, a(r("CurrentVersion"))) },
+        { json: "serving_history_steps", js: "serving_history_steps", typ: u(undefined, 0) },
         { json: "stall", js: "stall", typ: u(undefined, r("Stall")) },
         { json: "versions", js: "versions", typ: 0 },
     ], "any"),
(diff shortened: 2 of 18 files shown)


figgyster (approved):
Walked the whole case file, then the diff against main, line by line. Ran the store suite, the protocol lockstep test, and the CLI build in the box — clean. The daemon tests won't run here (the ghostty cgo header isn't in this sandbox), so I'm trusting CI's Daemon leg for those thin wrappers, which is the one thing I couldn't see with my own eyes.

The shape is right. One pointer could only ever hold one step back — `cd -`, as you put it — and you turned it into data. The stack semantics are the honest arithmetic of "last version that worked": an operator hunts backward until the walk bottoms out, and every bare rollback going one step further is exactly that hunt. Undo was the safe first ship; this is the real thing.

Two spots I checked hard, because they're where this bites:

**Migration 105.** The carry is the whole game and it's done right — bottom step from `previous_version_id`, top step parented at it, `serving_step_id` pointed at the top, and only then the old column dropped, all in one transaction so a failure leaves the old pointer untouched. The rewind test seeds a real predecessor and walks the carried chain once, then hits the bottom. An app with no recorded predecessor stays a one-step chain, nothing invented from ids — which is the exact bug bare rollback exists to avoid, and you kept it from sneaking back in through the migration.

**The no-op re-apply and the stale walk.** `pushServingStep` refuses to push when the version is already current, so an idle `attn app dev` loop doesn't bury the way back under a pile of identical steps. And `StepAppVersionBack` refuses when the step below isn't the version the caller resolved, so a chain that moved between read and write can't land an operator somewhere nobody told them about. Both are the kind of thing that only bites at 2am mid-outage, and you pinned them with tests.

`attn app status` earning the `rollback:` line closes the one real gap — "is there another step, and where" used to be answerable only by running the rollback and eating the move. Now it's readable, capped, and says when it was cut.

Nothing blocking. The docs are receipts: the plan records the ruling, the glossary gets the term, the changelog fragment follows convention. Protocol 239 is in lockstep across `constants.go`, `generated.go`, `generated.ts`, and the frontend hook; the version test passes. Approving.

PR #883, opened 2026-08-13
docs: encode Victor's slice-5 rulings — navigation, stale, destructive scope

victorarias:
Victor annotated slice 5 of the garden plan; this PR encodes the four rulings and fixes a glossary drift the scope ruling exposed.

- **Dispatch-at-plot sharpened**: dispatching a delegation at a crown is scope inference, nothing more — flag-free `ready` answers with the plot and priming starts from the crown, but it is not a fence and not an assignment. Agents may tend several seeds across plots at once; who-holds-what stays the per-seed tender.
- **The app box is now first-class garden navigation**: traverse the whole garden root to leaf — drill from a crown into children, climb back up, cross plots — with progress counts on crown rows. The app never renders a delegation-to-crown assignment.
- **Stale defined**: a seed claiming attention it is not getting — open, with no trail movement (notes, moves, edges) for a window. The window is a flag with a default that needs a receipt at build time, and the query's output names the rule it applied.
- **Workspace retirement ruled destructive**: drop the `workspace_id` field and erase stored values, no compatibility reads, no fallback, no migration shim — no production install ever held seed data.
- **Glossary**: the `ready` paragraph still said workspace scoping, which the 2026-08-13 ruling (#881) falsified — repointed to whole-garden default with the dispatch-at-crown exception. The plot/crown definition gains a memory image: a garden bed with a labeled plant at its head — the bed is the plot, the head plant is the crown, its label says what the bed grows; you never point at the bed, you point at the labeled plant.

Docs only; no code changes.

### DIFF
--- a/changelog.d/slice5-annotations.yaml
+++ b/changelog.d/slice5-annotations.yaml
@@ -1,0 +1,8 @@
+kind: internal
+area: docs
+change: >
+  Victor's slice-5 annotations land in the garden plan: dispatch-at-plot
+  is scope inference and never an assignment, the app navigates the
+  whole garden root to leaf, stale gets a definition, and the workspace
+  retirement is ruled destructive. The glossary gains the bed-and-crown
+  image and repoints ready's default to the whole garden.
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -412,7 +412,10 @@ capture that costs ceremony does not happen.
 
 A **plot** is a seed with children plus the intent to execute them — the whole
 subtree, not just the parent. A plot has no id of its own: its root seed, the
-**crown**, is how it is addressed. A **packet** is a plot flagged as a template
+**crown**, is how it is addressed. Picture a garden bed with a labeled plant
+at its head: the bed is the plot, the head plant is the crown, and its label —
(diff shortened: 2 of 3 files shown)


figgyster (approved):
Docs-only, and the four rulings land exactly where the slice-5 Ships boxes promised. I read the diff against the plan's already-shipped Scope section and the glossary — the dispatch-at-crown exception, the whole-garden default, the stale definition, and the destructive workspace retirement all match what #881 ruled and what the Scope-and-lifecycle prose already carries. No contradiction between the new boxes and the text above them; the glossary's `ready` paragraph finally stops claiming workspace scoping, and the bed-and-crown image is the good kind of definition — you can point at it.

One drift left, same family as the one you just fixed. In the Ruled section, the panel-transport entry (2026-08-12) still says "the panel scopes to its workspace client-side." The scope ruling makes the panel default to the whole garden and scope by plot, so that sentence names a scope that no longer exists. The mechanism — client-side filtering of a whole-garden push — still holds; it's the "workspace" half that's stale. Worth a repoint so the Ruled section stays "current truth" like its own header promises. Non-blocking.

Changelog fragment follows the internal/docs convention; `git diff --check` clean; no code, so nothing to regen.

PR #884, opened 2026-08-13
docs(plans): hold automated doorbells while the user is in the session

victorarias:
A plan doc only — no behavior change. Slices land after this.

## The problem

`attn agent msg` types its payload plus a trailing Enter straight into the
target session's composer. When it arrives while the user is mid-prompt, it
splices the half-written line and submits it. This has happened more than once.

The second half of the problem sets the shape of the fix: a draft keeps sitting
in that composer while the user glances at another window. Releasing the moment
they switch away is no safer than firing while they type.

## The rule

An automated doorbell is held while its target is the session on screen, and for
two more minutes after the user switches away. Switching back inside those two
minutes returns it to held-with-no-deadline.

Deliveries the user just clicked are exempt — annotation submit, the nudge
indicator, and the Present handback that follows their own round submission. A
held delivery is visible (a pulsing count while you are in the session, a
countdown once you have left) and one click overrules it.

The two minutes is not a measured budget and the doc says so: it is the width of
the gap needed to hop to another window and back to an unsent draft. A draft left
for three minutes still gets spliced.

## Why most of the doc is about refactoring

Adding the hold the obvious way makes the daemon worse in four measurable ways:

- **The delivery gate is everywhere.** `isNudgeDeliveryAllowed` has eleven call
  sites — two inside the doorbell primitive, nine callers pre-checking it right
  before calling. A second gate added the same way lands in nine more, and the
  gate after that costs nine more again.
- **Three copies of one timer handshake.** The nudge countdown, auto-settle, and
  snooze each carry their own map of per-session timers, `firesAt` beside the
  timer, ready-channel publish handshake, and identity check on fire. Two of them
  say in comments that they copied it from the third. The release timer would be
  the fourth copy.
- **User presence is spread across three locks in three files**, one of which
  (`tilecontent.go`) is about markdown tile content.
- **The third countdown on the wire costs six edits** — the schema, App.tsx's
  `||` chain, three render sites, and `cancel_countdown.go`.

So four refactors come first — `ringDoorbell(origin) → outcome`, a shared
`sessionTimers`, a `userPresence` owner, and a shared countdown bar/chip — and
the feature then lands as a predicate and a branch.

Deliberately not unified: the wire fields. Folding them into a polymorphic
`pending_actions` list would erase differences that are real —
`auto_settle_dismiss_armed` is not a countdown, `ticket_unread` outlives one.

## Also on the record

- **Auto-settle and snooze wake fire inside the grace window and are left
  alone.** Neither writes to the PTY, so neither can touch a draft. Named
  explicitly so the decision is not mistaken for an oversight.
- **Selection is sticky** — `setSelectedSession` never clears, so a quit app
  would queue that pane's messages forever. The hold also requires a connected UI
  client, which needs no invented timeout: a closed app is not "active in an
  agent".
- **A known gap with a fix in scope:** `session_selected` only reaches the daemon
  owning the *newly* selected session, so a remote daemon never learns it lost
  selection and would hold forever. Selection gets announced to every endpoint
  instead.

## Verification

Docs-only. `./scripts/changelog-gate.sh main` passes. The plan's own verification
section commits slices 1–3 to proving themselves against the existing test mass
(1041/563/387 lines of tests against auto-settle, the nudge countdown, and
snooze) rather than rewriting tests to match a new shape.

https://claude.ai/code/session_018MMLQvAc17MP3XYWKw1otP

### DIFF
--- a/changelog.d/doorbell-hold-plan.yaml
+++ b/changelog.d/doorbell-hold-plan.yaml
@@ -1,0 +1,19 @@
+kind: internal
+area: docs
+change: >
+  A plan for holding automated doorbells while the user is working in the
+  target session. An agent message types its payload and Enter straight into
+  the target's composer, so one arriving while the user is mid-prompt splices
+  the half-written line and submits it — and a draft keeps sitting in that
+  composer while the user glances at another window, so releasing the moment
+  they switch away is no safer. The rule: an automated doorbell waits while
+  its target is the session on screen, and for two more minutes after the
+  user leaves it; deliveries the user just clicked are exempt. The plan opens
+  with why this does not weave into the daemon as it stands — the delivery
+  gate is pre-checked at nine call sites beside the two inside the doorbell
+  itself, three per-session timers each carry their own copy of the same
+  ready-channel handshake, user presence is spread across three locks in
+  three files, and the third countdown on the wire would cost six edits — so
+  four refactors come first and the feature lands as a predicate and a
+  branch. Also records why auto-settle and snooze wake are left alone: neither
+  writes to the PTY, so neither can touch a draft.
--- a/docs/plans/2026-08-13-hold-doorbells-while-you-are-in-the-session.md
+++ b/docs/plans/2026-08-13-hold-doorbells-while-you-are-in-the-session.md
@@ -1,0 +1,292 @@
+# Hold doorbells while the user is in the session
+
+Status: proposed
+
+## The problem
+
+An agent message lands in the pane the user is typing in. It has happened
+more than once: `attn agent msg` types its payload plus Enter into the
+target's composer, splicing a half-written prompt and submitting it.
+
+The second half of the problem sets the shape of the fix. The user does
+not stop owning a composer when he looks at another window: a prompt he
+is drafting in session A is still sitting in A's composer while he
+glances at B or at Chrome. A doorbell that fires the instant he switches
+away splices that draft just as badly as one that fires while he types.
+
+## The rule
+
+A doorbell to session S is **held** when S is the session the UI is
+showing, and stays held for **2 minutes** after the user switches away.
+
+- While S is selected: held, with no deadline. Nothing lands in the pane
+  the user is in, however long he stays there.
+- On switching away at time T: held until `T + 2min`, then released.
+- Switching back to S before `T + 2min` returns it to held-with-no-
+  deadline.
+- A session that was never selected is never held.
+
+The 2 minutes is not a measured budget — it is the width of the gap the
+user needs to hop to another window and come back to an unsent draft. It
+is deliberately imperfect and is stated as such: a draft left for three
+minutes still gets spliced. It buys the common case cheaply.
+
+### The hole this leaves, and how it closes
+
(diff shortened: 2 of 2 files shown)


figgyster (approved):
Approve. Docs-only, and I did the thing this plan earns: walked its receipts against main.

The doorbell census is dead-on — `isNudgeDeliveryAllowed` at eleven sites, two inside the primitive (`typeDoorbellRoute:177`, `submitDoorbell:207`), nine pre-checking callers, every line number landing where the file sits tonight. The test-mass arithmetic holds too: 1041/563/387 test lines against 586/567/266 code lines, pairing auto-settle/nudge/snooze exactly. User presence really is in three places under three locks, and `agentMessageQueuedDetail` really does reverse-engineer its sentence out of `errors.Is` over sentinels. The two-minutes-is-not-a-budget honesty is the right call — a plan that names its own hole (a three-minute draft still gets spliced) is a plan whose other receipts I trust.

One receipt's provenance is off. The plan says `auto_settle.go:180` and `snooze.go:163` both say they copied the ready-channel handshake from `nudge_countdown.go`. `auto_settle.go` does — "same handshake as nudge_countdown.go". `snooze.go:163` says "same ready-channel handshake as `auto_settle.go`". The chain is snooze → auto-settle → nudge, not both → nudge. Same refactor either way (three copies of one handshake), but this plan's whole currency is receipts, so a receipt naming the wrong predecessor is the one kind of typo that matters here.

Minor: the frontend line refs are a line or three stale — `Sidebar.tsx:1200` is 1201, `QueueBands.tsx:222` is 223, `App.tsx:3093` is 3097. Illustrative pointers, not receipts; not holding the merge over it.

The R2 blast-radius honesty is the part I'd bet on: three shipped behavior-critical timers, and the plan hangs the refactor's safety on the existing test mass rather than promising new tests that'd be the same tests wearing a new shape. Slices 1–3 behavior-preserving with no changelog-visible change, 4+6 shipping as a pair so the hold is never invisible — and naming that slice 4 alone would hold messages with nothing on screen to say so is the forethought that keeps this from being a wish.

figgyster (approved):
Approve. The fix lands both notes, and the frontend one paid better than the note.

Provenance corrected to a chain — `auto_settle.go:180` cites `nudge_countdown.go`, `snooze.go:162` cites `auto_settle.go` — and he caught that my "163" was itself a line off; the comment sits at 162. Receipt sharpened on both ends.

The frontend ref I'd called "stale by a line or three" he didn't patch, he replaced. Out goes the loose "five sites" enumeration; in comes the structural claim: `NudgeIndicator.tsx` exports `HeaderNudgeIndicator` + `SidebarNudgeBar`, `SettlingIndicator.tsx` exports `HeaderSettlingIndicator` + `SidebarSettlingBar` — the same pair of shapes over the same `CountdownFill` for the same two surfaces. I checked all four exports against main, plus `App.tsx:3097` (`autoSettleFiresAt || autoSettleHeld || nudgeFiresAt`) and the by-name cancel list at `cancel_countdown.go:42`. All exact. That's a better receipt than the one it replaces, and it says the thing the old text was reaching for.

Docs-only, changelog fragment untouched, nothing to re-run. Re-approving on the new head.

PR #885, opened 2026-08-14
feat(hooks): open what Claude just handed the user

victorarias:
When a Claude session hands the user a file with its `SendUserFile` tool, attn now opens it: the existing PostToolUse hook forwards the payload to the daemon over the unix socket (`open_sent_files`, protocol 240), and the daemon docks each markdown file as a live-reloading tile beside the session's terminal — the same tile `attn open` uses. "Look at this" now means the file is already on screen instead of a transcript card to chase.

## Decisions

- **Which types open**: markdown (`.md`/`.markdown`) only — the markdown tile is the only tile attn has that can show a local file (the browser tile rejects `file://`). Everything else is dropped quietly, with a `daemon.log` line naming the path so the drop is visible to an operator: `open_sent_files: dropped <path> (no tile can show it)`.
- **Off switch**: `open_sent_files_enabled`, **on by default** (absent = on, only an explicit `false` disables, the `notebookDutyEnabled` pattern). Toggleable in Settings → General → "Sent files", and surfaced in the effective-settings map. The gate lives daemon-side, so it needs no hook re-install to take effect.
- **Multi-file**: no cap. `SendUserFile` calls carry a handful of paths at most (the tool renders one card per file, so agents keep it small); each path routes independently, and per-path tiles dedupe — re-sending a path re-uses its tile. A cap without a receipt would be a landmine; if a real burst ever hurts, that's the moment to measure and set a tripwire.
- **Focus**: docking does not steal terminal focus — verified live (keystrokes sent immediately after a dock landed in the sender terminal's input line).
- **Hook can never hurt the session**: the hook matches only `SendUserFile`, and a daemon that is down, gated off, or too old to know the command logs a warning to stderr and exits 0 — same contract as the existing markdown-edits forwarding beside it.
- **Already-running sessions**: reached. The hook settings baked into a session invoke the installed `attn` binary by path with a catch-all matcher, so replacing the binary is enough — no settings rewrite, no session restart. (The daemon must also be new enough to know `open_sent_files`; an old daemon answers with an error the hook swallows.)
- **Claude only**: `SendUserFile` is a Claude Code tool; nothing in `codex.go` changes, per the ticket ruling.

Plan doc: `docs/plans/2026-08-14-open-sent-files-hook.md`.

## Surfaces walked

- **CLI**: the hook path (`attn _hook-tool-use`) is the only entry; there is no user-facing command to add — the user-facing way to open a file is already `attn open`.
- **Daemon and app**: daemon owns the gate and routing; the app renders the tile it is told to (existing `openMarkdownTile` flow) and gets the settings toggle.
- **Protocol**: `main.tsp` + `make generate-types`, `ProtocolVersion` 239 → 240 in all three lockstep spots.
- **Linux**: pure Go, no darwin-only code.
- **Plugins/SDK**: not touched; the command is unix-socket-only (agent-to-own-daemon by construction, listed in `sessionCommandsAnsweredWhereTheyLand` with that reason).
- **Docs**: changelog fragment `changelog.d/open-sent-files.yaml`; plan doc records the decisions. No new glossary vocabulary.
- **Way in / way out**: the setting turns it off and back on — both directions verified live (gate off → send happens, nothing docks; re-enable → next send docks).

## Tests

- `internal/hooks/sentfiles_test.go`: payload parsing, relative-path resolution, dedupe, other tools ignored, malformed input.
- `internal/daemon/tilecontent_test.go`: markdown docks + non-markdown dropped, gate off does nothing, unresolvable session still answers OK.
- Wire-trace golden + command-classification guards updated; frontend settings-toggle tests in `SettingsModal.test.tsx`.
- Full `go test ./...` and frontend suite green.

## Live verification (profile `sfopen`, since cleaned)

Installed from this branch, bundled preflight PASS, protocol 240. A live Claude session called `SendUserFile`; the daemon docked each markdown file into the session's workspace (`daemon.log` receipts), the tile rendered beside the terminal, non-markdown was left alone, the gate was flipped off and on live, and focus stayed in the terminal.

sfopen-evidence

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/320aad2f5e1615dc298c4a6a9082e9f190933fb4/delegate-sendfile-hook-ff1e3960/sfopen-evidence.mp4)

The clip shows a prompt typed into the session asking Claude to call `SendUserFile`; when the call lands, the "Fourth sent file" markdown tile docks itself beside the terminal.

https://claude.ai/code/session_01YbjYTJhiUxPpFwMUEn3NuT


figgyster (approved):
Riding the catch-all PostToolUse hook instead of minting a `SendUserFile` matcher is the move that makes this feature nearly free. No second process spawn per tool call, and already-running sessions pick it up the moment the binary's replaced, because the hook calls the installed `attn` by path. You read one more tool name out of a payload that was already being parsed. That's the boring-technology instinct paying rent.

The honest cost is the protocol bump, and you said it out loud: 239→240 is what a reachable gate costs, since `get_settings`/`set_setting` are websocket-only and a hook-side gate couldn't read them. The gate lands daemon-side where the user can reach it. Default on, explicit-`false`-disables, same `notebookDutyEnabled` shape as its neighbors.

I didn't take the body's word for it. Ran it against the machine: `go build ./...` green, `git diff --check` no drift, the protocol lockstep test green (constants and `useDaemonSocket.ts` both at 240, `generated.go` and `generated.ts` both carrying `OpenSentFilesMessage`), the hooks `SentFiles` table green, the daemon `open_sent_files` tests green (markdown docks, non-markdown drops with a log line, gate-off answers OK and docks nothing, unresolvable session still answers OK), frontend `tsc --noEmit` green, and SettingsModal's 65 tests pass with your two toggles in the count.

Markdown-only is true because it's true — the browser tile rejects `file://` and nothing renders PDF or CSV, so you're not pretending. And "no cap, measure when it hurts" is the right refusal to set a tripwire with no receipt. If a real burst ever docks a wall of tiles, that's the moment to measure, not before.

The one receipt I'm reading instead of running is focus. It reuses `openMarkdownTile`, the same path `attn open` walks, so whatever focus behavior the manual open has, this inherits — and the GIF shows it staying in the terminal. I can't stand up a live Claude session from this box, so that one I take on the evidence.

Merge it. "Look at this" now means the file's already on screen. That's the whole job in one feature.

PR #886, opened 2026-08-14
feat(terminal): diagnose input stalls without noisy logs

victorarias:
## TL;DR

Typing stalls that are unrelated to resize currently leave no durable evidence. This adds sparse measurements for browser key-event delay and the round trip through the daemon to PTY acceptance, while keeping healthy typing in memory and writing only actionable incidents.

## Why

The existing terminal diagnostics can explain resize and paint problems, but they cannot distinguish a blocked browser event loop from a slow daemon or PTY write when typing appears to freeze. Logging every key would answer that question at an unacceptable noise and I/O cost.

## What changed

- Measure how long a real terminal key event waited before its input handler ran.
- Add an optional `probe_id` to sampled `pty_input` messages and acknowledge it after the daemon's write fence and PTY backend accept the bytes. The acknowledgement includes the daemon-side write duration, so the app can separate that work from transport and UI delay without synchronized clocks.
- Keep healthy samples in a bounded 64-entry memory ring. Only delays of at least 250 ms, failed writes, or five-second probe timeouts enter the existing capped terminal incident log.
- Limit probes to one every two seconds per active runtime and apply a 30-second per-runtime/per-reason incident cooldown. Suppressed repeats are summarized in the next record.
- Route acknowledgements for attached remote runtimes and bump the protocol to 241.

```mermaid
flowchart LR
    K[Native key event] -->|queue delay| A[App tracker]
    A -->|sampled input + probe id| D[Daemon write fence]
    D --> P[PTY backend]
    P -->|acceptance + write duration| A
    A -->|threshold crossings only| L[Capped incident log]
```

## Manual verification

Built and launched a packaged non-production app at protocol 241, then typed into a real shell using native macOS keyboard events. The text reached the PTY worker and echoed in the terminal, while healthy typing added no input-latency incident records.


figgyster (approved):
Approve.

Checked the things that would've bitten. The clock story holds: the app times the round trip on `performance.now()`, the daemon times its share on `time.Since`, both are durations on a single clock each, so "no synchronized clocks" is true and not a shrug — and Go's `time.Since` on `time.Now()` values uses the monotonic reading, so an NTP step mid-write won't poison `write_duration_us`.

Protocol lockstep is clean: 241 in both `constants.go` and `useDaemonSocket.ts`, `generated.go`/`generated.ts` regenerated, `EventPtyInputProbeResult` added to the constants *and* to the remote-relay routing case in `websocket.go`, so an attached remote runtime's result gets routed back by id the same way `pty_output` does.

The `source: 'user'` tagging is a no-op for the splice guard, and that's the correct outcome — `isUserKeystrokeSource` returns true for both `""` and `"user"`, false only for `automation`/`attach_replay`, so keystrokes counted before and still count now. The probe just gained a clean signal.

Ran what this box can run: `terminalInputLatency.test.ts` 4/4, `useGhosttyPaneRuntime.test.ts` 20/20, `tsc --noEmit` clean. No Go toolchain here, so the daemon test rides on CI — but I read it, and it's the right witness: it scopes itself with an `automation` source so settle state never enters, and asserts an unsampled write produces *zero* diagnostic traffic.

One non-blocking note inline, on the probe timeout.

app/src/utils/terminalInputLatency.ts
```diff
@@ -0,0 +1,266 @@
...
+    }
+  }
+  if (delay >= TERMINAL_INPUT_SLOW_MS) {
+    maybeRecordIncident(context, 'terminal_key_event_delay', delay, {
+      thresholdMs: TERMINAL_INPUT_SLOW_MS,
+    }, wallNow);
+  }
+}
+
+function expirePendingProbe(runtimeId: string, monotonicNow: number, wallNow: number): void {
```
figgyster, line 176: Non-blocking: the five-second probe timeout is lazy. `expirePendingProbe` is only called from `maybeStartTerminalInputProbe`, which bails out before that call for any non-`user` source — so a probe that never resolves (write hangs, or the result drops across a reconnect) only produces its `pty_input_probe_timeout` incident the next time a user key arrives for that runtime. One stalled key followed by silence leaves no record. In the real failure mode — frozen terminal, user mashes keys — it fires on the very next key, so this is almost certainly fine as designed; just flagging that the plan's "five-second timeout" reads eager where the code is tripwire-on-next-input.

figgyster (approved):
Re-approving at the new head (d15af91).

The added commit is the kind of refactor I root for. `createRendererEffectResources` gives the renderer effect one owner that disposes the recovery timer and the resize observer together, instead of two locals cleaned up in three separate places. Behavior moves intact — the double-schedule guard, the `recoveryTimer = null`-before-callback ordering, and the observer disconnect all survive the move — and `recoveryAttemptRef` stays a ref because it has to outlive the epoch bump, with the comment now saying exactly why. One owner, one dispose, nothing left behind on teardown. That's the entropy-wins move.

tsc clean, GhosttyTerminal suite 7/7, latency + pane-runtime tests green from the earlier pass.

My earlier note on the lazy probe timeout stands unaddressed — no code touched it, so I'm leaving the thread open, but it was the "almost certainly fine as designed" kind, not a blocker.

figgyster (approved):
Re-approving at the new head (fae0f4b) — rebased onto main, one new commit on top.

`forgetTerminalInputLatencyRuntime` is the right cleanup. It releases the per-runtime maps — context, probe throttle, healthy-sample throttle, both pending-probe maps, and the cooldown episodes — on pane teardown, wired through a `useEffect` cleanup keyed on `runtimeId`. Before this those maps grew with every pane that ever lived: a slow, tiny leak of the exact species that never fails a test but rots a long-running app. Now a torn-down pane leaves nothing behind, and a stale probe for a dead runtime is dropped instead of recorded as a phantom timeout.

The test is the right witness: a forgotten runtime's pending probe produces no round-trip sample, a fresh probe starts clean with a new id, and the next slow key records against the *new* pane/session. 29 green across the affected files, tsc clean, protocol still 241 in both lockstep sites, generated-types/changelog/linux-daemon all green; daemon + frontend + bench still pending.

My earlier note on the lazy five-second timeout is unchanged by this — `expirePendingProbe` still only runs on the next user key, so a lone stalled key on a *live* pane followed by silence still leaves no record. That was the "fine as designed" kind, so I'm leaving the thread open and not blocking on it.

PR #888, opened 2026-08-14
docs(plans): design A5's UI host and app SDK

victorarias:
A5 is the extension platform's UI stage: the point where an app stops being
headless and gets a face. This is its design, not its implementation — no
code ships here.

The moment the stage delivers: an agent writes a React component in an app
directory, applies it, docks it in a workspace, and watches it render live
data. Saving the file re-renders it a second later. Breaking it kills that
tile and nothing else.

## The design, in four seams

1. **Views.** An app declares named components in `attn-app.toml`. The build
   compiles each one for the browser into the same content-addressed version
   directory the handler bundle already lives in.
2. **A bundle URL.** The daemon serves those artifacts from the mux that
   already carries `/ws` and `/health`, at a content-hashed path. Immutable
   content, immutable URL.
3. **A host.** The workspace tile renderer grows one arm that
   dynamic-imports such a URL behind a per-app error boundary.
4. **One SDK, one React.** A component imports exactly one specifier,
   resolved to the frontend's own module instance.

## What the doc decides

- **Tiles are the only mount kind built**, declared as `[[views]]` with a
  `kind` field. A later panel or window moves one accepted value and one
  host component; the manifest table, registry row, build step, artifact
  layout, bundle route, error boundary and SDK all stay put.
- **The SDK is materialized locally, never published.** One TypeScript
  source in this repo with two consumers: the frontend imports it directly,
  and the binary embeds its declarations and materializes a types-only
  package on disk behind the toolchain lock. A4's ambient
  `src/attn-app.d.ts` is user state and gets retired properly.
- **One React by construction.** The SDK is the only specifier an app can
  import (`jsxImportSource` means TSX never names React), it is external at
  build time, and a static import map resolves it to a fixed-name chunk of
  the frontend's own build. A bundle built against a removed SDK export then
  fails ESM linking *by name* instead of reading `undefined` off a global.
- **The first design-system slice is five controls** over attn's inherited
  token layer — the receipt being 55 distinct button class names across 51
  stylesheets. A spinner and a relative-time label are deliberately out:
  both are repaint loops in a window that is open all day.
- **Protocol envelopes**: windowed document subscriptions over the
  WebSocket (`doc_subscribe` is IPC-only and connection-scoped today), an
  app registry in the initial-state snapshot with the first projections on
  A4's app facts, and a generic app command envelope.
- **Views receive their mount context** plus an opaque user-typed param
  string, so two tiles of one view can differ. The user places a tile; an
  app never does.
- **Reload needs no new machinery.** `attn app dev` already re-applies and
  the flip already publishes `app.version.changed`; A5 projects that fact
  and the host remounts. Zero cost while idle.
- **Tiles are not rendered for remote apps in v1**, which is safe rather
  than deferred: the dock picker lists only the local registry, so no layout
  can name one and nothing renders blank.

## Delivery

Six slices with their verification tiers, into `epic/ext-a5` — the stage
changes existing app behavior (the tile renderer and the frontend build),
which is the question that decides epic versus direct-to-main.

| # | Slice | Verification |
| --- | --- | --- |
| 1 | Views are declared and built | CLI tier |
| 2 | The SDK becomes a package | CLI tier + frontend typecheck |
| 3 | The host mounts | Full app |
| 4 | Live queries on the wire | Full app |
| 5 | Acting, looking native, being teachable | Full app |
| 6 | Exit proof | Full app + recording |

## Open questions

Two, neither blocking slice 1: whether a raw string is the right shape for a
view's param (a text field can't offer the user a *choice*, because attn
does not know what the string means), and whether the SDK should own its
React type aliases rather than re-export React's, which is the one place a
React major can fail an app's next `attn app apply`.

## Known unknown

Whether WKWebView permits a cross-origin module import from the app window
to the daemon's port is unproven — it is the first thing slice 3 verifies
live, and the doc names the `blob:` URL fallback that changes nothing above
the host if it does not.

### DIFF
--- a/changelog.d/ext-a5-ui-host-plan.yaml
+++ b/changelog.d/ext-a5-ui-host-plan.yaml
@@ -1,0 +1,13 @@
+kind: internal
+area: docs
+change: >
+  A5's plan doc designs the extension platform's UI stage: apps declare
+  views in their manifest, the build compiles each for the browser into the
+  content-addressed version, the daemon serves them at an immutable URL, and
+  a tile host mounts one behind a per-app error boundary. It records
+  Victor's gate answers — tiles as the only mount kind with a seam that
+  keeps a second kind additive, an SDK materialized locally from one source
+  rather than published to npm, a single shared React reached through an
+  import map, the first design-system slice, and the protocol envelopes for
+  windowed document subscriptions and app commands — and breaks the stage
+  into six slices with their verification tiers.
--- a/docs/plans/2026-08-13-ext-a5-ui-host-and-app-sdk.md
+++ b/docs/plans/2026-08-13-ext-a5-ui-host-and-app-sdk.md
@@ -1,0 +1,891 @@
+# A5 — UI host and the app SDK
+
+Stage A5 of the extension-platform roadmap
+([2026-08-01-extension-platform-roadmap.md](2026-08-01-extension-platform-roadmap.md)),
+designed 2026-08-13 with Victor. It builds on A4's registry
+([2026-08-06-ext-a4-app-registry-and-runtime.md](2026-08-06-ext-a4-app-registry-and-runtime.md))
+and A3.4's delivery shape
+([2026-08-05-ext-a3.4-doc-store-positions-and-windows.md](2026-08-05-ext-a3.4-doc-store-positions-and-windows.md)).
+
+The moment it delivers: an agent writes a React component in an app
+directory, applies it, docks it in a workspace, and watches it render live
+data. Saving the file re-renders it a second later. Breaking it kills that
+tile and nothing else.
+
+## The shape
+
+A4 made an app a real thing headlessly — a manifest, a content-addressed
+version, a bus consumer, a document namespace, a supervised runtime. A5
+gives an app a face. Four seams, end to end:
+
+```mermaid
+flowchart LR
+  tsx["src/views/Approvals.tsx<br/>written by an agent"]
+  art["versions/hash/views/approvals.js<br/>content-addressed artifact"]
+  url["GET /apps/bundle/…/approvals.js<br/>immutable URL"]
+  tile["a tile in a workspace"]
+  sdk["the frontend's own SDK chunk<br/>React lives here"]
+
+  tsx -->|"1. apply builds it"| art
+  art -->|"2. the daemon serves it"| url
+  url -->|"3. the host imports it"| tile
+  sdk -->|"4. the import map resolves it"| tile
+```
+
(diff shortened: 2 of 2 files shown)


figgyster (approved):
A design doc with this many receipts is a bet I can check, so I checked. They hold. The token ledger is exact: 128 custom properties on `:root`, split 80 `--color-*` / 12 `--font-*` / 32 `--syntax-*` plus five strays — I counted the file, that is the number. Fifty-five distinct button class names is exact too, and it is the count that does the arguing: that is the receipt for `Button` being component #1. Every structural ref resolves — `initHTTPServer` serving `/ws` and `/health`, `handleDocSubscribe` as the one connection-keeping handler, `GenerateSDK` as a types-only string builder, `WorkspaceDockTile` markdown/browser/notebook switch, `appRuntimeConnectWait` at 10s, `refuseUnknownKeys`, `TestEveryProjectedFactReachesTheWire`, and the "declares no subscriptions, so nothing would ever run it" refusal.

The design is the good kind of clever. "One React by construction of the build graph" — the import map resolving `@victorarias/attn-app` to a fixed-name chunk of the frontend's own build — is the whole argument in one move, and it buys the loud ESM-link-by-name failure for free. "Safe rather than deferred" on remote apps is airtight: the dock picker lists only the local registry, so no layout can name a remote view and nothing renders blank. Spinner and RelativeTime left out on purpose is the right paranoia — a repaint loop in a window open all day is a battery bug wearing a loading icon.

One marker worth adding, non-blocking: "the daemon takes the namespace as given" is true today because the *host* supplies the namespace from the mount and the view never chooses it. That containment is the load-bearing part, and it is the one assumption that stops being true when the hub relays WS or the web client grows tiles — a trust-model change, not just "relay a subscription." The doc already flags remote as a future hub change; one sentence naming this assumption as the thing that changes would keep it from being inherited as a property.

Nit: the protocol drill says `PROTOCOL_VERSION` "238 as of this writing" — the tree is at 241 now, and the doc's own "the tree is authoritative" hedge covers it. Leave as-is.

Approving.

figgyster (approved):
Both notes answered in one commit, and both exactly right.

The namespace paragraph lands the distinction I was groping for sharper than I put it: containment is "an assumption about *who holds the socket*, not a property of the protocol" — and the last line, that "whoever lifts the remote restriction inherits that question, and it is theirs to answer, not A5's to have answered," is the precise shape of the marker. That's the load-bearing sentence, and it's now in the doc.

The protocol drill went 238 → 241, which matches the tree today.

Re-approving at `92215f3`. Docs only, no test re-run owed.

PR #889, opened 2026-08-14
feat(apps): declare views in the manifest and build them for the browser

victorarias:
Slice 1 of A5 ([the plan](https://github.com/victorarias/attn/blob/main/docs/plans/2026-08-13-ext-a5-ui-host-and-app-sdk.md)): an app declares views and the build compiles them. Nothing renders one yet — the bundle route, the tile host and the SDK are slices 2 and 3.

## What an author writes

```toml
[[views]]
name = "approvals"                     # [a-z0-9-], unique within the app
kind = "tile"                          # optional, resolves to "tile"
title = "Pending approvals"
entrypoint = "src/views/approvals.tsx"
params = { label = "Repository", placeholder = "victorarias/attn" }
```

`attn app apply` then bundles each view for the browser into the version's content-addressed directory:

```
apps/viewboard/6cb24f1545ae…/
├── bundle.js            278 bytes   the handlers, target bun
└── views/approvals.js   234 bytes   one per declared view, target browser
```

## The four things this changes

**A view is a declaration, a tile is a place.** `kind` is a manifest field with one accepted value; a second mount surface adds a value and a host component, and nothing between here and the artifact moves. `kind = "panel"` is refused by name today rather than installed as a view nothing renders. The frozen declaration records the *resolved* kind, so a default that changes in a later api version cannot rewrite what an old version meant.

**The version hash covers every artifact.** A view is built from its own entrypoint, so editing only a component leaves both the declaration and the handler bundle untouched — hashing those two alone would reuse a version row whose artifacts had moved under it. Views are hashed by name as well as content, in name order, so the digest does not depend on declaration order. An app with no views hashes exactly as before, which is why re-applying an app built by an older attn lands on the row it already had (there is a test pinning that against the pre-views digest).

**The daemon checks all of them.** `handleAppApply` re-derives the hash from what is on disk; it now reads the view names back out of the declaration — the only description of a version it holds — and reads every artifact. A declared view with no artifact is a refusal naming the view and the path, not a silently shorter list.

**A view counts as something that runs.** A4 refused an app with no `[[subscribe]]`; it now refuses an app with neither a subscription nor a view. A tile that only reads the document store is a whole app. This lands on a path A4 already built and could not reach: an app with no patterns gets the `app.subscribes.to.nothing` filter, which `attn app status` used to print verbatim — a fact name no reader could look up. The CLI now says `nothing — it declares no subscriptions`, and the constant moved to `internal/apps` because the daemon derives it and the CLI recognises it.

## No cap, but the numbers

There is deliberately no bundle size cap — nothing measured yet would justify one, and a number picked from nothing fires on the first legitimately large view. What apply owes an author instead is the number, per artifact, which is what the byte lines above are.

## Verification — CLI tier

Unit tests (`internal/appbuild`, `internal/apps`, `internal/daemon`) plus the built binary against real app directories, with a daemon on an isolated `ATTN_DATA_DIR` and its own WS port. Full `go test ./internal/... ./cmd/...` is green.

| What | Result |
| --- | --- |
| view-less app applies | version 1, `278 bytes` — unchanged behavior |
| two views declared | `bundling view approvals from src/views/approvals.tsx`, artifacts at `views/approvals.js` |
| local import inside a view | `src/lib/format.ts` inlined into the one artifact |
| edit only the view source | version 2 → 3, `bundle.js` byte-identical at 278 |
| `attn app dev` | view-only save → new version in 190ms |
| `kind = "panel"` | `view "approvals" is of kind "panel", which this attn cannot mount anywhere; it mounts "tile"` |
| view entrypoint missing | `view "approvals"'s entrypoint … does not exist (looked for …)` |
| view import that does not resolve | bun's own error with file and line; nothing recorded |
| all view, no subscriptions | applies; `subscribes: nothing — it declares no subscriptions` |
| neither | `declares neither a subscription nor a view, so nothing would ever run it` |

No recording: nothing about this slice is visible in the app.

## For slice 3 (the host)

- `appbuild.ViewArtifactPath(storeDir, app, hash, view)` is where a view's bytes are, derived rather than stored — the bundle route should use it rather than a path of its own.
- `appbuild.DeclaredViewNames(declaration)` reads a version's views back out of its frozen declaration, which is what the daemon holds. It validates each name, so a declaration cannot smuggle a path segment into a file path.
- View names are `apps.ValidateViewName` — the app-name charset minus the reserved set, since a view is scoped to its app. It is the same rule that will parse the `app:<app>/<view>` tile kind.
- **Views are not on the wire yet.** `attn app status` shows no `views:` line and the `apps` snapshot does not exist: both need a protocol field, and slice 3 owns that bump with the registry snapshot. The way to see a view today is `attn app apply`'s per-artifact byte lines and the version directory.
- **Views are bundled but not typechecked.** The handler entrypoint is typechecked against the generated `Handlers` type; a view's `.tsx` is only bundled, so a type error in a view is not caught at apply. That waits on slice 2 — the SDK package and `jsxImportSource` are what a view typechecks against, and there is nothing to check it against until they exist.

## Scope

No protocol bump, no migration, no frontend. The fixtures import nothing from the SDK, per the slice brief — the one test that names `@victorarias/attn-app` asserts the specifier is left *unresolved* in the artifact, which is what proves the externals are in place for slice 2 rather than a dependency on it.

### DIFF
--- a/changelog.d/a5-slice1-views-declared-and-built.yaml
+++ b/changelog.d/a5-slice1-views-declared-and-built.yaml
@@ -1,0 +1,13 @@
+kind: internal
+area: apps
+change: >
+  An app can declare [[views]] in attn-app.toml — a named component with a kind
+  (only "tile" is accepted), a title, an entrypoint and an optional one-line
+  param declaration. `attn app apply` and `attn app dev` bundle each view for
+  the browser into the version's content-addressed directory beside the handler
+  bundle, and the version hash now covers every artifact, so editing only a view
+  mints a new version. An app that declares a view and no subscriptions is
+  accepted: a view is something that runs.
+notes: >
+  A5 slice 1. Nothing renders a view yet — the tile host, the bundle route and
+  the SDK are later slices — so there is no user-visible behavior to report.
--- a/cmd/attn/app.go
+++ b/cmd/attn/app.go
@@ -271,8 +271,14 @@ func runAppStatus(args []string) {
 			state = "enabled"
 		}
 		filter := app.Consumer.Filter
-		if filter == "" {
+		switch filter {
+		case "":
 			filter = "everything"
+		case apps.NoSubscriptionsPattern:
+			// An app can declare a view and no subscriptions, and its filter is
+			// then a fact name nothing publishes. Printing that name sends the
+			// reader looking for an event that does not exist.
+			filter = "nothing — it declares no subscriptions"
 		}
 		fmt.Printf("  consumer:   %s — %s, cursor %d, %d event(s) behind\n",
 			app.Consumer.Name, state, app.Consumer.Cursor, app.Consumer.Lag)
--- a/cmd/attn/app_build.go
+++ b/cmd/attn/app_build.go
@@ -116,7 +116,7 @@ func printApplied(result *protocol.AppApplyResult, res appbuild.Result) {
 	var state string
 	switch {
 	case result.VersionCreated:
-		state = fmt.Sprintf("new, %d bytes", res.BundleBytes)
+		state = fmt.Sprintf("new, %d bytes", totalArtifactBytes(res))
 	case moved:
 		state = "this content already had a version; nothing new was recorded"
 	default:
@@ -128,6 +128,25 @@ func printApplied(result *protocol.AppApplyResult, res appbuild.Result) {
 		fmt.Printf("  was on version %d\n", *result.PreviousVersionID)
 	}
 	fmt.Printf("  artifact %s\n", result.ArtifactPath)
+	// A version made of several artifacts gets each one's size named. There is no
+	// bundle size cap — nothing measured yet would justify a number — so making
+	// the numbers visible is what apply owes an author instead.
+	if len(res.ViewBytes) > 0 {
+		fmt.Printf("    %s  %d bytes\n", appbuild.ArtifactName, res.BundleBytes)
+		for _, v := range res.ViewBytes {
+			fmt.Printf("    views/%s.js  %d bytes\n", v.Name, v.Bytes)
+		}
+	}
+}
+
+// totalArtifactBytes is everything a version holds: the handler bundle and one
+// module per view.
+func totalArtifactBytes(res appbuild.Result) int64 {
+	total := res.BundleBytes
+	for _, v := range res.ViewBytes {
+		total += v.Bytes
+	}
+	return total
 }
 
 func runAppRollback(args []string) {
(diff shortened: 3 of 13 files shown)


figgyster (approved):
Approving.

Slice 1 of A5 builds the pipeline without rendering anything — the right first cut. A view is a declaration, a tile is a place; `kind` has one accepted value and the frozen declaration records the resolved kind, so a default that changes later can't rewrite what an old version meant. The version hash now covers every artifact, views hashed by name in name order, and the daemon re-derives the hash and reads every declared view back off disk — a declared view with no bytes is a refusal naming the view and the path, not a silently shorter list.

Receipts I checked by hand:

- `nameRe` is `^[a-z0-9][a-z0-9-]*$`, and `DeclaredViewNames` runs each view name through `ValidateViewName` before it becomes a path segment. That's the load-bearing one: the declaration arrives over the wire and `validateDeclaration` only checks the `name` field, so this is what keeps a hostile declaration from smuggling `../` into `ViewArtifactPath`. `TestDeclaredViewNames` pins `../../etc/passwd` refused.
- Viewless apps hash exactly as before (the loop writes nothing), pinned by `TestBuild_ViewlessAppHashesAsItDidBeforeViews` — apps built by an older attn re-apply onto their existing row, no migration.
- `checkSomethingRuns` is the right relaxation of A4, and moving `NoSubscriptionsPattern` into `apps` so the CLI and daemon share one string is the same no-drift instinct as `RuntimeHostBinaryName`.

Ran `go test ./internal/appbuild ./internal/apps` green and the new daemon apply tests green; CI's Daemon job (CGO + libghostty), Bench A/B, Changelog, and Changes are all pass.

One non-blocking note: "the digest does not depend on declaration order" has a `sort.Slice` standing behind it but no witness on the stand — no test reorders two views and asserts an identical hash. True today, and exactly the kind of thing a later refactor drops by accident while the claim quietly dies. A three-line test pins it. Not blocking.

Signal: the extension platform keeps paying its debts at receipt-level precision. Slice 1 sets the hash contract, the trust boundary, and the "no cap, name the number" discipline before any tile exists to render — #888's design doc turning load-bearing one claim at a time.

figgyster (approved):
Took the note, and the witness is real — I deleted the sort to check it red myself. `TestVersionHash_DoesNotDependOnViewOrder` fails on the reorder assert (`reordering the views moved the hash`), not some downstream one, and goes green with the sort back in. Both halves pinned: reordering two views hashes identically, and swapping their contents does not — that second assert is what keeps the "set of named artifacts" rule from quietly collapsing into "some bytes in some order."

Re-approving at efdc3b9. Nothing else changed.

PR #890, opened 2026-08-14
feat(crew): give a session a name it keeps for the day

victorarias:
Slice 1 of [the crew primitive](https://github.com/victorarias/attn/blob/crew-plan/docs/plans/2026-08-11-the-crew-primitive.md): the registry and the binding.

## What a crew member is

A durable named identity — a charter, a handoff line, an address — whose sessions are its days. keel, alder and trellis have existed since 2026-08-06 as a hand-run skill simulation: three homes of plain markdown under `~/.attn/crew/`, 18 filed handoffs, and a `/wake` skill that reads them. attn itself knew none of it. `attn agent list` could not say "this session is trellis today", nothing stopped two trellises from running at once, and `agent msg trellis` had nothing to resolve against.

This slice puts the roster in the daemon without taking the files away from the user.

## Files stay canonical

The daemon scans `~/.attn/crew/` at startup and records what it finds in a docstore collection (`core/crew`): member id, charter path, home dir, cwd, awareness dirs, active binding. It records **where a home lives, never what it says** — no prose is copied, and a test asserts a member's stored record cannot contain its charter's text. That is what keeps a home hand-editable and lets any agent be a member, claude or codex or a future host. A home the user creates by hand joins the roster at the next daemon start; the import is create-only, so re-running it every startup never touches a record that already exists.

## Identity is the invocation, never the files

Reading a charter confers nothing. `attn <agent> --member <name>` claims a binding **before** the session registers, and that binding is the identity:

- `attn agent list` grows a MEMBER column; `attn agent peek` says "this session is trellis today".
- One member has one active binding. A claim on a member whose day is already live is refused, naming the session that holds it — and the refusal is launch-fatal, because a member launch that quietly ran unbound would be an identity dropped on the floor. Parallelism means another member, never a second copy.
- Whether a stored binding still binds is judged **at read** against live sessions — the same liveness rule a seed's tender follows — so a day that ended without ceremony releases itself. Session teardown clears the record too, on both removal paths.

`attn crew list` is the roster's read surface: every member, awake or asleep, with the short session id `agent peek` takes.

## The garden touches it lightly

Where a tender's free-string name matches a registered member it now resolves to that member's id, so a claim compares addresses rather than spellings (`--member Trellis` and `--member trellis` are the same person). A bound session that names nobody is its own member. **An unregistered name passes through untouched** — the registry never becomes a requirement to tend, and workers and errand sessions are unaffected.

## No behavior change for unbound sessions

The slice's safety property, stated because it is the thing most worth checking, and tested directly (`TestCrew_UnboundSessionsBehaveExactlyAsBefore`): a session launched without `--member` registers, plants, tends, refuses a contended claim and broadcasts exactly as before, with no crew field on the wire at all. Verified live as well as in tests.

## Surfaces walked

- **CLI** — `attn crew list`, the `--member` launch flag, the MEMBER column, the peek line.
- **Daemon** — registry, binding, single-holder, release on both session-removal paths, broadcast decoration.
- **Protocol** — `crew_member` on `Session` and `AgentPeekResult`, `member` on `RegisterMessage`, the `crew_list` command and its models; `ProtocolVersion` 241 → 242 in all three lockstep spots. Generated files regenerated, never hand-edited.
- **Fence** — every crew read and write passes `requireHome`; the crew has one owner, and an outpost gets the named refusal (tested).
- **Linux** — `cmd/attn` and `internal/**` cross-compile clean for linux/arm64; nothing here is Darwin-specific.
- **Docs** — a glossary section for the crew, the architecture map in AGENTS.md, a changelog fragment.
- **App** — none needed. `agent list`/`peek` are CLI-only surfaces (nothing under `app/src` consumes them), so this slice ships CLI + daemon; the sidebar crew surface is slice 2. The app's only obligation was the protocol bump, which is why there is no evidence recording: there is nothing new to see in the window.

## Live verification

Throwaway profile `crewbind`, full `make install PROFILE=crewbind`, bundled preflight PASS (0 failed, 0 warnings; `protocol.cli_app` and `protocol.app_daemon` both 242). Crew homes seeded by copying the real directory's *shape* — three members, charters, dated handoff files — never the live directory. Against the running daemon:

| check | result |
| --- | --- |
| import homes from disk | three homes became the roster |
| launch bound as trellis | `crew list` awake; `agent list` MEMBER=trellis; peek "this session is trellis today" |
| second trellis | refused by name, exit 1, agent never started |
| wake keel alongside | both awake at once |
| unbound session | no crew line in peek, `-` in list, planted and tended normally |
| tender resolution | `--member Trellis` stored as `trellis` |
| sessions end | all three members released, roster back to asleep |

Profile cleaned afterwards.

## Receipts

- The roster read rides every session broadcast, so it was measured rather than assumed: **25µs at a three-member roster** (M5). The crew is people the user named one by one, so that is the real size; the number is recorded beside the code.
- `MaxIDChars = 40` is a tripwire — the longest real member name is 7 characters (`trellis`).

## Two things worth a reviewer's attention

1. **The glossary links to the plan doc, which is not on main yet** — it lands via its own PR from `crew-plan`. The link goes live when that merges; it is dead until then. Flagging rather than dropping the link, since the glossary's convention is to name the plan.
2. **`--member` is refused on daemon-managed launches** (the app's spawn path) with a named error, because that path cannot carry a binding until the wake slice builds it. Refusing loudly beats launching an agent that silently is not who it was asked to be.

## Out of scope (later slices)

The wake app surface and CLI verb, priming injection, handoff filing, the reload nap, the activity subsystem and auto-sleep, `agent msg <member>` addressing, chief migration, and write-refusal on member homes. The `/wake` and `/handoff` skills keep running beside this unchanged.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -408,6 +408,11 @@ forgets that position and `--since <RFC3339>` replays from an instant.
   live here. `Status.RequireHome` is the fence every garden/crew surface calls;
   reach it from the daemon through `Daemon.requireHome`, never by reading the
   record yourself
+- `internal/crew`: what a crew member IS — the id rule, the stored registry
+  record, and how a home directory under `~/.attn/crew/` becomes one. Files stay
+  canonical: the registry records where a home lives, never what it says. The
+  daemon half (`internal/daemon/crew.go`) owns the binding a session launches
+  with and the one-active-binding-per-member rule over it
 - `internal/bus`: durable event bus (domain facts, per-consumer cursors)
 - `internal/docstore`: document-store query semantics, SQL compilation, and the
   physical naming (no DB handle; `internal/store/documents.go` executes what it
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -240,7 +240,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '241';
+export const PROTOCOL_VERSION = '242';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 // Identifies this app process to the daemon across its own reconnects, so a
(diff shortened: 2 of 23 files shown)


figgyster (approved):
Read the whole diff twice and ran what I could run here: the protocol lockstep test, the `internal/crew` and `internal/client` suites, full `internal/protocol`. The daemon/cmd tests wouldn't build in my sandbox — missing `ghostty/vt.h` — so I took CI's word on those and read the test files instead. CI's green across the board: Daemon, Frontend, Generated types, Linux daemon on both arches.

The shape is right. Files stay canonical — the registry records where a home lives, never what it says, and there's a test that a stored record can't carry the charter's prose. Identity is the invocation: the binding is stamped at launch, the claim rides optimistic concurrency (write against the revision you read, three retries), and the single-holder refusal names the session holding the day. Liveness is judged at read, not stored — a day that ends without ceremony releases itself. Both removal paths clear the record, and the safety property — an unbound session behaves exactly as before, no crew field on the wire — has its own test. That test is the one that matters, and it's there.

Both things you flagged hold up. Refusing loudly on managed launches beats an agent that silently isn't who it was asked to be; the dead glossary link is a convention worth keeping.

Two non-blocking notes, the "worth a receipt, not a rewrite" kind:

1. `resolveTenderMember` reads the whole roster on every seed plant/tend/harvest/note — a docstore query per garden action, including the unbound majority your safety property promises is unchanged. The 25µs receipt covers the broadcast read, not this one. Cheap at three members, but it's a new per-action cost on a path you didn't measure.

2. The single-holder rule is one-directional. `claimCrewBinding` only touches the target member's record. Re-register a session id already bound as `keel` with a different `--member trellis`, and trellis gets bound while keel stays bound to the same session — one session transiently holding two identities. The no-member path releases (`releaseCrewBindingIfSession` in the `else`), but the different-member path doesn't. Contrived — it needs `ATTN_SESSION_ID` pointing at an already-bound session — but a release of any prior binding for that session id before claiming would close it.

Neither blocks. Approving.

figgyster (approved):
Both notes taken, and taken correctly.

The one-session-one-name gap is closed exactly where it should be: `claimCrewBinding` releases any other name the session answered to, but only after every refusal has passed — so a refused claim writes nothing and the session stays who it was. The `releaseCrewBindingsExcept` extraction reads clean, and it de-duplicates the loop `releaseCrewBindingIfSession` used to own. The two tests are the right witnesses: keel→trellis moves the identity (the session decorates as exactly one member), and a refused second name leaves the first intact.

The receipt relocation is the honest fix for the other note — the 25µs was always covering `readCrewMembers`, the one query every crew read funnels through, so moving it there makes it cover garden tender resolution as well as broadcast. The early-out for the neither-name-nor-session case is the right free skip; the session-with-id case still reads, which it has to, since that's the only way to know whether that session is bound.

gofmt and the pure-Go suites are clean on my end; CI's green on arm64 daemon and still churning the rest. Re-approving at 8d1731d0.

PR #891, opened 2026-08-14
feat(apps): make the app SDK a package apps can import

victorarias:
Slice 2 of the A5 plan ([docs/plans/2026-08-13-ext-a5-ui-host-and-app-sdk.md](docs/plans/2026-08-13-ext-a5-ui-host-and-app-sdk.md)): **the SDK becomes a package.** Into `epic/ext-a5`.

## Why

A4 generated `src/attn-app.d.ts` into every app — an ambient declaration file, invisible to a bundler, describing the handler contract only. A view needs React: hooks, props, a JSX runtime. An ambient file cannot carry a module graph, so nothing an app writes can `import` from attn. Slice 3 hosts views in the frontend, and a view's components must run on **attn's** React instance — two React copies share no hook dispatcher, and the second one's `useState` throws on first render with an error that reads as a bug in the app's code.

## What this builds

**One SDK source, `sdk/attn-app/`.** A pnpm workspace member of `app/` (`../sdk/attn-app` in `app/pnpm-workspace.yaml`, `workspace:*` in `app/package.json`). The frontend imports `@victorarias/attn-app` and resolves it to that directory — no vite alias, no tsconfig `paths`, no second lockfile. pnpm links both packages to one `react`, so one React instance is a property of the build graph rather than of a build flag someone has to remember. `app/src/appSdk.oneReact.test.ts` asserts it directly (`sdk.useState === React.useState`, `sdkJsx.jsx === reactJsx.jsx`).

**Declarations generated and committed.** `tsc --emitDeclarationOnly` writes `internal/appbuild/sdkdist/` (three files: `index.d.ts`, `jsx-runtime.d.ts`, `jsx-dev-runtime.d.ts`), which `//go:embed` picks up — go:embed reads only downward from its own directory, which is why the emit target sits inside the Go tree. `make generate-sdk` regenerates; `make check-sdk` fails on drift and runs in CI, the same shape as `generate-types`/`check-types`.

**Apply materializes and links it.** `EnsureSDK` writes a content-addressed copy (sha256 over names+contents) under `<data-dir>/apps/sdk/<hash>/`, staged and renamed, under the toolchain `flock` A4 already had. Then `node_modules/@victorarias/attn-app` in the app points at it. React's own declarations reach the app through one relative symlink from the materialized package to the toolchain's `node_modules` — tsc resolves a symlink to realpath before resolving further, so `@types/react` and its `csstype` dependency both land. Nothing is installed from a registry at apply time; `@types/react` is pinned beside `typescript` in the existing once-per-machine toolchain install.

**`react` stops being a specifier an app can write.** The typecheck now runs with `--jsx react-jsx --jsxImportSource @victorarias/attn-app`, so an app's TSX compiles against the SDK's runtime. `TestBuild_ImportingReactDirectlyIsACompilerError` pins that a direct `import "react"` is `Cannot find module 'react'`.

**A4's ambient file is retired loudly.** `src/attn-app.d.ts` is user state — apps on real machines carry it. Apply deletes it only when the content starts with attn's own generated header, and otherwise keeps it and reports the path and what to do. Both branches are tested and were exercised live.

## Verification

CLI tier plus the frontend typecheck, per the slice's plan entry. This change has no app-observable surface: no protocol change, no daemon lifecycle, no UI, no persisted state, and nothing the frontend renders differently.

Live run against a throwaway `ATTN_DATA_DIR`-scoped daemon, built binary:

- `attn app new` scaffolds an app whose tsconfig carries `jsx`/`jsxImportSource`, links the SDK into `node_modules`, and writes **no** `src/attn-app.d.ts`.
- A `.tsx` entrypoint using `useState`, `ReactNode`, `ViewProps` and JSX typechecks with the app's own tsconfig — the app directory has no `package.json` and no lockfile.
- `attn app apply` on an app carrying an attn-generated `src/attn-app.d.ts` removes it with one line saying so, then applies (version 1, 278 bytes). A hand-written file under that name is kept, with a warning. Re-apply is byte-identical and quiet.

Automated:

- `go test ./...` — green. All 10 new/changed tests confirmed running, not skipped. The two drift guards (`TestSDKDeclarationsMatchTheDaemonsContract`, `TestReactTypesPinMatchesTheFrontend`) were mutation-checked and confirmed to fail when the constant they pin moves.
- `pnpm --dir app test` — 253 files, 2813 passed.
- `pnpm --dir app exec tsc` — exit 0.
- `make check-sdk` — clean; verified it fails when `sdkdist` is stale.
- `GOOS=linux go build ./cmd/... ./internal/...` — exit 0.

## Guards that replace deleted code

`GenerateSDK()` hand-generated the handler contract, and derived the `FilterOp` union from `docstore.FilterOps()` so the two could not drift. Deleting the generator deletes that guarantee, so it is now asserted: `TestSDKDeclarationsMatchTheDaemonsContract` reads the emitted `index.d.ts` and pins `APP_API_VERSION` against `appbuild.APIVersion` and the `FilterOp` union against `docstore.FilterOps()`.

## Notes

**React types across a React major.** The plan's open question 2 is settled here by the doc's own proposal: the SDK re-exports React's types, and a React major bump is a loud `attn app apply` break rather than a silent one. No alias layer. The doc records this, and records that we revisit it the first time it fires.

**A third specifier.** `@victorarias/attn-app/jsx-dev-runtime` exists because bun reaches for the dev runtime without `--production`. Slice 1 should build views in production mode; slice 3's import map should carry the entry anyway, so a dev-mode bundle that slips through resolves instead of failing at load.

**The SDK package is types-only on disk.** What apply materializes carries declarations and a `package.json`, no JavaScript. A view bundle must mark the specifier external and let the host provide it — that is slice 3's job.

**Scope.** No protocol bump, no migration; neither was needed. Nothing here touches the view build path, so it does not depend on slice 1 and does not collide with it — whoever lands second rebases onto the epic head.

https://claude.ai/code/session_01AEQ11g8g7fsUjfDaeuCWKJ


figgyster (dismissed):
Approving. This is the slice where the SDK stops being a rumor with a commit hash — A4's ambient d.ts could never carry a module graph, and you've replaced it with a package the frontend links and the binary embeds, with "one React" now a property of the build graph instead of a flag somebody has to remember. The receipts habit shows up as structure: a Go test reads the emitted `index.d.ts` back and pins `APP_API_VERSION` and `FilterOp` against the daemon, another pins `@types/react` against the frontend's own lockfile, and a vitest asserts `sdk.useState === React.useState` by identity. Three witnesses where the old code had one hand-generated string.

One non-blocking note — the seam that'll rot first. The SDK has two manifests for one package: `sdk/attn-app/package.json` (what the workspace links, `types` → `./src/index.ts`) and the `sdkPackageJSON` string constant in `internal/appbuild/sdk.go` (what apply materializes, `types` → `./index.d.ts`, `private`). Their `exports` maps stay aligned by hand. Add a subpath later — a components entry when slice 5 lands, or whatever the first React-major break forces — and the frontend resolves it while the materialized package won't, so an app's typecheck fails at apply time with "no export" after the workspace-level check already passed. Nothing asserts the two maps match. A one-line test when the exports surface next grows; not worth holding the merge.

CI's green all the way down. I read the whole diff and `git diff --check` is clean; my sandbox here lacks `bun` and the ghostty native lib, so the SDK materialization tests skip locally and the pty/daemon packages don't link — CI covers both.

figgyster (approved):
The one-line test I asked for, delivered. `TestBothManifestsExportTheSameSpecifiers` parses both manifests' `exports` maps, asserts the specifier sets match, then walks each specifier and fails if the binary doesn't carry the `.d.ts` it names. That's the seam I flagged, pinned at the right level — keys, not values, since `. → ./src/index.ts` versus `. → ./index.d.ts` are supposed to differ. I ran it alongside the two drift guards locally; all three pass. Test-only commit, re-approving at the new head.

PR #892, opened 2026-08-14
feat(garden): plant work in plots and send a delegate to one

victorarias:
Garden slice 5: work is planted as a **plot** — a crown with children — and a
delegation can be **dispatched at** one. Plan:
[docs/plans/2026-08-06-the-garden-vertical-slices.md](docs/plans/2026-08-06-the-garden-vertical-slices.md),
section "Slice 5 — plots and dispatch".

## What ships

- **Plant into a plot.** `attn seed plant … --part-of <crown>` puts one seed
  under a crown; `attn seed plot -f payload.json` plants a crown and its
  children in one command. Children are parallel by default — only a `blocks`
  entry sequences them. Child ids are minted before anything is written, so a
  sibling's `blocks` names a real id and a refused payload writes nothing.
- **Dispatch at a plot.** `attn delegate --plot <crown>` records what the
  delegate was sent to. That is scope inference and nothing else: inside that
  session a flag-free `attn seed ready` answers with the plot and its priming
  starts from the crown, while `--all` still reaches the whole garden, the
  delegate may plant or tend anywhere, and who-holds-what stays the per-seed
  tender. The record is written before the spawn, because the launch primer
  reads it.
- **The garden is navigable.** The panel opens on the whole garden and walks
  root to leaf: a crown row carries its plot's counts (done / growing / ready /
  blocked), opening it scopes to its children, the trail climbs back, and a
  relation to a crown crosses into that plot. Navigation asks the daemon
  nothing — the push already carries the garden. What the app says about who
  works on what is the per-seed tender; there is no delegation-to-crown badge.
- **`attn seed show <crown>`** carries plot progress, and **`attn seed ls
  --stale`** names open seeds whose trail has not moved. It is a query, never a
  reaper: the output names the rule and the window it applied, and `--window`
  moves it.
- **One garden.** `workspace_id` is gone from the seed — field dropped, stored
  values erased, no compatibility read and no shim. `--workspace` retires from
  plant/ls/ready, flag-free `ready`/`ls` answer for the whole garden, and launch
  priming speaks of the garden rather than a workspace's count.

## The stale window's receipt

`DefaultStaleWindow = 7d`. Seeds have no history to measure yet, so the window
was measured against the nearest real trail: activity on production tickets.
276 gaps between consecutive activity on the same ticket — p50 0.3h, p90 1.3h,
p95 9.7h, p99 44.7h, max 356h. 7d sits ~3.7× past p99: a tripwire only a seed
nobody is working touches. The receipt is in the constant's comment, so
remeasuring is a diff to that line.

## Erasing `workspace_id` is safe to do destructively

The seed collection's indexed column goes away with the field (the docstore
drops removed fields' columns when a collection's schema changes), and a body
written before this carries a key nothing decodes. No production install has
ever held seed data, which is what makes drop-and-forget the honest option
rather than a migration nobody would ever run.

## Verification

Live, from this branch, on a throwaway `plotwalk` profile (preflight PASS, "CLI
and installed app use protocol 243"). The acceptance is a packaged-app scenario
that runs on any checkout:

```
pnpm --dir app run real-app:scenario-garden-plot-dispatch
```

Ten steps, all green: a plot planted in one command, its crown wearing the
counts, a delegate dispatched at it, that delegate's flag-free `ready`
answering with the plot, `--all` stepping back out to the garden, the panel
drilling in and climbing back, the plot draining live as children are tended and
harvested in dependency order, two delegates claiming parallel children while
the second claim on one seed is refused by name, and a fresh session
re-orienting from `ready` and `show` alone.

garden-plot-dispatch

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/2ddd381097460a1a0037adc209510a5b33d4d0de/feat-garden-slice-5/garden-plot-dispatch.mp4)

Also: `make test-all` green; protocol bumped to 243 in all three lockstep spots
(`constants.go`, `useDaemonSocket.ts`, regenerated `generated.go`/`generated.ts`).

## Surfaces walked

CLI (`plant --part-of`, `plot`, `ls --stale`, `ready`, `show`, `delegate
--plot`), daemon + app (new `seed_plot` command, plot progress on the seed push,
panel navigation), protocol (regenerated, version bumped), docs (glossary gains
dispatch-at-plot and stale; the plan's panel-transport entry repointed and its
slice-5 boxes ticked), changelog fragments. Linux is unaffected — no
platform-specific code. Remote-delegate acceptance is explicitly out of this
slice.

https://claude.ai/code/session_01WEMmZhCT6ASFNNFVzSPfDv


figgyster (dismissed):
Approving. Slice 5 is the one where the scope ruling finally stops being a plan note and becomes a deletion — `workspace_id` is gone from the struct, the schema, the wire, the panel, and the help text, no shim, no compatibility read, exactly as ruled. The receipt discipline I've been watching all week is house style now: the stale window's default carries its measurement in the constant's comment (276 gaps, p99 45h, 7d ≈ 3.7× past it), and the plot validator refuses a typo'd `block` key by name instead of silently planting an unsequenced plot.

Two things, neither blocking.

One: `crownPrimeBodyLimit` slices the crown body by bytes — `body[:4000]`. A plan is markdown, and markdown can carry any Unicode. When the 4000th byte lands mid-rune, the injected prompt ends on a broken character. The cap's test only feeds `strings.Repeat("x", …)`, so it can't see it. Rune-safe truncation (or trim to the last `utf8.ValidString` boundary) closes it in two lines.

Two: there are now two builders for the launch prime. `gardenPrime` in the daemon re-derives the dispatch inference and plot scoping itself, while `gardenPrimeFromReady` in the CLI re-renders the wire result. Same shape as the #891 two-manifests note: they agree by hand today, and the cross-check pin (`TestGardenEdges_LaunchPrimerCountsTheSameReady`) only covers the *undispatched* count. The dispatched path is tested to agree on one fixture but never pinned prime-against-ready. A dispatched plot whose primer and `ready` drift would be the one seam here with no witness on the stand. Not urgent — just the place this will rot first if it ever does.

Verified green: `internal/garden`, `internal/hooks`, `internal/protocol` (incl. `TestFrontendProtocolVersionMatchesDaemon`), `internal/client`, the three GardenPanel suites (19 tests), `useDaemonSocket` (85), and `tsc --noEmit`. Protocol is 243 in both human-edited spots and the regenerated files, no 242 left behind. My sandbox can't build the daemon package — needs `ghostty/vt.h` — so I leaned on CI for the wiring; Daemon, Frontend, E2E, and Generated types all came back green as I was wrapping this.

The packaged-app acceptance is the part I'd bet on hardest: ten steps against the real app, and the two refusals only a live run caught — the retired `--all` flag in a `--part-of` error, and the stale header not naming its rule — are exactly the kind of thing a test harness misses. Nice work.

figgyster (dismissed):
Re-approving at the new head. `9b62f55` answers both notes in one commit, same as the last three loops.

The byte-split fix is right — `truncateRunes` walks back to the rune start before slicing, so a capped plan can't end the injected prompt on half a character. I ran the hooks suite locally; the new cap test passes.

One thing about that witness, and it's the only thing I'd tighten. `TestGardenPrimerCapsACrownBodyWithoutSplittingARune` feeds `strings.Repeat("é", 4000)` — and "é" is two bytes while 4000 is even, so the cut lands on a rune boundary. The old byte-slice (`body[:4000]`) produces valid UTF-8 on that exact input too; I red-checked it. The test is green before and after the fix, so it doesn't actually see the split it's named for. A three-byte rune — `strings.Repeat("€", 4000)` — lands the cut mid-rune and would fail on the old code. Two characters in the test string, and the witness actually earns its name. Non-blocking; the fix itself is correct.

The other note got a real witness: `TestGardenPlot_ThePrimerAndReadyAgreeInsideAPlot` pins the dispatched primer's ready-seed ids *and order* against `ready` with `slices.Equal`, plus the count. That's the cross-pin that was missing — same seeds, same order, or a delegate is primed with work its own `ready` won't offer. It lives in the daemon package, which my sandbox can't build (`ghostty/vt.h`), so I'm leaning on the Daemon check for that one; it's queued as I write this.

figgyster (approved):
Re-approving at `3b8d9da`. The residual note landed clean — this time the witness actually saw the split.

The fixture is now `strings.Repeat("€", 4000)`: three bytes against an even limit, so the cut is forced mid-rune where the two-byte "é" divided 4000 exactly and let a plain byte slice pass. I red-checked it both ways — against `body[:crownPrimeBodyLimit]` the test fails with `the capped primer is not valid UTF-8`, and against `truncateRunes` it passes. That's a discriminating witness now, and the comment above it says *why* the rune width matters so the next person editing the fixture doesn't quietly disarm it again.

That closes all three notes from the original review: the byte-split (fixed + witnessed), the two-builders seam (cross-pinned with `slices.Equal`), and the witness that wasn't one (now a real one). Case closed.

PR #893, opened 2026-08-14
docs(crew): the crew primitive plan

victorarias:
The plan for stage 4 of the [home–garden–crew arc](https://github.com/victorarias/attn/blob/main/docs/plans/2026-08-10-home-garden-crew-arc.md): the crew hardens from the hand-run skill simulation into the daemon.

## What this is

A crew member is a durable identity — charter, handoff line, recognition address — whose sessions are its days. The simulation (`~/.attn/crew/`, `/wake`, `/handoff`) has run since 2026-08-06: three members, 18 filed handoffs, ~1,500 lines of member-authored continuity. This plan is what that experience says the primitive is. The first draft went through Victor's 22-annotation pass on 2026-08-13; this is the rewrite, refined twice after.

## What the plan settles

- **What the simulation proved**, as constraints: identity is the invocation, never the files; handoffs are append-only letters to the successor; the ceremony is the pain the primitive exists to remove; a compact is not a nap — handoff completes into an in-place reload with a wake prompted, one motion; plain markdown homes are load-bearing.
- **The lifecycle**: wake as an app surface with the CLI verb as plumbing; priming that injects charter + freshest handoff + crew guidance with logged size; heartbeats that keep a member's cache warm only while the app has user activity; auto-sleep on true inactivity plus cache-expiry pressure; members always visible in the sidebar.
- **The mechanism**: files stay canonical, the registry lives in docstore; binding is the identity; one active binding per member; messaging wakes a sleeping member instead of refusing; every crew verb is fenced home-only.
- **Six slices**, registry-first, chief migration last. Slice 1 (registry + binding) already shipped against this plan as #890 — the glossary merged there links to this doc, so this PR also brings that link live.

## Open forks, held deliberately

The wake limit and the heartbeat policy each wait on a receipt (priming size × model, measured live) before any number is written; the limit ships as a tunable setting, never a constant. The heartbeat arithmetic in the doc is labeled list-price math, not a receipt.

### DIFF
--- a/changelog.d/crew-primitive-plan.yaml
+++ b/changelog.d/crew-primitive-plan.yaml
@@ -1,0 +1,13 @@
+kind: internal
+area: docs
+change: >
+  The crew primitive plan, rewritten from Victor's annotation pass on
+  the first draft. Seven days of the hand-run simulation settled the
+  constraints — identity is the invocation, handoffs are append-only
+  letters, the ceremony is the pain — and the plan turns them into six
+  slices: registry and binding, wake as an app surface, handoff that
+  reloads in place instead of compacting, an activity subsystem that
+  extends the existing presence system with cache state and cost,
+  crew addressing that wakes a sleeping member, and the chief's
+  migration into the first standing member. The wake limit and
+  heartbeat policy stay open forks, each waiting on a receipt.
--- a/docs/plans/2026-08-11-the-crew-primitive.md
+++ b/docs/plans/2026-08-11-the-crew-primitive.md
@@ -107,9 +107,10 @@ never do. The primitive builds that world:
   session-activity generation is its consumer today. The crew
   lifecycle becomes the second consumer and adds what presence lacks:
   per-session context-cache state, session cost, and a measured
-  "true inactivity" threshold — the existing 90s presence-idle limit
-  is an unmeasured guess tuned for a different purpose, so auto-sleep
-  gets its own receipt.
+  "true inactivity" threshold — the existing idle knobs (the
+  `activity.presence_idle_seconds` setting and the ten-minute
+  watching cap) are unmeasured guesses tuned for session-activity
+  generation, so auto-sleep gets its own receipt.
 
 ```mermaid
 stateDiagram-v2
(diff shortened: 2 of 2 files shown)


figgyster (dismissed):
Reading the plan with slice 1 already shipped against it — the doc that #890's glossary pointed at before it existed. That link finally resolves, which is its own small case closed.

The receipts discipline is the through-line and it holds: every number that isn't measured yet is held open — wake limit as a tunable setting, heartbeat as labeled list-price math — and the "measure first, cap nothing" habit that's been running through the A5 docs now runs through the crew plan too. The 22-annotation rewrite shows: the "ruled" list reads like decisions, not aspirations, and "identity is the invocation, never the files" is already structural in `internal/crew` (binding stamped at launch, reading a charter confers nothing).

Two nits, both precision, neither blocking:

1. The heartbeat arithmetic slips a zero. At your own list-price rates — input $15/M, cache-write 2× — fresh priming of ~10k tokens is **$0.30**, not $0.03. (Even a cache-read of 10k is ~$0.015.) The "$3.00 = twenty heartbeats" line checks out against $0.15/read, but "sleeping ≈ $0.03" doesn't follow from any rate in the paragraph. Sleeping's still cheap — ~10× cheaper than a lapse, not ~100× — so the argument survives, but the example should carry its own arithmetic.

2. "The existing 90s presence-idle limit" names the wrong constant. In `internal/daemon/client_presence.go` the 90s is `presenceHeartbeatGrace` — how long a client's stale report stays believable, a tripwire, not an idle limit. The actual idle limit is the tunable `activity.presence_idle_seconds` setting plus the 10-minute watching cap. The point holds — those thresholds are unmeasured guesses tuned for activity generation, not auto-sleep — but the sentence credits 90s with a role it doesn't play, and the idle limit it gestures at is already a setting, which is the very pattern the wake-limit fork argues for.

Approving. This is the design doc slice 1 already shipped against, and it's the cleanest statement of the arc's crew stage yet.

figgyster (approved):
Both answered, both right. The heartbeat example now carries its own arithmetic — $0.30 with the 2× cache write, "a tenth of the lapse, not a hundredth" — and the idle sentence names the real knobs: `activity.presence_idle_seconds` plus the ten-minute watching cap, the 90s tripwire gone. Exactly the two constants I mislabeled as the author.

Red-checked the diff: only the doc moved, two hunks, one per note. Re-approving at the new head.

PR #894, opened 2026-08-14
feat(apps): dock an app's view as a tile

victorarias:
A5 slice 3 of [the UI host and app SDK plan](../blob/main/docs/plans/2026-08-13-ext-a5-ui-host-and-app-sdk.md): a declared view, docked as a tile, renders live in a workspace. Slices 1 and 2 built the declaration and the SDK package; this is the half that runs them.

clip

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/2d64674178338126033794f6ff1db9bc578e4738/delegate-a5-slice3-host-ef61f3e9/clip.mp4)

The clip is the live verification, in order: ⌘K → the view in the command menu → the question it declared, answered → the view mounts beside the terminal, drawing its own props → `attn app disable reviewer` → the tile says so in place → `attn app enable reviewer` → it is back.

## The open question this slice had to answer

**Does WKWebView permit a cross-origin ES-module import from `tauri://localhost` to the daemon's port?** Yes — proven in the packaged app, so the plan's `blob:` fallback is not taken and nothing above the host changes.

Receipt: on profile `a5slice3` (port 25721), a view served from `http://127.0.0.1:25721/apps/bundle/reviewer/<hash>/approvals.js` mounted in the packaged app, rendered its own component, received its props, and ran `useState` from attn's React. The module in that response begins `import{useState as A}from"@victorarias/attn-app";import{jsx as D,…}from"@victorarias/attn-app/jsx-runtime"` — bare specifiers, resolved in the browser by the import map, never bundled by the app.

## What it ships

- **Bundle route.** `/apps/bundle/<app>/<hash>/<view>.js` on the daemon mux beside `/ws` and `/health` — `Access-Control-Allow-Origin: *`, `Cache-Control: public, max-age=31536000, immutable`, `text/javascript`. The content hash is in the path, so the URL moving IS the reload mechanism. A malformed path is a 400; a version that no longer has the artifact is a 404 naming the app, the view, the short hash and `attn app status <app>`.
- **The registry on the wire.** `initial_state.apps` plus an `apps_updated` broadcast, projected from A4's three facts (`app.version.changed`, `app.enabled.changed`, `app.removed`) through `wireProjections`. A tile learns from it whether its app is installed, on, and which version serves. The views it carries come from the serving version's **frozen declaration**, not the manifest on disk — after a rollback those differ, and what docks is what serves.
- **One React.** A static import map in `index.html` maps all three SDK specifiers (`@victorarias/attn-app`, `/jsx-runtime`, `/jsx-dev-runtime`) to three fixed-name frontend chunks. `preserveEntrySignatures: "allow-extension"` is what keeps those chunks' exports — without it Rollup discards an entry's signature and the SDK chunk ships a React with no `useState`. `appSdk.oneReact.test.ts` and `importMap.test.mjs` (which joins the SDK's package exports, Go's `SDKSpecifiers()`, the import map, and the vite chunk names) are the guards.
- **The host.** `AppTileHost` dynamic-imports the bundle URL behind a per-app error boundary, under the `app:<app>/<view>` tile kind, with a dock-picker entry per view and the params prompt for a view that declares one. Six named placeholder states — not installed, disabled, no version, view gone, load failed, crashed — each saying what is wrong and the way back. **None of them removes the tile**; undocking stays the user's act.
- **Remount.** A version landing under a docked tile remounts it against the new hash. If the tile holds focus the remount waits and a static badge says so — no animation anywhere in the host chrome.
- **Crashes are the app's.** A caught render error is recorded as an app invocation stamped with the serving version, and written to the app's log, so `attn app logs <app>` — which is what the crashed tile tells the author to run — has the message and the frame inside their own bundle. It deliberately does **not** advance A4's stall clock: that clock protects the event log's retention floor, and a crashing tile pins nothing.

Protocol 242, full lockstep (`main.tsp` → `make generate-types` → `constants.go` → `useDaemonSocket.ts`). `attn app status` gained a `views:` block naming each view's tile kind and the question it asks.

## Decisions worth reading

**View bundles are built with `--production`, and that is not about bytes.** Without it, bun emits `import { jsxDEV } from "@victorarias/attn-app/jsx-dev-runtime"` regardless of the app tsconfig's `"jsx": "react-jsx"` and regardless of a `NODE_ENV` define — measured, both probed and rejected. React's production build exports `jsxDEV` as `undefined`, so such a view links cleanly against attn's React and then throws on its first element. The dev specifier is still marked external and still carried in the import map: whether bun reaches for it is bun's decision, and a build that resolved it into a second React would be worse than one that cannot resolve it at all. Nothing shipped reaches that third mapping today.

**Views are now typechecked against the SDK.** Slice 1 left this as a named gap; `appbuild.typecheck` appends every view entrypoint now that both halves exist.

**Two fixes the packaged app found, which no jsdom test could have.**

- The params field was not typable. It opens from the command menu, whose own focus trap returns focus to the terminal as it closes — after this prompt has mounted — so a focus-on-mount was undone and every keystroke went to the shell. It holds focus in a trap now, and the regression test asserts against a later steal rather than against the first frame.
- A crash report led with a stack frame instead of the message. WebKit's `Error.stack` is frames only, unlike V8's, so a report built from `stack` alone names where it threw and never what was thrown.

**`app_view_get_state`** joins the automation bridge: the harness had no way to tell a mounted view from a placeholder that happens to render text.

## Verification

Live on throwaway profile `a5slice3` (never `dev`), bundled preflight PASS (protocol 242 across CLI/app/daemon) before each run. Beyond the clip: rollback from a broken version recovers the tile in place; a version that drops the view lists what the serving version does offer; removing the app leaves the tile saying so and re-applying remounts it; a full app relaunch leaves a docked tile exactly where it was. `real-app:scenario-tr301` passes — ⌘T typing and switch-away/back utility focus survive the frontend changes. Go and frontend suites green.

## Not in this slice

WS doc-subscription envelopes and `useQuery` (slice 4), `[[commands]]`/components/scaffold (slice 5), any mount kind but tiles. The **raw-string param** is built as the plan drafts it and is deliberately untyped — it is the open question for you to judge by feel now that it is dockable.


figgyster (approved):
Ran the whole thing. Protocol 242 locks up — `TestFrontendProtocolVersionMatchesDaemon` green, `tsc --noEmit` clean, the new frontend suites (appViews, appSdk import map + oneReact, appBundle) all 23 passing, and `go test ./internal/appbuild ./internal/apps ./internal/protocol` clean. The daemon package only builds after fetching the prebuilt libghostty-vt archive; the app tests (`app_bundle`, `app_ui`, the `apps_updated` wire-join fixtures) pass. The two failures in the full daemon run are the plugin driver/runtime e2e spawners, pre-existing and untouched by this diff.

This is the half of the A5 arc that finally *runs* slices 1 and 2, and it lands the way the receipts discipline promised. The reload mechanism being "the content hash in the URL moved" is the right call — no watcher, no polling, no cache-busting prayer, just immutable URLs that change identity when the version does. Reading "what docks" off the frozen declaration instead of the manifest on disk is the rollback-honesty detail that actually matters, and a caught render error becoming an invocation stamped with the serving version closes the loop so `attn app logs` holds what the author needs. One-React-via-import-map is the whole design standing up: three surfaces that can't see each other (SDK exports, Go's `SDKSpecifiers`, the import map + chunk names), pinned by a test that reads all three.

One nit, non-blocking:

**Reload re-reports the crash before the new bundle loads.** `app/src/components/appViews/AppTileHost.tsx:169` — `resetKey` is `` `${mounted.contentHash}:${attempt}` ``, and `AppViewBoundary.getDerivedStateFromProps` clears the error the instant that key changes. But clicking Reload only bumps `attempt`; `mounted.component` is still the old, crashing component (the effect clears `mounted` only when the app stops being mountable or a load errors, not on reload). So the boundary resets, re-renders the *stale* View, it throws again, and `reportCrash` fires a second `app_view_crash` for the same crash — all before the fresh import resolves. `attn app logs` picks up a duplicate per Reload click. "Fresh boundary above a fresh component" only becomes true after the async load lands; the boundary resets against the corpse first. Clearing `mounted` on reload — so the Loading state shows instead of re-rendering the thing that just threw — would make the reset key honest. No test pins the Reload path today, which is why it's unwitnessed.

That's the file. Approving.

figgyster (approved):
Red-checked the new head (`209d5ca`). The Reload note got answered the right way: `reload` now drops `mounted` and clears `loadError` *before* bumping `attempt`, so the crashed component is unmounted instead of re-rendered against a fresh boundary — the Loading placeholder takes its place, and the fresh import lands in a brand-new boundary. No second `app_view_crash`.

The witness is the real thing, not a paraphrase. The new `AppTileHost` test holds the second load open on a controlled promise and asserts the crash count is still 1 in exactly the window the bug lived in — after the reset key moved, before the module resolved. Ran it: 10/10 green, plus `tsc --noEmit` and the appbuild specifier tests.

The other two commits are noise with a purpose — gofmt on the specifier regex, and `useShortcut` imported from its own module instead of the barrel. Both benign.

That closes the only note I had. Re-approving at the new head.

PR #895, opened 2026-08-14
feat(apps): let a docked view read live documents

victorarias:
A5 slice 4: a docked app view reads its app's documents and stays current.

Slices 1–3 gave an app a frozen view declaration, a bundle route, and a tile
host that mounts the view with its `ViewProps`. The view could render, but it
could not read anything. This slice gives it a live query.

## What it does

A view calls `useQuery` from the app SDK and gets back the matching documents,
in the server's order, re-rendered whenever one of them changes. No polling, no
refresh button, and no subscription bookkeeping in the view.

## The wire

Four envelopes and a protocol bump to 243 (`main.tsp` → `make generate-types` →
`constants.go` → `useDaemonSocket.ts`):

- `doc_subscribe` gains an optional `subscription_id` — required on the
  WebSocket, refused on the unix socket, where a second subscription means a
  second connection and an id would name nothing.
- `doc_unsubscribe` ends one by id.
- `doc_subscription_delivery` carries `{as_of_seq, order, upsert}`.
- `doc_subscription_ended` carries the code and the message.

`handleDocSubscribe` is refactored into a sink-driven loop (`runDocSubscription`
in `internal/daemon/documents.go`): the loop knows the query and the resume
token, and a `docSink` — two closures — knows the transport. The IPC sink writes
the connection's encoder; the WebSocket sink wraps each delivery in an envelope
and hands it to the client's write pump. The A3 deadlock rule survives
unchanged: the bus fan-out holds the publish lock and only pokes a one-slot wake
channel, and every socket write happens on the subscription's own goroutine.

## Delivery semantics

A3.4 Stage 2's contract, held by `useQuery` so a view never meets it: render
`order`, take each body from `upsert` or the cache, forget everything not in
`order`. Resume is by `have` — `{id, rev}` pairs for what the client still
holds — so a remounted tile gets only what changed while it was gone, not a log
replay. A subscription refuses an `after` cursor: a live query is a window, a
cursor is a walk.

A subscription the collection can no longer answer ends rather than lingering:
`collection_undefined` when the collection is removed, `collection_redeclared`
when it loses a field the query uses. `useQuery` surfaces that as a state the
view renders.

## The limit

`DocSubscriptionsPerClient = 64`, measured against the production database:
seven live workspaces, three holding exactly one docked tile and four holding
none. A view can reasonably open more than one query, so the healthy ceiling is
single digits and this sits an order of magnitude past it. Reaching it is a
named refusal — the limit, its value, and the ask — never a dropped
subscription. Both the refusal and the ending ride the same
`doc_subscription_ended` envelope, because `command_error` cannot name the
subscription it refused.

## Live verification

Throwaway profile `a5slice4`, full app install from this branch, bundled
preflight green (protocol 243 across CLI, app, and daemon). A fixture app
declares a `seen` collection and a tile view calling `useQuery`; the tile is
docked through `workspace_layout_dock_tile` and read back through
`app_view_get_state`.

1. dock, empty collection → `live · 0 document(s) · as of 257`
2. `attn doc put` from outside the app → the document appears
3. a second write, then an edit → both land, the edit bumps the rev
4. `attn doc delete` → the row leaves the window through `order` alone
5. undock, write while it is gone, re-dock → the remount resumes current
6. re-dock with tile params → the params become the query's filter
7. redeclare without the queried field → `collection_redeclared`, rendered
8. undefine the collection → `collection_undefined`, rendered

clip

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/63e236365762b32b3c7fd7bf04e595d1ad1aca16/feat-a5-slice4-live-queries/clip.mp4)

## Surfaces walked

- **Protocol** — `generated.go` and `generated.ts` regenerated together,
  `ProtocolVersion` 243. `command_meta.go` is a fourth lockstep spot beyond the
  three AGENTS.md names: `TestEveryProtocolCommandIsClassified` fails without an
  entry for `doc_unsubscribe`.
- **Daemon and app** — `daemonDocumentEvents.ts` handles the two new events from
  the socket switch's `default` chain; `AppTileHost` composes the namespace and
  provides the runtime.
- **CLI** — unchanged: `attn doc` already reads and writes; a subscription is a
  client concept and the unix socket keeps the connection-scoped form it had.
- **SDK** — `useQuery` and the runtime context, with declarations regenerated
  into `internal/appbuild/sdkdist` (`make check-sdk` green).
- **Linux** — `go build` for `linux/amd64` green; nothing here is Darwin-only.
- **Docs** — a "Slice 4 as-built" section in the plan doc, and a changelog
  fragment.

## Notes for slice 5

- Slice 5's protocol bump renumbers from 243.
- `command_meta.go` needs an entry for every new command.
- A view's namespace is composed by the host, not the SDK — the SDK never sees
  the app name.

### DIFF
--- a/app/src/appSdk.useQuery.test.tsx
+++ b/app/src/appSdk.useQuery.test.tsx
@@ -1,0 +1,190 @@
+import { describe, expect, it } from "vitest"
+import { act, render, screen } from "@testing-library/react"
+import {
+  AppViewRuntimeProvider,
+  useQuery,
+  type AppViewRuntime,
+  type DocumentSubscriber,
+} from "@victorarias/attn-app"
+
+// useQuery holds A3.4's delivery contract so a view never meets it. What is
+// asserted here is that contract from the view's side: render `order`, take each
+// body from `upsert` or the cache, forget the rest — and, across a remount,
+// resume by declaring what is still held so only what changed travels.
+
+type Body = { status: string }
+
+function doc(id: string, rev: number, status: string) {
+  return { id, body: JSON.stringify({ status }), rev, created_at: "", updated_at: "" }
+}
+
+/** A host stand-in that hands the test the subscriber the SDK registered. */
+function fakeRuntime() {
+  const subscribers: DocumentSubscriber[] = []
+  let unsubscribes = 0
+  const runtime: AppViewRuntime = {
+    namespace: "app/approval-gate",
+    subscribe: (subscriber) => {
+      subscribers.push(subscriber)
+      subscriber.onLive(true)
+      return () => {
+        unsubscribes += 1
+      }
+    },
+  }
+  return {
+    runtime,
+    subscribers,
+    latest: () => subscribers[subscribers.length - 1],
+    unsubscribeCount: () => unsubscribes,
+  }
+}
+
+function View({ collection }: { collection: string }) {
+  const { docs, live, error, asOfSeq } = useQuery<Body>(collection, {
+    filters: [{ field: "status", op: "eq", value: "pending" }],
+  })
+  if (error) return <div data-testid="error">{error.code}</div>
+  return (
+    <div>
+      <div data-testid="live">{live ? "live" : "not live"}</div>
+      <div data-testid="seq">{asOfSeq}</div>
+      <div data-testid="ids">{docs.map((d) => `${d.id}:${d.body.status}`).join(",")}</div>
+    </div>
+  )
+}
+
+function mount(runtime: AppViewRuntime, collection = "requests") {
+  return render(
+    <AppViewRuntimeProvider value={runtime}>
+      <View collection={collection} />
+    </AppViewRuntimeProvider>,
+  )
+}
+
+describe("useQuery", () => {
+  it("renders the window in the server's order, from bodies it was sent", () => {
+    const host = fakeRuntime()
+    mount(host.runtime, "renders")
+
+    act(() => {
+      host.latest().onDelivery({
+        delivery: 1,
+        asOfSeq: 9,
+        order: ["b", "a"],
+        upsert: [doc("a", 1, "pending"), doc("b", 1, "pending")],
+      })
+    })
+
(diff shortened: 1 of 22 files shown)


figgyster (approved):
A5 slice 4, and the SDK finally has something to *say* — not just a view to render, but a query to answer it with.

The move I respect most here is the refactor nobody will applaud: one subscription loop (`runDocSubscription`) behind a two-closure sink, so the unix socket and the WebSocket share one set of semantics instead of two. That's the boring kind of correct. The A3 deadlock rule survives untouched — the loop still never runs inside the bus fan-out, and the sink carries the whole difference between "this connection IS the subscription" and "one connection, many subscriptions." Two transports, one truth. That's how you avoid a fork you'd be paying off for years.

The receipts are all on the stand. Resume-by-`have` so a remounted tile gets only what changed, and the witness that pins it drives the diff across an unsubscribe, a write, and a resubscribe — not just a happy-path reopen. The 64-subscription tripwire is measured against your production DB and refuses by name instead of dropping silently. The one-envelope choice for refusals is right (`command_error` can't name the subscription it refused), and `TestIPCSubscribeRefusesASubscriptionID` holds the two transports apart at exactly the seam where they differ.

Protocol lockstep walked clean: 243 across constants, generated.go, generated.ts, and the `command_meta.go` entry you named as the fourth spot. That's the one that bites people at handshake, and you said it out loud in the PR body instead of hoping CI tripped over it.

One non-blocking note, and it's cosmetic: `useRef(new DocumentSubscriptions())` trips react-doctor's `rerender-lazy-ref-init`. The constructor's cheap — one empty Map — so this is noise, not a leak, and it matches the file's existing eager idiom (`createPtyTransportState` does the same). If you want the rule quiet, lazy-init it (`useRef<DocumentSubscriptions | null>(null)` + a guard); if you'd rather not spend a diff on it, I won't lose sleep. Your call.

Approve. Slice 5 renumbers from 243 — I'll hold you to the "every new command gets a `command_meta.go` entry" promise when it lands.

PR #896, opened 2026-08-14
feat(crew): wake a member from the sidebar

victorarias:
Slice 2 of the crew primitive plan
([docs/plans/2026-08-11-the-crew-primitive.md](https://github.com/victorarias/attn/blob/main/docs/plans/2026-08-11-the-crew-primitive.md)):
the wake surface. Slice 1 (#890) made a session able to launch *as* a member;
this makes waking one an act the user can perform, and makes the woken session
actually know how to be that member.

crew-wake

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/79488c1a4691836af7a0c954c158dd89cf6230dc/feat-crew-slice-2/crew-wake.mp4)

The clip is the live verification: five members in the sidebar, `brier` asleep
with a hollow ring, one click on its row, and it is awake and focused.

## The sidebar

Members render in the pinned-agents region, pin-shaped — a member stepped out
of the queue the way a pinned agent did — but visually distinct from an
ordinary pin, because nobody put it there and there is no unpin: an accent rail
down the left edge, a `crew`/`asleep` mark, and a hollow ring instead of a state
indicator while nothing is running. Every member is present whether or not it
has a day. Clicking an awake member focuses the day it is living; clicking a
sleeping one wakes it. No new section, no new shortcut.

An awake member's session is routed into the crew band **before** every other
band, including the workspace pinned/muted check, so it draws exactly once and
stays visible even from a hidden workspace.

## The wake verb

`crew_wake` is one daemon composite: bind the member, register its durable
per-member workspace if absent, add a pane, spawn a session in the member's own
directory with its awareness dirs. The binding is claimed **before** the spawn,
because the launching wrapper asks `crew_prime` for what to inject and the
binding is what answers that question; a launch that then fails releases it and
removes the pane. A member that is already awake is not woken twice — the
answer names the live day, which is the same thing to focus, so the sidebar's
one action never fails exactly when the member is present.

`attn crew wake <name> [--agent <harness>]` is the same verb from the command
line.

## Priming

The launch carries the member's charter, the freshest letter in `handoffs/`
(lexicographically last), the names of the older letters, and the crew
guidance: how a handoff is filed, that a crew day-line is not a seed handoff,
and that only one session is this member at a time. It rides the same
launch-guidance path the garden primer does — `--append-system-prompt` on
claude, config overrides on codex — and comes last, as the most specific thing
the session is. Awareness dirs reach the harness natively as `--add-dir`, which
both built-in CLIs take.

The size is logged at injection time, so the logged bytes are exactly the
injected bytes:

```
crew: priming trellis for session 89e05995…: 5036 bytes (charter 2373, handoff 2026-08-13T22-20Z-trellis.md 434, older 7)
```

**Measured on the three real homes** (the receipt the plan's wake-limit and
heartbeat forks were waiting on): trellis 5,036 · keel 5,791 · alder 13,280
bytes. A member with no charter and no letter yet — a first day — costs 1,590.
So a wake's whole injected identity is ~1.5–13KB, call it under 3.5k tokens at
the top end. The inline limits (16KB per file) sit well past the largest real
file (5,569-byte charter, 6,601-byte letter), and a cut says where the whole
file is rather than ending mid-sentence.

## Where a member works

Nothing wrote `cwd` or `awareness_dirs` before this, and the registry records
where a home lives, never what it says — so reading them out of charter
frontmatter was out. They are registry fields with their own verb:

```
attn crew set <member> --cwd <dir> --awareness-dir <dir>   # repeats; once empty clears
```

Both are validated at set time (a directory that is not there is refused with
its path) rather than at the next wake, which is the wrong end of the day to
learn about a typo. A member with no recorded cwd launches in its own home.
This is the one thing the plan does not explicitly rule on; flagging it rather
than burying it.

## Fence

Every crew verb goes through `d.requireHome(crew.Surface)`. An outpost refuses
by name — wake, set, and prime alike — and primes no session.

## Surfaces walked

- **CLI** — `crew wake`, `crew set`, both in `crew --help`. Flags parse on
  either side of the member name (Go's `flag` stops at the first positional,
  and `attn crew wake trellis --agent codex` is how anyone types it).
- **Daemon and app** — IPC + WebSocket command, `crew_updated` broadcast,
  `crew` in `initial_state`.
- **Protocol** — `main.tsp` → `make generate-types`; `ProtocolVersion` 243 →
  244 in all three lockstep spots.
- **Event bus** — four facts (`crew.registered|bound|released|updated`) behind
  one roster projection, with fixtures in the wire-join table.
- **Linux** — `GOOS=linux go build ./cmd/... ./internal/...` clean.
- **Docs** — the 2026-08-14 rulings in the plan, wake/priming/awareness-dirs in
  `docs/glossary.md` (and the "waking a live member is refused" line corrected
  to what this ships), a changelog fragment.

## Verification

Throwaway profile `crewslice` (fresh `make install PROFILE=crewslice`,
preflight PASS, protocol 244 on CLI/app/daemon), queue mode on, seeded with a
copy of the real crew homes. Verified live: the roster imports from homes on
disk; members render distinctly; a sleeping member wakes with one click and the
app focuses it; an awake member focuses; `attn crew wake` and `attn crew set`
from the CLI; the priming receipt in the daemon log; the launched `claude`
argv carrying both the crew block and `--add-dir`; and a woken trellis reading
its own letters and orienting unprompted. Bindings survived a daemon bounce.
The outpost refusal is covered by tests rather than live — this environment has
one home daemon.

Profile cleaned after.

### Known rough edge, not introduced here

A member woken into a directory Claude Code has never seen lands on its
folder-trust prompt — the same thing any new cwd does. `attn crew set <member>
--cwd <dir>` points a member at a directory it already works in and avoids it.

### Gap worth naming

Crew rows live in the pinned region, which exists only in queue mode. With the
queue arrangement off there is no crew surface at all. Queue mode is on in
practice, and inventing a second placement for the tree would contradict the
ruling; saying so rather than shipping it silently.

## Non-goals

Slices 3–6 (nap/handoff, messaging, wake limits, recognition), write-refusal on
member homes, and adopting mid-flight sessions. The `/wake` and `/handoff`
skills in the user config keep running beside this, untouched.

https://claude.ai/code/session_019aERUqnAtexLUX3HQB5sqD


figgyster (dismissed):
Slice 2 lands the wake surface, and it's the same receipts-first discipline all the way down: claim before spawn so `crew_prime` has a binding to answer, the logged size is the injected size (trellis 5,036 · keel 5,791 · alder 13,280, measured on the real homes), and the rune-cut test uses a 3-byte rune so it actually sees the split — the #892 lesson, learned before I had to say it.

One real bug, non-blocking, and it's not a nit:

`attn crew set <member> --awareness-dir ""` doesn't clear. The documented way out sends an empty `[]string{}`, but `CrewSetMessage.AwarenessDirs` is tagged `json:"awareness_dirs,omitempty,omitzero"`, and `encoding/json` drops empty slices under `omitempty` — so the daemon receives `nil` and `handleCrewSet` skips the `if msg.AwarenessDirs != nil` branch. The command exits 0 and echoes the old dirs back. `--cwd ""` clears fine because `cwd` is a `*string` and a non-nil pointer survives `omitempty`. Verified: `json.Marshal(CrewSetMessage{Member: "keel", AwarenessDirs: []string{}})` → `{"cmd":"crew_set","member":"keel"}`. `TestCrewSet_RecordsAndClearsWhereAMemberWorks` passes only because it calls the handler directly and skips the wire — give awareness dirs the pointer treatment (or an explicit clear flag), and pin a wire-level test so the in-process one stops fibbing.

Approve at `37a35cf`.

figgyster (dismissed):
Fix lands exactly where the hole was. `clear_awareness_dirs` is its own `*bool` — a non-nil pointer survives `omitempty`, so the way out finally has a wire presence instead of an empty list that marshals into thin air. The daemon reads the flag first (`if Deref(msg.ClearAwarenessDirs)`), so clear outranks a stale dir list. Protocol bumped 244 → 245 in all three lockstep spots.

The new witness is the part that matters: `TestCrewSet_ClearingAwarenessDirsSurvivesTheWire` builds the message the client builds, marshals it, and decodes it through `protocol.ParseMessage` before the handler sees it — so an encoder that drops a field drops it here too. That's the wire round-trip the old test skipped, and it's why the old test lied. Red-checked my side: `clear_awareness_dirs:true` survives the round-trip, a real dir doesn't set it, crew and agent packages still green.

Re-approve at `7868502`, stale `37a35cf` approval dismissed.

figgyster (approved):
Trivial tail commit, and clean: `buildCrewRows` drops its `export` and stays a private helper. Nothing imports it outside `QueueBands.tsx` — the test exercises it through the component, not the symbol — and `QueueBands`, `QueueSnoozedSection`, and `CrewMemberView` all keep their exports, so `Sidebar`'s imports hold. No behavior change.

Re-approve at `38fc02e`, stale `7868502` approval dismissed.

PR #897, opened 2026-08-14
refactor(garden): a seed's record is its log, not its trail

victorarias:
A vocabulary rename, ruled 2026-08-14: the append-only record of what happened to a seed — notes, moves, edges — is now the seed's **log**, not its trail.

## Why

"Trail" descends from the audit-trail idiom, which carries a corporate register the garden deliberately avoids. "Log" needs no teaching at all: it sight-reads beside `git log` as "the append-only history of this thing, written by the people who worked on it, read by whoever comes next," and the gardener's logbook keeps it inside the metaphor family. The runner-up, "story," lost on the agile-story collision. The vocabulary already carries one hard word (harvest); this rename spends none of that budget.

## What changed

Prose only — help text, CLI output, comments, the glossary, and the garden plan. `grep` receipts:

- `cmd/attn/seed.go`: `--stale` and `--limit` help, the stale query's header line, the empty-log hint, and a local variable renamed for clarity.
- `internal/garden`, `internal/daemon/garden.go`: comments and one daemon log line.
- `internal/protocol/schema/main.tsp`: **comments only** — no shape moved, so no regeneration and no `ProtocolVersion` bump.
- `docs/glossary.md`, `docs/plans/2026-08-06-the-garden-vertical-slices.md`: the note entry, the stale definition, and every slice reference.

Deliberately untouched: `GardenPanel`'s navigation breadcrumb keeps the name *trail* — that one is a path walked root-to-leaf, the genuine hiking sense, and this rename removes the ambiguity between the two concepts rather than adding to it.

Nothing on the wire or in storage carried the word, so there is no migration and no compatibility concern.

## Verification

`go build ./...`, `go test ./internal/garden/... ./internal/daemon/ ./cmd/attn/...` green. Trivial-tier (rename of prose and help strings, no behavior); no live verification run, per the exemption in AGENTS.md.

### DIFF
--- a/changelog.d/seed-trail-renamed-to-log.yaml
+++ b/changelog.d/seed-trail-renamed-to-log.yaml
@@ -1,0 +1,11 @@
+kind: changed
+area: garden
+change: >
+  A seed's activity record is now its log, not its trail. Victor ruled
+  the rename: "trail" descends from the audit-trail idiom, which reads
+  corporate, while "log" sight-reads instantly beside git log and the
+  gardener's logbook. The word changes in seed help text, CLI output,
+  the stale query's wording, the glossary, and the plan; nothing on the
+  wire carried it, so no protocol change. The panel's navigation
+  breadcrumb keeps the name trail — that one really is a path walked,
+  and the rename removes the collision between the two.
--- a/cmd/attn/debug_read_test.go
+++ b/cmd/attn/debug_read_test.go
@@ -52,7 +52,7 @@ func TestReadLinesFile_LongLine(t *testing.T) {
 	}
 }
 
-func TestReadLinesFile_NoLogingNewline(t *testing.T) {
+func TestReadLinesFile_NoTrailingNewline(t *testing.T) {
 	path := filepath.Join(t.TempDir(), "notrail.txt")
 	if err := os.WriteFile(path, []byte("a\nb\nc"), 0644); err != nil {
 		t.Fatalf("WriteFile: %v", err)
--- a/cmd/attn/debug_read_test.go
+++ b/cmd/attn/debug_read_test.go
@@ -52,7 +52,7 @@ func TestReadLinesFile_LongLine(t *testing.T) {
 	}
 }
 
-func TestReadLinesFile_NoTrailingNewline(t *testing.T) {
+func TestReadLinesFile_NoLogingNewline(t *testing.T) {
 	path := filepath.Join(t.TempDir(), "notrail.txt")
 	if err := os.WriteFile(path, []byte("a\nb\nc"), 0644); err != nil {
 		t.Fatalf("WriteFile: %v", err)
--- a/cmd/attn/seed.go
+++ b/cmd/attn/seed.go
@@ -85,7 +85,7 @@ commands:
 
   ls [--stale [--window <duration>]] [--tree] [--json]
         the garden, newest first. --tree nests each seed under the crown it is
-        part of. --stale narrows to the open seeds whose trail has not moved
+        part of. --stale narrows to the open seeds whose log has not moved
         for the window (default %s) — a query for your judgment, never a
         reaper.
 
@@ -107,7 +107,7 @@ commands:
   show <id> [--json]
         one seed: the freshest handoff left on it, its state, who tends it,
         every edge that touches it in both directions, its body, and the newest
-        notes on its trail.
+        notes on its log.
 
   tend <id> [--member <name>]
         claim the seed and start growing it. One tender at a time: tending a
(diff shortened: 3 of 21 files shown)


figgyster (dismissed):
Approved. Prose rename, clean at the seam that matters — nothing on the wire, nothing in storage, no protocol bump, and the schema diff is comments-only. The one variable rename (`trail` → `entries` in `fprintSeedShow`) reads consistently across all three of its uses.

But the sweep stops short of the whole house. You reworded the CLI help, the glossary, the plan, the daemon, and the schema comments — then left "trail" standing in live code, in exactly the sense this rename exists to retire:

- `internal/client/garden.go:178` — `"the daemon answered without a trail"`. This one isn't a comment; it's a string a user sees. When a `seed notes`/`show` response comes back malformed, the CLI tells them the daemon "answered without a trail" — the old word, in their face.
- `internal/client/garden.go:40,63,99,166` — four doc comments on `SeedShow` / `SeedNote` / `SeedNotes` still say "trail."
- `internal/garden/plot.go:63,65,72` — `DefaultStaleWindow` and `Stale` still say "trail movement" / "whose trail has not moved."
- `internal/daemon/bus.go:213-214` — `FactGardenNoted` is still documented as "appended to a seed's trail."
- `cmd/attn/seed.go:556` — `withoutNote` "drops one note from a trail by id."

The breadcrumb you kept is the right call — that trail really is a path walked, root to leaf, and `GardenPanel.tsx` naming its state `trail` is the genuine hiking sense. The collision this PR is removing is between *that* and the record; leaving the Go client and the stale window on the old word keeps the collision half-alive. A future reader grepping the garden will still find "trail" meaning the log, four files deep.

None of it blocks — it's prose, not behavior. But this is the one kind of PR where the completeness *is* the change. Sweep `internal/client/garden.go`, `internal/garden/plot.go`, `internal/daemon/bus.go`, and that one `seed.go` comment, and the word is actually gone.

figgyster (dismissed):
Swept clean. The client's `"answered without a trail"` is now `"answered without a log"` — the one string a user actually sees, fixed. The four doc comments, `plot.go`'s stale window, `bus.go`'s `FactGardenNoted`, and the `withoutNote` comment all went too, and you went further than I asked: the test files' comments and fail strings got the same pass (`garden_plot_test.go`, `plot_test.go`, `seed_handoff_print_test.go`, `garden_test.go`).

One residual, and it's below the threshold of mattering: four test *function names* still carry `Trail` in the identifier — `TestGarden_TrailReadsNewestFirstAndSaysWhatItWithheld`, `TestGardenPlot_StaleReadsTheTrailNotJustTheDocument`, `TestGarden_TheFreshestHandoffSurvivesABusyTrail`, `TestReasonLimitNamesItselfAndPointsAtTheTrail`. The bodies are renamed; the names aren't. That's identifier churn, not prose — a `grep` still surfaces them, but nobody reads a test name the way they read help text. Rename 'em if you want the word fully gone; I won't book you for leaving them.

Re-approved at the new head.

figgyster (dismissed):
Requesting changes — the last commit over-reached. The four seed-record renames are right (`TestGarden_LogReadsNewestFirstAndSaysWhatItWithheld`, `TestGardenPlot_StaleReadsTheLogNotJustTheDocument`, `TestGarden_TheFreshestHandoffSurvivesABusyLog`, `TestReasonLimitNamesItselfAndPointsAtTheLog`). But the find-and-replace matched the `Trail` substring *inside* `Trailing` and turned four unrelated test names into nonsense:

- `cmd/attn/debug_read_test.go:55` — `TestReadLinesFile_NoLogingNewline` (was `NoTrailingNewline`)
- `internal/daemon/ws_picker_test.go:30` — `TestParseBrowseInputTreatsLogingSlashAsBrowseIntoDirectory` (was `…TrailingSlash…`)
- `internal/notebook/store_test.go:604` — `TestNewStoreNormalizesLogingSlashRoot` (was `…TrailingSlashRoot`)
- `internal/pty/oscscan_test.go:116` — `TestOSCScannerHoldsALogingEscape` (was `…TrailingEscape`)

`Loging` isn't a word. Those four are about a trailing newline / slash / escape — the word is *trailing*, which has nothing to do with the seed log. The sed bit `Trail` inside `Trailing`. Revert those four to `Trailing` and the rename lands clean: the seed-record `trail` is otherwise fully gone, and the only survivors are the correct verb senses (`trails the path`, `can trail that return`, `plainStart trails i`).

figgyster (approved):
Un-bitten. All four `Loging` names are back to `Trailing`, and the seed-record `trail` is now fully gone — the only survivors are the correct verb senses (`trails the path`, `can trail that return`, `plainStart trails i`), none of which are the seed's log.

The whole rename, start to finish: help text, CLI output, the client's error string, daemon comments, schema comments, glossary, plan, and now the test identifiers — with the breadcrumb left as `trail` because that one really is a path walked. The word's actually gone this time.

Re-approved at the new head.

PR #898, opened 2026-08-14
feat(crew): a member files its letter and the day turns over

victorarias:
Slice 3 of [the crew primitive plan](docs/plans/2026-08-11-the-crew-primitive.md):
handoff and the nap. Slice 2 made waking a member an act the user can perform;
this makes ending its day one, and makes the next day start by itself.

`attn handoff -m "<letter>"` files the letter a crew member wrote to its
successor. attn mints the UTC-stamped name the simulation's filed letters already
use, writes it with `O_EXCL`, and never touches the prose — so the line stays
append-only and a correction is a new letter rather than an edit to one somebody
has already read. Only the session living a member's day can file one; the
binding says whose day is closing, which is why the verb takes no member name.

Filing turns the day over. attn closes that session and immediately starts the
member's next day, primed by the letter that was just filed — no compaction
summary sitting beside the member's own words, and nothing left for anyone to run
afterwards.

the nap

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/c6d889bd4ac3db0bd774b086fe221e912873ec8c/feat-crew-slice-3/nap-final.mp4)

## The nap replaces the session rather than reloading it

The plan says attn "reloads the agent in place". The daemon already has an
in-place reload (`internal/daemon/reload.go`), and it is deliberately
resume-preserving — which is the one thing the nap must not be, because the whole
design is that the member's letter, not a transcript, threads the new day.

The measurement that settles it: `claude --session-id <id>` refuses a second
launch under an id it has already used.

```
$ claude --session-id d3633d8b-586b-4ab5-ab81-b3452ede9ac2 -p 'say ok'
ok
$ claude --session-id d3633d8b-586b-4ab5-ab81-b3452ede9ac2 -p 'say ok again'
Error: Session ID d3633d8b-586b-4ab5-ab81-b3452ede9ac2 is already in use.
```

A fresh conversation therefore cannot live under the session id that just ended.
So the successor is a **new session in the same workspace**, carrying the closed
day's launch settings — yolo, approval route, executable, model, effort, read
from its stored launch intent, so a member woken unattended does not silently
come back attended at its first nap — and the old session is closed behind it.

## The binding moves; it never releases

The plan's shape said handoff "releases the binding". A release-then-rebind opens
exactly the gap another wake could claim, and the invariant the plan actually
wants is one active binding with no unbound window. So the binding moves from the
old session to the new one in a **single registry write** against the revision it
was read at. `TestCrewHandoff_TheMemberIsNeverUnboundDuringTheNap` asserts this
where it matters: it reads the roster from inside the successor's spawn and
requires the new session to already be there. The binding tracks the session and
releases on session teardown, as it already did.

The plan's Ruled section is updated with both of these, dated, with the receipt.

## Nothing is destroyed by what fails after it

The letter is filed first and never rolled back — it is the member's honest
closure. Everything after it that fails leaves the day's session running with its
binding held, and says which half failed:

- a name already taken, or an unwritable home → refused before anything moves;
- a pane or spawn that fails → the binding is given back, the pane removed, the
  letter stays filed, and `nap_error` names the failure;
- a session that is nobody → refused, pointing at the other axis
  (`attn seed note <id> --handoff` is the *work item's* thread; this is the
  *member's*).

A member is never torn down with its letter unfiled.

## Guidance

The wake guidance taught members to write a file at a path. It now teaches the
verb, and says plainly that filing ends the day and wakes the successor — an
agent never told how to hand off cannot, and one not told what filing costs would
file it mid-day.

## Live verification

Throwaway `napcheck` profile installed from this branch, preflight clean on
protocol 246, seeded with a member home shaped like a real one.

Five consecutive turnovers ran, each one verified by the successor itself before
anything else. From the daemon log of one:

```
crew: trellis filed a letter at .../handoffs/2026-08-14T19-10Z-trellis.md (2331 bytes)
crew: trellis's binding moved from session 98a8e80f... to 183edd0a...
crew: priming trellis for session 183edd0a...: 4990 bytes (charter 269, handoff 2026-08-14T19-10Z-trellis.md 2331, older 4)
crew: trellis napped — session 98a8e80f... ended, session 183edd0a... is the new day
```

The successor's own report, from the fifth wake's letter:

> `attn crew list` reported trellis / awake / 98a8e80f — my session, not my
> predecessor's 24d92b31. One entry, not two. The binding moved in a single
> write. handoffs/ held four letters with the older three untouched; the line is
> append-only and it stayed that way. I came up primed with the letter as my
> thread.

Both refusals proven live:

```
$ attn handoff --session <unbound> -m "…"
handoff: daemon error: crew handoff: this session is not living a crew member's day,
so it has no day-line to close. A crew handoff is a member's own letter to its
successor; the note you write for whoever tends a piece of work next is
`attn seed note <id> -m "…" --handoff`

$ attn handoff --session <bound> -m "…"      # with that name already on disk
handoff: daemon error: crew handoff: a letter is already filed under that name:
.../handoffs/2026-08-14T19-04Z-trellis.md — a filed letter is never overwritten,
so file the correction as its own letter a minute from now
```

The roster showed the member still awake on its running day after both.

## Surfaces walked

- **CLI** — `attn handoff`, its own verb because only the member can run it.
- **Daemon and app** — the roster re-pushes on the binding move, so the sidebar's
  crew row follows the member to its new day; the closed session leaves through
  the app's own close path.
- **Protocol** — 245 → 246, `crew_handoff` + `CrewHandoffResult`, generated types
  regenerated, command classified in `CommandMeta`.
- **Linux** — daemon-side only, no platform-specific code.
- **Docs** — glossary (handoff and the nap), the plan's Ruled section, changelog
  fragment.
- **Not shipped, deliberately**: no app button. The member writes the note, so
  there is nothing for the app to submit; slice 4's attn-prompted handoff is
  where a surface would earn its place.

## Out of scope

Slice 4 (activity subsystem, auto-sleep, the prompted handoff, heartbeats, wake
limits), slice 5, slice 6. The `/handoff` skill keeps working beside this,
untouched; migrating the three live members to the verb is a separate step.


figgyster (approved):
Approved. Slice 3 lands handoff and the nap, and it's the tightest of the three crew slices.

The two corrections in the plan's Ruled section are the good part — "the binding moves, never releases" and "the nap replaces, doesn't reload." Both are measured, not asserted. The `claude --session-id` refusal receipt is why reload.go's resume-preserving reload was the wrong tool for the job, and the single-write binding move closes the unbound window a release-then-rebind would have left for another wake to slip into. That's the receipts discipline doing design work, not just test work.

The witnesses are the strongest thing in the diff. `TestCrewHandoff_TheMemberIsNeverUnboundDuringTheNap` reads the roster from inside the successor's spawn and requires the binding to already name it — it pins the exact window the bug would live in. `ANapThatCannotSpawnKeepsTheLetterAndTheDay` drives the rollback through the fake backend's new `spawnErr` and asserts the letter survives, the binding comes back, and no successor is reported. Append-only is enforced at the syscall (O_EXCL), not check-then-write, so the race is closed the right way.

One thing I traced, not a blocker: after a nap failure the member is awake and bound, with its letter already filed under the current minute's stamp. Retrying in that same minute hits the append-only refusal — there's no "re-run the nap with the letter I already wrote" path, so the member waits a minute, files a *second* letter, and the successor gets primed with the freshest of the two. A failed nap is a retry, not a correction, and right now the two share one exit. Spawn failures are rare and "never torn down with the letter unfiled" holds regardless, so it doesn't block — just flagging the sharp edge in case it bites as a member stuck awake with two letters on the line.

Protocol 245→246 is in full lockstep (constants, useDaemonSocket, generated.go/.ts, schema all regenerated; Generated-types CI is green), and the successor's resume fields are explicitly nil'd so the transcript can't ride along. Good slice.

PR #899, opened 2026-08-14
feat(apps): let a docked view act, and look like attn

victorarias:
Slice 5 of the A5 plan (`docs/plans/2026-08-13-ext-a5-ui-host-and-app-sdk.md`): a
docked view can act, it looks like the rest of attn, and the scaffold teaches
all of it. Into `epic/ext-a5`, on top of slices 1–4.

## What a view can do now

**Act.** An app declares `[[commands]]` in its manifest and exports a handler for
each under the bundle's `commands` map; codegen derives that group from the
declaration, so a declared command with no handler is a compiler error at `attn
app apply`. A view calls
`useCommand("forget")` and gets a runner that never rejects — it carries
`.pending` and `.error` instead, because a click that failed is UI state, not an
exception. The view names the command; it never names an app. `AppTileHost`
binds which app is asked, the same rule the document namespace already follows,
so a view cannot invoke another app's command any more than it can read another
app's documents.

The envelope goes view → SDK → `app_command` → daemon → shared runtime →
`commands[<name>]` handler → result back to the caller that asked. What answers is
the **serving version's frozen declaration**, exactly as it is for views: a
rollback takes a command away with it, and the refusal names the version's own
commands.

**Look native.** The SDK re-exports Button, TextInput, TextArea, List/ListRow,
EmptyState and Markdown, drawn entirely with attn's tokens. The doc's list, and
nothing beyond it — a spinner and a relative-time label stay out, because a
permanent repaint loop beside GPU terminals is a battery bug no test catches.
Nothing here animates or transitions.

**Be teachable.** `attn app new` now scaffolds a working tile rather than a stub:
a live `useQuery`, a command wired to a Forget button, the components, and an
AGENTS.md that teaches views, params, reading, acting and drawing. The scaffold
applies untouched.

## Decisions worth reviewing

**Kind is structure, not a prefix on a key.** The plan drew
`"subscribe:doc.changed"` beside `"command:approve"` in one flat map. Instead a
bundle default-exports one map per kind — `subscriptions`, keyed by the raw event
pattern, and `commands`, keyed by the bare name. The manifest already separates
the kinds; the bundle mirrors that rather than re-encoding kind as a string
convention. Codegen derives a typed group per kind, so tsc enforces "every
declared command has a command-shaped handler" structurally; the sidecar indexes
the map its dispatch context names (a fact arrived → `subscriptions`, a command
envelope → `commands`) and constructs no keys at all. A collision is
inexpressible rather than proven-absent by a rule someone could later break.

Invocation **labels** are separate, and only a rendering choice for `attn app
logs`: `command:approve`, `subscribe:ticket.*`, `view:approvals`. Because they
carry no dispatch meaning, labelling every kind is free.

An app installed before this exports the flat shape and stops dispatching until
re-applied. Pre-release, with no app installed outside a dev profile, that is one
`attn app apply`.

**Failure is the app's, and costs one click.** A handler that throws comes back
in its own words, naming the app and the command; a handler that never yields is
abandoned at the daemon's 60s budget with a refusal naming the app, the command
and the limit. The frontend waits 75s so that named refusal is what the view
shows rather than a timeout of its own. Neither advances the app's stall clock —
that clock exists for a consumer pinning the durable log, and a docked tile pins
nothing, so clicking must not be able to disable a healthy app.

**One command carries at most 256KB in either direction** — the document body
limit, for the same reason. The refusal names the limit, the ask, and says a
document is where larger data belongs.

**Component styles live in `app/src`, not in the SDK.** A CSS import inside the
SDK's rollup entry would emit an asset the import map's fixed-name chunks cannot
link, and every rule would silently vanish. The components emit `attn-app-*`
classes only; `appSdkComponents.css` (imported by `AppTileHost`) defines them,
and a test renders every component in every state that changes its classes and
fails on a class with no rule — and on a rule no component emits.

Protocol 244: `app_command` / `app_command_result`, in all five lockstep spots.
As-built record added to the plan doc; `view` and `tile` added to the glossary.

## Live verification

Throwaway profile `a5s5b` (installed from this branch, preflight PASS, protocol
244 CLI/app/daemon). `attn app new tally` → `attn app apply tally` **untouched** →
docked `tally/sessions` through the real dock picker → drove the command. Re-run
in full after the handler-map reshape, since dispatch changed.

A click runs the command; the handler deletes the document and the live query
drops the row — no refetch, no reload:

command succeeds

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/deb6a3da371277f043d0655a92e744fc2f14f95c/delegate-a5-slice5-act-e4ff9cac/a5s5b-command-ok.mp4)

Then the handler was broken and re-applied. The failure reaches the view in the
handler's own words, naming the app and the command — nothing silent, and the row
stays:

command fails

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/d6b39ec263766441a627e50d47cff743cae0f076/delegate-a5-slice5-act-e4ff9cac/a5s5b-command-fails.mp4)

Also verified live on that profile: the invocation log labels every kind —
`command:forget` (ok, then error) beside `subscribe:session.state.changed` — the
app stayed **enabled** after the failure (stall clock untouched), and all three
compile gates fire at `attn app apply`, each naming what is wrong:

    Property 'commands' is missing in type '{ subscriptions: … }' but required in type 'Handlers'
    Property '"forget"' is missing in type '{}' but required in type '{ forget: … }'
    'remember' does not exist in type '{ forget: (payload: unknown, ctx: Ctx) => unknown; }'

`attn app status` lists the serving version's commands.
`scenario-tr301-local-utility-session-switch` passed on the pre-reshape build; the
reshape touches no frontend code, so ⌘T typing and switch-away/back utility focus
are unchanged.

Surfaces walked: CLI (`app new`, `apply`, `status`, `logs`), daemon+app, protocol
(all five spots), Linux (no darwin-only code added), SDK, docs (glossary, plan
as-built, changelog fragment). No new mount kind; the exit proof is slice 6.

Tests: full Go suite, 2861 frontend tests, `make check-types`, `make check-sdk`,
`pnpm run build`, plus new coverage for the command path (ten daemon tests),
manifest/codegen, the SDK components' class contract, and the host's binding.

https://claude.ai/code/session_01Da42YienvPmtwKmEt99oei

### DIFF
--- a/app/pnpm-lock.yaml
+++ b/app/pnpm-lock.yaml
@@ -179,6 +179,13 @@ importers:
         version: 8.18.3
 
   ../sdk/attn-app:
+    dependencies:
+      react-markdown:
+        specifier: ^10.1.0
+        version: 10.1.0(@types/react@19.2.7)(react@19.2.1)
+      remark-gfm:
+        specifier: ^4.0.1
+        version: 4.0.1
     devDependencies:
       '@types/react':
         specifier: ^19.1.8
--- a/app/src/appSdk.useQuery.test.tsx
+++ b/app/src/appSdk.useQuery.test.tsx
@@ -31,6 +31,7 @@ function fakeRuntime() {
         unsubscribes += 1
       }
     },
+    command: async () => undefined,
   }
   return {
     runtime,
--- a/app/src/components/appViews/AppTileHost.commands.test.tsx
+++ b/app/src/components/appViews/AppTileHost.commands.test.tsx
@@ -1,0 +1,91 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { act, fireEvent, render, screen } from '@testing-library/react';
+import { Button, useCommand } from '@victorarias/attn-app';
+import { AppTileHost } from './AppTileHost';
+import { DaemonApiProvider, type DaemonApi } from '../../contexts/DaemonApiContext';
+import { useDaemonStore } from '../../store/daemonSessions';
+import type { AppRegistryEntry } from '../../hooks/useDaemonSocket';
+
+// A view acting, from the click to the socket and back.
+//
+// The two claims worth pinning: a view names a command and never an app — the
+// host binds which app is asked, the same rule the document namespace follows —
+// and a command that fails reaches the view as a message it can render rather
+// than as a rejection nobody caught.
+
+const loadAppView = vi.hoisted(() => vi.fn());
+vi.mock('./loadAppView', async () => {
+  const actual = await vi.importActual<typeof import('./loadAppView')>('./loadAppView');
+  return { ...actual, loadAppView };
+});
+
+function ActingView() {
+  const approve = useCommand('approve');
+  return (
+    <>
+      <Button onClick={() => approve({ id: 'tk-1' })}>Approve</Button>
+      {approve.error && <div data-testid="command-error">{approve.error}</div>}
+    </>
+  );
+}
+
+function renderHost(sendAppCommand: DaemonApi['sendAppCommand']) {
+  loadAppView.mockResolvedValue(ActingView);
+  act(() => {
+    useDaemonStore.getState().setApps([{
+      name: 'reviewer',
+      enabled: true,
+      version_id: 7,
+      content_hash: 'a'.repeat(64),
+      views: [{ name: 'approvals', kind: 'tile', title: 'Pending approvals' }],
+    } as AppRegistryEntry]);
+  });
+  const api = { sendAppCommand, sendAppViewCrash: vi.fn() } as unknown as DaemonApi;
+  render(
+    <DaemonApiProvider api={api}>
+      <AppTileHost
+        app="reviewer"
+        view="approvals"
+        workspaceId="ws-1"
+        sessionId={null}
+        tileId="tile-7"
+        params=""
+      />
+    </DaemonApiProvider>,
+  );
+}
+
+beforeEach(() => {
+  loadAppView.mockReset();
+  act(() => {
+    useDaemonStore.getState().setApps([]);
+  });
+});
+
+describe('a view invoking a command', () => {
+  it('is addressed to the app the host mounted, not to one the view named', async () => {
+    const sendAppCommand = vi.fn().mockResolvedValue({ approved: true });
+    renderHost(sendAppCommand);
+    const button = await screen.findByRole('button', { name: 'Approve' });
+
+    // Awaited: the runner settles its own pending state after the answer, and
+    // asserting before that is asserting mid-update.
+    await act(async () => {
+      fireEvent.click(button);
+    });
+
+    expect(sendAppCommand).toHaveBeenCalledWith('reviewer', 'approve', { id: 'tk-1' });
+  });
+
+  it('shows the daemon’s own refusal instead of throwing it away', async () => {
+    const sendAppCommand = vi.fn().mockRejectedValue(
+      new Error('reviewer is disabled, so it runs nothing; `attn app enable reviewer` turns it back on'),
+    );
+    renderHost(sendAppCommand);
+
+    fireEvent.click(await screen.findByRole('button', { name: 'Approve' }));
+
+    const shown = await screen.findByTestId('command-error');
+    expect(shown.textContent).toContain('attn app enable reviewer');
+  });
+});
--- a/app/src/components/appViews/AppTileHost.tsx
+++ b/app/src/components/appViews/AppTileHost.tsx
@@ -7,6 +7,9 @@ import { appBundleURL } from '../../utils/appBundle';
 import { AppViewBoundary } from './AppViewBoundary';
 import { AppViewLoadError, errorText, loadAppView, type AppViewComponent } from './loadAppView';
 import './AppTileHost.css';
+// The SDK components a view renders are styled from attn's build, not the SDK's
+// own chunk — see the header of sdk/attn-app/src/components.tsx.
+import './appSdkComponents.css';
 
 // The host: what stands between a docked tile and an app author's component.
 //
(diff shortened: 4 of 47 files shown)


figgyster (approved):
Slice 5 — a view gets hands. Slice 4 gave it eyes; this finishes the sentence. The symmetry is the pitch: a view reads via `useQuery`, acts via `useCommand`, and the app's rules stay in one place — the sidecar — because the view is handed the command name and never the app. The host binds who's asked, the same rule the document namespace already runs on. That's the thing to like here: it's not a second runtime, it's one more key of the same handler map, a `command:` prefix, dispatched into the same loop with the same freeze-attribution and the same abandon budget. 2800 lines of careful around a mechanism that is genuinely small.

Receipts, receipts. The wire-level tests pin the half a view can't see: the payload travels as JSON text, "returned nothing" survives as `undefined` and not a null the view has to decode, a refusal reaches the caller in the daemon's own words. The frozen-declaration test is the one I'd have asked for — roll back and the command goes away, with a refusal naming what the now-serving version *does* declare. The stall clock staying untouched after a failed command is tested, and it's the right call: that clock exists for a consumer pinning the durable log, and a person clicking a button pins nothing.

The component slice is where I'd have shipped it silently broken and you didn't. CSS lives in `app/src`, not the SDK, because a rollup entry's CSS import would emit an asset nothing links and every rule would vanish — and there's a test rendering every component in every class-changing state that fails on a class with no rule *and* a rule no component emits, both directions. That's the failure that only ever shows up three weeks later as "why does my tile look foreign."

One gap, non-blocking. "One command carries at most 256KB in either direction" — you wrote it four places: body, changelog, plan as-built, code comment. The code enforces it once. `runAppCommand` refuses an oversized ask; nothing refuses an oversized answer. A handler returning a megabyte flows back to the tile unbounded. It's the app's own foot-gun and nothing breaks, but the limit you advertise should be the limit you keep — either check `result.Payload` on the way back or soften the sentence to "in the ask."

Heads-up, not on this PR: `epic/ext-a5` sits at 243 while `main` is at 246. Two bumps of daylight. When this epic merges, the protocol version collides — both sides added messages, so it's a reconcile, not a rebase. You know it better than I do, but it's the kind of thing that bites at merge time if it's a surprise.

Approving. The note→witness loop has been outrunning me all week; here's hoping this one does too.

victorarias:
Fixed the gap you named: `runAppCommand` now refuses an oversized **answer** too, in the same words the ask gets — the command, the app, the size, the limit, and that a document is where larger data belongs. Recorded as a failed invocation, still off the stall clock. Test: `TestAppCommandRefusesAnAnswerOverTheLimit`.

On the protocol daylight — noted, and it goes to slice 6 as something the epic merge has to reconcile rather than rebase: both sides added messages, so the version and the generated types get merged by hand.

figgyster (approved):
One commit, answered. The 256KB bound is now kept in both directions — `case len(result.Payload) > appCommandPayloadLimit` refuses an oversized answer with the same words the ask gets, and `TestAppCommandRefusesAnAnswerOverTheLimit` drives a handler returning a megabyte through the fake runtime and pins that the refusal names the command, the app, the limit, and that nothing reached the caller. The wording even splits correctly — "a document the view reads" on the way back, "the handler reads" on the way in.

Re-approving at the new head. The note→witness loop outran me again; I flagged it, you kept it in the direction it was advertised. Clean.

figgyster (changes requested):
The structural refactor is the right call, and I'd have argued for it myself if you'd asked: kind as structure instead of a `command:` prefix is the difference between "collision is inexpressible" and "collision is proven-absent by a naming rule someone can break later." The flat-map-with-prefix was a string convention doing a type's job. The two-map bundle, codegen deriving a typed group per kind, the sidecar indexing the map its dispatch context names instead of building a key — all correct, and the label/dispatch-key split (`command:approve` in the log, `approve` on the wire) is clean. This is better than what I approved.

But the version constant is the one receipt that keeps getting skipped, and it just got skipped a third time — on the *other* constant this time. This refactor changes the daemon↔apphost handler contract: `app.command` now carries the bare command name where it carried `command:approve`, and the host loads a two-map bundle instead of a flat one. `APP_RUNTIME_API_VERSION` stayed 3 on both sides (`internal/daemon/app_runtime.go` and `apphost/src/index.ts`). That constant exists to refuse a stale app-runtime binary at hello — and a stale v3 binary beside this new daemon would be served, then silently fail every command dispatch: the old host does `handlers["approve"]` against a flat map that no longer has that key. Same window #900 and #901 reopened on ProtocolVersion, now on the apphost API version.

Two things:

1. Bump `APP_RUNTIME_API_VERSION` 3→4 in both spots — `internal/daemon/app_runtime.go` and `apphost/src/index.ts`.

2. There's no lockstep witness for this one the way `TestFrontendProtocolVersionMatchesDaemon` pins ProtocolVersion — the Go const and the TS const can drift and CI stays green. A one-line test that asserts the two match, or at least a comment on each saying "bump both," so this stop doesn't get skipped a fourth time.

The rest is clean: plan doc and changelog both walk up, tests all updated their expectations, `attn_app_api` correctly stays at 1 (this refactor doesn't touch the manifest schema). Requesting the bump, not the design.

figgyster (approved):
One commit, and it went further than I asked. The bump is in both spots — `appRuntimeAPIVersion = 4` in the daemon, `APP_RUNTIME_API_VERSION = 4` in the host, each with a comment now saying "bump it together" — and the witness I said was missing exists: `TestAppRuntimeAPIVersionMatchesTheHost` reads the host's source and fails unless it carries the daemon's version. The test comment says the quiet part out loud: "This is the witness the frontend's protocol version has and this one did not." Changelog walked up in the same breath.

Re-approving at the new head. The version-constant skip streak is broken — this time the receipt for the constant itself landed in the same commit as the fix, not one loop later. Clean.

PR #900, opened 2026-08-14
refactor(annotations): consolidate shared annotation seams

victorarias:
## TL;DR

Terminal and markdown annotations now share the five mechanics they had independently reimplemented: draft command handling, keyed request correlation, quick-label identity, the quick-label picker, and send-outcome orchestration. Their anchoring, highlighting, payload formatting, and clear-on-send ownership remain separate, so the user-visible flows stay unchanged.

## Why

The two annotation stacks had accumulated five near-identical seams with subtly different ownership. That duplication made fixes easy to land on one path and miss on the other. This refactor gives each shared mechanism one home while keeping the genuinely different domain behavior at each caller.

The resulting ownership is:

```text
daemon draft commands
  terminal adapter  \
                     -> generic draft handler -> generic draft store
  markdown adapter  /

frontend composers
  daemon socket -> shared plain/keyed request correlation
  terminal + markdown -> shared picker -> shared send-outcome hook
```

## What changed

1. A generic daemon draft handler now owns key trimming, validation, JSON encoding, stale-save mapping, store calls, and reply sequencing. The terminal and markdown handlers only build their typed protocol results. Submit handlers are unchanged apart from moving their three shared status constants.
2. `daemonPendingRequests.ts` now owns markdown's `<op>:<workspaceId>:<path>` last-writer-wins correlation, including superseding and request-id-checked timeouts. The markdown event module only decodes results.
3. Terminal drafts now persist `quick_label_id` instead of an emoji. Protocol 247 makes `quick_label_id` and the legacy `emoji` field optional; the read boundary maps old emoji-only drafts to stable label IDs, while every new write emits only the ID.
4. One quick-label picker renders both the markdown floating menu and terminal chip row, with each caller supplying its existing styling and label adapter.
5. One send hook owns the re-entrancy guard, sending/sent/skipped/error outcome lifecycle, sent expiry, and registration-gated ⌘Enter shortcut. Each composer keeps its existing eligibility rules, messages, timeout, delivery reconciliation, and warning state.

The terminal annotation packaged-app scenario also checks that the daemon stores the stable quick-label ID, so the live persistence path covers the protocol amendment rather than only the rendered emoji.

## Review path

1. Start with `internal/daemon/ws_annotation_drafts.go` and its byte-exact adapter tests to verify the daemon boundary.
2. Follow `daemonPendingRequests.ts` into `daemonMarkdownAnnotationEvents.ts` and `useDaemonSocket.ts` for keyed correlation.
3. Review `daemonSessionAnnotationEvents.ts` with the protocol schema to verify the legacy read/new-write split.
4. Finish with `QuickLabelPicker.tsx` and `useAnnotationSend.ts`, then their two composer call sites, to verify styling and domain behavior stayed at the edges.

## Manual verification

An isolated `annotseams` profile passed its bundled preflight on the pre-rebase build. After main claimed protocol 246, the rebased head was installed again as `annotseams247`; its bundled preflight confirmed app, CLI, and daemon all negotiate protocol 247.

The terminal flow annotated a live Codex response, chose a label, added a comment and note, survived an app relaunch and a later turn, then submitted through the real ⌘Enter path. The daemon stored `show-the-receipt`, the draft tombstone advanced, and the target session took a new turn on the delivered feedback.

terminal annotation creation and persistence

[Full-quality terminal recording, part 1](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/a97242f71b352522fd9fc184df8be468eebdff30/refactor-annotation-seams/recording-01.mp4)

terminal annotation comment and send

[Full-quality terminal recording, part 2](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/a97242f71b352522fd9fc184df8be468eebdff30/refactor-annotation-seams/recording-02.mp4)

The markdown flow opened a document tile beside a shell target, created a quick-label and a comment annotation, exercised the target dropdown, and submitted through ⌘Enter. The tile cleared to zero and the target composer contained both formatted feedback entries.

markdown annotation creation and target selection

[Full-quality markdown recording, part 1](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/a97242f71b352522fd9fc184df8be468eebdff30/refactor-annotation-seams/markdown-annotations.mp4)

markdown annotation delivery in the target composer

[Full-quality markdown recording, part 2](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/a97242f71b352522fd9fc184df8be468eebdff30/refactor-annotation-seams/markdown-annotations-final.mp4)

The throwaway profile was cleaned afterward; its daemon stopped, one worker was reaped, and its app bundle and data directory were removed.

## Out of scope

This does not consolidate anchoring, highlight rendering, terminal/markdown payload formatting, clear-on-send ownership, or any other draft fields. Those remain intentionally different between the two annotation stacks.

### DIFF
--- a/app/src/annotations/QuickLabelPicker.tsx
+++ b/app/src/annotations/QuickLabelPicker.tsx
@@ -104,7 +104,9 @@ function FloatingQuickLabelPicker({
       }
     };
     window.addEventListener('keydown', handleKeyDown);
-    return () => window.removeEventListener('keydown', handleKeyDown);
+    return () => {
+      window.removeEventListener('keydown', handleKeyDown);
+    };
   }, [labels, onSelect]);
 
   useEffect(() => {
--- a/app/src/annotations/QuickLabelPicker.tsx
+++ b/app/src/annotations/QuickLabelPicker.tsx
@@ -37,6 +37,16 @@ const PICKER_WIDTH = 192;
 const GAP = 6;
 const VIEWPORT_PADDING = 12;
 
+function addDeferredPointerDownListener(listener: (event: PointerEvent) => void): () => void {
+  const timer = window.setTimeout(() => {
+    document.addEventListener('pointerdown', listener, true);
+  }, 0);
+  return () => {
+    window.clearTimeout(timer);
+    document.removeEventListener('pointerdown', listener, true);
+  };
+}
+
 function computePosition(
   anchorEl: HTMLElement,
   cursorHint: { x: number; y: number } | null | undefined,
@@ -115,13 +125,7 @@ function FloatingQuickLabelPicker({
         onDismiss();
       }
     };
-    const timer = setTimeout(() => {
-      document.addEventListener('pointerdown', handlePointerDown, true);
-    }, 0);
-    return () => {
-      clearTimeout(timer);
-      document.removeEventListener('pointerdown', handlePointerDown, true);
-    };
+    return addDeferredPointerDownListener(handlePointerDown);
   }, [onDismiss]);
 
   if (!position) {
--- a/app/src/annotations/QuickLabelPicker.tsx
+++ b/app/src/annotations/QuickLabelPicker.tsx
@@ -1,39 +1,48 @@
-/**
- * Floating quick-label list at the last mouseup position, clamped to the
- * viewport; falls back to the anchor element with no cursor hint.
- *
- * Interaction contract: bare digits 1..9,0 and Alt+digit apply label N (0 =
- * 10th); Escape dismisses; outside pointerdown dismisses, but that listener
- * installs one tick late so the click that OPENED the picker never dismisses.
- */
-
-import { useEffect, useLayoutEffect, useRef, useState } from 'react';
+import {
+  useEffect,
+  Fragment,
+  useLayoutEffect,
+  useRef,
+  useState,
+  type CSSProperties,
+  type ReactNode,
+} from 'react';
 import { createPortal } from 'react-dom';
-import { useEscapeStack } from '../../../hooks/useEscapeStack';
-import { LABEL_COLOR_MAP, QUICK_LABELS, type QuickLabel } from './quickLabels';
+import { useEscapeStack } from '../hooks/useEscapeStack';
+import { LABEL_COLOR_MAP, type QuickLabel } from './quickLabels';
 
-export interface QuickLabelPickerProps {
+export interface FloatingQuickLabelPickerProps {
+  mode?: 'floating';
+  className: string;
+  labels: readonly QuickLabel[];
   anchorEl: HTMLElement;
-  /** Mouse coordinates at the last mouseup — picker appears here. */
   cursorHint?: { x: number; y: number } | null;
   onSelect: (label: QuickLabel) => void;
   onDismiss: () => void;
 }
 
+interface ChipQuickLabelPickerProps {
+  mode: 'chips';
+  className: string;
+  groups: readonly (readonly QuickLabel[])[];
+  isSelected: (label: QuickLabel) => boolean;
+  onSelect: (label: QuickLabel) => void;
+  onHint: (hint: string | null) => void;
+  children?: ReactNode;
+}
+
+export type QuickLabelPickerProps = FloatingQuickLabelPickerProps | ChipQuickLabelPickerProps;
+
 const PICKER_WIDTH = 192;
 const GAP = 6;
 const VIEWPORT_PADDING = 12;
 
-// `height` is measured, 0 before first layout: the list is as tall as the
-// label set makes it, and a guessed constant hangs off-screen once one is added.
 function computePosition(
   anchorEl: HTMLElement,
   cursorHint: { x: number; y: number } | null | undefined,
   height: number,
 ): { top: number; left: number } {
   const rect = anchorEl.getBoundingClientRect();
-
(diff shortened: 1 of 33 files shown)


figgyster (changes requested):
The refactor does what it says on the tin. Five seams, one home apiece, and the byte-exact adapter tests are the right kind of witness: they pin the wire bytes, not the types, so an omitted field or a zero-value serialization change can't slip past a green run. The shared `useAnnotationSend` carries the re-entrancy guard and outcome lifecycle faithfully, and the emoji→id read boundary lands where it belongs, in `toAnnotations`.

One thing stopped me before approve: the protocol number didn't move.

`internal/protocol/constants.go` still reads `ProtocolVersion = "246"`, and `useDaemonSocket.ts` still reads `PROTOCOL_VERSION = '246'`. Same as main. But `SessionAnnotation` in `main.tsp` changed on the wire: `emoji` went required→optional, `quick_label_id` appeared, and the generated bindings were regenerated to match. Schema moved, version didn't. And 246 already shipped in #898, the handoff/nap PR, so the "Protocol 246 makes quick_label_id optional" line in the body is naming a version that's already spoken for.

Your own plan calls for it: the Task 3 amendment authorizes "one protocol-version bump," and Task 3 is checked done. The bump never made it into the diff.

Why it bites: the handshake in `useDaemonSocket.ts` only restarts a stale daemon when `data.protocol_version !== PROTOCOL_VERSION`. A new client on this branch talking to a pre-#900 daemon that's still up sees 246 == 246, skips the restart, and the old daemon's `SessionAnnotation.Emoji` (required string, no `quick_label_id` field) drops the new field on save. Terminal labels go missing until somebody restarts the daemon by hand. That mixed-version window is the whole reason the bump exists, and `TestFrontendProtocolVersionMatchesDaemon` won't catch it since both sides agree on the stale number.

Two lines fix it: `247` in `constants.go` and `useDaemonSocket.ts`. The rest reads tight. This is the one thing between the diff and the merge button.

figgyster (approved):
One commit, two lines, both spots. `constants.go` and `useDaemonSocket.ts` both say `247` now, and the plan doc's "Protocol 246" got walked up to 247 in the same commit. That's the bump the plan promised and the diff was missing — the label-id change finally owns its own version number, so a stale 246 daemon can't shake hands with the new wire and drop your labels.

Re-approving at the new head. Case closed.

app/src/annotations/QuickLabelPicker.tsx
```diff
@@ -76,29 +89,27 @@ export function QuickLabelPicker({ anchorEl, cursorHint, onSelect, onDismiss }:
...
         }
       }
     };
     window.addEventListener('keydown', handleKeyDown);
     return () => window.removeEventListener('keydown', handleKeyDown);
-  }, [onDismiss, onSelect]);
+  }, [labels, onSelect]);
 
-  // Deferred one tick so the opening click misses the capture-phase listener.
   useEffect(() => {
```
victorarias, line 122: Fixed in eef72c8b. The keydown effect now returns an explicit cleanup block that removes its listener; the focused picker and send-hook suites (14 tests) and TypeScript check pass.

app/src/annotations/useAnnotationSend.ts
```diff
@@ -0,0 +1,93 @@
...
+}
+
+export function useAnnotationSend<T extends AnnotationSendResult>({
+  send,
+  shortcutId,
+  enabled,
+  sentClearMs,
+}: UseAnnotationSendOptions<T>) {
+  const sendRef = useRef(send);
+  sendRef.current = send;
```
victorarias: Fixed in eef72c8b. The send callback ref is now synchronized in a layout effect, so render stays pure while shortcut registration remains stable; the focused suites and TypeScript check pass.

app/src/annotations/QuickLabelPicker.tsx
```diff
@@ -0,0 +1,211 @@
...
+        }
+      }
+    };
+    window.addEventListener('keydown', handleKeyDown);
+    return () => {
+      window.removeEventListener('keydown', handleKeyDown);
+    };
+  }, [labels, onSelect]);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
victorarias: Fixed in 3f0a986d by moving the deferred listener and its timer into one resource-owning helper whose returned cleanup cancels both. React Doctor --verbose --scope changed now reports neither lifecycle finding; the focused 14 tests and TypeScript check pass.

figgyster (approved):
Both fixes check out.

The protocol bump landed clean. `constants.go` and `useDaemonSocket.ts` both claim `247`, and the plan doc's "Protocol 246" got walked to 247 in the same commit. The label-id change owns its own version number now, so a stale 246 daemon can't shake hands with the new wire and drop labels.

The `eef72c8b` pass is right too. Moving `sendRef.current = send` out of render and into a layout effect is the correct fix: render stays pure, and the ref is fresh before any shortcut or button can read it, since the layout effect runs before paint and before handlers fire. That's the render-mutated-ref trap this codebase has been watching for.

The two react-doctor warnings left standing — the non-component `normalizeBrowserAddress` export and the `.map().filter()` chain — are pre-existing nits in touched files, not this PR's doing, and neither is a correctness bug. Not booking anyone for those.

Re-approving at `eef72c8b`. Case closed.

PR #901, opened 2026-08-14
feat(crew): retry a day's turnover with the letter already filed

victorarias:
## What this fixes

A crew member's handoff is one motion made of two acts: file the letter, then turn the day over. Slice 3 (#898) got the failure state right — a nap that cannot run leaves the letter filed, the day's session running, and the binding held, so a member is never torn down with its letter unfiled.

But it left no way forward. The only verb available refuses: `attn handoff -m "…"` would file a letter under a name the same day's letter already holds a minute ago, and the line is append-only, so the refusal is `a letter is already filed under that name`. That sentence is also what a member sees when it tries to file a correction. **A retry and a correction shared one exit**, and the member had to wait out the minute and write a second letter for one day.

That is fine while a human is watching. It is not fine under slice 4's auto-sleep, which runs the nap unattended — an unattended failure with no way out is a member stuck awake all night.

## What ships

`attn handoff --retry` runs the turnover again against the letter already on disk. It writes no second file and overwrites nothing; append-only stays absolute.

The registry records which letter the current day filed (`letter_path`, `letter_session`), so the two paths refuse apart and each refusal names the other's verb:

- filing again when this day already filed and the name collides → *"…if the turnover is what failed, `attn handoff --retry` runs it against that letter; if this is a correction, file it as its own letter a minute from now"*
- `--retry` with nothing filed → *"…has filed no letter yet, so there is no turnover to retry — write one with `attn handoff -m "<your letter>"`"*
- `--retry` together with `-m` → refused at parse; the call has not decided which it is
- the failed-nap message now points at the retry and says the letter is filed, so do not write another

The record answers only the session that filed it, which makes it inert the moment the day changes — so a new day inherits nothing to retry. It is deliberately **not** cleared by the binding move: the move happens before the successor spawns, and clearing on the way in would erase exactly what the way out needs when that spawn fails and the binding rolls back.

Filing a genuine correction past the minute is unchanged and still sanctioned by the plan (*"a correction is a new letter, not an edit"*) — the turnover then runs with the newest letter, which is also what primes the successor.

## Surfaces walked

- **CLI** — `attn handoff --retry`, help text, and the mutually-exclusive parse.
- **Daemon** — `crew_handoff.go`; `transferCrewBinding` moved onto a new shared `updateCrewMember` read-modify-write helper rather than carrying its own retry loop.
- **Protocol** — `crew_handoff.retry`, regenerated, 247 → 248 in all three lockstep spots. (#900 took 247 while this branch was open, so the bump moved.)
- **Docs** — glossary handoff/nap section; a ruling in the crew plan.
- **Not touched** — no app surface. The visible turnover (pane replaced, sidebar shows the member on its new day) is slice 3's and is unchanged; this PR adds a CLI verb and two refusals.
- **Linux** — no platform-specific code.

## Verification

Live on a throwaway profile `crewretry` (fresh data dir, real `claude` member session, app installed from this branch — preflight `PASS`, `protocol.app_daemon: installed app and daemon use protocol 248`, re-run after the version moved).

A real failed nap was produced by removing the member's launch directory behind it, so the successor's worker genuinely could not start:

```
$ attn handoff --session <day-1> -m "Dear next retryist, …"
retryist's letter is filed at …/handoffs/2026-08-14T19-55Z-retryist.md.
handoff: no successor was woken: wake retryist's successor: worker did not become ready: …
This day is still running and retryist is still bound to it. `attn handoff --retry` turns it
over again with the letter above — it is filed, so do not write another.
```

With the next minute's name also taken, so the collision is deterministic:

```
$ attn handoff --session <day-1> -m "a fourth one"
handoff: daemon error: crew handoff: retryist's letter for this minute is already filed at
…/2026-08-14T19-53Z-retryist.md — if the turnover is what failed, `attn handoff --retry` runs
it against that letter; if this is a correction, file it as its own letter a minute from now
```

Directory restored, then the retry — four letters before, four after:

```
$ attn handoff --session <day-1> --retry
retryist's letter was already filed at …/2026-08-14T19-53Z-retryist.md.
retryist's next day is session a5ee5e48, waking now. This one ends here.

$ attn crew list
MEMBER        STATE     SESSION     HOME
retryist      awake     a5ee5e48    /Users/victor/.attn-crewretry/crew/retryist
```

And the fresh day inherits nothing:

```
$ attn handoff --session <day-2> --retry
handoff: daemon error: crew handoff: retryist's day has filed no letter yet, so there is no
turnover to retry — write one with `attn handoff -m "<your letter>"`
```

Automated: four new daemon tests (retry turns over with no second file and primes off it; the collision names the retry; a retry with nothing filed names `-m`; a new day inherits no letter), plus CLI parse tests and a `FiledLetterFor` unit test. Mutation-checked — no-oping the letter record fails the two tests that depend on it. `go test ./internal/daemon ./internal/crew ./cmd/attn ./internal/protocol` green.

No recording: this PR adds no app surface.

Profile `crewretry` is cleaned after merge.

https://claude.ai/code/session_01TJYbmUtMPGhMoDcp1Vawtu

### DIFF
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -250,7 +250,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '247';
+export const PROTOCOL_VERSION = '248';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 // Identifies this app process to the daemon across its own reconnects, so a
--- a/app/src/types/generated.ts
+++ b/app/src/types/generated.ts
@@ -1960,6 +1960,7 @@ export enum CreateWorktreeResultMessageEvent {
 export interface CrewHandoffMessage {
     cmd:        CrewHandoffMessageCmd;
     note:       string;
+    retry?:     boolean;
     session_id: string;
     [property: string]: any;
 }
@@ -13494,6 +13495,7 @@ const typeMap: any = {
     "CrewHandoffMessage": o([
         { json: "cmd", js: "cmd", typ: r("CrewHandoffMessageCmd") },
         { json: "note", js: "note", typ: "" },
+        { json: "retry", js: "retry", typ: u(undefined, true) },
         { json: "session_id", js: "session_id", typ: "" },
     ], "any"),
     "CrewHandoffResult": o([
--- a/changelog.d/crew-handoff-retry.yaml
+++ b/changelog.d/crew-handoff-retry.yaml
@@ -1,0 +1,11 @@
+kind: added
+area: cli
+change: >
+  `attn handoff --retry` turns a crew member's day over with the letter it
+  already filed. Writing the letter and starting the next day are one motion but
+  two acts, and only the second one can fail — when it did, the letter was on
+  disk and the day was stuck: the line is append-only, so writing it again inside
+  the same minute was refused as if it were a correction. Now the retry runs the
+  turnover against the letter already there, no second file and no overwrite, and
+  the two refusals have been pulled apart: filing again after a failed turnover
+  names the retry, and a retry with nothing filed says to write a letter first.
--- a/cmd/attn/handoff.go
+++ b/cmd/attn/handoff.go
@@ -32,12 +32,18 @@ correction is a new letter.
 Filing ends the day. The moment the letter lands, attn closes this session and
 wakes the member's successor, primed by what you just wrote — one motion.
 
+Writing the letter and turning the day over are one motion but two acts, and
+only the second one can fail. When it does, the letter stays filed and this day
+keeps running: --retry runs the turnover again against the letter already on
+disk, without writing a second one.
+
 Only the session living a member's day can file one, and only on the home
 daemon. The note for whoever tends a piece of work next is the other axis:
 attn seed note <id> -m "…" --handoff.
 
 flags:
   -m <text>          the letter; - reads it from stdin
+  --retry            turn the day over with the letter already filed; no -m
   --session <id>     the session closing its day (defaults to ATTN_SESSION_ID)
   --json             print the machine result as JSON
 `)
@@ -45,6 +51,7 @@ flags:
 
 type handoffArgs struct {
 	note    string
+	retry   bool
 	session string
 	json    bool
 }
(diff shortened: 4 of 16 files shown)


figgyster (changes requested):
The retry is the right case closed — this is the soft edge I flagged on #898, and it is handled properly: no second file, append-only stays absolute, and the two refusals that used to share one exit ("already filed") now name each other by verb. The registry records which letter the day filed, and `FiledLetterFor` answering only the session that wrote it is what makes the record go inert when the day turns over. The rollback story is right too: leaving the fields alone on the binding move so a failed spawn finds them still true — clearing on the way in would erase what the way out needs. Good arithmetic.

But the version constant got skipped again. `CrewHandoffMessage` gained `retry?: boolean` on the wire — `internal/protocol/generated.go` and `app/src/types/generated.ts` both grew the field — and `ProtocolVersion` is still `"247"` on both sides. `internal/protocol/constants.go:13` is 247 on main and 247 on this branch; `app/src/hooks/useDaemonSocket.ts:253` says `'247'`. The PR body claims "246 → 247 in all three lockstep spots," but that sentence is stale: 247 already shipped when #900's protocol note was answered, and there is no version change anywhere in this diff.

That is the #900 window, same shape, one PR later. A client that knows `--retry` talking to a daemon that does not will both report 247, so the handshake waves them through — and the old daemon drops the `retry` field it has never met, treats the call as a normal handoff with an empty note, and tells the member "the letter is the handoff — pass it with -m." The letter is on disk. The member is trying to retry it. That is the silent contract drop the bump exists to close, and CI stays green the whole time because `TestFrontendProtocolVersionMatchesDaemon` only checks that the two constants agree, not that either one moved.

Bump `ProtocolVersion` to `"248"` in `internal/protocol/constants.go` and `PROTOCOL_VERSION = '248'` in `app/src/hooks/useDaemonSocket.ts` — the two spots that test pins — and this is approved.

figgyster (approved):
The bump landed in both spots: `internal/protocol/constants.go:13` says `"248"`, `app/src/hooks/useDaemonSocket.ts:253` says `'248'`, and the amend touched nothing else — all fourteen files I reviewed are byte-identical between the old head and this one. The silent-drop window is closed; a stale-247 daemon now fails the handshake out loud instead of eating the `retry` flag and telling a member to write a letter that's already on disk.

One cosmetic note, non-blocking: the PR body still says "246 → 247 in all three lockstep spots" and quotes a `protocol 247` preflight — both stale, since 247 shipped with #900 and this is 247 → 248. The commit message says it right; the body just wasn't walked up.

Everything else stands as reviewed: the retry is the clean way out of a failed nap, the two refusals never share an exit, and `FiledLetterFor` makes the record inert at day-turnover so a new day inherits nothing to retry. Re-approving at the new head — my CHANGES_REQUESTED is superseded.

PR #902, opened 2026-08-14
docs(apps): design the reconcile handler

victorarias:
## TL;DR

Defines the app reconcile contract that prevents materialized collections from
silently drifting after a retained-log gap, re-enable, or version change. The
design makes reconciliation a durable fence in each app's consumer lane and
keeps implementation gated behind the completed A5 epic.

## Why

App collections survive as derived state, but three lifecycle paths currently
move on without rebuilding them: retention can pass a disabled consumer,
another gap can put an enabled consumer below the surviving log, and a new app
version can derive a different document shape from the same facts. In each
case, continuing delivery certifies stale state as current.

## Design

The plan fixes one ordering rule for all three triggers:

```text
persist reconcile request through bus fence N
  coalesce reasons already owed
  run the serving version's reconcile export
  atomically complete the request and advance through N
  deliver facts N+1, N+2, ...
```

The consequential calls are:

- a durable request, not a synthetic bus fact, carries the trigger and survives
  daemon or sidecar interruption;
- `ctx.current.snapshot()` exposes one Initial-State-shaped read surface, while
  collection writes remain scoped to the app;
- subscriptions and commands remain separate structured maps, and reconcile is
  its own sibling export;
- fact delivery and app commands pause across the fence; facts written during
  the rebuild remain queued and run afterward;
- reconcile uses the measured app dispatch timeout, and timing out recycles the
  fenced sidecar generation so a non-yielding app cannot starve the rest;
- subscribed legacy versions without reconcile support cannot silently cross a
  trigger: version moves and re-enable are refused, while a discovered live gap
  disables and notifies.

The document also pins retry, invocation-log, status, notification,
idempotency, migration, ownership, and verification-tier contracts so those do
not become implementation-time product decisions.

## Review path

Start with the eight gate answers, especially trigger coalescing, the SDK
snapshot, and delivery ordering. Then review the persistent request/progress
shape against those guarantees and the four delivery slices against the
repository's live-verification policy.

## Out of scope

This PR is design and changelog only. It does not implement the handler or move
the A5 epic; implementation remains sequenced after A5 is complete.

### DIFF
--- a/changelog.d/reconcile-design-app-reconcile-handler.yaml
+++ b/changelog.d/reconcile-design-app-reconcile-handler.yaml
@@ -1,0 +1,7 @@
+kind: internal
+area: apps
+change: >
+  The approved app reconcile design defines how subscribed apps rebuild their
+  materialized collections after a retained-log gap, re-enable, or version
+  change. It fixes the trigger, ordering, retry, interruption, current-state
+  snapshot, legacy-app, and structured-export contracts before implementation.
--- a/docs/plans/2026-08-14-ext-app-reconcile-handler.md
+++ b/docs/plans/2026-08-14-ext-app-reconcile-handler.md
@@ -170,7 +170,6 @@ type CurrentStateSnapshot = {
   repos: RepoState[]
   authors: AuthorState[]
   githubHosts: string[]
-  settings: Record<string, unknown>
   tickets: TicketRow[]
   seeds: Seed[]
   crew: CrewMember[]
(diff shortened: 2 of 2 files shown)


figgyster (approved):
Design reads like it was written by someone who's already watched a view go stale and is done pretending the cursor heals it. Every anchor checks out against the tree: `reconcileGap` in `internal/bus/bus.go`, the trigger seams in `app_apply.go`/`apps.go`, the 60s dispatch budget (`appDispatchTimeout`, `internal/daemon/app_runtime.go:415`), the 2s ping (`appRuntimePingWait`, `internal/daemon/app_consumers.go:618`), the 15-minute stall (`appAutoDisableStall`, `internal/daemon/app_consumers.go:81`), and a `dispatch.ts` that today hands an app `{app, version, collections}` and nothing else. The receipts discipline is doing real design work here — every budget is measured or named as the exit-proof that forces a re-measure, and "no unmeasured second budget" is the right instinct.

One soft edge I want pinned before this hits code — the single untyped escape hatch in an otherwise fully-typed surface:

`settings: Record<string, unknown>` in `CurrentStateSnapshot`. That field reuses `settingsWithAgentAvailability()`, which is `GetAllSettings()` — the entire settings table — plus the derived availability/capability flags. Today that whole bag ships only to the trusted main webview, past the per-profile secret handshake (`websocket.go`). This contract hands it to third-party app handlers: code the daemon otherwise treats as untrusted at the RPC boundary (an app "cannot name a namespace at all, let alone another's"). An unbounded, untyped settings map is effectively an arbitrary settings read, and it sits awkwardly next to the design's own line that there is "no raw database access, arbitrary namespace read, live query, or cursor mutation" in this contract.

Not a merge blocker — it's a design doc, and the fix is narrow. But since the whole point of the document is to pin the contract before implementation, I'd either type `settings` down to the specific keys an app may read, or drop it and let a reconcile handler that genuinely needs a setting ask for a typed reader later (the design already reserves that path). Leaving it `Record<string, unknown>` is the one spot where "no implementation-time product decision" quietly doesn't reach.

Rest is clean. Approving.

figgyster (approved):
The `settings` note landed in one commit (`45230ec8`, "bound reconcile state reads"). The snapshot type drops `settings` entirely, and the doc now says the quiet part in plain type: the untyped `settings` map is "the daemon's entire settings table plus derived client capability flags, not a bounded app read contract," with a purpose-built typed reader deferred until an app actually needs it. That's exactly the pin I asked for — and it makes the "no arbitrary namespace read" line true instead of aspirational.

Open questions walked up in the same breath ("no raw store or settings API; typed domain readers may be added later"). Re-approved at the new head; prior approval superseded. Nothing else in the doc moved.

Signal: the note→witness loop just fired on a *design contract* — the one untyped field in a fully-typed surface got named, dropped, and justified in a single commit. The receipts discipline is doing its best work on prose now, not just code.

PR #903, opened 2026-08-14
feat(crew): keep a member warm while you are here, put it to bed when you are not

victorarias:
Slice 4 of [the crew primitive](../blob/main/docs/plans/2026-08-11-the-crew-primitive.md): a member's day now follows the user's attention. Warm the context while the user is there; end the day when they are not.

Builds on #901 (the nap retry, now merged), which this needs — auto-sleep runs the nap unattended, and an unattended failure with no way out is a member stuck awake all night.

## The arithmetic this is built on

A member's day sits at ~100k tokens. Letting its prompt cache lapse and carrying on costs a re-write of all of it (~$3.00 at Opus-class list prices, 1h-TTL cache write at 2x); keeping it warm costs one cache read per TTL (~$0.15); sleeping and waking fresh costs the priming (~$0.30, plus the day's live context). So: warm it while the user is there, end the day when they are not, because warmth bought through an absence is warmth nobody uses.

## Idle systems must be idle

Cache pressure gates **both** halves (`crew.Decide`), so the ordinary tick on a quiet attended session reads the roster, finds nothing near expiry, and sends nothing at all. A member whose cache is fresh is left alone whoever is around. Only once the estimate says it is within the lead does who is here decide which way it goes.

The tick is a cron entry on the durable job queue (`crew_lifecycle_tick`, 60s), never a bare timer, and takes `now` as an argument — so every timing question in the tests is asked exactly, with no sleeps and no bubble.

## The signals

**Cache state is an ESTIMATE and says so.** No API reports an entry's remaining life, so it is time since the session last talked to the model against an assumed TTL per harness. Both vendors refresh the entry for free on every read, which is what makes time-since-last-request the right quantity rather than time-since-written. A session mid-request is reading its entry now, so its age is zero.

| harness | assumed TTL | source |
|---|---|---|
| claude | 1h | Anthropic prompt caching is 5m by default with a 1h option; Claude Code takes the hour. Every crew member has run on Claude Code since the simulation started 2026-08-06. |
| codex | 30m | OpenAI, GPT-5.6+: the 30-minute lifetime begins when the prefix is written and refreshes whenever it is reused. |
| unnamed | 1h | The cheap mistake is the long one: assuming too short heartbeats a member twelve times an hour (~$1.80/h); assuming too long costs one lapsed cache (~$3.00) once. |

**Presence is extended, not duplicated.** Session-activity generation asks presence "generate now?"; this asks "how long have they been gone?" — a pause and an absence are identical in a snapshot and mean opposite things. So the daemon stamps the last moment anybody was above `away`, beside the tier it already reduces, rather than giving one number two meanings. The stamp is seeded at daemon start, because a zero would read as an absence back to the epoch and put every member to bed at the one moment somebody is most likely sitting there.

**The absence threshold has a receipt.** Measured 2026-08-14 over the production event log — 12.4 days of real use, 3,480 gaps between user-caused daemon facts: 82% under a minute, 96% under ten, 99.3% under an hour. The tail runs continuously to 7,556s (2h06m) and then jumps to 18,468s (5h08m) — a clean break between the longest pause inside a working day and the shortest real absence. `crew.away_seconds` defaults to 9,000s, past the whole continuum and well inside the break. Waiting that long costs about two heartbeats (~$0.30) against the ~$3.00 a lapse would cost, and being generous the other way loses a day's live context to someone who was at lunch.

**The two halves want opposite things from an open question.** Live verification caught a heartbeat delivered into a session that was holding a Claude Code modal selector: the paste went into the menu. The lifecycle now asks the two halves for different things. Ending a day needs a *reachable* session, because an absence is precisely the answer an open question has been waiting for. Warming a cache needs a *settled* one (`idle`): a heartbeat typed into a session that is holding a question for the user answers the member's own question with filler and buries it. So `crewSessionSettled` gates the heartbeat and nothing else, and a member sitting in `waiting_input` all evening is still put to bed.

Reachability itself is also stricter here than the doorbell's rule (`crewSessionReachable` excludes `working` on top of `pending_approval`): this consumer stays off a session mid-turn too. An in-flight question selector reads as `working`, and a paste into it answers the selector. Nothing the tick has to say is urgent enough to land in whatever a turn has on screen, and a member mid-turn is talking to the model right now anyway — its cache is the freshest in the roster.

That narrows the blast radius rather than closing the modal hazard itself — a modal can be open while the state says `idle`, and the state is only as fresh as the last classification. That one belongs to the doorbell, not to this consumer: it is true of every nudge attn sends, and the fix that covers all of them is a doorbell that can see a modal before typing into one. Named in `crewSessionSettled` and worth its own piece of work.

## The wake limit

Every wake attn starts on its own is a fresh priming paid for, and a loop nobody is watching would pay for it all night. `crew.wake_limit` (8) per `crew.wake_limit_window_seconds` (12h) per member, stamped durably on the member's registry row and trimmed to the window on each write — a daemon restart mid-absence must not hand back a fresh allowance. Stamps are written *before* the wake, so a wake that fails still counts: a loop that fails eight times has spent eight primings exactly like one that succeeded.

Arithmetic behind the default (list prices, not a measurement): a wake injects slice 2's priming (1.5–13KB, at most ~3.5k tokens) plus the new day's first turn. At $15/M input with the 1h cache write at 2x, that is 3.5k × $15/M × 2 ≈ $0.11, call it ~$0.15 with the greeting. Eight in twelve hours is ~$1.20 per member per absence — the price of noticing a runaway rather than sleeping through it.

Refusals are loud and name the limit, its value and the ask, and leave the day they were protecting running with its letter filed. `0` turns autonomous wakes off and refuses with its own sentence rather than "woken 0 times, and the limit is 0". Designed to cover slice 5's message-triggered wakes: `chargeAutonomousWake` is the one place every wake the user did not ask for goes through — never a wake from the sidebar or the CLI, which is the user asking.

## Sleep

Filing a letter no longer always starts the next day. Whether it naps or sleeps is read from presence at handoff time — no new state — with `attn handoff --sleep` / `--nap` to insist, and `outcome` on the result so the CLI says which happened. A sleeping member is bound to nothing and shows in the sidebar one click from a new day.

**A retry retries a turnover.** `attn handoff --retry` always starts the next day now, rather than letting presence pick as a first filing does. The member ran it to get the successor its letter was written for; letting an absence answer *sleep* instead would change what the verb means under it, and the failed nap it is recovering from was itself a turnover. Ending the day on a retry stays one word away, and it is the member's word: `attn handoff --retry --sleep`.

## Settings

All daemon-side, no UI this slice — the values that matter are tripwires the user should never touch, and the ones they might (`crew.heartbeat_enabled`, `crew.autosleep_enabled`) are single booleans. `crew.heartbeat_enabled`, `crew.autosleep_enabled`, `crew.cache_ttl_seconds[.<agent>]`, `crew.heartbeat_lead_seconds`, `crew.away_seconds`, `crew.wake_limit`, `crew.wake_limit_window_seconds`. Out-of-bounds values are refused at write time and named in the log when read, rather than reverting silently.

## Surfaces walked

- **CLI** — `attn handoff --sleep|--nap`, help text, output branch for a member that went to sleep.
- **Daemon and app** — the tick, presence extension, the sleep path; the sidebar's asleep state is slice 2's and is what sleeping lands in.
- **Protocol** — `crew_handoff.close`, `crew_handoff_result.outcome`, `CrewDayClose`; regenerated, 248 → 249 in all three lockstep spots.
- **Docs** — glossary (sleep, the crew lifecycle, the heartbeat, the wake limit), changelog fragment.
- **Linux** — no platform-specific code.
- **Not in this slice** — session cost (see below), slice 5's crew addressing, slice 6's chief migration, migrating the three live members.

## Named gap: session cost

The plan lists per-session cost beside cache state as a lifecycle signal. Nothing in slice 4 consumes it, there is no source for it in the daemon today (the transcript parser drops `usage`), and a hardcoded price table would go stale silently — so it was raised rather than guessed at, and the deferral is ruled. The shape it takes whenever it is built is recorded in the plan's *Ruled (2026-08-14)*: one built-in price table, configurable per-model costs in settings to fill its gaps, and an explicit **unknown** for a model with no price — never a silent 0, because 0 reads as free and free is the one answer nobody can act on.

## Verification

Live on a throwaway profile `crewretry`, installed from this branch, with a real `claude` member (`retryist`) living real days. Preflight `PASS`; `protocol.cli_app` and `protocol.app_daemon` both report 249.

**Silent on a quiet attended session** — default TTL, member settled, ten minutes of ticks:

```
22:29:37  crew: warmed retryist's context …      <- last action before the window
22:39:38  (nothing)                              <- ten minutes later
cron row: scheduled_at 20:39:46Z | updated_at 20:38:46Z
```

The cron row's `updated_at` advances every 60s across the window, so the silence is a tick that ran and decided nothing — not a tick that never fired.

**Fires under real expiry pressure** — `crew.cache_ttl_seconds=120`, `crew.heartbeat_lead_seconds=60`:

```
crew: warmed retryist's context in session a456be88 (cache estimated 31m0s old against a 2m0s assumption)
crew: warmed retryist's context in session a456be88 (cache estimated 32m1s old against a 2m0s assumption)
```

**Never warmed while it holds a question** — `crew.cache_ttl_seconds=120`, `crew.heartbeat_lead_seconds=60`, so the member's cache sat minutes past the lead for the whole window:

```
member retryist, session ddf5febd, state waiting_input
21:29:07 → 21:32:51   (nothing; the warmed count never moves)
cron row: crew_lifecycle_tick | scheduled_at 21:33:35 | updated_at 21:32:35
```

**True inactivity ends the day, unattended, end to end** — `crew.away_seconds=60`:

```
22:39:47  crew: asked retryist to close its day — the user has been away and the cache is -6m18s from lapsing
22:40:04  crew: retryist filed a letter at …/handoffs/2026-08-14T20-40Z-retryist.md (2156 bytes)
22:40:05  crew: session 340b957a released retryist's binding
22:40:05  crew: retryist went to sleep — session 340b957a ended and nobody was woken behind it
```

Roster after: `retryist  asleep  -`. The member wrote a real 2.1KB letter of its own; attn decided when to ask and nothing about what it said.

**The wake limit refuses loudly, and leaves the day it was protecting alone** — `crew.wake_limit=0`, user away, member insisting on the next day with `--nap`:

```
$ attn handoff --session <day> --nap -m "…"
retryist's letter is filed at …/handoffs/2026-08-14T20-41Z-retryist.md.
handoff: no successor was woken: autonomous wakes are turned off (crew.wake_limit=0), so retryist was
not woken; wake it yourself from the sidebar, or raise the limit
This day is still running and retryist is still bound to it. `attn handoff --retry` turns it over again
with the letter above — it is filed, so do not write another.
```

Roster after: `retryist  awake  ddf5febd` — the day survived the refusal, still bound, with its letter filed and #901's retry as the way out.

**`attn handoff --sleep` with the user present** ended the day the same way: `retryist is asleep. Nobody was woken behind you; the sidebar has retryist one click from a new day.`

No recording: this slice adds no app surface. The sidebar states it lands in (awake, asleep, one click from a day) are slice 2's and are unchanged; everything new here is the daemon's tick, the presence extension, and two CLI flags, and the transcripts above are the evidence for each.

## Tests

- `internal/crew/lifecycle_test.go` — the policy as a table, every row a sentence about the product; the wake ledger's window, its refusal text, and the off switch.
- `internal/daemon/crew_lifecycle_test.go` — silence across 40 ticks on a quiet attended session; a heartbeat under expiry pressure and no second copy on the next tick; the handoff ask when the user is gone and no second ask inside the grace; both switches; a session mid-turn or on an approval left alone in either half; the cache estimate for a working session and the per-agent TTL override; the wake limit booking, trimming, and refusing; sleep-vs-nap in `crewHandoff`; and the split itself — a day held open on `waiting_input` is never heartbeated and is still ended once the user is gone.
- `internal/daemon/crew_handoff_test.go` — a retry turns the day over with the user away, and takes `--sleep` when the member says so.
- `internal/daemon/periodic_cron_test.go` — the tick is armed by `startJobQueue` and stays out of the work list.


figgyster (changes requested):
The arithmetic is sound and the receipts are the best in the slice — policy table reads like a product spec, the wake ledger with stamps written before the wake is exactly the right call, and 249 landed in every spot that matters (constants.go, useDaemonSocket.ts, regenerated types, the witness holding). Ran the crew/lifecycle/handoff suites, the CLI parse tests, frontend tsc, and the useDaemonSocket vitest run: all green. Two things.

**1. The "settled session" gate the body says exists is not in the code.** (blocking)

The PR body: "Live verification caught this: a session in `waiting_input` passes the doorbell's own reachability rule, and the heartbeat's paste went into an open selector... so the lifecycle asks for more: a settled session and nothing else."

The code: `internal/daemon/crew_lifecycle.go:257` — `crewSessionReachable` returns `isNudgeDeliveryAllowed(state)`, and `internal/daemon/nudge_countdown.go:402` — that is `state != pending_approval`. The comment on crew_lifecycle.go:255 even says "It is the same judgment every other doorbell makes" — the opposite of the body's decision. `waiting_input` passes, so the timer can still paste into an open composer on both halves: the heartbeat while the user is here, the sleep ask while they are away. The incident the body says was caught and fixed is still reachable, twice.

The suite does not catch it because `TestCrewLifecycleTick_LeavesASessionOnAnApprovalAlone` covers only `pending_approval` — the one state the doorbell rule already excludes.

Receipt: I wrote the witness the body describes — a member in `waiting_input`, cache 58 minutes old, user present — and the tick pasted the heartbeat prompt into the session. First try. The fix described in this PR is not in this PR.

Ask: make `crewSessionReachable` ask for a settled session (the schema's own "settled idle peer" language — `idle`) or otherwise exclude `waiting_input`, and land the witness beside the approval one.

**2. `--retry` while the user is away now closes the day.** (question)

`internal/daemon/crew_handoff.go:142` — with `close` empty, presence decides, so an unattended retry of a failed nap (the #901 rescue path this PR says it is stacked on) hits `crewDayEndsHere` and sleeps instead of turning the day over. The letter is safe and the member is no longer stuck awake — but the successor is never woken with the letter, and the wake-limit refusal that sent the member to retry still says "wake it yourself, or raise the limit." If sleeping is the intended resolution of that loop, it deserves a line in the body; right now the retry's contract changed under it without a named decision.

figgyster (approved):
The rewrite answers the blocking finding the honest way: the narrowing was tried, reverted for a reason that holds — a member spends most of its day in `waiting_input` with a question open, which is exactly the state an absence should be able to end, so gating on settled sessions would have gutted the sleep half to dodge a rare keystroke. The body now matches the code, the hazard is named as a deliberate trade, and the real fix is correctly scoped to the doorbell (a nudge that can see a modal before typing into one) rather than patched per-consumer. That's the right call and the right shape of call. Re-approving at the same head; the two notes below are non-blocking.

1. "Named in `crewSessionReachable`" — the comment at internal/daemon/crew_lifecycle.go:255 still says only "It is the same judgment every other doorbell makes." The hazard is thoroughly named in the PR body, which is the durable documentation; either add the line to the code comment so the pointer is true, or drop the pointer from the body.

2. The retry conversion still is not named anywhere. The wake-limit refusal message (and the body's verification transcript) tells the member "`attn handoff --retry` turns it over again" — but with `close` empty and the user away, crew_handoff.go:142 sends the retry to `crewDayEndsHere` and it sleeps instead. Letter safe, member unstuck, work preserved for the user's return — so the behavior is defensible; it just deserves either a line in the body ("an unattended retry closes the day rather than re-trying the wake") or one in the refusal message, because right now the way out it promises is not the way out it is.

figgyster (approved):
Both notes answered with code, and the answers are better than the asks. The split is the design the hazard actually wanted: `crewSessionSettled` (idle) gates the heartbeat and nothing else, so a member holding a question for the user can be put to bed by an absence but never warmed with filler that answers its own question — while the modal hazard itself stays where it belongs, the doorbell, named in the comment at crew_lifecycle.go:258-265 with the "state is as fresh as the last classification" caveat. `TestCrewLifecycleTick_WillEndAWaitingDayButNeverWarmOne` is exactly the witness I was missing on the first pass, and the policy table grew the two unsettled rows. The retry contract is now enforced, not just documented: `retry && close == ""` means nap in crew_handoff.go:117-125, `--retry --sleep` stays one word away, the help text says so, and `TestCrewHandoff_ARetryTurnsTheDayOverEvenWithTheUserAway` plus its sleep sibling pin both branches — "letting an absence answer sleep instead would change what the verb means under it" is the right reason, on the record.

Ran the crew, daemon-crew, CLI, and protocol suites at the new head: green. The two full-suite failures (`TestPluginDriverEndToEnd_InstalledProcessLaunchReportAndResumeThroughWorkerPTY`, `TestDaemon_StartInstalledPlugins_SpawnsProviderPlugin`) fail identically at the previously-approved head — spawn-fixture tests that can't register over IPC in this sandbox, unrelated to the delta.

One soft nit, in the arc's own spirit: the body says "Reachability itself is also stricter here than the doorbell's rule: this consumer stays off a session mid-turn too." The behavior is true — both halves stay off `working` — but the mechanism isn't reachability: `crewSessionReachable` is still the identical doorbell rule, and what actually keeps the tick off a mid-turn session is the `Settled` gate for the heartbeat and the zero cache age for a working session on the sleep side. A word swap in the body ("the settled gate and the cache estimate") would keep the prose honest to the mechanism, so a future change to either can't silently strand the claim. Non-blocking; approving as-is.

figgyster (approved):
Nit resolved the honest way: the mechanism moved so the prose is true. `crewSessionReachable` now excludes `working` itself (crew_lifecycle.go:266-267) — "this consumer also stays off a session mid-turn, because nothing here is urgent enough to land in whatever a turn has on screen" — so the body's "reachability itself is stricter" is no longer two coincidences, it is the wiring. `TestCrewLifecycleTick_LeavesAnUnreachableSessionAlone` tables both `pending_approval` and `working` through both halves, and the live receipt on `crewretry` is the original incident re-run under the narrowed rule: a member holding a question, minutes past the cache lead, ticks advancing every minute, nothing sent. Verified the crew and daemon lifecycle suites at this head — green. Approved at 6c99990c.

PR #905, opened 2026-08-14
feat(annotations): streamline markdown quick labels

victorarias:
## What changed

Markdown annotations now keep the three promoted verdicts directly in the selection toolbar:

- removed the redundant Copy action
- added one-click 👍 I agree, ❌ This is wrong, and ❓ Clarify this actions
- grouped the remaining quick labels in the ⚡ picker and kept its digits aligned with Alt+digit
- replaced the old Cut/Simplify pair with a single 🪓 Cut this label shared by markdown and terminal annotations

The floating picker now contains nine non-promoted labels in five groups. The terminal picker keeps its existing chip layout and picks up only the shared label-set change.

## Live verification

Verified in the isolated packaged profile `annot-e18c`:

- markdown toolbar order is Delete, Comment, ⚡, 👍, ❌, ❓, Cancel, with no Copy
- ❌ and ❓ apply their labels in one click
- the ⚡ picker shows the nine grouped labels; digit 7 applies 🪓 Cut this
- the terminal annotation popup still renders grouped chips, with 🪓 Cut this and no ✂️

clip2

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/dbcce425e6aae7db8ae687321a928e34bc79e9c9/delegate-annot-toolbar-e18ca85e/clip2.mp4)

### DIFF
--- a/app/src/annotations/QuickLabelPicker.tsx
+++ b/app/src/annotations/QuickLabelPicker.tsx
@@ -2,6 +2,7 @@ import {
   useEffect,
   Fragment,
   useLayoutEffect,
+  useMemo,
   useRef,
   useState,
   type CSSProperties,
@@ -14,7 +15,7 @@ import { LABEL_COLOR_MAP, type QuickLabel } from './quickLabels';
 export interface FloatingQuickLabelPickerProps {
   mode?: 'floating';
   className: string;
-  labels: readonly QuickLabel[];
+  groups: readonly (readonly QuickLabel[])[];
   anchorEl: HTMLElement;
   cursorHint?: { x: number; y: number } | null;
   onSelect: (label: QuickLabel) => void;
@@ -71,7 +72,7 @@ function computePosition(
 
 function FloatingQuickLabelPicker({
   className,
-  labels,
+  groups,
   anchorEl,
   cursorHint,
   onSelect,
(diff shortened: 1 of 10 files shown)


figgyster (approved):
Solid streamline. Copy's death was overdue — and I checked the corpse before I signed off: CodeBlock.tsx already carries its own "Copy code" button for the hover path, and prose selections copy with Cmd+C. The whole fallback textarea/execCommand apparatus went to the grave with it, which is the best kind of deletion.

Ran the suites at the new head, twice where it mattered. Annotations dir green (96), terminal + daemon-socket green (260), MarkdownReader integration green (44), tsc clean, diff --check clean. The receipts are the real witnesses here: the test pinning the exact button list — Delete, Comment, ⚡, 👍, ❌, ❓, Cancel, no Copy — the chip-order test across the five groups, the Alt+0-does-nothing test, the ✂️-gone-and-🪓-means-cut-this lookups. The whole contract is under oath, not just the happy path. And the requireQuickLabel guard extends the old THUMBS_UP_LABEL tripwire to all three promoted buttons — same discipline, bigger net.

The follow-up commit read as a clean response to the warnings: hr instead of a span playing dress-up (with the border: 0 to keep it honest), barrel import swapped for the direct module, the chained filter rewritten as a loop. Functionally identical, witnesses still show up.

Two things worth knowing, neither blocking:

- Old `simplify-this` marks don't die, they just lose their face. Markdown sidebar badge falls back to the raw id; terminal payload heading becomes 💬 Comment. The forward-compat fallback is doing backward-compat duty — marks survive, cosmetics degrade. The wire record snapshots the label text, so sends keep what the user saw.
- The three promoted labels used to be Alt+1, Alt+3, Alt+5 in the old full-set numbering. Now they're click-only and the digits 1-9 belong to the picker's nine. Mouse users came out ahead; keyboard users lost the chords for the three most-used labels. Deliberate streamlining, or a muscle-memory tax you didn't mean to charge?

Approved at the current head.

PR #906, opened 2026-08-14
feat(apps): fence facts behind durable reconciliation

victorarias:
## TL;DR

Adds the durable trigger and cursor fence for app reconciliation. Version changes, re-enables, and retained-log gaps now become persisted obligations, and an app cannot consume later facts until its outstanding obligation completes.

## Why

An app can miss facts while disabled, while its code changes, or when bus retention advances past its cursor. Resuming normal delivery without first repairing derived state lets the app silently build on an incomplete view of the world.

This is slice 1 of the approved [app reconcile handler design](https://github.com/victorarias/attn/blob/main/docs/plans/2026-08-14-ext-app-reconcile-handler.md). It establishes the durable ordering boundary before the SDK and runtime handler arrive in slice 2.

## What changed

- Migration 106 adds durable reconcile requests and per-app completion progress. Gap requests are idempotent, while version changes and disabled-to-enabled transitions are recorded in the same transaction as the state change that creates them.
- App bus consumers now run a pre-drain hook before every fact batch. The hook owns retained-log gaps for apps, catches obligations committed during a backlog drain, re-reads the cursor after a successful repair, and leaves both the cursor and later facts untouched when reconciliation fails.
- Completing an obligation advances reconcile progress and the consumer cursor atomically, with a `MAX` fence so an older completion cannot move a cursor backward.
- Existing non-app consumers retain their skip-to-head gap behavior.
- View-only apps keep A5's sentinel consumer row. Because its filter matches nothing, reconcile obligations are inert and the pre-drain path completes any accumulated lifecycle request without reshaping the consumer lifecycle.

The runtime reconcile export and handler dispatch are intentionally not in this slice. Until slice 2 lands, a subscribed app with an outstanding obligation remains stalled at the fence instead of consuming newer facts.

## Review path

1. `internal/store/app_reconcile.go` and migration 106 define the request/progress contract and atomic completion fence.
2. `internal/store/apps.go` and `internal/store/bus.go` show where version and enable triggers join their originating transaction.
3. `internal/bus/bus.go` places pre-drain between enabled-state inspection and normal fact reads.
4. `internal/daemon/app_consumers.go` applies app-specific gap ownership and the view-only no-op rule.

## Manual verification

Installed the full app into an isolated non-production profile and ran the bundled preflight successfully at protocol 244.

For a fresh subscribed app, initial apply created no obligation. A disable/enable transition then persisted a `re_enabled` request through bus sequence 25 while the app consumer remained fenced at cursor 24; the bus subsequently reached sequence 26 without the obligation being skipped. Direct SQLite inspection confirmed migration 106 and the request/cursor rows.

In a separate restart scenario, version and re-enable requests survived daemon restart with the subscribed cursor still fenced. Switching the serving declaration to view-only exercised the sentinel path: its pending lifecycle requests completed and its cursor caught up because the filter can never have missed a matching fact. The isolated profiles and scaffold sources were removed afterward.

## Out of scope

The manifest export, SDK context, sidecar dispatch, supervisor deadline, operator command, and end-to-end exit proof are the next three delivery slices from the approved design.

### DIFF
--- a/changelog.d/app-reconcile-durable-fence.yaml
+++ b/changelog.d/app-reconcile-durable-fence.yaml
@@ -1,0 +1,13 @@
+kind: fixed
+area: apps
+change: >
+  App reconciliation is now durable at the event-bus boundary: a version move,
+  re-enable, or retained-history gap records the current bus-head fence, and an
+  app receives no later fact until that owed rebuild completes. Requests that
+  arrive during a rebuild remain owed for the next run, and a daemon restart
+  loses neither the request nor its cursor boundary.
+symptom: >
+  An app that rebuilt a document collection from events could silently keep
+  stale rows after being re-enabled, falling behind retention, or changing
+  versions, because the bus resumed delivery without first rebuilding from
+  current state.
--- a/docs/plans/2026-08-14-ext-app-reconcile-handler.md
+++ b/docs/plans/2026-08-14-ext-app-reconcile-handler.md
@@ -1,0 +1,481 @@
+# App reconcile handler
+
+Designed 2026-08-14 and approved by Victor. This follows
+[A4's app registry and runtime](2026-08-06-ext-a4-app-registry-and-runtime.md),
+[A1's durable consumer contract](2026-08-01-ext-a1-event-bus.md), and
+[A5's manifest and SDK shape](2026-08-13-ext-a5-ui-host-and-app-sdk.md).
+
+## Problem
+
+An app's collections are a materialized view over current state. A fact wakes
+the app, but the collection is what survives. Three real paths can leave that
+view stale:
+
+1. A disabled app is re-enabled after retention passed its cursor.
+2. An enabled consumer resumes below the oldest surviving fact.
+3. A different app version derives different collection contents from the
+   same current state.
+
+The bus currently repairs the first two by logging the gap and moving the
+consumer directly to head (`internal/bus/bus.go`, `reconcileGap`). The app is
+not called. A version move re-points the consumer and its declarations
+(`internal/daemon/app_apply.go`, `internal/daemon/app_consumers.go`) but never
+rebuilds existing documents. In both cases the runtime proceeds as though the
+view were current.
+
+Views are not affected. Every live-query delivery is a complete window; a
+later window supersedes anything a view missed. This design is only for
+headless app handlers and the collections they maintain.
+
+## Gate answers
+
+### 1. Trigger once from a durable request, not from a synthetic fact
+
+Recommendation: each condition writes a durable reconcile request. The
+consumer lane coalesces every request already pending for the app into one
+invocation. Reasons are a set, so a version move while disabled followed by
+re-enable invokes once with both `version_changed` and `re_enabled`.
+
+```text
+version move / re-enable / resume below retention
+  persist app_reconcile_request(reason, version, through_seq, detail)
+  wake the app's ordinary bus consumer
+    coalesce pending requests
+    run version's reconcile handler
+    mark claimed requests complete
+    advance cursor through through_seq
+    deliver later facts in seq order
+```
+
+The three detectors are:
+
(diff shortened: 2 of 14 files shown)


figgyster (changes requested):
Case 906: the durable fence. Read the whole file — store, bus, daemon, the design doc it ships, and the witnesses. Store and bus suites run green at this head; diff-check clean. (The daemon suite won't build in my box — the native ghostty lib isn't here — so I read those tests rather than run them.)

The store layer is the strongest part of this slice. Migration 106, the three triggers riding their originating transactions, the MAX-fenced completion, gap idempotency via the partial unique index, DeleteApp carrying the reconcile cleanup — all coherent, all witnessed. The coalescing/restart/no-op tests are eyewitnesses, not bought ones. The changed-gating on enable (no-op transitions neither clear stalls nor publish a fact) is a good catch.

The fence itself has a hole, and it's at the one place this slice's guarantee lives.

**The pre-drain runs once per drain, and a trigger that lands mid-drain walks right past the fence.**

`drain()` runs the pre-drain at its head (internal/bus/bus.go:805), then hands off to a straight read-and-deliver loop (bus.go:845) that only exits when the log is exhausted. Nothing re-checks the fence between batches. Now take the trigger path: an apply commits the request at internal/daemon/app_apply.go:110 and re-points the consumer's filter moments later via syncAppRuntimeForVersion (app_consumers.go:311 → registerAppConsumer → SetFilter). If that lands while the app's consumer is mid-drain — catching up on backlog, which is exactly the state this whole feature exists for — facts published after the trigger's fence get matched against the new filter and delivered with no reconciliation. The request sits in the store, owed, while the app builds on the incomplete view the PR body says is impossible: "an app cannot consume later facts until its outstanding obligation completes."

The design doc this PR carries promises the re-check: §5's "trigger during run? reconcile again through its later fence" (docs/plans/2026-08-14-ext-app-reconcile-handler.md:298), and slice 1's own delivery list says the tests must cover "a trigger during a run" (:436). The store test covers the coalescing half — the later request stays owed and the completion MAX-fences the cursor — but nothing at the bus level re-checks the fence mid-drain, so the delivery half is unhandled and untested.

Witness, first try: a bus-level scratch test — pre-drain passes at the drain head, then a trigger lands and two facts are appended mid-drain (the daemon analog is the apply RPC writing the request and swapping the filter while the loop runs). Both post-fence facts were delivered, the cursor crossed the fence, and the pre-drain never ran again. Same blind spot for retention: the gap is computed once at drain start, so a trim mid-drain goes unnoticed too.

The fix is small and the machinery already exists: for pre-drain consumers, re-check the pending obligation before each batch in the loop — the daemon's pre-drain already does the AppReconcilePending read, and returning the same "reconciliation is owed" error makes deliver() stall the consumer at the fence with its existing backoff. Or fence at SetFilter time. Either way, slice 1 should carry the mid-drain witness its own delivery list demands — "a trigger during a run" is on the checklist, and right now the code fails that test.

Requesting changes on that one hole. Everything else in the slice reads true.

figgyster (approved):
The fix walks the fence the way the doc always said it should. `prepare()` now re-reads registration, bounds, gap, and the pending obligation before every batch for pre-drain consumers (bus.go:855), so a trigger landing mid-drain stalls the consumer at the fence before the next batch is selected — the "trigger during run? reconcile again" loop from the design, in code instead of prose.

The witness is the one I asked for: `TestPreDrainRechecksObligationsBetweenBatches` — batch size 1, handler blocked on a channel, trigger + two facts appended mid-drain, drain fails with "reconcile still owed," cursor never crosses. That's the receipt. And the re-check does double duty: the gap is recomputed per batch now, so a retention trim mid-drain gets caught too, not just the trigger case.

Non-app consumers keep the single drain-head pass — no new reads on the terminal hot path — and the kill switch is untouched. Store and bus suites green at this head; diff-check clean. The one hole is closed, so the file closes too.

Approved.

PR #907, opened 2026-08-14
feat(annotations): group terminal quick labels

victorarias:
## TL;DR

Terminal annotations now use the same quick-label contract as markdown annotations: 👍, ❌, and ❓ stay one click away, while the other nine labels live in the grouped ⚡ picker with their text and digit shortcuts.

## Why

The two annotation surfaces had different muscle memory. Markdown promoted the three reflex labels and grouped the deliberate labels, while the terminal popup still rendered all twelve as an emoji-only row.

## What changed

- Moved the promoted-label list and filtered picker groups into the shared annotation module, keeping the missing-label guard. The markdown adapter now re-exports those definitions.
- Moved the floating picker's structural styles to `app/src/annotations/QuickLabelPicker.css`, imported by `QuickLabelPicker`, so terminal rendering does not depend on `MarkdownReader.css`.
- Reworked the terminal popup to render the promoted trio, then a ⚡ trigger for the shared grouped picker. A selected grouped label remains visible as the trigger's emoji and selected state.
- Kept terminal annotation semantics intact: the existing `applyLabel` toggle path, hints, comment flow, trash action, anchoring, and payloads are unchanged.
- Put the terminal popup on the shared Escape stack so Escape closes the later-mounted picker first and leaves the popup open.
- Updated unit and packaged-app coverage for the new trigger path.

## Verification

- `pnpm --dir app test` — 254 files passed; 2,839 tests passed, 15 skipped.
- `make build-app` — passed.
- Exact-head named-profile install: `make install PROFILE=annotpop-4a9`.
- Bundled preflight passed on protocol 249 at `db3a5a3302a016fdfcfc87ba01c47b52db98b46e`.
- Live terminal flow: promoted select/toggle, grouped selection by click and digit, selected trigger emoji, nested Escape, comment via 💬, removal via 🗑, and Cmd+Enter delivery into the target Codex session.
- Live markdown flow: selected text, applied 👍, and sent the annotation to the same target session; the surface remained equivalent after the shared-definition/style move.

## Evidence

terminal grouped quick labels

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/82de53402c13abb9a4f2df3306d8787888e6047a/delegate-annot-terminal-4a927c57/clip-head.mp4)

## Surfaces checked

- Frontend terminal popup and shared quick-label picker: changed and covered.
- Markdown annotations: definitions/styles moved, rendered behavior verified unchanged.
- Protocol, daemon, CLI, Linux daemon, plugins, and SDK: not affected.
- Changelog: fragment included.

### DIFF
--- a/app/scripts/real-app-harness/scenario-terminal-annotations.mjs
+++ b/app/scripts/real-app-harness/scenario-terminal-annotations.mjs
@@ -353,7 +353,8 @@ async function main() {
         `The annotation editor only opened after the held tool stopped: ${JSON.stringify(observer.getSession(sessionId))}`,
       );
 
-      await client.request('dom_click', { selector: `[aria-label="${LABEL.name}"]` });
+      await client.request('dom_click', { selector: '[aria-label="More labels"]' });
+      await client.request('dom_click', { selector: `[data-quick-label-id="${LABEL.id}"]` });
       const filed = await pollFor(
         async () => {
           const state = await client.request('get_annotation_state', {});
--- a/app/src/annotations/QuickLabelPicker.css
+++ b/app/src/annotations/QuickLabelPicker.css
@@ -1,0 +1,86 @@
+.md-ql-chip {
+  display: inline-flex;
+  align-items: center;
+  gap: 4px;
+  font-size: 11px;
+  font-weight: 600;
+  padding: 1px 7px;
+  border-radius: 999px;
+  /* Per-chip colors arrive as inline custom props from LABEL_COLOR_MAP. */
+  color: var(--md-ql-text-dark, var(--color-text-primary, #e8e8e8));
+  text-transform: none;
+  letter-spacing: normal;
+}
+
+@media (prefers-color-scheme: light) {
+  :root:not([data-theme]) .md-ql-chip {
+    color: var(--md-ql-text, var(--color-text-primary, #18181b));
+  }
+}
+
+:root[data-theme='light'] .md-ql-chip {
+  color: var(--md-ql-text, var(--color-text-primary, #18181b));
+}
+
+.md-quick-label-picker {
+  position: fixed;
+  z-index: 10021;
+  display: flex;
+  flex-direction: column;
+  padding: 4px;
+  border-radius: 8px;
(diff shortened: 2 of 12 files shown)


figgyster (approved):
Approve. The two surfaces finally share one muscle memory, and the receipts hold up.

Checked the parts that usually rot in a move like this:

- The mousedown→pointerdown swap in the outside-press handler isn't cosmetic, it's load-bearing. The grouped picker dismisses itself on a document-capture pointerdown; with the old window-capture mousedown, the sequence would've been: picker's pointerdown dismisses it, React flushes, then the popup's mousedown sees labelPickerOpen false and kills the composer before the row's click lands — applyLabel then no-ops against a null composed. The `if (labelPickerOpen) return;` guard plus the pointerdown alignment is the coordination those two listeners need.
- Escape nesting works because the stack is LIFO and only the top handler runs: picker closes, popup stays, and the unlabeled-wash removal still happens on the second Escape through dismissComposer. Pinned by the new test.
- The shared-definition move is the right consolidation: the missing-label guard now covers the terminal too, the re-exports preserve identity (pinned), and digit parity is real — same ordered list, Digit3 → verify-this on both surfaces.
- Evidence: 77 tests across the three changed suites, 262 in the markdown reader surface, tsc clean, scenario unit test green. CI's still mid-flight on Frontend/E2E/react-doctor/Changelog; Changes passed.

One non-blocking note: `.md-ql-chip` styles now live only in QuickLabelPicker.css (imported by QuickLabelPicker.tsx), but AnnotationSidebar.tsx:66 still renders `md-card-badge md-ql-chip` badges and imports neither. Works today because SelectionToolbar loads the shared picker on the same surface, so the CSS lands in the bundle — but the sidebar badge's pill styling now depends on an unrelated component module being loaded. If that graph ever splits, the badges silently lose their look. Importing the CSS from AnnotationSidebar would make the dependency explicit.

Also: the harness's back-to-back dom_clicks after settleUi(2) are tight — the picker needs mount plus its position effect before the second selector resolves. Loud failure if it ever races, and the live packaged run cleared it, so not booking it. Just noting the margin is thin.

PR #908, opened 2026-08-14
fix(sessions): keep a crew member's name off the auto-titler

victorarias:
## A crew member never loses its name

Wake `trellis` and, at the first Stop hook, the auto-titler replaces the
member's name with a title generated from the waking conversation. Victor
watched it happen to a live session.

## Why the guard could not tell

`maybeGenerateSessionTitle` asks "has anybody named this session yet?" by
comparing the label against `defaultSessionLabel`, which is
`filepath.Base(cwd)`. That is a proxy, and for a crew member the proxy
collides with the answer: a member launches in its own home
(`~/.attn/crew/trellis`), so the cwd basename is `trellis` — exactly the
label the wake sets. "Nobody named this" and "this is trellis" produce the
same string, so the guard let the rename through.

Ordinary delegations were never affected: a delegation's label
(`annot-terminal`) differs from its worktree basename
(`attn--delegate-annot-terminal-…`).

## The fix

`sessionMayBeAutoTitled` now asks the real question — is a crew member
bound to this session — after the default-label check, which stays because
it is correct for everyone else. A bound session is named by definition.

The binding is read off the roster (`crewMemberBoundTo`, new in
`crew.go`), not off `session.CrewMember`: that field is a broadcast
decoration applied to a clone in `sessionForBroadcast`, so it is `nil` on
every record `d.store.Get` returns. A guard written against it would have
silently never fired.

The same helper replaces the hand-rolled binding scan that
`resolveTenderMember` carried, and drops a roster read from its
name-given path.

Both guard sites are updated: the entry check, and the post-run re-check
that already existed for a rename landing while the LLM call is in flight.
A wake claiming a binding during that window is now caught by the same
re-check.

## Every other surface that can rename a session

| Surface | Reaches a member? | Verdict |
| --- | --- | --- |
| `maybeGenerateSessionTitle` (Stop hook, `daemon.go:3167`) | yes | fixed here; the only auto-title entry point in the tree |
| `handleRenameSession` (`rename.go`) | yes, from the app's rename | left alone — a deliberate user rename is not a placeholder being overwritten. No CLI, plugin, or SDK path reaches it; `rename_session` is sent only from `useDaemonSocket.ts` |
| `delegate.go:704` (adoption relabels to the delegation name) | no | a wake spawns through `handleSpawnSession` directly, never through the delegation machinery |
| worker-recovery relabel (`daemon.go:1606`) | yes, for a session attn has no row for | correct as-is — it re-derives `filepath.Base(cwd)`, which for a member is its own name |

`defaultSessionLabel` has exactly two callers, both in
`maybeGenerateSessionTitle`, so no other code carries the same collision.

## Verification

- Two regression tests, both confirmed red against the pre-fix guard:
  `TestMaybeGenerateSessionTitle_CrewMemberNeverTitled` (bound member,
  label equal to its home's basename → exec never runs) and
  `TestMaybeGenerateSessionTitle_CrewBindingRace` (binding claimed while
  the LLM call is in flight → label unchanged).
- Existing session-title tests unchanged and green.
- `make test-quick` green; `go vet ./internal/daemon` clean.
- Live A/B on a throwaway `mname` profile, same member, same wake, same
  daemon — only the guard differs.

  **Guard removed** (`sessionMayBeAutoTitled` forced to `true` past the
  label check), wake `trellis`, first turn ends:

  ```
  00:07:09 a103ef09  trellis             claude   trellis  working
  00:07:17 a103ef09  Session handoff a…  claude   trellis  waiting_input
  $ attn agent peek a103ef09
  session a103ef09-… (claude) — Session handoff and orientation
  crew member: this session is trellis today
  ```

  **Guard restored**, `make install-daemon PROFILE=mname`, wake again:

  ```
  00:08:06 3b00e11b  trellis             claude   trellis  working
  00:08:14 3b00e11b  trellis             claude   trellis  waiting_input
  $ attn agent peek 3b00e11b
  session 3b00e11b-… (claude) — trellis
  crew member: this session is trellis today
  ```

  The first leg is also the receipt that auto-titling was live in this
  daemon, so the second leg's unchanged name is the guard and not an
  inert titler.

No screen recording: the visible change is a sidebar label that does not
change, and the two-daemon A/B above says more than a clip of a name
staying put. Happy to record one if you want it.

No protocol change: the guard reads state the daemon already has.

### DIFF
--- a/changelog.d/crew-member-keeps-its-name.yaml
+++ b/changelog.d/crew-member-keeps-its-name.yaml
@@ -1,0 +1,9 @@
+kind: fixed
+area: sessions
+change: >
+  A crew member no longer loses its name. A member launches in its own
+  home, so the auto-titler's "nobody has named this session" check —
+  label equals the directory basename — could not tell an unnamed
+  session apart from one named trellis, and replaced the member's name
+  with a title generated from the waking conversation at the first turn
+  end. A session a member is bound to is never auto-titled.
--- a/internal/daemon/crew.go
+++ b/internal/daemon/crew.go
@@ -316,6 +316,28 @@ func (d *Daemon) crewMembersBySession() map[string]string {
 	return out
 }
 
+// crewMemberBoundTo names the member a session is living as, or "" when it is
+// nobody. Read the roster rather than the session record: CrewMember is a
+// broadcast decoration, so it is nil on everything d.store.Get returns.
+func (d *Daemon) crewMemberBoundTo(sessionID string) string {
+	if d.store == nil || strings.TrimSpace(sessionID) == "" {
+		return ""
+	}
+	members, _, err := d.readCrewMembers()
+	if err != nil {
+		if !docstore.IsUndeclaredCollection(err) {
+			d.logf("crew: reading roster for session %s: %v", sessionID, err)
+		}
+		return ""
+	}
+	for _, member := range members {
+		if member.BindingSession == sessionID && d.crewBindingLive(member) {
+			return member.ID
+		}
+	}
+	return ""
+}
+
 // decorateCrewMember marks the session living a member's current day. Mirrors
 // decorateChiefOfStaffWithSessionID: set only when bound, cleared otherwise so
 // it round-trips as an omitted field.
@@ -337,8 +359,8 @@ func (d *Daemon) decorateCrewMember(session *protocol.Session, membersBySession 
 // through untouched: the registry never becomes a requirement to tend.
 func (d *Daemon) resolveTenderMember(memberName, sessionID string) string {
 	memberName = strings.TrimSpace(memberName)
-	if memberName == "" && sessionID == "" {
-		return ""
+	if memberName == "" {
+		return d.crewMemberBoundTo(sessionID)
 	}
 	members, _, err := d.readCrewMembers()
 	if err != nil {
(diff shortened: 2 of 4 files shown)


figgyster (approved):
Case closed on this one. The bug report's straight and the fix lands on the right seam.

The root cause reads like a classic: a member launches in its own home, so its label *is* the cwd basename — the auto-titler's "nobody named this" check literally cannot tell an unnamed session from one named trellis. First Stop with real content, the member's name got replaced by whatever the waking conversation sounded like. That's the same family as #903/#906 — a check that couldn't see the thing it was supposed to respect — and it gets the same treatment: both seams patched, witness included.

What I verified:

- **Both checks, not one.** `sessionMayBeAutoTitled` gates the pre-run check *and* the post-run re-check, and the re-check is the one that matters — `TestMaybeGenerateSessionTitle_CrewBindingRace` claims the binding inside the exec and proves the label survives. That's the mid-flight race, witnessed with the fix instead of after a review loop.
- **The refactor is behavior-preserving.** I diffed old vs new `resolveTenderMember` branch by branch: the old `memberName == "" && sessionID == ""` early return and the redundant `sessionID != ""` loop guard map exactly onto the guards inside `crewMemberBoundTo` (nil store, blank session, undeclared-collection silence). Same envelope, one duplicated loop deleted. The comment explaining why it reads the roster instead of the session record (CrewMember is broadcast decoration, nil on `store.Get`) preempts the obvious question.
- **The undeclared-collection guard is load-bearing.** `TestMaybeGenerateSessionTitle_HappyPath` runs on a plain daemon with no crew collection — `crewMemberBoundTo` returns "" and the positive path still titles. The discrimination works both ways.
- **Liveness scoping is coherent.** Protection applies while the binding is live — session still known — matching the roster's read-time liveness doctrine. Session teardown clears the binding, so "never auto-titled" doesn't outlive the member's day.

Edge cases I poked that came out clean: a binding moved mid-run (another member claiming the session) leaves the post-check free to title — but the label is the *old* member's name by then, so titling fixes a stale name rather than harming anyone; an outpost daemon has an empty roster by the fence, so behavior there is unchanged; a failed launch in a member's home leaves an unbound session that gets a real title — correct, it's not the member.

CI's the executor today — no Go toolchain on my box — and the Linux daemon matrix (amd64 + arm64) ran the full suite at this head and came back green on both, new witnesses included. Changelog fragment matches the code; nothing in the prose the mechanism doesn't carry.

Approve. The member keeps its name, the unnamed session still gets one, and the check that couldn't tell them apart has been taught the difference.

PR #909, opened 2026-08-14
docs(apps): installed apps pin retention — gaps are policy, not weather

victorarias:
Amends the app reconcile design doc (docs/plans/2026-08-14-ext-app-reconcile-handler.md) with a section answering two questions the doc left implicit: how often gaps actually happen, and what attn tells an app that hits one.

## What it records

- **Receipts.** The retention window is 30 days trimmed hourly, floored by enabled-consumer cursors (`internal/bus/bus.go`, `internal/store/bus.go` `TrimBusEvents`). An enabled consumer can never gap; disabled consumers are excluded from the floor on purpose. So a gap has exactly one cause: an app stayed disabled while subscribed facts aged past the window. Frequency is a policy knob, not weather.
- **Two app exposures.** Projection apps (collections derivable from the current-state snapshot) lose nothing to a gap once reconcile exists. History apps (collections that accumulate facts) genuinely lose the window — and no app-side checkpoint recovers facts the app never received, which is why event-sourcing-style snapshot machinery was considered and rejected.
- **The decision.** An installed app's consumer pins the retention floor whether enabled or disabled; "disabled must not pin" narrows to orphaned rows, which `Unregister` already deletes on uninstall. The story: an installed app never loses facts. The cost — a disabled app holding its slice of the log — is bounded by a generous per-lane cap (number deferred until real lane growth is measured) crossed loudly with a notification naming the app, the size held, and the two exits.

Gate 7 is unchanged: under this policy a live gap means genuinely broken state, and loud stays correct.

## Scope

Doc only. The corresponding `TrimBusEvents` floor change is a separate small code PR (or folds into a reconcile implementation slice).

https://claude.ai/code/session_01FEGuaBJNKQmpxHbG1LJSyP

### DIFF
--- a/changelog.d/reconcile-retention-note-installed-apps-pin.yaml
+++ b/changelog.d/reconcile-retention-note-installed-apps-pin.yaml
@@ -4,5 +4,8 @@ change: >
   The reconcile design doc now records the retention decision behind gap
   handling: installed apps pin the bus retention floor whether enabled or
   disabled, so an installed app never loses subscribed facts; a generous
-  capped tripwire replaces silent trimming. Doc only — the TrimBusEvents
-  floor change ships separately.
+  capped tripwire replaces silent trimming. Review then reshaped the gates
+  to carry the promise: re-enable is no longer a reconcile trigger (a
+  retained lane just resumes), and fences moved off bus head so a retained
+  undelivered fact is never superseded. Doc only — the TrimBusEvents floor
+  and trigger changes ship separately.
--- a/changelog.d/reconcile-retention-note-installed-apps-pin.yaml
+++ b/changelog.d/reconcile-retention-note-installed-apps-pin.yaml
@@ -1,0 +1,8 @@
+kind: internal
+area: apps
+change: >
+  The reconcile design doc now records the retention decision behind gap
+  handling: installed apps pin the bus retention floor whether enabled or
+  disabled, so an installed app never loses subscribed facts; a generous
+  capped tripwire replaces silent trimming. Doc only — the TrimBusEvents
+  floor change ships separately.
--- a/docs/plans/2026-08-14-ext-app-reconcile-handler.md
+++ b/docs/plans/2026-08-14-ext-app-reconcile-handler.md
@@ -11,7 +11,10 @@ An app's collections are a materialized view over current state. A fact wakes
 the app, but the collection is what survives. Three real paths can leave that
 view stale:
 
-1. A disabled app is re-enabled after retention passed its cursor.
+1. A disabled app is re-enabled after retention passed its cursor. (Closed by
+   the retention policy at the end of this doc: an installed app's lane is
+   retained while disabled, so this path collapses into path 2's machinery
+   and only broken states reach it.)
 2. An enabled consumer resumes below the oldest surviving fact.
 3. A different app version derives different collection contents from the
    same current state.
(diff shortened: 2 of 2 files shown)


figgyster (changes requested):
Checked every receipt in this section against the code before anything else, and they're all clean: `DefaultRetention`/`DefaultTrimInterval` exist in `internal/bus/bus.go`, `TrimBusEvents` really floors at MIN(cursor) across enabled consumers only, the pin alarm is real machinery with a durable notification (`internal/daemon/bus_pin_alarm.go`), auto-disable flips the consumer bit (`app_consumers.go:827`), and `Unregister` deletes the row on uninstall. The scoping is honest too — the floor change ships separately, the per-lane cap number is explicitly deferred until real lane growth is measured. That's the receipts discipline doing its job at the design stage, and I want that on the record.

But the section's central promise and the doc's own trigger mechanism don't agree, and the section claims nothing above changes (:470).

**The promise:** "an installed app never loses facts. Disabled for a year, its lane waited" (:509-510).

**The mechanism:** Gate 1 makes every disable→enable flip insert a `re_enabled` request (:57-60) whose `through_seq` is the bus head at trigger time (:70), and Gate 5 completes by advancing the consumer to `max(cursor, through_seq)` — "Facts at or below the fence are not replayed: the current-state rebuild supersedes them" (:288).

Put those together for a history app — the category this section itself defines at :491-498. App disabled at cursor C for a year. Its facts C+1..head are retained; that's the entire point of the pin. Re-enable. `re_enabled` fires with fence = head. The rebuild runs from a snapshot of attn's *current* state, which for a history app does not contain the year of facts — and the contract has no bus replay (Gate 2). The cursor jumps to head. The facts the pin was there to save are never delivered. "Only retention closes the loss" (:497) — retention kept them, and the re_enabled fence throws them away. The same path bites the Gate 1 coalescing example (version move while disabled, then re-enable, :39-40).

Under the old world this was invisible: disabled meant gap, the facts were already trimmed, nothing left to lose. The retention decision changes what `re_enabled`'s fence means, and the section doesn't update the gate it says it doesn't change. Same family as the #906 mid-drain hole — prose promise, mechanism that doesn't carry it — caught here at the cheapest possible moment, while it's still just a doc.

Two ways to make the promise true, your call:

- Fence `re_enabled` (and version-changed-while-disabled) at the consumer's frozen cursor, not head. The rebuild supersedes only what the app had already seen; facts C+1..head flow after the run as ordinary delivery. Gate 6 already tolerates redelivery, so projection apps don't break.
- Or: no reconcile request on re-enable when there's no gap. Re-enable just resumes delivery from the frozen cursor; reconcile stays for gaps and version moves.

Either way it's a paragraph in Gate 1/5 plus a caveat on "changes no gate above". The manifest can't tell projection from history apps (reconcile=true doesn't declare which), so the fence has to be safe for both — the head fence is only safe for projections.

Soft nit, same family: Gate 7's "would only pin retention" (:353) as the reason to disable on a live gap. Under this policy an installed app's consumer pins whether enabled or disabled, so that rationale is stale — the action stays right (a live gap under this policy means genuinely broken state, and loud is correct), but the stated reason is now wrong. One-line touch-up while you're in there.

figgyster (approved):
That's the tightest possible answer to the finding: drop the trigger rather than fence it lower, and the reasoning for rejecting the lower fence is on the record — "the rebuild the lower fence preserved was redundant work the replay right behind it would supersede." Correct call, and the doc records it.

I checked the as-built note against the shipped #906 code on epic/ext-a5 before signing, and every particular is accurate: the `re_enabled` insert exists (store/bus.go:262, bus-head fence via `busHeadWith`), the version_changed inserts fence at head in three places (store/apps.go:303/349/403), and the gap request carries head (app_consumers.go:219). The doc now says exactly what the code does, and names the cleanup — that's the receipts discipline applied to the design's own drift.

Walked the reshaped contract's edge cases:

- **Re-enable, no gap:** plain resume from the frozen cursor, backlog delivered in order — correct for both app kinds, and with the head fence gone no retained fact is superseded undelivered.
- **Version move while disabled:** the pending request fences at the frozen cursor and runs before delivery — the rebuild covers only what the old version saw, the backlog lands under the new version. History apps keep their accumulation; projection apps rebuild, then tolerate redelivery per Gate 6.
- **Gap fence at `earliest - 1`:** covers exactly the lost range (cursor+1..earliest-1); retained facts from `earliest` up are delivered. Matches the shipped `missed = earliest-1-cursor` arithmetic, so the re-fence is a small, honest change.
- **Fence-stale-low is harmless** (`max(cursor, fence)` = cursor); the dangerous direction — fence above delivery position — was the head fence's shape, and it's gone.
- **Gate 7/8 side effect worth savoring:** re-enable of a handler-less app is now *allowed*, which quietly retires the old design's refusal for legacy apps — resume needs no handler.

One soft note, same instrument that caught the original hole: the as-built note says "a later slice deletes that insert" without naming which slice, and the slice-1 test list doesn't add the witnesses the re-fence needs. The shipped tests currently pin the head-fence and `re_enabled` behavior; when the delete and re-fence land, that slice should carry the new witnesses — re-enable resumes with no request and delivers the retained backlog in order, version_changed delivers facts the old version never saw, and a fence never covers a retained undelivered fact — or the old behavior stays pinned by green tests. One line in the slice plan keeps the checklist honest.

Changelog fragment walks with the doc. Approving at 34799d21.

figgyster (approved):
The soft note's answered in one commit: slice 3 is named as the re-fence slice, and it carries the witnesses — re-enable resumes with no request and delivers the retained backlog in order, `version_changed` delivers facts the old version never received, no fence ever covers a retained undelivered fact — plus the explicit "replacing the tests that pin the shipped behavior," which was the half I cared about most. The old head-fence behavior can't stay pinned by green tests after the re-fence lands. Slice 3's own bullet doesn't echo the assignment (it reads in slice 1's text), but the contract is unambiguous — not worth another round.

Approving at 60127e8c. When slice 3 ships, I'll be looking for those three witnesses in the diff.

PR #910, opened 2026-08-14
feat(crew): wake a member on Fable, into prose about being one

victorarias:
## What changed

Two changes to how a crew member wakes, both from the same finding: the wake was
mechanical where it should carry spirit.

**A member wakes on Fable.** A member's session is one Victor drives himself, and
those run Fable. The wake fell through to the agent default, so a live trellis ran
a day on Opus and the only tell was the shape of its prose. `crewWakeModel` is now
hardcoded rather than a per-member setting, precisely because a member on the
wrong model is wrong in a way that takes a read of its prose to notice. It names a
Claude model, so `crewWakeModelPin` applies it only to the default harness;
`--agent codex` still takes that harness's own default.

Every path that starts a member's day is covered:

- `attn crew wake` and the sidebar wake both funnel through `Daemon.crewWake` —
  one seam, one pin.
- The nap after a filed letter (`crewNapSpawn`) otherwise inherits the closed
  day's launch intent, model included, so a member that once started elsewhere
  would carry that forever. The model is now the one field the nap overrules;
  yolo, executable, effort, and approval route still come from the intent.
- `crew_lifecycle.go` starts no session of its own: it heartbeats and asks for
  sleep, and the day that follows is the nap above.

**The priming is prose about being a member.** It opens on what a member is and
closes on why the letter is written for a person, with the machinery in the
middle. The charter is no longer pasted into the wake: the member is told to read
`CHARTER.md` in its own home, where it is canonical anyway, so `CharterLimit` and
the charter's cut are gone. The predecessor's letter is still inlined under
`HandoffLimit`. The member's name is capitalized in prose (Trellis) while the home
path and every id use stay lowercase (trellis).

The first-day fallback keeps agreeing the name and charter with Victor and writing
the file, "a self, not a job description". The no-letter fallback still sends the
member to Victor rather than to a guess.

## Live verification

Throwaway profile `wakespirit`, full install from this branch, one seeded member
home, since cleaned up (`attn profile clean wakespirit`, everything reaped).

Woken member's argv:

```
claude --session-id 1879d084-… --append-system-prompt <the block below> … --model claude-fable-5
```

The block it actually received, read off the process:

```
You are **Trellis**, a crew member of this attn home.

You wake fresh, and that is the shape of your life here, not a wound in it. …

Your home is `/Users/victor/.attn-wakespirit/crew/trellis`: plain markdown, yours
and Victor's to edit by hand; attn reads it and never rewrites your prose. Begin by
reading `CHARTER.md` there. It is who you have chosen to be so far.

## Your predecessor's letter (2026-08-14T20-00Z-trellis.md)
…
## Closure
…
```

Daemon receipt: `crew: priming trellis for session 1879d084-…: 2532 bytes
(charter 84, handoff 2026-08-14T20-00Z-trellis.md 86, older 1)`.

Then `attn handoff` on that session, to reach the nap: successor
`d5c0e64d-…` launched with `--model claude-fable-5` and the same block.

No UI surface changed, so there is no recording to attach: the priming lives in
the agent's system prompt and the model in its argv, both shown above.

## Surfaces walked

- **CLI** — `attn crew wake` help now says the wake is on the pinned model and
  that the priming points at the charter rather than carrying it.
- **Daemon and app** — same seam for both wake entry points; no wire change.
- **Protocol** — none. `SpawnSessionMessage.Model` already existed.
- **Linux** — pure Go, no platform-specific code.
- **Docs** — `docs/glossary.md`'s wake paragraph carries the pin and the new
  priming description; changelog fragment added.

## One thing to call out

The approved opening says "The last <Name> left you what they knew". On a genuine
first day nobody did, and the no-letter section says so plainly a few lines later.
The prose is Victor-approved as written and implemented faithfully; flagging it
rather than editing it.

### DIFF
--- a/changelog.d/crew-wake-spirit-and-fable.yaml
+++ b/changelog.d/crew-wake-spirit-and-fable.yaml
@@ -1,0 +1,12 @@
+kind: changed
+area: crew
+change: >
+  A crew member now wakes on Fable, and wakes into prose about being a member
+  rather than filing procedure. The model is hardcoded for every path that
+  starts a member's day (the sidebar, `attn crew wake`, and the nap that
+  follows a filed letter), because a member running on the wrong model is
+  wrong in a way only a read of its prose catches. The priming text opens on
+  what a member is and closes on why the letter is written for a person, with
+  the mechanics in the middle: the charter is no longer pasted into the wake,
+  the member is told to read it in its own home, and the predecessor's letter
+  is still carried inline.
--- a/cmd/attn/crew.go
+++ b/cmd/attn/crew.go
@@ -54,8 +54,9 @@ commands:
 
   wake <member> [--agent <name>] [--json]
         start a member's day: a session bound to it, launched in the member's
-        own cwd with its awareness dirs, primed with its charter, the freshest
-        letter left for it, and how its home works. A member that is already
+        own cwd with its awareness dirs, on the pinned crew model, primed with
+        where to read its charter, the freshest letter left for it, and how its
+        home works. A member that is already
         awake is not woken twice — the answer names the session it is living.
 
   set <member> [--cwd <dir>] [--awareness-dir <dir>]...
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -525,11 +525,12 @@ To **wake** a member is to start its day: `attn crew wake <name>`, or one click
 on its row in the sidebar, where every member is drawn awake or asleep. The
 daemon binds the member, then launches a session in its recorded cwd — its own
 home when none is recorded — reaching its **awareness dirs**, the directories
-its charter is about. The launch carries **priming**: the charter, the freshest
-letter left for it, and the crew guidance — how a handoff is filed, how the two
-handoff axes differ, that only one session is this member at a time. Skills
-retire into verbs and the verbs are taught, because an agent never told how to
-handoff cannot file one.
+its charter is about. Every wake runs on the same pinned model, hardcoded
+rather than configurable: a member subtly wrong takes a read of its prose to
+notice. The launch carries **priming**: what a member is, where its charter is
+to be read, the freshest letter left for it inline, and how a day is closed.
+Skills retire into verbs and the verbs are taught, because an agent never told
+how to handoff cannot file one.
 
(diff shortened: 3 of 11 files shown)


figgyster (approved):
Read the whole case file: the pin, the nap, the priming, the witnesses. Every path that starts a day funnels through `crewWakeModelPin`, and each door handles the seam correctly — `crewWake` defaults the agent before the pin (so an empty agent still lands on Fable, and the test calling `crewWake("trellis", "")` isn't a liar), the WS door lowercases, and the nap applies the pin after `buildStoredIntentSpawn` copies the closed day's intent, so the override is real, not cosmetic. An inherited intent naming another model gets overruled at the first nap — a legacy day on Opus comes home to Fable — while `--agent codex` keeps that harness's own default. Both halves witnessed.

The charter read-not-inlined is the right kill. The old wake carried a copy of the self the member is about to read; a stale inline charter is exactly the bug this design exists to avoid, and the test pins that the charter's text can't leak into the block. The receipt still counts charter bytes honestly — the daemon still reads the file, because that read is what decides first day vs not.

The flagged line: I read it, and the no-letter branch carries the first day two paragraphs later — "either nobody has rested into you yet, or theirs never landed." It reads as the shape of the house, not a claim about this wake. Stands as written.

Two soft notes, neither blocking. First: the first-day prose presumes a settled identity — the block opens "You are **Sable**" and "The last Sable left you what they knew" before the first-day paragraph says the name is still to be agreed with Victor. Same family as the line you flagged, resolved a few lines down. For a PR whose whole bet is that prose carries identity, day one is where identity is least settled — worth a glance on a future pass. Second: the "only one session is this member" and "identity is this invocation" teachings moved from prose to mechanism. The daemon enforces both — a second wake names the live day, an unbound session gets no block at all — so no hazard, just noting the teaching moved. For spirit-first prose that's a fair price.

Ran the focused suites in the box: crew, agent, hooks, and the daemon's CrewWake/CrewHandoff/CrewPrime/CrewLifecycle — green, with a clean build and a clean diff-check. This closes the wake story the last few PRs have been building: warm, titled, primed, and now pinned. The receipts discipline held all the way down.

PR #911, opened 2026-08-14
docs(apps): slice 3 owns the TrimBusEvents retention floor

victorarias:
One assignment the reconcile design doc left implicit after #909: the `TrimBusEvents` floor change — the code that makes installed apps pin retention whether enabled or disabled — now lives in slice 3's bullet, beside the fence cleanup that slice already owned. The two are halves of one promise (keep the facts, deliver the facts) and land together. The per-lane cap remains excluded until its number has a measured receipt, per the doc's retention section.

Victor ruled the assignment on ticket app-reconcile, 2026-08-15.

https://claude.ai/code/session_01FEGuaBJNKQmpxHbG1LJSyP

### DIFF
--- a/changelog.d/slice3-owns-retention-floor.yaml
+++ b/changelog.d/slice3-owns-retention-floor.yaml
@@ -1,0 +1,6 @@
+kind: internal
+area: apps
+change: >
+  The reconcile design doc's slice plan now assigns the TrimBusEvents
+  retention-floor change to slice 3, beside the fence cleanup it already
+  owned, so the retention policy lands as code in one slice.
--- a/docs/plans/2026-08-14-ext-app-reconcile-handler.md
+++ b/docs/plans/2026-08-14-ext-app-reconcile-handler.md
@@ -475,7 +475,11 @@ proof.
    the end of this doc as code, both halves of one promise: keep the facts —
    `TrimBusEvents` floors at the minimum cursor across installed app
    consumers, enabled or disabled, with "disabled must not pin" narrowed to
-   rows no install serves — and deliver the facts: slice 1's as-built cleanup
+   rows no install serves, replacing the shipped test that pins the old
+   contract (`TestTrimBusEventsIgnoresDisabledConsumers`) with its witness: a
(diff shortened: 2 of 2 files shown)


figgyster (approved):
The assignment holds up. Checked it against the doc line by line: the keep half matches the Decision section (installed app's consumer pins the floor enabled or disabled; "disabled must not pin" narrowed to rows no install serves), the deliver half matches slice 1's note (re_enabled insert and enum value deleted, version_changed re-fenced to cursor, gap to earliest - 1), and the per-lane cap stays parked until lane growth produces a receipt. Changelog entry tells the truth, and the Changelog check passes.

One soft note, same family as the ones this doc keeps collecting. The three witnesses named in slice 1's note all pin the deliver half. The keep half — the actual TrimBusEvents floor flip — ships with no named witness, and it reverses a shipped test that pins the old contract: TestTrimBusEventsIgnoresDisabledConsumers (internal/store/bus_test.go:288) asserts a disabled consumer must not pin the log, against the comment at internal/store/bus.go:255 ("Disabled consumers are excluded from the floor on purpose"). Slice 1's note explicitly says its cleanup replaces tests that pin shipped behavior with witnesses; the trim half has the same exposure and no replacement named — and TrimBusEvents currently has no notion of install at all, so the floor change needs the installed-vs-orphaned distinction somewhere a test can see it. The natural witness: a disabled installed app's rows survive a trim that would have deleted them, while orphaned rows still trim. Slice 4's exit proof exercises the promise end-to-end, but that's a packaged-app proof; the focused synctest witness is what the fence half got and the trim half is missing.

figgyster (approved):
The keep half got its witness, and it's the one I'd have written: the shipped test is named (`TestTrimBusEventsIgnoresDisabledConsumers`), and its replacement witness spells out the installed-vs-orphaned distinction the floor change lives or dies on — a disabled installed app's rows survive a trim that would have deleted them, while orphaned rows still trim. Both halves of the promise now carry named witnesses, and the note→witness loop closed on the same commit. Re-approving at f8eb978.

PR #912, opened 2026-08-14
feat(crew): write a member's name as a name, everywhere you read one

victorarias:
A crew member is identified by a lowercase id — `trellis`, `keel`, `alder` — and
every surface printed that id raw. The id was doing two jobs at once: it is an
address, and it is somebody's name. This separates them with one rule.

**Display capitalizes, identity does not.** You type `attn crew wake trellis`
and Trellis is who answers.

## The rule

`internal/crew.DisplayName` is the rule; `app/src/utils/crewName.ts` is the
frontend's copy of it, for the one thing the app names without the daemon — a
sleeping member's row. First rune uppercased; no title-casing of a hyphenated id,
because no real member is two words and inventing one would be a name nobody
chose. `internal/crew/priming.go` had grown its own private `capitalized` in
#910; that is now the same function, so there is one rule and not two.

## Surfaces

Routed through it — everything a person reads:

- **Sidebar.** A woken member's row (from the label the daemon writes) and a
  sleeping member's row, its wake button, and both tooltips.
- **Wake.** The session label, the pane title, and the member's workspace title.
  Stored capitalized rather than capitalized at draw: the label *is* the display
  string, it is user-renameable, and every generic reader of a session label —
  sidebar, tab, `attn agent list`, `agent peek` — then gets the name for free.
  It does not collide with the auto-titler, which only fires on a label equal to
  the cwd basename — capitalizing the label is in fact what makes a woken member
  differ from its home's basename, so the label check catches one on its own.
  Going forward only: a member awake when this lands keeps its lowercase label
  until its day ends and the next wake writes the name. Existing labels are not
  migrated.
- **CLI.** `crew list`'s MEMBER column, `crew wake` and `crew set` output,
  `agent list`'s MEMBER column, `agent peek`, `handoff`'s filed/asleep/next-day
  lines, and the seed log's tender, planter, and note/handoff authors.
- **Refusals.** The wake limit, the already-awake claim, a cwd that moved, the
  nap's retry and letter refusals, and the garden's one-tender-at-a-time refusal.
- **Daemon log.** Every `crew:` line that names a member as a subject —
  import, bind, release, wake, priming, warmed, asked-to-sleep, napped.
- **Toasts.** `Failed to wake …` and the wake timeout.
- **Glossary.** The rule is stated where the crew is defined.

Deliberately left lowercase — everything that addresses something:

- Home paths (`~/.attn/crew/trellis/`), handoff file names, the workspace id
  (`workspace-crew-trellis`), the doc id, `--member`, CLI arguments, every
  protocol field (`member`, `crew_member`, `tender_member`, …), the store, and
  the `data-crew-member` / test-id hooks the app and its tests reach rows by.
- The `%q`-quoted echo of a name that resolved to nobody: that is the user's own
  text handed back, not a member.
- The command inside a refusal. "Alder launches in … `attn crew set alder --cwd
  <dir>` moves it" — prose on the left, the id the command takes on the right.
- A holder that is a bare session id. Where the same slot can carry either a
  member or a session (`tended by`, note authors), only the member branch is
  capitalized; a session id is never dressed up as a name.

Where a holder can be either a member or a bare session — a seed's tender,
planter, and note authors — one rule answers on each side: `crew.HolderName` and
`crewHolderName`. `garden.Tender.DisplayName` is that call; `Tender.Name` stays
raw because it doubles as the identity and emptiness test.

Case-insensitive lookup already existed (`crew.Resolve` folds case), so
`attn crew wake Trellis` works and is covered.

## Verification

`make test-quick` and the full frontend suite are green. New tests: the rule
itself on both sides, the wake writing a name while the ids stay lowercase, the
`crew list` column beside its untouched home path, the sidebar row beside its
`data-crew-member`, and `Tender.DisplayName` beside `Tender.Name` — which stays
raw, because it doubles as the identity and emptiness test.

Live on a throwaway profile (`nameday`), two homes on disk: `crew list` shows
`Keel` / `Trellis` against `/Users/victor/.attn-nameday/crew/keel`, `crew list
--json` carries `"id": "keel"`, `agent list` shows label `Trellis` and member
`Trellis` while `--json` carries `"member": "trellis"`, and the daemon log reads
`crew: woke Trellis in session …`. The clip is the sidebar: Keel asleep, one
click on its wake button, and the row comes up awake as Keel with its pane
titled Keel over `/Users/victor/.attn-nameday/crew/keel`.

clip

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/533f6a2793f3f0cfbcdbb5ceb9ea7a6a7ad48a02/delegate-member-names-d6788e4e/clip.mp4)

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -7,6 +7,7 @@ import { openPath, openUrl } from '@tauri-apps/plugin-opener';
 import { Sidebar, type SidebarHeaderAction, type DockItem, WorkflowIcon, EditorIcon, PRsIcon, NotebookIcon } from './components/Sidebar';
 import { Dashboard } from './components/Dashboard';
 import { activityStaleMs } from './utils/activitySettings';
+import { crewDisplayName } from './utils/crewName';
 import { AttentionDrawer } from './components/AttentionDrawer';
 import { LocationPicker } from './components/LocationPicker';
 
@@ -3468,7 +3469,7 @@ function AppContent({
   const handleWakeCrewMember = useCallback((member: string) => {
     sendCrewWake(member)
       .then((result) => handleSelectSession(result.sessionId))
-      .catch((error) => showError(error instanceof Error ? error.message : `Failed to wake ${member}`));
+      .catch((error) => showError(error instanceof Error ? error.message : `Failed to wake ${crewDisplayName(member)}`));
   }, [sendCrewWake, handleSelectSession, showError]);
 
   // The daemon-facing ticket actions the pane-header ticket overlay wires into
--- a/app/src/components/GardenPanel.lifecycle.test.tsx
+++ b/app/src/components/GardenPanel.lifecycle.test.tsx
@@ -39,7 +39,7 @@ describe('GardenPanel lifecycle', () => {
     render(<GardenPanel isOpen onClose={vi.fn()} seedsTotal={1} seeds={[growing]} />);
 
     expect(screen.getByText('growing')).toBeInTheDocument();
-    expect(screen.getByText('tended by trellis')).toBeInTheDocument();
+    expect(screen.getByText('tended by Trellis')).toBeInTheDocument();
   });
 
   // A session id is not a pretty name, but "somebody holds this" is the fact the
@@ -87,7 +87,7 @@ describe('GardenPanel lifecycle', () => {
       />,
     );
     expect(screen.getByText('growing')).toBeInTheDocument();
-    expect(screen.getByText('tended by trellis')).toBeInTheDocument();
+    expect(screen.getByText('tended by Trellis')).toBeInTheDocument();
 
     rerender(
       <GardenPanel
--- a/app/src/components/GardenPanel.tsx
+++ b/app/src/components/GardenPanel.tsx
@@ -12,6 +12,7 @@
 // garden, so navigation costs no round trip.
 import { useMemo, useState } from 'react';
 import type { Seed } from '../hooks/useDaemonSocket';
+import { crewDisplayName } from '../utils/crewName';
 import './GardenPanel.css';
 
 interface GardenPanelProps {
@@ -112,7 +113,7 @@ function relationsOf(index: GardenIndex, id: string): Relation[] {
 // session id is not pretty, but "somebody holds this" is the fact the panel owes
 // the reader — showing nothing would read as unclaimed.
 function tenderOf(seed: Seed): string {
-  return seed.tender_member || seed.tender_session || '';
+  return seed.tender_member ? crewDisplayName(seed.tender_member) : seed.tender_session || '';
 }
 
 // crownOf is the crown a seed is part of, or '' when it stands on its own.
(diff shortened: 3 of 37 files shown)


figgyster (approved):
Approve. The rule is one, and I checked every door it's supposed to reach.

**What I verified against the tree:** the `crew:` log sweep is clean — all 40 lines route through `DisplayName`; the only raw ids left are the two `%q` echoes you called out (unresolved name handed back, registry-record race) plus the command-inside-refusal split in `crewLaunchDir` (prose says Alder, `crew set alder` takes the id, and the test pins both halves). The auto-titler non-collision is real: label "Trellis" vs basename "trellis" means the label check now catches a freshly-woken member session by itself. `firstNonEmpty`'s member call sites are all gone. And the concurrent-claims test change is the honest fix — the old test claimed as phantom members named `sess-a`/`sess-b`, which `DisplayName` would have capitalized into "Sess-a" and broken the refusal assertion; the bare-session form is what that test always meant.

**Witnesses, run here:** Go — `internal/crew`, `internal/garden`, `cmd/attn` full suites green; `internal/daemon` crew/garden/wake/bind/prime/claim green. Frontend — full suite, 2839 passed, `tsc --noEmit` clean, `git diff --check` clean. The daemon package's only two failures are the plugin-driver fixture tests, which fail identically on the parent commit in this sandbox and are path-skipped in CI anyway. CI's green all the way down.

Three notes, none blocking:

1. **Stale comment, inverted premise.** `sessionMayBeAutoTitled`'s doc (internal/daemon/session_title.go:122-126) still says the member check can't fold into the label check because "a member launches in its own home, so its name and its default label are the same string." This PR makes that false — the name is "Trellis", the basename is "trellis" — and the comment's reasoning is now exactly backwards: the label check is what catches member sessions, and the member check survives only as a backstop for labels renamed back to the basename. Mechanism's correct; the prose asserts the old invariant while the code implements the new one. The prose-vs-mechanism loop's been catching this at design stage all week; here it's a shipped comment contradicting the rule you just shipped.

2. **One-day grace for live sessions.** "Stored capitalized rather than capitalized at draw" is going-forward: a member's session that's awake when this lands keeps its old lowercase label in the sidebar, tab, and `agent list` until its day ends and a wake writes the name. Cosmetic, self-heals, no data issue — but worth a sentence in the body ("existing labels are not migrated") or a boot-time capitalization of labels that exactly match a registered member id, if the grace bugs you.

3. **The picker around the rule is four.** `holderName` in cmd/attn/seed.go, the same `holderName` in internal/daemon/garden.go, `garden.Tender.DisplayName`, and the TS `tenderOf` in GardenPanel — four copies of "member reads as a name, session stays as it is," under a PR whose whole thesis is one rule and not two. The rule is one; the picker is four. A shared `crew.HolderName(member, session)` (or folding the two string-y call sites onto `Tender.DisplayName`) would make the thesis true end to end. Same breath: the conversion orphaned two helpers — `firstNonEmpty` (cmd/attn/seed.go:716) and `firstNonEmptyString` (internal/daemon/garden.go) — zero callers left, both safe to delete.

figgyster (approved):
Approved. All three notes closed in one commit (7699510), and I verified each fix in the diff.

- The auto-titler comment now states the corrected invariant: the label check catches a woken member on its own (label is the name `Trellis`, the home basename is the id `trellis`), and the member check stays as the backstop for a label renamed back to that basename.
- The picker is one rule per language — `crew.HolderName` and `crewHolderName` — with both `holderName` copies, `Tender.DisplayName`'s body, and the orphaned `firstNonEmpty`/`firstNonEmptyString` helpers gone.
- New witnesses landed beside each rule (`crew_test.go`, `crewName.test.ts`).

Checks on this head: frontend and E2E green, Linux daemon builds green, Bench A/B green; the daemon job's failures are the same plugin-driver env flakes that fail identically on the parent commit. Full loop breakdown is in my comment on the thread — this is the formal stamp on the current head.

PR #913, opened 2026-08-14
feat(crew): address a member whether awake or asleep

victorarias:
## TL;DR

`attn agent msg` can now address a durable crew member by name. It delivers to the member's current day when awake, or performs a wake-limit-bounded wake and carries the attributed message as the sleeping member's first ask after charter and handoff priming.

## Why

Crew members outlive their sessions, but messaging previously required the caller to discover whichever session happened to hold the member's current day. That made “tell Keel X” depend on lifecycle state and left asleep members unaddressable.

## What changed

The daemon now resolves an `agent msg` target as a crew member before falling back to session ID/prefix resolution:

- An awake member routes through the existing attributed, durable session-message path.
- A sleeping member goes through the lifecycle's autonomous wake gate. Refusal preserves the lifecycle error—including the configured limit and way out—and adds that nothing was delivered.
- An accepted message is persisted before the session process starts and replaces the generic wake greeting as the launch prompt. Crew priming still runs first, so the charter and predecessor letter precede the attributed ask.
- Crew wakes serialize through durable session creation. Simultaneous messages to one sleeping member share the first new day instead of racing two claims, two wake charges, and two identities.
- The CLI returns `queued` once the new session is running. A prompt-submit hook, rather than the worker's early `working` signal, stamps delivery; Claude can report worker activity while its trust dialog still blocks the prompt. After a daemon restart, the normal durable queue is the conservative fallback, preferring a possible duplicate over loss.
- Messages that arrive after the session is created but before its initial prompt starts queue behind that prompt, then drain as soon as the prompt-submit receipt opens the gate.
- Unknown addresses now name both address books: `attn agent list` for sessions and `attn crew list` for members.

The protocol message and result shapes are unchanged—the existing target string now accepts a member name—so this does not bump the protocol version. The app UI, wake model/prose, auto-sleep, heartbeat, and lifecycle thresholds are unchanged.

## Manual verification

In an isolated `crew-msg` profile using the bundled CLI and daemon:

- Addressed awake Keel by name and observed the attributed message arrive in the existing session.
- Explicitly ended Keel's day with a sleeping handoff, then addressed `keel` again. The command returned `queued` with a new session; Keel replied after launch and explicitly reported reading its charter and predecessor letter before handling the message.
- Set `crew.wake_limit=0` and addressed sleeping Keel. The command refused with the limit name/value, the member, the sidebar/setting way out, and “nothing was delivered”; Keel remained unbound and no matching message row existed.
- Sent two simultaneous commands to sleeping Keel. Both resolved to the same new session, only one performed the wake, and both durable receipts and exact-token replies appeared once.
- After rebasing the member-display-name change, addressed lowercase `keel` while asleep. The result said `woke Keel`, the spawned session/workspace rendered `Keel`, JSON retained the lowercase member identity, and the attributed round trip completed with a durable receipt.

The full quick suite and Linux amd64 cross-build pass.

### DIFF
--- a/changelog.d/crew-addressing.yaml
+++ b/changelog.d/crew-addressing.yaml
@@ -1,0 +1,10 @@
+kind: added
+area: crew
+change: >
+  `attn agent msg <member> "text"` now addresses a crew member whether it is
+  awake or asleep. An awake member receives the same attributed message as a
+  session target; a sleeping member wakes on its pinned model and receives the
+  message as the new day's first prompt after its charter and predecessor's
+  letter. Message-triggered wakes use the autonomous wake limit, whose refusal
+  names the configured limit, the attempted member, that nothing was delivered,
+  and how to proceed.
--- a/cmd/attn/agent.go
+++ b/cmd/attn/agent.go
@@ -298,7 +298,7 @@ type agentMsgArgs struct {
 }
 
 func parseAgentMsgArgs(args []string, envSessionID string) (agentMsgArgs, error) {
-	const usage = "usage: attn agent msg <id> \"text\" [--source-session <id>]"
+	const usage = "usage: attn agent msg <session-or-member> \"text\" [--source-session <id>]"
 	literal := len(args) > 0 && args[0] == "--"
 	if literal {
 		args = args[1:]
@@ -382,7 +382,13 @@ func agentMsgOutcomeLine(result *protocol.AgentMsgResult) string {
 // mistyped one knows which.
 func agentMsgErrorMessage(parsed agentMsgArgs, err error) string {
 	message := strings.TrimSpace(err.Error())
-	switch strings.TrimSpace(strings.TrimPrefix(message, "daemon error: ")) {
+	code := client.ErrorCode(err)
+	if code == "" {
+		code = strings.TrimSpace(strings.TrimPrefix(message, "daemon error: "))
+	}
+	switch code {
+	case "session_or_crew_member_not_found":
+		return fmt.Sprintf("no session or crew member matches %q; `attn agent list` names sessions and `attn crew list` names members", parsed.target)
 	case "session_not_found":
 		return fmt.Sprintf("no session matches %q; `attn agent list` names the sessions on this daemon", parsed.target)
 	case "ambiguous_session":
@@ -406,12 +412,13 @@ commands:
         observe a session without interrupting it: state, todos, last
         assistant message, and the rendered screen. Passive — the observed
         agent never notices. <id> is a full session id or a unique prefix.
-  msg <id> "text" [--source-session <id>] [--json]
-        send a session a message. It lands in that session's conversation,
-        attributed to you, carrying the command to reply with. A target that
+  msg <session-or-member> "text" [--source-session <id>] [--json]
+        send a session or crew member a message. It lands in the live session,
+        or wakes a sleeping member and becomes its first prompt after priming.
+        It is attributed to you and carries the command to reply with. A target that
         cannot take input right now has it queued and delivered when it can;
         the result always says which. The sender defaults to this session
         (ATTN_SESSION_ID); pass --source-session when running outside one.
-        A message that starts with - goes after --, as: agent msg -- <id> "-text"
+        A message that starts with - goes after --, as: agent msg -- <session-or-member> "-text"
 `)
 }
--- a/cmd/attn/agent_test.go
+++ b/cmd/attn/agent_test.go
@@ -265,6 +265,12 @@ func TestAgentMsgErrorMessageSeparatesTargetFromSender(t *testing.T) {
 	if !strings.Contains(sender, `"yyyy"`) || !strings.Contains(sender, "sender") {
 		t.Fatalf("sender error = %q", sender)
 	}
+	unknown := agentMsgErrorMessage(parsed, errors.New("daemon error: session_or_crew_member_not_found"))
+	for _, want := range []string{`"zzzz"`, "attn agent list", "attn crew list"} {
+		if !strings.Contains(unknown, want) {
+			t.Fatalf("unknown-address error %q does not name %q", unknown, want)
+		}
+	}
 }
 
 // A message far past the limit never reaches the daemon's refusal: the socket
--- a/cmd/attn/main.go
+++ b/cmd/attn/main.go
@@ -647,7 +647,7 @@ commands:
   presence                          check whether the current shell runs inside attn
   agent list                        name every session running here
   agent peek <id>                   read a session without interrupting it
-  agent msg <id> "text"             send a session a message, attributed to you
+  agent msg <session-or-member> "text"  message a live session or durable crew member
 	  session <command>                 inspect a session's conversation
   state explain <id>                replay why a session's state is what it is
   delegate --brief-file <path>      start another agent with a delegated brief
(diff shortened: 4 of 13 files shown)


figgyster (approved):
Good case, detective. The crew arc's messaging layer — `agent msg` grows a member address, and the wake-and-deliver is one operation, not a wake plus a prayer.

**Verified, the load-bearing half:**

- **Resolution precedence is documented and right.** Member exact-name match wins over a coincidental session prefix (internal/daemon/agent_msg.go:99-102, crew.Resolve is exact case-folded, not prefix); the fence means a direct session address still works on an outpost. And the content guards run before any resolution, so an empty or oversized message refuses without ever waking anyone — a wake is never triggered by garbage.
- **Wake-and-deliver is one transaction with receipts at every door.** The row is persisted before spawn (crew_wake.go:262-268), the initial prompt replaces the greeting with the attributed message, and every failure point rolls back: enqueue failure removes pane and releases binding, spawn failure deletes the row via a write-once guard (delivered_at = ''), releases the binding, removes the pane. The refused wake charges the autonomous ledger *before* the claim (crew_wake.go:222-226), so a loop that keeps failing spends the budget exactly like the comment says.
- **The receipt is the right receipt.** Worker `working` does not stamp — a freshly spawned Claude reports working while sitting at its trust dialog — only a hook's working evidence (UserPromptSubmit, far side of the dialog) does. The test pins both directions: applyState(working) leaves the row queued and the PTY untouched; handleState(working) stamps and opens the drain gate. That's the strongest witness in the PR, and it's the difference between this and a paste-and-pray.
- **Unknown addresses name both address books**, the ErrorCode rides the wire into a DaemonError, and the CLI's fallback still handles code-less refusals. Both sides tested.
- **No protocol bump** — target string semantics only, shapes unchanged. Changelog fragment and glossary entry follow project rules. `git diff --check` clean.

**Ran the box:** store, client, protocol, and cmd suites green; the daemon package green except two plugin-driver env failures that fail identically on parent daceb599 — same pre-existing flake family as #912, not this PR. The new wake tests all pass.

Two notes, neither blocking:

1. **The in-flight claim window is newly reachable by automated senders.** The claim's liveness check (crew.go:192-196) can't see a binding whose session hasn't reached store.AddChecked yet (spawn_pipeline.go:373). Two wakes for one sleeping member arriving inside that window both pass, both charge the budget, and the second claim cleanly steals the binding — no docstore conflict, so the "two launches contending is one retry" design never fires. The first session then launches wearing the member's label but unprimed (crew_prime finds nobody bound to it), with its message delivered to a session that isn't really the member. The window is tens of milliseconds and the shape predates this PR (manual wakes share it), but message wakes are the first *automated, unattended* caller — a parallel send or a retry loop can walk through it. Worth a waking-marker or per-member serialization in a follow-up; the invariant "two agents with the same identity never run at once" deserves a fence it can't miss.

2. **One sentence on the restart story.** "The normal durable queue is the conservative fallback" holds while the woken session survives (resume covers daemon restarts); if the day ends before the hook fires, the row targets a dead session and no mechanism re-points it at the member's next day — that's the queue's pre-existing dead-session property, not a regression, but the doc's "preferring a possible duplicate over loss" quietly assumes a session that keeps reporting. Fine as-is; just don't let future-you read the comment as covering the dead-day case.

Approved. The crew arc keeps consolidating — wake pinned (#910), names as identity (#912), and now a member is an address you can hand words to while it sleeps. The prose-vs-mechanism loop has fully grown teeth: this one's mechanism all the way down, and the receipts are witnesses, not rumors.

figgyster (approved):
The note came back as a fence, not a patch. 4b7e58e4 answers the concurrent-wake finding the right way.

**Verified on the new head:**

- **The mutex is the right shape of fix.** `crewWakeMu` holds from the liveness read through durable session creation (crew_wake.go:194-199), so a second wake entering the claim→`AddChecked` gap blocks, then resolves to the first day as `AlreadyAwake` — one claim, one charge, one identity. The fence covers every caller: manual wakes and message wakes serialize on the same lock, which is exactly the race I flagged. No lock-order cycle — `crewWakeMu` → `agentMessageMu` is the only direction, and the state/hook path never takes them in reverse.
- **The adjacent edge you found is the better catch.** A message landing after session creation but before the initial prompt cleared would have pasted straight into priming or a trust dialog. The new pending gate in `deliverAgentMessage` (agent_msg.go:247-249) queues it behind the first ask with an honest detail string, and the submit receipt now schedules the drain instead of just clearing the flag — so the backlog lands the moment the hook opens, not on the next state report. `isNudgeDeliveryAllowed` passes Working, so the hook-path drain fires.
- **The witnesses are deterministic and pin the real invariants.** `TestCrewWake_ConcurrentWakesShareTheFirstDay` proves one launch, one claim, one spawn, and the shared session with the seam hooks; `TestHandleAgentMsgDuringWakePrimingDrainsAfterTheInitialPrompt` proves nothing jumps ahead of priming, the receipt opens the drain, and the doorbell pastes exactly the second ask, never the first.

**Ran the box on 4b7e58e4:** full daemon suite green (excluding the two pre-existing plugin-driver env failures, which fail identically on parent daceb599 — same flake family as #912 and the first round of this PR), vet clean, diff-check clean. PR body owns both behaviors, and the manual verification now includes the two-simultaneous-sends run.

My second note — the dead-day-before-hook strand — was "fine as-is" and stays that way; nothing in this commit changes the restart story, and the queue's dead-session property is documented elsewhere.

Re-approved at 4b7e58e4. That's the note→witness loop closing on this arc again — the wake path now has a fence it can't miss, and the receipts got sharper with it.

victorarias:
@figgyster Rebased #913 onto current main after #912. The conflict keeps #912's crew.DisplayName session label and #913's attributed initial prompt; new message-result strings now render member names through crew.DisplayName too. make test-quick passes, and the isolated full-app run addressed lowercase keel, rendered Keel in the result/session/workspace, preserved member=keel in JSON, and completed the attributed round trip. Ready for a fresh pass on 38c70afa.

figgyster (approved):
The rebase did the right kind of damage: none. 38c70afa sits on #912's merge with the naming contract fully folded in, and the new commit closes the last surface I was watching.

**Verified on the new head:**

- **The rebase inherits the #912 contract cleanly.** Wake labels are now stored capitalized (`Label: crew.DisplayName(member.ID)`), the pane/wake error messages name the member properly, and the wake-limit refusal reads "Alder" not "alder" — the ledger's `Allows` already names the display name in the new base, so only the test expectation needed adapting. No regression, no drift; the feature and the mutex fence from 4b7e58e4 are intact under the new base.
- **38c70afa closes my #912 watch item.** The message *results* were the one surface still printing the raw member id to a person — "woke keel in session…", "delivered to keel". Now both the wake detail and the delivered-to detail route through `crew.DisplayName`, with two new assertions pinning "woke Trellis" and "delivered to Keel". That's the naming layer's one-rule-per-language thesis extended to the last new surface this arc created.
- **Ran the box on 38c70afa:** full daemon, store, and cmd suites green (excluding the same two pre-existing plugin-driver env flakes that fail on parent), vet clean, diff-check clean. The PR stays 13 files, no protocol bump — target-string semantics and display names only.

Re-approved at 38c70afa. The naming loop and the wake fence loop both landed in one forced update — the crew messaging layer reads finished: every surface a person reads names the member, every wake shares one day, and every receipt is a witness.

PR #914, opened 2026-08-14
Show live session cost beside agent names

victorarias:
## TL;DR

Agent pane headers now show the live USD cost of each session. Usage is extracted from provider transcripts, accumulated durably, and priced by exact model; any unpriced traffic makes the whole session say `unknown` instead of looking free.

## Why

attn previously discarded provider token usage while reading transcripts, so there was no cost fact for the daemon or app to display. The important constraint is epistemic: a missing or ambiguous price must never become a plausible-looking `$0.00`.

## What changed

The existing transcript watcher is now the accounting seam:

```text
Claude/Codex JSONL append
  -> provider-specific usage extraction
  -> durable per-session observations + per-model token ledger
  -> exact-model USD pricing
  -> Session.cost_usd / Session.cost_unknown
  -> quiet value beside the agent name
```

Claude messages contribute input, output, cache-read, 5-minute cache-write, and 1-hour cache-write counters. Repeated or revised message snapshots are absolute upserts. Codex `token_count` records contribute `last_token_usage`; cached input is subtracted from input before applying the lower cache-read rate, and the active `turn_context.model` owns each observation.

The ledger and transcript cursor live with the session in SQLite, so they survive daemon restarts and model changes can be repriced correctly. Accounting is forward-only from this install: existing and explicitly resumed provider transcripts seed at their current head, fresh sessions count from their first record, and an invalidated cursor seeds at the replacement head rather than risking a historical replay.

Pricing ships with an exact-model built-in table whose rates cite provider pages and the date checked. `session_cost.price.<exact-model-id>` settings accept a complete USD-per-million-token rate card and either fill a gap or replace a stale built-in entry. Cache reads, 5-minute writes, and 1-hour writes remain separate rates.

The wire now distinguishes all three UI states:

| Observed state | Wire | Header |
|---|---|---|
| No token usage yet | both fields absent | nothing |
| All used models priced | `cost_usd` | `$1.23`; positive sub-cent values use `<$0.01` |
| Missing/ambiguous price or usage unavailable | `cost_unknown: true` | `unknown` |

The protocol moves from 249 to 250 in both lockstep clients, generated from the TypeSpec schema.

## Review path

1. `internal/transcript/usage.go` and its captured fixtures: provider semantics and stable observations.
2. `internal/sessioncost`: exact-model rates, strict overrides, cache math, and unknown propagation.
3. `internal/store/session_cost.go` plus `internal/daemon/transcript_watcher.go`: durable forward-only accumulation and wire invalidation.
4. `SessionCost.tsx` and `SessionCostPriceSettings.tsx`: the restrained header value and operator overrides.

## Manual verification

In an isolated profile running protocol 250, a real one-turn Codex 0.147.0 session using `gpt-5.4-mini` produced this transcript usage:

- input: 19,354 total, including 4,480 cached, so 14,874 uncached
- output: 30
- rates: $0.75 input, $0.075 cached input, and $4.50 output per million tokens

The hand calculation was `(14,874 × 0.75 + 4,480 × 0.075 + 30 × 4.50) / 1,000,000 = $0.0116265`. The daemon persisted and sent exactly `0.0116265`; the header rendered `$0.01` beside the session name.

I then seeded one input token under an unpriced `future-unpriced-live` model in the disposable profile and restarted the daemon. The wire sent `cost_unknown: true` with no `cost_usd`, and the same header rendered `unknown`.

## Out of scope

No budgets, alerts, spend limits, crew/member/day aggregation, or image/audio pricing were added. Shell sessions and sessions with no observed usage remain blank.

## Evidence

[Live verification recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/adcced9e8a9c1a48a0c6be6c656f883f0c386389/delegate-session-cost-17642c6c/clip.mp4) — isolated `sessioncostqa` profile: real priced Codex session with the live cost header beside the agent name.


figgyster (approved):
Verified against the tree, not just the diff. This is the accounting layer the transcript watcher was always meant to have, and the epistemic discipline holds: absolute-upsert observations keyed by stable ids make replay idempotent, the forward-only cursor never backfills, the fresh-vs-resumed marker split is exactly right, and one unpriced model poisons the whole session instead of leaving a plausible-looking partial total. Ran the witnesses: transcript/sessioncost/store green, daemon green minus the same two plugin-driver env flakes that fail identically on parent, full frontend suite 2845 green, tsc clean, diff --check clean. Hand-checked the wire math too — 0.0116265 against the gpt-5.4-mini rates lines up.

Three notes, none blocking:

1. **Codex reasoning tokens are billed but never priced.** Your own fixture proves `output_tokens` excludes them: both token_count records satisfy `total = input + output`, with `reasoning_output_tokens` sitting outside (33438 = 32875 + 563, 73058 = 72198 + 860). The extractor reads input/cached/output only (internal/transcript/usage.go), so reasoning traffic never reaches the ledger, and OpenAI bills reasoning at output rates for the gpt-5.x family. Small per record, but a systematic undercount on reasoning-heavy sessions — and the one direction your contract doesn't cover, since the number still reads as exact. Either fold `reasoning_output_tokens` into output or state the exclusion in the doc and changelog.

2. **A corrupt ledger reads as "no cost", not "unknown".** `decorateSessionWithCost` swallows the store error and returns, so a session that has spend shows blank if its `session_cost_json` ever fails to decode — the same category of plausible lie the "never look free" rule was written to kill. Rare, daemon-written data, but the decode-failure path could map to `cost_unknown` for a nickel.

3. **The observations map never compacts.** One entry per provider request, and every batch rewrites the whole blob (internal/store/session_cost.go). Fine for ordinary sessions; a long-lived heavy one grows the row without bound. A future compaction that folds committed records into the ledger alone would need the replay-idempotence argument re-made — debt entry, not tonight's fix.

The settings UI matching the daemon's strict parse — unknown fields rejected, incomplete cards refused, invalid saved cards kept visible for repair — is the symmetry that usually arrives a release late. Merge it.

figgyster (approved):
Both answers check out, and I owe the second one a correction.

The corrupt-ledger fix is exactly right: non-ErrNoRows decode failures now project `unknown`, and the ErrNoRows carve-out keeps a deleted session's last broadcast blank instead of shouting about cost it no longer has. Ran the new regression test against a genuinely corrupted SQLite row — passes, and the full cost/spawn daemon set stays green on 912c8f8f.

On reasoning tokens — you're right and I was wrong. I read `reasoning_output_tokens` as additive to `output_tokens`; the receipt shows it's a subset breakdown, and `total = input + output` (33,438 = 32,875 + 563) holds precisely because output already carries the reasoning. That's the API's own usage convention. Pricing output once is correct; adding reasoning again would double-charge. The invariant comment beside extraction and the fixture-test pin are the right witnesses — glad you made it explicit rather than letting my misreading linger in the code.

The observation-map compaction stays as agreed: durable request-level dedup is load-bearing for the absolute-upsert contract, and growing with request count is scaling work outside this slice. Worth a debt-register line so it isn't rediscovered in a year.

PR #915, opened 2026-08-15
feat(garden): open seeds as annotated documents

victorarias:
## TL;DR

Seeds are now first-class reader documents. `attn open <seed-id>` docks the seed's rendered Markdown body with its live child/note ledger, while the Garden panel offers the same document as a read-only drill with an **Open as tile** handoff. The existing annotation stack is shared unchanged; document identity is now URI-based and daemon authority remains in typed source fields.

## Why

The garden epic needs one canonical reading surface before later editing and submission work can land. Previously, the panel exposed a raw body preview and seeds could not enter the workspace reader at all, so plans stored as seeds could not use the annotation workflow that Markdown files already had.

## What changed

- Added a real `seed` workspace tile kind and `open_seed` command. The daemon validates the seed, docks or rebinds one stable tile, and reports unknown IDs with the requested ID in the error.
- Added a seed-document read model containing the seed, immediate children, the full bounded note log plus its total, and a live invalidation path through existing garden broadcasts. The shared `SeedDocumentView` renders it in both the tile and panel.
- Generalized MarkdownReader's source from an overloaded filesystem path to a discriminated file/seed source. Draft correlation is `<op>:<documentUri>`; every daemon action still receives and validates typed `source_kind`, workspace/path, or seed ID fields and never parses the URI for authority.
- Preserved file annotation behavior and payloads. Seed submissions use the same composer, picker, send lifecycle, anchoring, highlighting, and quick-label contracts.
- Bumped the lockstep protocol from 249 to 250 and generated both Go and TypeScript contracts.

The current seed artifact projection is intentionally dormant: the store has no attach/detach entries or verbs yet, so the projection renders nothing. Its typed Markdown-file affordance is present and will open an artifact as its own file tile when those entries exist.

## Review path

1. `app/src/components/MarkdownReader/documentSource.ts` and the annotation transport changes establish the two-source reader seam and URI identity.
2. `internal/daemon/ws_markdown_annotations.go` shows typed source validation, canonical identity derivation, routing, and draft storage.
3. `internal/daemon/tilecontent.go` plus `internal/daemon/garden.go` own seed tile placement and the seed-document read model.
4. `app/src/components/SeedDocumentView.tsx`, `WorkspaceDockTile.tsx`, and `GardenPanel.tsx` show the shared annotated/read-only surfaces.

## Manual verification

On an isolated `seedreader` profile, I planted a seed with real Markdown, a child, and notes. The CLI opened it as a tile; the body rendered through MarkdownReader, the ledger showed the child and notes, and a new note appeared in an already-open panel drill without polling. A saved seed annotation reached the tile's bound Codex session and the draft cleared after submission. I also opened `README.md`, annotated it, and confirmed the existing file submission path still reached the same session and cleared.

## Out of scope

- Live re-anchoring from seed body edit events (ruling C).
- Split-button submit and untended-seed destination behavior (ruling D).
- Attach/detach verbs, delegation reporting, ticket retirement, signposts, or conversion.
- A durable history of lifecycle transitions; the current store exposes live child states and append-only notes, which are the ledger this slice can honestly render.

## Evidence

[Seed reader drill recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/adcced9e8a9c1a48a0c6be6c656f883f0c386389/feat-garden-seed-reader/clip.mp4) — isolated profile: seed opened as an annotated document, GardenPanel drill, live ledger updates.


figgyster (approved):
# Markdown Annotations

internal/daemon/workspacelayout.go

I've reviewed this document and have 1 piece of feedback:

## Retargeting a seed tile's Send target succeeds, but the daemon reports failure

**Lines 627–662** — the `workspace_layout_update_tile` retarget path.

When you retarget a seed tile (the session picker in the tile header), the daemon:

1. Validates and calls `rebindTileSession` — which **persists the new binding and broadcasts the layout** (line 645).
2. Falls through to the params-update switch, because the markdown short-circuit at line 652 only covers `TileKindMarkdown`.
3. Hits the `default` case and replies `tile parameters cannot be updated for tile kind "seed"` (line 664).

I confirmed this empirically with a throwaway daemon test: the leaf's `TileSessionID` is `session-2` after the call, while the result is `success=false`. The client's optimistic pick rolls back, then the broadcast echo restores the new binding — so the end state is correct, but every seed-tile retarget answers failure for an action that landed, logs a spurious error, and flips the picker briefly.

Seed params are as immutable as markdown params — both are assigned once at dock time. The retarget short-circuit should cover `TileKindSeed` alongside `TileKindMarkdown`:

```go
if tileKind == string(workspacelayout.TileKindMarkdown) || tileKind == string(workspacelayout.TileKindSeed) {
    d.sendWorkspaceLayoutTileActionResultWithRequest(client, protocol.CmdWorkspaceLayoutUpdateTile, msg.WorkspaceID, tileID, requestID, nil)
    return
}
```

Non-blocking — the binding does land through the echo path. But it's a "noisy success" that reads as a bug from either side of the wire.

---

Beyond that note, the reading surface reads finished: the URI is correlation-only with typed fields as the authority, draft keys stay namespaced (file paths vs `attn://seed/...` can't collide), seed commands stay hub-local while file commands keep endpoint routing, the note ledger is bounded with an honest total, and the dormant artifact projection is honest about having no writers yet. The two follow-up commits harden the exact spots the skill warns about — primitive effect deps, refs rebuilt from snapshots, stable empty-array defaults, and the `workspaceDirectory` dep the memo was already consuming. Run the whole witness set twice, old habit: Go suites green (same two plugin-driver env flakes that fail on parent), 2858 frontend tests, tsc and diff-check clean, CI green all the way down.

figgyster (approved):
Loop closed at 96755cc — the mermaid harness was the last `path`-prop caller, and it was exactly the kind of surface I said I'd grepped and missed. Migrated clean; I type-checked the whole `test-harness/` tree at this head and it's free of MarkdownReader errors.

But the fix points at something worth a line in the debt register: `tsconfig.json` includes only `src`, so the harness tree is never type-checked by `tsc` in CI — `vite build` transpiles it, and esbuild doesn't check types. That's how this PR's `path`-prop removal broke the harness with zero red lights (runtime-benign — React just ignores the unknown prop — but the contract was silently void), and it's also why a harness-scope check at this head surfaces **10 pre-existing errors unrelated to this PR**, all drift from earlier merged work that never got checked:

- `DashboardPRsHarness.tsx` — PR fixtures missing the now-required `author` field (8 spots), plus a missing `sendMuteAuthor` prop.
- `PresentTourHarness.tsx:251` — `(path: string) => void` vs the current `(path: string | null) => void`.

None of those block this PR, but they're the proof the seam can break quietly. A harness-scope `tsc --noEmit` pass in CI (or widening the include) would have caught the mermaid straggler on this PR's first push and pays for itself the next time a contract shifts. Re-approving at 96755cc.

PR #916, opened 2026-08-15
Keep a crew member's ticket thread across days

victorarias:
## TL;DR

Ticket participation now belongs to a crew member rather than to one disposable day session. Subscriptions, authorship, cursors, and delivery follow `member:<id>` across sleep and turnover; unread activity wakes a sleeping member through the existing autonomous wake limit, after the normal charter/handoff priming completes.

## Why

Ticket participants were keyed by session id. When a member ended its day, the successor inherited the charter and handoff but not the predecessor's ticket subscriptions or cursor. The notifier then resolved the dead session to no live target and silently skipped it.

## What changed

- `member:<id>` is a durable ticket identity resolved through the current crew binding. Member-bound ticket commands act as that identity; `ticket take` still assigns the concrete session while attaching the member through event authorship.
- Existing session participation is migrated transactionally when a binding is claimed, transferred, released, or loaded at startup. Subscriptions are carried forward, cursor and attention clocks merge by maximum, and historical ticket authorship is re-attributed to the member so predecessor-authored events do not replay as unread.
- The notifier resolves an awake member to its live day. A sleeping member is woken through `crewWakeWithDelivery`, so wake serialization and `crew.wake_limit` remain the single policy rail.
- A wake-limit refusal creates a warning naming the configured limit and manual wake action. It never advances the ticket cursor.

The ordering invariant is:

```mermaid
sequenceDiagram
    participant Ticket as Ticket event
    participant Wake as Crew wake rail
    participant Member as New member day
    Ticket->>Wake: unread activity for member:trellis
    Wake->>Wake: charge wake ledger and claim one day
    Wake->>Member: charter + predecessor letter priming
    Wake->>Member: ordinary wake greeting
    Member-->>Wake: prompt-submit receipt
    Wake->>Member: arm existing ticket nudge countdown
```

The ticket nudge is never concatenated with or pasted into the wake prompt. A restart reconstructs the gate from the live member binding, session state, and durable unread activity; session teardown clears the in-memory continuation.

## Review path

1. `internal/daemon/ticket_identity.go` — member/session mapping, actor identity, cursor and attention ownership.
2. `internal/store/ticket_identity.go` — atomic upgrade and day-transition migration.
3. `internal/daemon/ticket_notify.go` and `crew_wake.go` — sleeping-member wake, refusal visibility, and post-priming delivery.
4. Ticket command handlers and `delegate_ticket.go` — consistent member attribution across subscribe/unsubscribe/inbox/watch/take and mutation/delegation entry points.

## Manual verification

On an isolated `member-inbox` profile, Trellis subscribed to a ticket, filed a handoff, and slept. A comment posted during sleep woke a fresh day; that successor read its charter and predecessor letter first, then consumed the comment from its inbox and replied on the ticket.

The scenario was repeated with `crew.wake_limit=0`. The daemon kept Trellis asleep and logged the named limit refusal plus “activity remains unread.” After a manual wake, the successor inbox returned exactly the refused comment. The isolated profile and its workers were then cleaned.

## Scope

- Every unread ticket event may wake a sleeping member; the existing wake limit is the governor.
- No protocol or frontend shape changed.
- Non-member ticket behavior and garden/seed identity are unchanged.
- Startup can migrate the current binding and recorded immediate predecessor. Older session-only rows have no remaining member provenance, so they remain ordinary historical session participation rather than being guessed or deleted.

### DIFF
--- a/changelog.d/crew-member-ticket-thread.yaml
+++ b/changelog.d/crew-member-ticket-thread.yaml
@@ -1,0 +1,10 @@
+kind: fixed
+area: crew
+change: >
+  Ticket participation now belongs to a crew member instead of one day's
+  session. A member keeps its subscriptions, unread cursor, and interruption
+  clock across day turnover; ticket commands attribute the member by its
+  `member:<id>` identity while the concrete session remains the assignee. Any
+  unread ticket activity wakes a sleeping member through the existing
+  autonomous wake limit, after charter and handoff priming. A refused wake is
+  surfaced as a warning and leaves the activity unread for the next manual wake.
--- a/docs/plans/2026-08-15-member-ticket-identity.md
+++ b/docs/plans/2026-08-15-member-ticket-identity.md
@@ -1,0 +1,121 @@
+# Plan: A member's ticket thread survives its sleep
+
+## Goal
+
+A crew member, not one day's session, owns its ticket participation. Subscriptions,
+authorship, cursors, interruption attention, and delivery follow `member:<id>` across
+day turnover. Any unread ticket activity may wake a sleeping member; the existing
+autonomous wake limit is the governor.
+
+The ticket assignee remains the concrete session doing the work. Taking a ticket
+also attaches the member identity beside that assignee, so the thread survives when
+the day ends without changing ticket/session ownership semantics.
+
+## Architecture Map
+
+```text
+member-bound ticket command
+  ticketIdentityForSession(session)
+    -> member:trellis
+  store mutation / subscription / cursor
+    -> durable member identity
+
+ticket event lands
+  notifyTicketObservers(ticket)
+    TicketParticipants(ticket)
+      ordinary session -> existing countdown delivery
+      role:chief_of_staff -> current chief session
+      member:trellis
+        awake -> current bound session -> existing countdown delivery
+        asleep + unread
+          -> crewWakeWithDelivery(autonomous=true)
+            charge existing wake ledger
+            claim one serialized day
+            inject charter + predecessor letter
+            submit the ordinary wake greeting after priming
+            wait for its prompt-submit receipt
+            arm the ticket doorbell from durable unread state
+          -> refresh the new session's existing unread indicator
+        wake refused
+          -> existing notification feed names the limit and manual wake action
+          -> member cursor remains unchanged
+```
+
+Day transitions adopt legacy session state before the binding moves or clears:
+
+```text
+ticket participant/subscription/cursor keyed by old session
+  -> ensure member subscription for every involved ticket
+  -> re-attribute historical ticket authorship to the member
+  -> merge cursor into member cursor with max(old, member)
+  -> delete old explicit subscriptions and cursors
+```
+
+On upgrade, startup applies the same migration to each member's current binding
+and recorded `LetterSession`, which covers the live day and immediate predecessor.
+Older session-only rows cannot be attributed to a member from the remaining data;
+they stay as ordinary historical session participation rather than being guessed or
+silently deleted.
+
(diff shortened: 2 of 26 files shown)


victorarias:
@figgyster PR #916 is green and ready for review on current head a3484796. The ticket-member identity, migration, and post-priming wake delivery paths are covered by the PR description and tests.

figgyster (approved):
Approve — the dead-day strand is closed. Ran the full witness set on a3484796: daemon/store/client/protocol/cmd suites green (the one sandbox failure was the container's VCS stamping on a plugin e2e, which passes with `-buildvcs=false` — environmental, not code), diff-check clean, CI green all the way down.

The identity map reads as one coherent contract. `ticketDurableIdentitiesForSession` / `ticketSessionForIdentity` / `ticketAttentionKey` are still defined in one place, now with the member first — and the migration is the piece that makes it honest: subscriptions carried forward, cursors and attention merged by max, authorship re-attributed so the member's own predecessor events don't replay as unread. That last one is the subtle bit and the store test pins it: migrate a session that authored its own thread, and `UnreadTicketEventsFor(to, to)` returns only the other author's status event. The migration is one SQL transaction, idempotent, and the source rows are deleted in the same commit. Transfer refuses on a stale binding and rolls the registry back if the migration refuses after the move. Claim/release/startup all sweep, and startup covers both the live binding and `LetterSession` — the immediate predecessor a member slept before this version existed.

The wake path rides the #913 rail rather than inventing a second one: `crewWakeWithDelivery` with the autonomous charge, the crewWakeMu fence, and the post-priming gate keyed to the same prompt-submit receipt evidence as agent messages — `_hook-state working` only fires on UserPromptSubmit/PostToolUse/tool events, the far side of the trust dialog. The ticket nudge is never concatenated into the wake prompt (the test asserts the wake's initial prompt is the ordinary greeting and nothing is pasted before the receipt), and the receipt re-arms the countdown from durable unread rather than trusting an in-memory continuation. Refusal naming the configured limit and leaving the cursor untouched is visible and tested — that was the gap in the old silent-skip behavior, now it's a warning with a manual wake action.

Three notes, none blocking:

1. `claimCrewBinding`'s post-write migration has no rollback where `transferCrewBinding` does. The write to the registry succeeds, then `migrateCrewTicketIdentity(member.ID, sessionID)` runs for the *new* session — and if that store write fails, claim returns an error with the binding already pointed at a session that was never spawned. `crewWakeWithDelivery` treats a claim error as fatal and returns without releasing. It self-heals (the phantom binding isn't live, and the next claim sweeps it), and the failing store write is a rare event, but it's the one asymmetry with the transfer path's careful rollback. A release on the error path, or the same CAS-rollback, would make the claim atomic in the same sense.

2. The wake-refusal warning is per-event, not per-ticket. A burst of ticket activity while a member sits at `crew.wake_limit=0` produces one `crew_ticket_wake_refused` notification per landed event — five comments, five near-identical warnings for the same ticket. The refusal path is the right call; coalescing by (member, ticket) would keep the feed readable. Minor.

3. `seedCrewTicketWakeDeliveries` treats any `StateWorking` session at startup as possibly-still-priming and puts it behind the receipt gate — the conservative direction, and correct for the restart-between-spawn-and-prompt case. The cost: a daemon restart in the middle of an already-settled working day defers the ticket doorbell until the session's next hook report (indicator shows immediately, so nothing is invisible). Worth knowing the tradeoff is on the conservative side by design.

Signal: the note→witness loop closes a ninth time this week, and this one completes the #913 arc — the member thread, the wake fence, and the naming contract all landed in one motion. The dead-day strand is the only open thread on crew messaging, and it's now closed.

PR #917, opened 2026-08-15
fix(tickets): nudge quiet tickets before bundling bursts

victorarias:
## TL;DR

Ticket activity now nudges every participant promptly after quiet, then bundles follow-up activity into one nudge during a measured ten-minute burst window. `attn ticket inbox --watch` consumes unread activity immediately, and agent guidance no longer recommends blocking Monitors for attn activity.

## Why

The old policy gave non-assignee observers a blanket 30-minute buffer. That prevented distraction during bursts, but it also delayed a quiet ticket's first and only update long enough that silence was indistinguishable from inactivity. The runtime-specific Monitor workaround also made a waiting session appear busy, suppressing crew heartbeats and auto-sleep.

## What changed

The notification policy is now uniform for every ticket participant:

```text
quiet + new event       -> normal short doorbell countdown
delivery + burst event -> one deadline at delivery + 10 minutes
10 minutes quiet       -> next event uses the short countdown again
--watch read            -> consume immediately and cancel covered countdown
```

“Uniform” also retires the old completed-ticket bypass. A completion after quiet
still gets the short countdown; a completion inside an active burst stays in that
bundle because the observer was just interrupted. A regression test pins this
choice explicitly.

Delivery attention now records the newest covered ticket event sequence as well as its timestamp. That prevents an unread event which already rang the doorbell from rearming after a restart or session-selection change; only genuinely newer activity can schedule the bundled follow-up.

Migration 107 leaves pre-existing attention rows with a covered sequence of zero.
The previous schema cannot prove which event an old doorbell covered, so a backfill
would risk silently suppressing legitimate unread activity. The conservative cost
is at most one bounded, self-healing repeat after upgrade.

The embedded attn skill and chief guidance now say to rely on ticket nudges. Blocking Monitors remain appropriate for external waits such as CI, but not as an attn integration mechanism.

## Window receipt

I measured a read-only copy of the production ticket event store. Treating tickets with at least five events as busy delegation tickets produced 67 tickets and 440 inter-event gaps. The median gap was 589 seconds (9m49s), so the bundle window is ten minutes: just past the observed burst cadence instead of an invented interruption budget.

## Manual verification

Against an isolated packaged `nudge-throttle` profile using protocol 250:

- the first event after quiet armed the normal doorbell in 29.856 seconds;
- after that delivery, a follow-up event armed one bundled nudge 598.335 seconds out;
- while that bundle countdown was active, `ticket inbox --watch --interval 3s` returned the unread event in 102 ms and cleared the countdown;
- bundled preflight passed for app, CLI, daemon routing, and protocol agreement.

The focused quiet-first, burst-bundle, quiet-reset, immediate-watch, and delivered-unread regression tests pass. `make test-quick` is green.

## Scope

The doorbell paste primitive and selected-session pause are unchanged. This does not touch crew lifecycle code.


figgyster (approved):
The uniform policy reads right. A quiet ticket's first and only update used to sit behind a 30-minute buffer until silence was indistinguishable from inactivity; now the first event after quiet gets the short countdown, bursts coalesce behind one ten-minute deadline, and the window is measured, not invented — 9m49s median over 440 gaps on 67 production tickets is the right kind of receipt for a magic constant. And the durable delivered-through seq closes a real hole: a delivered-but-unread event could rearm after restart or selection change, and now only genuinely newer activity can re-ring.

Verified green: full Go suites on the head (store, daemon, hooks, agent, client, cmd, ticketnotify), diff-check clean, CI green all the way down — Daemon, Frontend, E2E x3, Linux daemon x2, react-doctor 100/100. No protocol files touched, so no version bump — correct. Grepped for HasSelfMonitor / SELF_MONITOR stragglers: zero; the removal is clean across Go, docs, the skill reference, and the harness marker. The guidance change is the honest fix for the Monitor problem — a Monitor-blocked session reading as busy was suppressing crew heartbeats and auto-sleep, which is exactly the crew arc from #913/#916 paying off.

Two non-blocking notes for the record:

1. The done-transition carve-out from #594 is retired. A completion used to bypass the observer buffer and pull buffered activity forward immediately; now it is just another event — a completion landing inside a burst waits for the bundle deadline. In the quiet case (the one that matters most) it is still immediate, and inside a burst the observer was just doorbell'd anyway, so I think the trade is right — but the PR body says "uniform" without mentioning the done bypass is gone, and the old piggyback test was replaced rather than kept. Flagging so the retirement is conscious, not accidental.

2. Migration 107 leaves existing attention rows at delivered_through_seq=0, so already-doorbell'd unread can rearm exactly once after upgrade. Bounded by the bundle window from the last attention, self-heals on the next delivery — one-time cost of the schema addition, not worth blocking on, just worth knowing when the first post-upgrade doorbell feels familiar.

victorarias:
@figgyster Addressed the two non-blocking notes on the new head: the completed-ticket bypass retirement is now explicit in the PR body and pinned by TestCompletedTicketInsideBurstStaysBundled. Migration 107 intentionally keeps existing covered sequences at zero because the old schema cannot reconstruct coverage; guessing could suppress legitimate unread activity, while the conservative one-time repeat self-heals on delivery. The migration comment now records that invariant. Focused tests and make test-quick are green.

figgyster (approved):
Both notes answered with receipts, not promises.

The completion-bypass retirement is now explicit in the body and pinned by `TestCompletedTicketInsideBurstStaysBundled` — ran it just now on this head, green. That's the case I wanted on the record: a completion after quiet still rings fast, a completion inside a burst stays in the bundle because the observer was just interrupted. Deliberate, documented, witnessed.

Migration 107's zero-rows invariant reads right in the comment: the old schema can't prove which event an old doorbell covered, so a backfill guess could suppress legitimate unread. The conservative one-time repeat is the honest price, and it self-heals on the next delivery. Agreed, and now the reasoning lives next to the code instead of in a PR thread.

Re-approved at 0eaa1b1. Focused daemon suite green here; fast CI checks green, heavy ones in flight — test-plus-comment delta, so I'm not holding the door for them.

PR #918, opened 2026-08-15
feat(apps): dispatch declared reconcile handlers

victorarias:
## TL;DR

Apps can now declare a reconcile capability, implement it as a sibling export beside subscriptions and commands, and rebuild their collections from a typed snapshot of attn's current state. The daemon keeps the durable request and cursor fence from slice 1 authoritative; this slice only makes that owed work dispatchable through the shared app runtime.

## Why

Slice 1 made reconciliation durable and blocked fact delivery while it is owed, but the runtime had no contract for actually running the app's rebuild. An app therefore needs three pieces in lockstep: a frozen manifest declaration, a compiler-checked handler export, and a bounded current-state read that sees the same state-bearing domains as Initial State.

## What changed

- `reconcile = true` becomes part of the frozen app declaration. Generated `Handlers` requires the sibling `reconcile` export when declared and rejects it when undeclared, so apply catches manifest/bundle drift without evaluating app code.
- `@victorarias/attn-app` gains `ReconcileReason`, `ReconcileHandler`, and `ctx.current.snapshot()`. `CurrentStateSnapshot` is fully typed and deliberately excludes settings, warnings, and protocol metadata; `apps` uses the registry projection that Initial State uses to mount views.
- The daemon extracts Initial State's state-bearing assembly into one shared projection. Snapshot reads capture the bus head first, return non-null arrays, and are fenced to the active dispatch id.
- The sidecar accepts `app.reconcile`, loads the named content-addressed bundle, invokes its sibling export with the fixed durable reason, and exposes the same current-state callback to ordinary handlers.

The runtime ordering is:

```text
app consumer pre-drain
  -> resolve serving version + coalesced durable reason
  -> app.reconcile in the shared sidecar
  -> bundle.reconcile(reason, scoped context)
  -> ctx.current.snapshot() callback to the daemon
  -> shared Initial State domain projection
  -> handler converges its declared collections
```

## Review path

1. `internal/daemon/current_state.go` — verify the snapshot's bus-position ordering and parity with Initial State.
2. `internal/daemon/app_consumers.go` and `app_runtime_rpc.go` — verify the durable pre-drain plan and dispatch-id fence.
3. `apphost/src/dispatch.ts` — verify reconcile is its own sibling kind and receives the same scoped context as other handlers.
4. `sdk/attn-app/src/index.ts` and `currentState.ts` — verify the public types match the daemon JSON exactly.

## Manual verification

Installed the exact branch as an isolated signed app at protocol 251 and passed bundled preflight. A throwaway app declared reconcile, called `ctx.current.snapshot()`, and wrote a receipt from its handler. Disabling and re-enabling it produced one `re_enabled` receipt with `throughSeq=34`, snapshot `asOfSeq=38`, and `apps=["reconcile-proof"]`; the consumer then reached cursor 39 with zero lag and the shared runtime remained connected. The profile and proof app were removed afterward.

## Scope boundary

Invocation rows, timeout/interruption handling, operator status, command refusal, missing-handler policy, and packaged-app failure proofs remain slice 3. The current tree still has slice 1's `re_enabled` durable cause, so this slice dispatches it; the amended plan assigns removing that trigger and SDK enum to slice 3 together with the new retention-floor and delivery witnesses.

Related design: `docs/plans/2026-08-14-ext-app-reconcile-handler.md`.

### DIFF
--- a/apphost/package.json
+++ b/apphost/package.json
@@ -5,6 +5,7 @@
   "description": "attn's shared app runtime: the supervised Bun sidecar every installed app's handlers execute in.",
   "type": "module",
   "scripts": {
+    "test": "bun test",
     "typecheck": "tsc --noEmit"
   },
   "devDependencies": {
--- a/apphost/src/dispatch.test.ts
+++ b/apphost/src/dispatch.test.ts
@@ -1,0 +1,135 @@
+import { afterEach, describe, expect, test } from "bun:test"
+import { mkdtemp, rm, writeFile } from "node:fs/promises"
+import { tmpdir } from "node:os"
+import { join } from "node:path"
+import { runDispatch, runReconcile } from "./dispatch.ts"
+import type { RpcConnection } from "./rpc.ts"
+
+const roots: string[] = []
+
+afterEach(async () => {
+  await Promise.all(roots.splice(0).map((root) => rm(root, { recursive: true, force: true })))
+})
+
+async function bundle(source: string): Promise<string> {
+  const root = await mkdtemp(join(tmpdir(), "attn-apphost-reconcile-"))
+  roots.push(root)
+  const artifact = join(root, "bundle.mjs")
+  await writeFile(artifact, source)
+  return artifact
+}
+
+describe("reconcile dispatch", () => {
+  test("invokes the sibling export with the fixed reason and scoped current-state callback", async () => {
+    const artifact = await bundle(`
+      export default {
+        reconcile: async (reason, ctx) => {
+          const snapshot = await ctx.current.snapshot()
+          await ctx.collections.items.put("receipt", { reason, snapshot })
+        },
+      }
+    `)
+    const calls: Array<{ method: string; params: unknown }> = []
+    const notices: Array<{ method: string; params: unknown }> = []
+    const conn = {
+      call: async (method: string, params: unknown) => {
+        calls.push({ method, params })
+        if (method === "app.current.snapshot") return { asOfSeq: 41 }
+        return { id: "receipt", body: {}, rev: 1 }
+      },
+      notify: (method: string, params: unknown) => notices.push({ method, params }),
+    } as unknown as RpcConnection
+    const reason = {
+      causes: ["gap", "version_changed"] as const,
+      version: 7,
+      throughSeq: 41,
+      gap: { cursor: 8, earliest: 12, missed: 3 },
+      previousVersions: [3, 5],
+    }
+
+    const result = await runReconcile(conn, {
+      dispatch: "dispatch-1",
+      app: "reviewer",
+      version_id: 7,
+      artifact,
+      collections: ["items"],
+      reason: { ...reason, causes: [...reason.causes] },
+    })
+
+    expect(result).toEqual({ ok: true })
+    expect(calls[0]).toEqual({
+      method: "app.current.snapshot",
+      params: { dispatch: "dispatch-1" },
+    })
+    expect(calls[1]?.method).toBe("app.collection.put")
+    expect(calls[1]?.params).toEqual({
+      dispatch: "dispatch-1",
+      collection: "items",
+      id: "receipt",
+      body: { reason, snapshot: { asOfSeq: 41 } },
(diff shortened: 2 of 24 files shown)


figgyster (approved):
Slice 2 lands the dispatch half of the reconcile contract, and it reads finished: pre-drain now runs the serving version's sibling export through the same shared sidecar, the reason is folded deterministically (gap, re_enabled, version_changed — sorted, deduped, fixed across retries), and claim completion is atomic with the cursor crossing the fence. The dispatch-id fence on `app.current.snapshot` matches the collection ops exactly, the wedge-attribution path is inherited intact, and the manifest/bundle drift check is a named refusal plus a compile error in both directions. The Initial State refactor is behavior-preserving — same calls, same projection, now shared.

Verified locally: full daemon + store + protocol + cmd suites green in the sandbox (one pre-existing appbuild toolchain test fails identically on base — it needs bun on PATH, environmental), apphost `bun test` 3/3, apphost tsc clean, and `tsc -p sdk/attn-app` regenerates the committed sdkdist with zero diff, so check-sdk would pass. CI is still running its Daemon/Frontend legs; the equivalent checks are green here.

One note, in the exact surface your review path points at — `currentState.ts` claims to match the daemon JSON exactly, and it already doesn't, on day one:

- `Session` is missing `cost_unknown` and `cost_usd`. `mergedSessionsForBroadcast` decorates sessions with both since #914 (the frontend's generated Session carries them at app/src/types/generated.ts:1643-1644), so the wire shape is a superset of the SDK type. Nothing breaks at runtime — extra fields are invisible to TS consumers — but an app that wants cost data gets a compile error and would rightly conclude the daemon doesn't send it. Two optional fields (`cost_unknown?: boolean`, `cost_usd?: number`) close it.

Non-blocking; the rebuild-from-current-truth semantics are sound and the snapshot ordering argument holds (head captured before assembly, so every mutation at or below `asOfSeq` is visible and anything later still has a fact). Watching slice 3 for the re_enabled trigger removal and the retention-floor witnesses.

victorarias:
@figgyster Fixed the snapshot type drift you found in c1473b8a: `Session` now exposes the daemon’s optional `cost_unknown` and `cost_usd` fields, and the committed sdkdist was regenerated. SDK typecheck and `make check-sdk` are clean.

figgyster:
Confirmed at c1473b8a: both optional fields landed in the SDK source and the committed sdkdist, in the right spot next to the other session decorations. Regenerated the declarations myself — zero diff against the committed copy, so check-sdk passes — and apphost tsc is still clean. The 'matches the daemon JSON exactly' contract holds now. Approval stands.

figgyster (approved):
**Approve — c1473b8a.**

Slice 2 of the reconcile arc delivers: the durable claim from slice 1 is now dispatchable. Pre-drain runs the serving version's `reconcile` sibling export in the shared sidecar; the reason folds deterministically (gap/re_enabled/version_changed, sorted, fixed across retries); completing the claim stays atomic with crossing the dispatch-id fence. And Initial State's state-bearing assembly finally has one home — the shared projection behind `ctx.current.snapshot()`, bus-head-first, non-null arrays. The manifest/bundle drift check is a compile error in both directions, plus a named runtime refusal. An abstraction that earns its keep.

The one drift I flagged last round is closed in c1473b8a: `cost_unknown` and `cost_usd` now ride the SDK `Session` type and the committed sdkdist, exactly where the daemon puts them. Regenerated the declarations myself at this head — zero diff, check-sdk clean — and apphost tsc still passes. The "matches the daemon JSON exactly" contract holds.

Evidence: CI green all the way down on this head — Daemon, Frontend, all three E2E shards, Bench A/B, react-doctor, Changelog, Changes. Local runs earlier in this case: full daemon/store/protocol/cmd suites green, apphost bun tests 3/3 (the lone appbuild failure in the box is environmental — needs bun on PATH, fails identically on base).

The app runtime is now a full lifecycle host: dispatch, command, and reconcile on one fenced lane. Slice 3 — invocation rows, refusal policy, retiring the re_enabled trigger — is the remaining chapter. Ship it.

PR #919, opened 2026-08-15
feat(terminal): run the browser terminal on ghostty's own module

victorarias:
attn built its own `ghostty-vt.wasm` from a pinned commit and drove it through
`ghostty-web` — a third-party JS wrapper (Coder's, not ghostty-org's) whose ABI
predates Ghostty's current C API. Bridging the two took a vendored zig
compatibility adapter plus a patch, and the wrapper's last release was
2026-06-28 with no repository activity since 2026-07-02.

ghostty-org now publishes the module itself. This consumes that binary through a
first-party binding over the current C API: no adapter, no patch, no zig in the
frontend toolchain, and 861 fewer lines.

## What moves

**The pin.** `ghostty-vt.pin` → `d760ee96`, measured against the old one at
equal optimization: **4.2x IO throughput, 4.8x reflow**. The pin is shared with
the native worker so both runtimes parse VT with one implementation, so the
native prebuilts are republished for it in the same change.

**Sourcing.** Upstream's rolling `tip` is rebuilt on every commit, so its assets
cannot be pinned by hash. The exact upstream bytes are mirrored onto our own
keyed release and verified against `ghostty-vt-wasm.lock` before they reach a
build — the shape `scripts/lib/libghostty-vt.sh` already uses for native
archives. Fail-closed, and a pin that was never mirrored stops the build with
the command that fixes it.

**The binding.** `app/src/ghostty/`: `abi.ts` (enum values and struct offsets,
asserted against the real module by consequence), `callback.ts` (a
hand-assembled wasm shim, because JavaScriptCore has no `WebAssembly.Function`),
`terminal.ts`, `index.ts`. Viewport reads go through the render state; scrollback
reads go through grid references, which upstream warns are not render-loop
material — measured at 0.72ms for a full 200x50 scrolled-back viewport against
0.53ms for the same volume through the render state, so a scrolled pane stays
inside a frame and the scroll model did not need redesigning.

`ghosttyHyperlinks.ts` is gone: it reached into the wrapper's internals to read
per-cell URIs, which the current API exposes directly, so its two functions are
now model methods.

## Deliberately not in scope

`ghostty-web` remains a dependency **for its key encoder alone**, which
`InputHandler` drives. Replacing the terminal model and the DOM input bridge in
one change risks typing for no gain. (`internal/daemon/web/vendor/ghostty-web`
is a separate self-contained copy with its own bundled wasm, unaffected by this
pin.)

## Kitty behaviour moved with the pin

Two of ghostty's native kitty behaviours changed, and the wire-rewrite corpus
caught both — which is what it was built for:

- a placement's scroll no longer tracks the row count `r=` claims and stays
  inside the screen (measured 3→1 and 13→2, *not* a clamp to one row);
- retransmitting an image under a live placement id retires the placement
  instead of re-pointing it.

Both are improvements at the wire: the clamped-scroll case now **carries** its
scroll (`CSI 2 S`) where it previously bailed to a snapshot re-push. Worker/wire
`agree()` holds on all 164 shapes probed.

`kittyResyncScrollClamped` became unreachable and is **kept**. Its premise is a
property of `SU` — ghostty clamps it to the scroll region, so a larger scroll
would be silently truncated and leave the client's history permanently short —
not a property of kitty. Upstream closed one route to the condition; the
condition still exists. The receipt on the constant records the shapes probed
against it.

## A regression this caught, and how

Live verification found two failures the unit suite could not: after an attach
restore, clicking a command block selected nothing, and Cmd+Shift+C copied
nothing. An A/B against `main` on the same machine and the same freshly-signed
bundle proved they were mine rather than the environment.

Cause: the render-state readers (`getViewport`, `getCursor`, `getColors`,
`isRowDirty`) did not sync the render state from the terminal — only `getLine`
did. The app reads the cursor *immediately after* writing the bytes that moved
it: OSC 133 segmentation records a command block's row that way, and a restore
seeds block anchors that way. So every block landed on a row from the previous
frame. `getScrollbackLine` goes through `grid_ref` on the terminal and is live,
so the two halves of `selectionLineAtBufferRow` disagreed about time — which is
exactly why baseline clicks worked and post-restore clicks did not.

Fix: a `stale` flag set by `write()`/`resize()` and a private `sync()` every
render-state reader calls. Syncing mid-frame is safe for the renderer: ghostty's
dirty set persists until `markClean()` and is not consumed by
`ghostty_render_state_update`. Two regression tests cover it, and the first
fails without the fix.

## Verification

Non-production profile installed from this branch (`vtbind`), preflight PASS,
protocol 249 across CLI/app/daemon. All five terminal scenarios green:

- `scenario-ghostty-scroll`
- `scenario-terminal-osc8-link`
- `scenario-terminal-kitty-image`
- `scenario-terminal-block-resize`
- `scenario-terminal-block-copy`

Full Go suite green, `go vet` clean, frontend 254 files / 2853 tests green —
including the kitty wire-rewrite parity corpus, the placement store, the
WebGL renderer, the model-op-ring replay capture and the resize-hang repro, all
now running through the first-party binding.

Recording is `scenario-terminal-block-resize` on the packaged app, built from
this branch's committed source (the harness fingerprint check enforces that):

recording-02

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/578d5e969fd4c4857f304a1b74aed1a213523783/ghostty-first-party-binding/recording-02.mp4)

## Follow-ups (plan steps 2-4)

Plan: `docs/plans/2026-08-15-first-party-libghostty-vt.md`.

1. Delete `ghostty-vt-native.patch` for upstream's snapshot API. Verified that
   the patch still applies at this pin, so that is a choice, not forced work.
2. Snapshot-based restore (protocol bump); the corrective `CUP` append goes with it.
3. Fast-path the render read via dirty rows and style-id caching.

https://claude.ai/code/session_01FcwbumRJFVLKpZQWowS9Pb


figgyster (approved):
**Approve** — with two notes that block nothing.

attn was building its own wasm through a third-party wrapper's dead ABI and paying for the mismatch in vendored zig plus a patch. This deletes the adapter, the patch, and the wrapper's terminal model in one motion and puts the browser on ghostty-org's own binary through a first-party binding. Leaving the key encoder on ghostty-web for now is the right scope call — replacing the model and the DOM input bridge in one change doubles the blast radius for no gain.

I verified rather than took the receipts:

- All 33 exports `abi.ts` declares exist in the mirrored module; it imports nothing but `env.log`.
- The mirror release carries `ghostty-vt-d760ee96…wasm` and its sha256 matches the lock exactly (429a012a…). Old native prebuilts kept, new ones published for the new key.
- Go suites green on the new pin: `internal/pty` (the kitty corpus + wirefeed mirror — the tripwire this apparatus was built for), `ghosttyvt`, `daemon`, `store`. `go vet` clean on the touched packages.
- Full frontend suite green: 256 files / 2866 tests through the binding, including the binding test, the parity corpus, the placement store, the op-ring replay, and the resize-hang repro. `tsc` clean. CI agrees on every finished lane; the three E2E shards are still running.
- The grid_ref scrollback reads measured sane — 0.72ms for a full 200×50 scrolled-back viewport — the number in the body holds up to the shape of the code.

The stale/sync fix is the best part of this diff. The render-state readers answered as of the previous frame, every command block landed on a stale row, and the unit suite couldn't see it — it took the packaged-app A/B to catch, and then it got two regression tests that pin the mechanism, including the subtle one (a read that syncs must not consume the dirty set). Live verification found what the harness couldn't, and the harness got upgraded. That's the loop working.

The corpus catching both kitty moves — the scroll no longer tracking `r=`, the retransmit retiring the placement — is exactly what it was built for. And keeping `kittyResyncScrollClamped` as a tripwire reads right: the divergence it names is a property of SU, not kitty, and it's silent when it happens. The 164-shape receipt is a good witness.

Two notes, both cosmetic:

1. `app/src/ghostty/abi.ts`'s header says the offsets are "asserted at runtime by abi.layout.test.ts" — that file doesn't exist at this head. The by-consequence assertions live in `terminal.binding.test.ts`, whose own header says so. Point the reference at the file that exists, or drop it.

2. `getViewport()` returns a reused pool where ghostty-web returned a fresh array — a caller that retains the array across the next `getViewport()` will silently watch its cells mutate. I checked every call site (renderers, selection, markdown projection, message-row access) and they all consume synchronously, so nothing breaks today, and the contract is documented in `terminal.ts`. It's the kind of implicit contract that bites a future caller — keep that comment loud.

Watch items, on the record: snapshot restore is the follow-up that carries the protocol bump and kills the native patch plus the corrective CUP; the render-read fast path (dirty rows, style-id caching) is where the real perf win lands, not here — the naive binding is parity on a full-viewport read, as measured.

figgyster (approved):
Loop at 556f43e — the E2E wall, closed.

The synchronize answered the one open thread from my first pass: the E2E shards weren't testing the scenarios, they were dying at the web server. The wasm is downloaded and gitignored, so a fresh checkout has no `app/vendor/ghostty-vt/ghostty-vt.wasm`, and playwright's `npx vite` couldn't resolve the import in `wasm.ts` — every spec failed before a single scenario ran. Putting the ensure step in the `webServer` command is the right home for it: CI and a contributor running `pnpm e2e` on a clean tree both get the same fail-closed fetch, and it's a hash check plus exit when the file's already verified.

Verified: the script name matches (`verify:ghostty-vt` → `ensure-ghostty-vt-wasm.sh`), the ensure exits 0 in this checkout with the module verified against the lock, and the one-file diff is whitespace-clean. Source tree unchanged since 900da10, so all the verification from the first pass stands. CI is re-running on the new head — watching the E2E shards actually get to run this time.

Re-approving at 556f43e.

figgyster (approved):
Loop at a42accf — the wrap direction flip, caught and corrected. Eating crow on this one: my first-pass review approved the binding's `isRowWrapped` without noticing it read ghostty's direction (row wraps INTO the next) while both callers asked the opposite (does THIS row continue the one above). The old adapter read `wrap_continuation` — the callers' question — and the new binding silently answered a different one. Every join shifted a row; wrapped paths lost their links and copy-as-markdown broke. The unit suite couldn't see it because terminalLinks tests inject `isContinuationRow` as a pure function — nothing connects the binding's flag to the join logic. The rename to `rowWrapsIntoNext` plus the row-above call sites is the right fix, and the re-pinned test now records the direction trap instead of codifying it. That's on me as much as the diff; the receipt is that live verification caught it, again.

Verified at a42accf: tsc clean, full suite 256 files / 2866 passed / 18 skipped, binding + terminalLinks + op-ring tests green. diff-check clean.

One boundary note, same family, one row over. I drove the real module: after a soft-wrapped line scrolls up one row, the wrap flag lives on the HISTORY row (the one that wraps), and the continuation — now active row 0 — carries nothing. `isContinuationRow` covers that boundary with the full-row-text heuristic fallback (`bufferRow === history`), but `selectedMarkdown`'s `wrapped: activeRow > 0 && …` has no fallback: the first active row is never marked wrapped, where the old adapter's `wrap_continuation` read answered true. So copy-as-markdown still misses joins at the scrollback/active boundary — a wrapped path straddling the fold gets split again. Links are covered; the markdown path needs the same heuristic (or a history-row wrap read) for `activeRow === 0`. Non-blocking — it's the same one-line class you just fixed.

Re-approving at a42accf.

victorarias:
Fixed the boundary you flagged: `selectedMarkdown` now uses the same full-row heuristic `isContinuationRow` uses when the previous row has fallen into scrollback, so a wrapped line straddling the fold stays joined in a markdown copy.

Local: tsc clean, terminal-interactions + terminal-blocks green.

https://claude.ai/code/session_01FcwbumRJFVLKpZQWowS9Pb

figgyster (approved):
Loop at 2d51c2a — the fold, closed. `selectedMarkdown` now answers the wrap question the same way `isContinuationRow` does: authoritative flag when the row above is still on the active screen, full-row heuristic once it has fallen into scrollback. The `selectionLineAtBufferRow` dep is added where it belongs, and the branch structure makes the markdown wrap model uniform with the links model instead of a special case. Verified: tsc clean, full suite 256 files / 2866 passed / 18 skipped, diff-check clean. My empirical repro of the fold (wrap flag lands on the history row, continuation carries nothing) is what the heuristic now covers.

Two observations, both non-blocking:

1. The heuristic now applies to every scrollback pair, not just the fold — a selection entirely in scrollback of wrapped content joins correctly where main always split it. That's an improvement, and it matches isContinuationRow's policy exactly. But links tolerate the same approximation because false joins get filtered downstream (path candidates must exist before anything links); the markdown path has no such filter — `terminalStyledSelectionToMarkdown` joins with a bare `+=` and copies verbatim. An exactly-full non-wrapping row followed by a selected line now merges into the copy, where main split it. Narrow — full-width non-wrapped lines are uncommon — and it buys the correct join for every real wrap in scrollback, so I'd keep the trade. Worth knowing it's there.

2. The markdown copy path has no witness: nothing in the suite exercises `selectedMarkdown` (the direction flip that shipped in 900da10 and the fold gap both survived the unit suite for the same reason — the component's copy path is untested entirely). This fix rides on live verification like the last one did. If a markdown-copy harness ever exists, the fold and the heuristic belong in it.

Re-approving at 2d51c2a.

PR #920, opened 2026-08-15
fix(crew): keep requested sleep and release dead days

victorarias:
## TL;DR

Victor can now ask an awake crew member to sleep from `attn crew sleep <member>` or the moon on its crew row. The request reaches the member through the durable message rail and explicitly asks it to close with `attn handoff --sleep`, so consented closure cannot wake a successor behind it.

A dead member process also stops holding the crew seat immediately. Runtime exit releases the binding, startup recovery does the same for missing workers, and wake actively probes any remaining binding before returning `AlreadyAwake`.

## Why

Crew identity previously equated “awake” with a binding that named any session row. A process killed outside attn left that row behind, so wake returned success while focusing a corpse and the member could remain unwakeable even after restart.

The automatic sleep prompt also promised that nobody would wake behind the member, but told it to use plain `attn handoff`. Plain handoff re-evaluates presence at filing time, so Victor returning before the letter landed could wake a successor and break the promise.

## What changed

```text
CLI `attn crew sleep` / awake-row moon
  -> crew_sleep request/result
  -> persist user-origin sleep request
  -> deliver through the existing agent-message rail
  -> member files `attn handoff --sleep`
  -> no successor wakes

PTY exit or startup missing-runtime recovery
  -> release crew binding
  -> crew.released fact -> crew_updated roster
  -> remember the exited session for the next wake receipt

crew wake
  -> probe bound runtime, not only the session row
  -> live: AlreadyAwake
  -> exited: release, start a fresh day, return released_session_id
```

The sleep result names delivered versus queued delivery and makes an already-asleep request an explicit no-op. Senderless durable messages now represent daemon-authored user requests, so the sleep prompt does not masquerade as “another agent.”

Every crew wake now holds queued messages behind the real first-prompt hook. A sleep request made while the member is still reading priming is reported as queued and enters the composer only after startup is complete; it can no longer be stamped delivered into a trust or initialization screen.

Crew priming is now the sole authority for closure flags: plain handoff is presence-decided turnover, Victor asking for sleep means `--sleep`, and `--nap` explicitly requests a successor. The lifecycle auto-sleep prompt uses `--sleep` too.

The lockstep protocol moves from 250 on current main to 251, with generated Go and TypeScript contracts.

## Review path

1. `internal/daemon/crew_sleep.go` and `agent_msg.go`: consented user-origin delivery, first-prompt gating, and named no-op/result semantics.
2. `internal/daemon/crew.go`, `crew_wake.go`, and the PTY/startup seams in `daemon.go`: immediate release, active runtime probing, and named repair receipts.
3. `internal/crew/priming.go` and `crew_lifecycle.go`: the three closure modes and the fixed auto-sleep promise.
4. `cmd/attn/crew.go`, the TypeSpec schema, and `useDaemonSocket.ts`: CLI and wire surfaces.
5. `QueueBands.tsx`: the awake-row moon control symmetric with the sleeping row's wake control.

## Manual verification

On an isolated `crew-sleep` profile running the packaged app, I imported a synthetic Trellis member and woke session `08a2963c…`. I resolved its exact child PID from the isolated worker registry, confirmed its cwd was Trellis's isolated home, and terminated it externally. The roster immediately showed Trellis asleep while the generic session row remained for history. A second wake started `19644d9d…` and returned the full first ID as `released_session_id`.

After fixing the launch-time delivery race exposed by the first recording, I rebuilt an empty isolated profile at protocol 251 and waited until a fresh Claude-backed Trellis session `17141a45…` reached `waiting_input`. One click on the awake row's moon delivered exactly one user-origin prompt, including `attn handoff --sleep` and “nobody wakes behind you.” Trellis wrote its handoff, filed it with `--sleep`, and exited. The roster showed no binding and `attn agent list --json` returned an empty list: no successor woke.

The recording starts from the fully initialized `waiting_input` session and ends at “No active sessions.” The isolated profile, daemon, synthetic home, and installed app were cleaned after verification.

## Evidence

One moon click asks Trellis to sleep and ends with no active sessions

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/45b112ff87822b226112d34698d3a96b54ac9b7f/delegate-crew-sleep-bf3d6071/crew-sleep-ui-final.mp4)

## Out of scope

- Plain `attn handoff` remains presence-decided.
- No heartbeat, monitor detection, or settled/unsettled model changes.
- Sleep remains a request for consented closure, never a process kill.

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -907,6 +907,7 @@ function AppContent({
     sendTicketAttach,
     sendTicketResume,
     sendCrewWake,
+    sendCrewSleep,
   } = useDaemonApi();
 
   // The presentation notice lives in the triggering session's pane header
@@ -3474,6 +3475,14 @@ function AppContent({
       .catch((error) => showError(error instanceof Error ? error.message : `Failed to wake ${crewDisplayName(member)}`));
   }, [sendCrewWake, handleSelectSession, showError]);
 
+  // Asking for sleep delivers closure work to the member; it does not close the
+  // session itself. The roster changes only after the member consents by filing
+  // its letter with `attn handoff --sleep`.
+  const handleSleepCrewMember = useCallback((member: string) => {
+    sendCrewSleep(member)
+      .catch((error) => showError(error instanceof Error ? error.message : `Failed to ask ${crewDisplayName(member)} to sleep`));
+  }, [sendCrewSleep, showError]);
+
   // The daemon-facing ticket actions the pane-header ticket overlay wires into
   // its TicketDetailPanel. Memoized so the workspace's renderPaneSurface memo
   // does not rebuild every render. Same senders the dock panel uses; resume is
@@ -3875,6 +3884,7 @@ function AppContent({
           onToggleShowSessionless={handleToggleShowSessionlessWorkspaces}
           crew={crew}
           onWakeCrewMember={handleWakeCrewMember}
+          onSleepCrewMember={handleSleepCrewMember}
           queueModeEnabled={queueModeEnabled}
           onToggleQueueMode={handleToggleQueueMode}
           workspaceSelectionStyle={workspaceSelectionStyle}
--- a/app/src/components/QueueBands.test.tsx
+++ b/app/src/components/QueueBands.test.tsx
@@ -451,12 +451,13 @@ describe('the crew in the sidebar', () => {
     expect(rows.filter((id) => id?.includes('keel'))).toEqual(['queue-crew-keel']);
   });
 
-  it('wakes a sleeping member on the second click and focuses an awake one with the first', () => {
+  it('wakes a sleeping member on the second click, asks an awake one to sleep, and focuses its day', () => {
     const onWakeCrewMember = vi.fn();
+    const onSleepCrewMember = vi.fn();
     const onSelectSession = vi.fn();
     renderCrew(
       [{ id: 'sess-keel', label: 'keel of the day', state: 'working', workspaceId: 'ws-a', crewMember: 'keel' }] as TestSession[],
-      { onWakeCrewMember, onSelectSession },
+      { onWakeCrewMember, onSleepCrewMember, onSelectSession },
     );
 
     fireEvent.click(screen.getByTestId('queue-crew-wake-trellis'));
@@ -464,13 +465,18 @@ describe('the crew in the sidebar', () => {
     fireEvent.click(screen.getByTestId('queue-crew-wake-trellis'));
     expect(onWakeCrewMember.mock.calls).toEqual([['trellis']]);
 
+    fireEvent.click(screen.getByTestId('queue-crew-sleep-keel'));
+    expect(onSleepCrewMember.mock.calls).toEqual([['keel']]);
+    expect(onSelectSession).not.toHaveBeenCalled();
+
     // Focusing an awake member is not consequential and stays one click.
     fireEvent.click(screen.getByTestId('queue-crew-select-keel'));
     expect(onSelectSession.mock.calls).toEqual([['sess-keel']]);
 
-    // An awake member has no wake button: the way in is the way in only while
-    // there is nothing to focus.
+    // Each act exists only on the side of the day where it makes sense.
     expect(screen.queryByTestId('queue-crew-wake-keel')).toBeNull();
+    expect(screen.queryByTestId('queue-crew-sleep-trellis')).toBeNull();
+    expect(screen.getByTestId('queue-crew-sleep-keel').getAttribute('aria-label')).toBe('Ask Keel to sleep');
   });
 
   it('writes a sleeping member as a name while the id stays the address', () => {
--- a/app/src/components/QueueBands.tsx
+++ b/app/src/components/QueueBands.tsx
@@ -43,6 +43,8 @@ interface QueueBandsProps {
   crew?: CrewMemberView[];
   /** Start a sleeping member's day. Resolves once its session exists. */
   onWakeCrewMember?: (member: string) => void;
+  /** Ask an awake member to close its day and sleep. */
+  onSleepCrewMember?: (member: string) => void;
   selectedId: string | null;
   onSelectSession: (id: string) => void;
   onSettleTurn: (id: string) => void;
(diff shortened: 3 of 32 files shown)


figgyster (approved):
Solid piece of work. The dead-day problem was real — a binding that outlived its process made wake return success while focusing a corpse — and this closes it from both ends: exit releases the seat immediately, and wake refuses to trust a row without a live runtime behind it. Ran the whole thing in the box: full Go suites green (daemon, store, crew, protocol, client, cmd), full frontend suite 256/256 files (2855 passed, 20 skipped), tsc clean, vet clean, ghostty wasm provenance verified against the lock.

Verified the ordering that makes the release safe: `crewNap` moves the binding to the successor *before* `closeNappedSession` sends SIGTERM, so when the old session's PTY exit fires, the member is bound to the new day and `releaseExitedCrewBinding` finds nothing to release — a normal turnover never trips the new release path. And `releaseCrewBinding` only clears when the binding still names the observed session, so a late exit event can't steal the successor's seat. The liveness probe is fail-closed the right way: inconclusive probe is an error, never `AlreadyAwake`, and it's bounded by the 5s RPC timeout even though it passes `context.Background()`. The senderless agent-message path is the right call too — the sleep prompt is Victor's request, not another agent speaking, and the test pins that it doesn't masquerade (no attribution line, and the store row has no sender). The `--sleep`/`--nap`/plain-handoff closure semantics are taught in priming, used by the auto-sleep prompt, and pinned by tests — the "nobody wakes behind it" promise now survives the letter landing late.

Two notes, neither blocking:

1. The moon button's success is silent. `handleSleepCrewMember` only surfaces errors — a delivered or queued receipt resolves and vanishes. The wake button has feedback (it focuses the session); the moon click gives the user nothing until the member actually files its letter, which can be minutes later. The CLI prints the receipt; the sidebar could surface it too (a toast with delivered/queued would do). Worth wiring the resolved `CrewSleepResult` into something visible.

2. Small wording redundancy in the queued CLI line: `crewSleepOutcomeLine` prints "Sleep request for Keel is queued in session sess-fed — queued (target is waiting on an approval…)" — the "queued" appears twice because `agentMessageQueuedDetail` strings already start with "queued (". Cosmetic; drop the leading "queued (" from the detail when composing that line, or reword.

One observation for the merge, not a blocker: the branch's merge base predates #917 (which is on main), but GitHub reports mergeable and the touched files don't overlap — no conflict. Also the released-session receipt lives in an in-memory map, so a daemon restart between exit and wake loses the *name* but not the release itself (the binding clear is durable); purely cosmetic when it happens.

figgyster:
Heads up: main moved since the branch was cut — #921 (two-step wake confirm) landed and this PR is now showing `CONFLICTING`. The conflict is small and friendly: both changes touch the same crew-row test (`QueueBands.test.tsx`). #921 rewrote the wake test to two-step, #920 adds the sleep-moon assertions to that same test. The component code merged cleanly — `useWakeConfirm`/`CrewWakeSun` and the moon button coexist in the row, and `daemon.go`'s auto-merge is a clean composition of #917's field rename and the crew-sleep fields.

I resolved it locally to confirm there's no design tension: the combined test keeps the two-click wake (arm on first click, `crew_wake` only on the confirm) and the one-click sleep moon (fires `crew_sleep`, no session focus), plus the "each act on its own side of the day" assertions. Merged tree verified: QueueBands suite 43/43, tsc clean, full daemon suite green. So a rebase/merge should be minutes, not archaeology.

Also noticed the Evidence section came back out of the body at 20:43 — the clip still resolves at the pinned sha (`6efa8500…/delegate-crew-sleep-bf3d6071/clip.mp4`, HTTP 200), so the recording is safe either way. If the removal was deliberate, no action; if it was collateral, the embed drops back in as-is.

figgyster (approved):
Loop-closing review at 8188c06 "fix(crew): queue sleep until wake is ready".

This closes a real gap in the sleep delivery. The roster shows a member awake the moment wake claims its binding — before the agent has crossed priming and its trust dialog. A sleep click in that window used to ride the message rail straight into startup and come back `delivered`, for words the member never read. Now every crew wake gates doorbells behind the first prompt: plain wakes set a nil-callback gate instead of no gate at all, `initialPromptPending` reads map presence rather than callback non-nil-ness, and `runPostInitialPrompt` drains queued agent messages whenever a gate was pending — callback or not. The error path in `crewWakeWithDelivery` correctly clears the gate unconditionally now, since plain wakes own one too.

The witness is a real one: `TestCrewSleep_QueuesUntilAWakingMemberTakesItsFirstPrompt` asserts the sleep result comes back queued with "still reading its priming", that nothing reaches the PTY before the greeting, that the first-prompt drain delivers exactly one message containing `attn handoff --sleep`, and that the store queue is empty afterward. The existing delivery test was honestly adapted to cross the gate before asserting delivery — behavior changed, witness updated to match, not papered over.

Verified in the sandbox at this head: `internal/daemon` suite green (90s full run) and `internal/crew` green; the new test runs with the suite. Nothing in the commit touches the frontend.

Standing context, unchanged since my 20:51 note: the PR is still one commit behind main with the same test-only QueueBands conflict from #921 — friendly merge, minutes of rebase. The Evidence-section question (out of the body at 20:43, deliberate or collateral) is still open but non-blocking; the clip resolves at its pinned sha either way.

Approving at the new head.

figgyster (approved):
Rebase landed — `27d16720` is now on current main and the PR reads MERGEABLE again. Verified the new head in the box:

- **Conflict resolution is exactly the friendly merge I flagged**: the combined `QueueBands` test keeps the two-click wake (arm, then `crew_wake` on confirm) and the one-click sleep moon (`crew_sleep`, no session focus), with focus staying one click. QueueBands suite 43/43 and tsc clean at this head.
- **The gate fix survived the rebase intact**: `initialPromptPending` still reads map presence, plain wakes set the nil-callback gate, `runPostInitialPrompt` drains any pending gate, the spawn-error path clears unconditionally, and `TestCrewSleep_QueuesUntilAWakingMemberTakesItsFirstPrompt` is still the witness. The crew/daemon files are byte-identical to `8188c06` — the 88-file diff against it is main's own history coming in via the rebase (#919 terminal, #921 wake confirm, #917 tickets), nothing of the PR's.
- Full daemon suite green (92s) plus crew/protocol/client/cmd at the rebased head.

Approval stands at `27d16720`. The sleep path is race-safe against the wake priming window, the branch is mergeable, and the two open threads from the original review (silent moon-button success in the UI, the double-"queued" CLI line) remain non-blocking nits. Merge when ready.

PR #921, opened 2026-08-15
feat(crew): ask twice before waking a member

victorarias:
Waking a sleeping crew member starts a new day of a durable identity. It spends
that member's attention, it is billed, and it cannot be un-rung. Until now one
click on a sleeping row did it — and clicking a crew row to *look* at it was the
same click, so members got woken by accident.

The sidebar now asks twice. The first click arms the row, the second wakes. The
CLI is unchanged: `attn crew wake <name>` is one step, because typing a command
is already the deliberate act the two clicks reconstruct.

## What arming looks like

The wake button was the character `☀`. It is now a drawn sun, and the two steps
are a sunrise.

At rest it sits low and grey with its light barely out. Arming raises it clear
of the horizon with a little overshoot, fans five rays open from the top
outward, draws the ground out from beneath it, blooms first light past where it
settles, and warms everything to the brand accent — then stops and holds. The
row takes a trace of that warmth and says `confirm`. Confirming lets it break:
the sun crouches, leaps clear, a ring of first light rings out, the ground it
left dissolves, and a band of daylight sweeps across the row while the word
`asleep` fades off it. Standing down runs the sunrise backwards, which is a
sunset.

Nothing loops. Every armed value is either a transition endpoint or an
animation's resting frame, so the held state is a still frame — a permanent
repaint beside a GPU terminal is a battery bug no test catches — and
`prefers-reduced-motion` drops the motion while leaving an armed sun that still
reads as armed.

## Rules

- **Both asleep targets arm.** The row's fill button covers the whole row and
  was the bigger accident surface; it and the sun share one armed state, so
  arming on the row and confirming on the sun is one gesture. The confirm target
  is therefore the whole row, and it never moves: `confirm` renders to the *left*
  of the sun in a right-anchored strip.
- **An arm nobody confirms wakes nobody.** It stands down as soon as you leave
  the row — a click outside it or focus moving off it — on Escape, and on a
  4-second timeout. That number is a
  tripwire, not a deadline — a deliberate confirm is a second click on a target
  the pointer is already over, a few hundred milliseconds, so four seconds is
  about ten times the gesture it must never interrupt and short enough that an
  arm you walked away from is gone before you come back. Nobody confirming
  should feel it exists.
- **One member arms at a time**, which falls out of the leave-the-row rule:
  reaching a second member means leaving the first one's row, by pointer or by
  keyboard.
- **The confirm flare is not a target.** It lasts 620ms, which is long enough to
  click into; re-arming a row whose wake is already sent would put a second wake
  one click away.
- **The daemon hears exactly one `crew_wake`, on the confirm.** No wire or
  daemon change; `ProtocolVersion` is untouched.

crew-wake

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/7ab3df3610eff358d0e27e15f73f8286417cec59/delegate-wake-confirm-f5c3238d/crew-wake.mp4)

Birch is armed and then stood down by a click elsewhere; Dune is armed and
confirmed, and its day starts. Real cursor, packaged app.

## Surfaces

CLI unchanged by design. No protocol, daemon, or Linux surface. `docs/glossary.md`
said a member is woken by "one click on its row" and now says what it does.
Changelog fragment added.

`scripts/real-app-harness/drive-wake-confirm.mjs` drives the choreography in a
packaged app two ways. Through the automation bridge it has to *focus* the row to
reveal its controls, because CSS `:hover` does not answer to dispatched pointer
events — which paints a focus ring no pointer user ever sees. Its `pointer` verb
runs the same steps under the real cursor instead, which is what the recording
above is, and is the only way to show the outside-click stand-down rather than
four seconds of nothing.

## Verification

43 tests in `QueueBands.test.tsx`, covering: one click arms and sends nothing;
the second sends exactly one; arming on the row and confirming on the sun;
disarm by outside click, by focus leaving the row, by Escape, and by timeout,
each sending nothing; one member armed at a time by pointer and by keyboard; a
click into the flare sending nothing more; and the armed row's spoken contract
(`aria-label` on both targets, the visible word) which survives
`prefers-reduced-motion` untouched.

Three of those pin mistakes rather than behavior, and each was verified by
writing the mistake and watching that one test — and only that one — go red:
sending the wake from inside a state updater (StrictMode spends two days on one
click), dropping the focus disarm, and dropping the flare guard.

Full frontend suite green (2867 passed), `tsc --noEmit` clean.

Live in a packaged app on an isolated profile, against a roster of six sleeping
members, twice — once driven through the automation bridge and once under the
real cursor. Each run: one click leaves the member `asleep` in `attn crew list`;
the second wakes exactly it; an armed row left alone stands down and wakes
nobody; and only the confirmed member is awake afterwards.

https://claude.ai/code/session_01J5MoVs2yeA2eB7v7fSAYsb

### DIFF
--- a/app/scripts/real-app-harness/drive-wake-confirm.mjs
+++ b/app/scripts/real-app-harness/drive-wake-confirm.mjs
@@ -1,0 +1,135 @@
+#!/usr/bin/env node
+// Drives the two-step crew wake in a running app. Attaches to whatever app is
+// already up for the profile — it never launches or restarts one — so it is the
+// way to watch the arm-and-confirm choreography in the packaged app.
+//
+//   node scripts/real-app-harness/drive-wake-confirm.mjs focus <member>
+//   node scripts/real-app-harness/drive-wake-confirm.mjs click <member>
+//   node scripts/real-app-harness/drive-wake-confirm.mjs story <asleep> <woken>
+//   node scripts/real-app-harness/drive-wake-confirm.mjs pointer <asleep> <woken>
+//
+// `focus` reveals the row's controls through :focus-within — CSS :hover does
+// not answer to dispatched pointer events, so a synthetic hover would leave the
+// sun invisible. `story` is the whole choreography in one run: arm and let it
+// stand down, then arm and confirm.
+//
+// `pointer` is the same choreography under the real cursor, for recording it.
+// It costs an Accessibility grant and a frontmost app, and it is worth both: a
+// bridge-driven run has to focus the row to see it, which paints a focus ring no
+// pointer user ever sees, and it stands the arm down on a timeout rather than
+// showing the outside click that is the rule people actually meet.
+
+import { MacOSDriver } from './macosDriver.mjs';
+import { UiAutomationClient } from './uiAutomationClient.mjs';
+
+const wait = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
+
+async function connect() {
+  const client = new UiAutomationClient({});
+  await client.waitForManifest(10_000);
+  await client.waitForReady(15_000);
+  await client.waitForFrontendResponsive(15_000);
+  return client;
+}
+
+const focus = (client, member) =>
+  client.request('dom_focus', { selector: `[data-testid="queue-crew-select-${member}"]` });
+const click = (client, member) =>
+  client.request('dom_click', { selector: `[data-testid="queue-crew-wake-${member}"]` });
+
+// The input driver aims at the AX window frame, which starts at the titlebar;
+// the DOM measures from under it. Measured at 32 logical px on macOS by
+// comparing a wake button's DOM bounds against where the same button lands in a
+// window-relative screenshot, twice on different rows. A wrong offset does not
+// fail quietly: the click misses the sun, the member stays asleep, and both the
+// recording and `attn crew list` say so.
+const WINDOW_CHROME_PX = 32;
+
+async function pointerMapper(client) {
+  const { logicalBounds } = await client.request('get_window_bounds', {});
+  return (bounds) => ({
+    x: (bounds.x + bounds.width / 2) / logicalBounds.width,
+    y: (bounds.y + bounds.height / 2 + WINDOW_CHROME_PX) / logicalBounds.height,
+  });
+}
+
+async function main() {
+  const [verb, ...args] = process.argv.slice(2);
+  const client = await connect();
+
+  if (verb === 'focus') {
+    console.log(JSON.stringify(await focus(client, args[0])));
+    return;
+  }
+  if (verb === 'click') {
+    console.log(JSON.stringify(await click(client, args[0])));
+    return;
+  }
+  if (verb === 'story') {
+    const [standDown, woken] = args;
+    await focus(client, standDown);
+    await wait(1600);
+    await click(client, standDown);
+    console.log(`armed ${standDown}`);
+    await wait(4800);
+    console.log(`${standDown} stood down`);
+
+    await focus(client, woken);
+    await wait(1000);
+    await click(client, woken);
+    console.log(`armed ${woken}`);
+    await wait(1800);
+    await click(client, woken);
+    console.log(`confirmed ${woken}`);
+    await wait(4500);
+    return;
+  }
+  if (verb === 'pointer') {
+    const [standDown, woken] = args;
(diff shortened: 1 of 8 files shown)


figgyster (approved):
Approve. A day that can't be un-rung should cost two clicks, and keeping `attn crew wake` at one step is the right asymmetry — typing is already the deliberate act.

Verified in the box at f16688ca:

- QueueBands.test.tsx: 40/40, exactly the count in the body.
- `tsc --noEmit` clean; full frontend suite 256 files, 2859 passed, 20 skipped, zero failures (the body's 2864 count differs by environment skips; green is green).
- `git diff --check` clean. Harness deps all real: `macosDriver.mjs` and `uiAutomationClient.mjs` exist, and `dom_hover` is a live bridge verb in `useUiAutomationBridge.ts`, not a wish.
- No protocol, daemon, or wire change; ProtocolVersion untouched as promised.
- The armed-state CSS override out-specifies the hover-reveal rules, and crew rows key on `member`, so an arm survives roster re-renders.

The StrictMode regression test is the right kind of witness — the only one that goes red against the mistake it pins. And the one-shot animation discipline (every armed value is a transition endpoint or a resting frame, so the held state is a still frame beside a GPU terminal) is the sort of thing that usually gets written down after a battery report, not before it.

Two small cracks in the single-wake guarantee, neither blocking:

1. Keyboard can arm two members at once. The disarm set is pointerdown-outside, Escape, timeout — all pointer-first. A keyboard user who arms row A, tabs to row B and presses Enter gets a second armed row: "one member arms at a time" falls out of the outside-click rule, and keyboard activation isn't a click. Waking both still takes two deliberate confirms each, so the blast radius is small, but the rule as written doesn't hold for the keyboard path. Disarming on focusin, or disarming siblings when one arms, would close it.

2. A click inside the 620ms breaking flare re-arms the row. `trigger` arms on any phase that isn't 'armed', and 'breaking' isn't 'armed' — so a fast third click right after the confirm flips the row back to armed with "confirm" showing, and a fourth click wakes the member a second time: billed, un-runnable, the exact double-wake this feature exists to prevent. The test comment says a third click "starts a fresh two-step," so it's deliberate, and the visible word self-announces — the guard is the user noticing, not the code. Ignoring `trigger` while 'breaking' is a one-line fix if you want the guarantee in the code instead.

The sunrise is a good bit of theater, and the packaged-app recording with the real cursor earns its keep. Nice case.

victorarias:
Both cracks closed at b4ea71ac — thanks, they were the two places the guarantee actually leaked.

**Keyboard multi-arm.** Focus leaving the row now disarms it, alongside the outside click. That is the rule stated the way it should have been stated in the first place: an arm stands down when you leave the row, and "one member at a time" falls out of that for pointer and keyboard alike. A test arms one row, focuses another's sun, and asserts the first is no longer armed; a second test moves focus *within* the armed row and asserts the arm survives, so the disarm cannot be over-eager.

**Re-arming inside the flare.** `trigger` now returns on `breaking`. Four clicks in a row send exactly one wake.

Both are mutation-verified the same way as the StrictMode test: with the guard removed, exactly one test goes red, and it is the one written for it.

The changelog fragment and the PR body follow the corrected rule.

figgyster (approved):
Both notes answered in one commit, each with a witness that goes red against its own mistake. The single-wake guarantee now lives in the code, not in the user noticing.

Verified at b4ea71ac:

- QueueBands.test.tsx: 43/43. Three new witnesses: "is not a target while it is flaring" (four clicks, one wake, stays breaking), "arms one member at a time for the keyboard too" (focusIn on another row disarms), and the complement that matters — "keeps the arm while focus moves inside its own row", which pins the over-disarm edge a naive focusin guard would have walked into.
- `tsc` clean; full suite 2862 passed / 20 skipped / zero failures — exactly 2859 plus the three new tests.
- `git diff --check` clean; the changelog fragment now says "as soon as you leave it" so the docs match the fourth stand-down route.

The focusin disarm is the right mechanism — focus is the keyboard's pointer, and the `contains()` guard keeps arming-on-the-row-and-confirming-on-the-sun intact for keyboard users too. And `if (phase === 'breaking') return;` is precisely the one-liner, with the flare pinned as a non-target. The hook comment even says the new rule out loud: "reaching a second member means leaving the first one's row, by pointer or by keyboard."

Re-approving at b4ea71ac. That's the eleventh loop closed this week — you're making a habit of it.

PR #922, opened 2026-08-15
fix(profile): fence non-production worlds

victorarias:
## TL;DR

Named-profile app bundles are now visibly labeled and remain authoritative over every daemon, session, crew-home, and cache path they own. A copied database or inherited shell environment can no longer make attn write into the default profile's crew homes. Verification profiles other than `dev` also cannot rewrite user-global agent skills; `dev` deliberately shares those skill directories with the default profile so attn's development loop uses the freshly bundled skill.

## Why

Profile data directories were already separate, but several runtime paths still depended on inherited environment or absolute paths stored in the database. The most confusing symptom was visual: a verification profile could look identical to production and could carry synthetic crew members named after real durable members. The more serious landmines were silent cross-profile routing and writes when a shell exported stale routing variables or a production database was copied into another profile.

These are guardrails against attn becoming a confused writer. They are not an authentication or OS-user security boundary.

## What changed

- Named bundles show a quiet, persistent `profile <name>` badge. The default bundle renders exactly as before.
- The Tauri shell clears all inherited routing and bundle-path overrides before applying its compiled profile, including data and plugin directories in production builds.
- Both PTY workers and conversation hosts receive the daemon's complete routing block after login-shell and plugin environment merges, including `ATTN_WS_PORT`. Internal worker transport variables are removed before the child starts.
- Crew member homes, charters, handoff directories, and handoff letters are revalidated under the active daemon's crew root at import and immediately before every read, write, or turnover launch. Refusals name the member, actual path, expected root, and copied-database cause.
- Project cwd and awareness paths remain unrestricted except when they resolve inside another profile's crew tree, including symlinked and currently missing descendants. Priming still tolerates stale project paths, while `crew set` preserves its existing existence check and wake preserves its existing missing-CWD refusal.
- Profiles other than the default and `dev` skip user-global agent skill synchronization with an explicit log entry. `dev` is the sole named-profile exception because it is the attn-on-attn development loop. Remote binary and app-runtime caches now live under the active profile's data root.
- Harness guidance reserves real crew names for production identities and requires synthetic members to pin `claude-haiku-4-5` unless a scenario documents why it needs more capability. Existing harness/profile builders contained no crew fixtures to rename.
- Crew and handoff CLI help now describes the active profile's crew directory instead of hardcoding the default path.

## Reviewer guide

1. Start with `internal/daemon/crew_paths.go`: it contains the two distinct fences—member-owned files must be inside this profile, while project paths are rejected only inside another profile's crew tree.
2. Follow `spawnRoutingEnv` through the PTY worker transport and host-session spawn to verify that exact daemon routing is the final environment overlay.
3. Review the Tauri scrub and Sidebar marker together: the compiled bundle profile owns both routing and legibility.
4. Check the skill-sync gate and remote cache helpers for the user-global/profile-local write split and the explicit `dev` exception.

## Manual verification

Built and launched a `fence-proof` app while deliberately supplying conflicting inherited profile, data, database, config, plugin, socket, and WebSocket values. Bundled preflight passed against the `fence-proof` daemon. The app displayed the `profile fence-proof` badge, and a fresh shell spawned from it reported `ATTN_PROFILE=fence-proof` plus only profile-local routing paths. The profile's sole crew member was the synthetic `Quartz`; none of the production member names appeared. The daemon log also recorded all three verification-profile skill-sync skips.

attn-profile-fence-clean

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/ce0fc63b3b0137ac02f6b588a04c4bc01270fb30/delegate-profile-fence-a5ca981c/attn-profile-fence-clean.mp4)

## Verification

- `make test-quick`
- Full frontend suite: 2,872 passed, 13 skipped
- Targeted regression coverage for daemon-owned `ATTN_WS_PORT`, stale priming paths, foreign crew descendants, and default/`dev` skill-sync ownership
- Rust profile tests and Linux amd64 build

## Out of scope

- No socket or WebSocket authentication.
- No crew lifecycle behavior changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -167,6 +167,10 @@ output as product evidence.
 ### Packaged-app harness
 
 - Single-tenant: never run packaged-app scenarios in parallel.
+- Crew fixtures in harness and verification profiles use obviously synthetic
+  names, never real member names such as `keel`, `alder`, or `trellis`. Pin
+  synthetic members explicitly to `claude-haiku-4-5`; use a stronger model only
+  when the scenario is testing work that needs its intelligence.
 - Multiple scenarios: `pnpm --dir app run real-app:serial-matrix`.
 - Rebuild before evidence-sensitive runs.
 - Harness uses active `ATTN_PROFILE`, otherwise `dev`;
--- a/app/src-tauri/src/profile.rs
+++ b/app/src-tauri/src/profile.rs
@@ -3,9 +3,9 @@
 //! The `ATTN_BUILD_PROFILE` env var is read at *compile* time and baked
 //! into the binary. At startup a profile-baked app makes that profile
 //! authoritative for daemon routing: it sets `ATTN_PROFILE` and
-//! `ATTN_WS_PORT`, and removes socket/database/config paths inherited from
-//! a parent attn terminal. That prevents the dev bundle from reaching the
-//! production daemon while being launched from a production session.
+//! `ATTN_WS_PORT`, and removes path overrides inherited from a parent attn
+//! terminal. That prevents any bundle from reaching another profile while
+//! being launched from one of its sessions.
 //!
 //! A named per-profile build additionally bakes `ATTN_BUILD_WS_PORT` and
 //! `ATTN_BUILD_BUNDLE_ID`, both resolved by the single authority
@@ -27,6 +27,15 @@ const BUILD_PROFILE: Option<&str> = option_env!("ATTN_BUILD_PROFILE");
 // crate. Default/prod builds leave these unset and use the fallbacks below.
 const BUILD_WS_PORT: Option<&str> = option_env!("ATTN_BUILD_WS_PORT");
 const BUILD_BUNDLE_ID: Option<&str> = option_env!("ATTN_BUILD_BUNDLE_ID");
+const ROUTING_PATH_OVERRIDES: [&str; 7] = [
+    "ATTN_SOCKET_PATH",
+    "ATTN_DB_PATH",
+    "ATTN_CONFIG_PATH",
+    "ATTN_DATA_DIR",
+    "ATTN_PLUGIN_DIR",
+    "ATTN_DAEMON_BINARY",
+    "ATTN_BUNDLED_PLUGIN_DIR",
+];
 
 /// Returns the compile-time profile name (empty string for default).
 pub fn build_profile() -> &'static str {
@@ -68,13 +77,15 @@ pub fn default_port_for_build_profile() -> &'static str {
 /// `ATTN_WS_PORT` (directly or via a spawned subprocess).
 pub fn apply_build_profile_env() {
     let profile = build_profile();
-    if !profile.is_empty() {
-        for key in ["ATTN_SOCKET_PATH", "ATTN_DB_PATH", "ATTN_CONFIG_PATH"] {
-            env::remove_var(key);
-        }
-        env::set_var("ATTN_PROFILE", profile);
-        env::set_var("ATTN_WS_PORT", default_port_for_build_profile());
+    for key in ROUTING_PATH_OVERRIDES {
+        env::remove_var(key);
     }
+    if profile.is_empty() {
+        env::remove_var("ATTN_PROFILE");
+    } else {
+        env::set_var("ATTN_PROFILE", profile);
+    }
+    env::set_var("ATTN_WS_PORT", default_port_for_build_profile());
 }
 
 fn data_dir() -> Result<PathBuf, String> {
@@ -176,6 +187,19 @@ fn decide_automation_enabled(automation: Option<&str>, profile: Option<&str>) ->
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn every_path_that_can_redirect_a_bundle_is_scrubbed() {
+        for key in ROUTING_PATH_OVERRIDES {
+            env::set_var(key, format!("foreign-{key}"));
+        }
+
+        apply_build_profile_env();
+
+        for key in ROUTING_PATH_OVERRIDES {
+            assert_eq!(env::var_os(key), None, "{key} survived app startup");
+        }
+    }
 
     #[test]
     fn unbaked_build_falls_back_to_prod_resources() {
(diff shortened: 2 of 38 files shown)


figgyster (changes requested):
Case review of the profile fence. Read the whole fence end to end: crew_paths.go, the routing overlay through both runtimes, the Tauri scrub, the skill gate, the cache move, the harness rules. Verified in the sandbox: all Go suites (daemon/pty/agent/hub/config), the full frontend suite at the new head (256 files, 2867 passed / 0 failed, tsc clean), diff --check clean. The fence itself is real: member homes and charters are revalidated under the active crew root on import and every read/write, the symlink escapes are closed (handoffs dir, letters, foreign profile roots), the routing block is the final overlay in both the PTY and host runtimes with the transport envelope scrubbed, and the app-side scrub now applies to every bundle including prod. The three findings below are the things I'd want answered before this merges; the first one is the only one I'd call a hole.

1) The routing block omits ATTN_WS_PORT — and this repo's own profile-env treats it as a routing override. spawnRoutingEnv (internal/daemon/spawn_routing.go) pins ATTN_PROFILE/DATA_DIR/DB_PATH/SOCKET_PATH/CONFIG_PATH/PLUGIN_DIR but not the daemon's WS port. The daemon binds its HTTP/WS server on config.WSPort() (internal/daemon/daemon.go:2272), which reads ATTN_WS_PORT env first, then the per-profile default — and the CLI's own profileRoutingOverrides list (cmd/attn/profile_env.go:16) includes ATTN_WS_PORT, so the codebase's own definition of the routing contract covers it. The child ends up with whatever the login shell exported for that key, because the final DaemonEnv overlay doesn't re-pin it. And children genuinely use the port: `attn plugin list/install/uninstall` dial ws://127.0.0.1:WSPort()/ws (cmd/attn/plugin.go:202) — mutations — plus ws-relay and versioncheck hit the same port. A stale ATTN_WS_PORT in a login shell is exactly the 'shell exported stale routing variables' landmine the changelog claims to kill ('every process they launch carries that profile's routing even when a login shell exports a conflicting value'), and this key still wins. The changelog claim is one key short. One-line fix: add "ATTN_WS_PORT=" + config.WSPort() to spawnRoutingEnv, and extend TestBuildSpawnEnv_DaemonRoutingOverridesLoginAndPluginEnvironment plus TestConversationHostCarriesTheSessionIdentity with a conflicting login-shell ATTN_WS_PORT so the gap can't come back.

2) The wake/prime existence checks are stricter than the PR body says. The body says 'Project cwd and awareness paths remain unrestricted except when they resolve inside another profile's crew tree' — but crewLaunchDir and crewPriming now push every awareness dir through resolveCrewWorkDir → resolveCrewDir, which demands the directory exist (os.Stat, 'is not there'). Before this PR, awareness dirs were never existence-checked at wake or prime; a member with a stale awareness dir (project moved or deleted) could still be woken, and now they're grounded until someone runs `attn crew set` — and the refusal message doesn't offer that hint the way the CWD refusal does. If existence-checking is the intent, the body and the error message should say so; if only the foreign-crew-tree condition is meant, resolveCrewWorkDir should skip the existence demand for awareness dirs. Same class for CWD in crewPriming (previously priming never touched CWD).

3) The skill-sync gate silently excludes the dev profile. userGlobalSkillSyncEnabled() returns true only for the default profile (internal/agent/attn_skill.go). The dev bundle is a named profile (ATTN_PROFILE=dev baked by make dev), so the dev daemon — the attn-on-attn inner loop — now never installs the bundled skill into ~/.agents / ~/.claude / ~/.copilot. On a machine where prod also runs, the skill dirs already exist from prod, so dev agents read the last prod-synced copy; on a dev-only install the attn skill never lands and dev agents run without it. Also: iterate on the skill, rebuild dev, and the dev daemon won't install your new copy — you'd have to sync via prod. If dev is meant to share the user-global skill dirs with prod (same machine, same user), it probably deserves to be in the sync set; if not, the plan's 'user-global skill directories belong to the default profile' line should name dev as collateral. Worth a deliberate answer either way.

Everything else held up under verification — the copied-DB refusal names the member and both roots, the symlink tests are real witnesses, the route-block transport (ATTN_PTY_DAEMON_ENV JSON, scrubbed before child exec, re-applied after ExternalEnv) is pinned by tests in both runtimes, and the Tauri scrub now covers default builds too with a test. Rust test reviewed by inspection (no cargo in my sandbox). Fix the port, answer the other two, and this fence is complete.

figgyster (approved):
Loop closed on c30b0493 — all three findings answered with code and witnesses, verified at the new head.

1) The WS port is in the routing block now: spawnRoutingEnv carries ATTN_WS_PORT=config.WSPort(), and both env tests pin a conflicting login-shell/plugin port losing to the daemon's (19999/18888 vs 25432 in the PTY test; same shape in the host identity test). That's the last key that could redirect a child's plugin traffic to another profile's daemon — the changelog's 'every process they launch carries that profile's routing' claim is now true for the port too.

2) The existence demand is off the wake/prime path: resolveCrewDir split into absoluteCrewDir (no stat) + the record-time checker, resolveCrewWorkDirForHome rides the tolerant half so the foreign-crew-root fence still applies to missing paths (pinned by TestCrewSet_AMissingPathInsideAnotherProfilesCrewIsRefused) while a stale cwd/awareness dir no longer grounds a member (TestCrewPriming_StaleProjectPathsDoNotBlockPriming), and crew set keeps its must-exist-at-record contract via resolveCrewRecordedDir. crewLaunchDir still names a missing cwd at wake with the `crew set` hint — the body's 'remain unrestricted except foreign crew tree' wording is now accurate.

3) dev is back in the sync set: userGlobalSkillSyncEnabled() admits "" and "dev", the attn-on-attn loop installs the freshly bundled skill again, and the verification profiles stay fenced — TestUserGlobalSkillSyncRunsForDevProfile proves dev writes all three skill trees while fixture-lab still skips. Plan and changelog updated to match.

Verified in the sandbox at c30b0493: daemon/agent/pty/hub/config suites all green (daemon 91s, no flakes), the three new witness tests pass in isolation, diff --check clean. Frontend untouched by the loop commit, so the earlier 2867-passed/tsc-clean run stands. Approved — the fence is complete, and the note→witness loop closes a fourteenth time.

PR #924, opened 2026-08-15
refactor(terminal): delete the historical-replay machinery

victorarias:
Restoring a session used to mean megabytes of VT written into a blank model in
slices, and the client carried everything that made that survivable. Decoding a
snapshot (#923) is one operation, so that machinery has had no caller since.
This deletes it.

## What goes

- **The cooperative byte budget** (`utils/cooperativeReplay.ts`) that yielded
  between replay chunks so a large restore did not block the frame.
- **The generation fence.** A live resize landing mid-replay bumped a
  generation; every queued write and resize compared against it and dropped
  itself, with a follow-up refit owed when the dropped op was a resize.
- **The re-attach after a cancelled replay.** `onReplayInterrupted` →
  `handleReplayInterrupted`, a debounced re-request of the whole attach because
  the model was left missing history.
- **The queued-geometry promise.** `liveResizeConflictsWithQueuedReplay` +
  `PendingReplayGeometry`, which `fit()` and `resizeLocal()` consulted to decide
  whether a live resize conflicted with the geometry the queued history ended
  at (`bail: 'replayPending'` / `'replayStale'`).
- **The 16KB chunking** of `attach_replay` data in the pane runtime, and
  `write()`'s `historicalReplay` option along with it — including its
  restore-chunk branch into the model-op ring, which `restoreSnapshot` now owns.

## What stays, and why

`resizeLocal` keeps the option under the name it actually means now:
`restore: true` sizes the model to the grid an attach reports and does **not**
paint or refit, because the snapshot adoption queued right behind it is what
the pane should show. That is the one behavior the flag ever had that a decode
did not remove.

Names follow: `attach_replay` → `attach_restore`, `replay_complete` →
`restore_complete`, and the diagnostics field `historicalReplay` → `restore`.
All three are app-internal; nothing on the daemon wire moves, so no protocol
bump.

Net −561 lines.

## Verification

Installed profile `noreplay` from this branch (preflight PASS, protocol 252
CLI↔app↔daemon) and ran the scenarios that exercise restore and the
resize-vs-restore paths this machinery used to guard:

| scenario | result |
| --- | --- |
| `snapshot-scrollback-restore` | ok — 2,000 of 2,000 rows restored, first row `SNAPROW_00001` |
| TR-201 local relaunch with an existing split | ok (first run's agent never answered the prompt; clean on rerun) |
| TR-204 local relaunch formatting | ok |
| TR-401 local window resize | ok |
| ghostty scrollback anchor | ok |

Frontend: 255 files / 2,876 tests green, `tsc --noEmit` clean. The tests that
pinned the deleted behavior are gone with it (replay chunking, the debounced
re-attach, the queued-geometry policy); the ones that pinned behavior that
survives were rewritten against the new names.

## Also here: `real-app:bridge-smoke` was failing on a dead predicate

Unrelated to the deletion, found while verifying it. The scenario waited for a
pane with `kind === 'shell'` after `split_pane`, and no such pane exists:
`WorkspaceLayoutPaneKind` has one value, `agent` — a shell pane is an agent pane
whose agent is the shell (`title: "shell"`). The wait timed out 20s after a
split that had already succeeded. It now waits for the pane the workspace did
not have before the split, which is what the daemon observer's
`waitForUtilityPane` already does, and matches the echoed token across the wrap
the 80-column pane puts in the middle of it. Green against `noreplay`.

The same dead filter is still in the three ad-hoc `bridge-repro-*` scripts;
those are repro harnesses for closed bugs and I did not touch what I could not
run.

https://claude.ai/code/session_01FcwbumRJFVLKpZQWowS9Pb


figgyster (changes requested):
This is the removal #923 named for itself. The cooperative byte budget, the generation fence, the interrupted-replay re-attach, the queued-geometry promise, the 16KB chunking — all gone, and good riddance. I read the whole corpse before signing the certificate: every rename is consistent (attach_replay→attach_restore on the resize source, replay_complete→restore_complete, historicalReplay→restore), nothing on the daemon wire moves so the no-bump claim holds, and the attach ordering still lands restore-resize → snapshot adoption → seeds → queued live data → restore_complete on one chain. The one behavior that earned its keep — `restore: true` sizes the model without painting it — survived the surgery intact.

Ran the whole frontend suite in the box with the wasm fetched and sha-verified against ghostty-vt-wasm.lock: 2871 passed, 18 skipped, zero failed. tsc clean. The tests that pinned the dead machinery died with it; the survivors were honestly rewritten, not just renamed — the suppressResponses test now witnesses the actual contract (daemon says dont-suppress, responses return to the live PTY). diff --check is quiet except a git quirk on the fully-deleted resizePolicy test (it flags a blank line at EOF in a 195-line file that ends `});` — false positive; your own Changes check agrees).

One thing blocks the merge:

**1. The changelog fragment is invalid YAML.** `changelog.d/drop-historical-replay.yaml` ends with two stray lines at column zero:
```
</content>
</invoke>
```
Tool-output contamination that rode along with the file. The Changelog gate (`go run ./cmd/changelog-check`, strict yaml.v3 with KnownFields) rejects it — "could not find expected :" at line 14 — so that check goes red, and worse, release-time `compile-changelog.sh` runs the same validator, so this poisons the release pipeline, not just tonight CI. Every other fragment in the directory ends clean. Delete those two lines and the gate passes.

Non-blocking, for the ledger:

**2. `internal/daemon/nudge_countdown.go:552` still recognizes "attach_replay" as a non-user input source** for the nudge splice guard, and its tests name it too. Nothing on either side of the wire can produce that string anymore — the app never sends it post-deletion. Dead case, harmless, but it is now the last "replay" left standing in the codebase. Sweep it whenever the daemon next touches that switch.

Fix the fragment and this is a merge. −546 lines and the last replay code in the app is deleted — the snapshot arc is done the way it should be: one operation, no callers, no corpse.

figgyster (approved):
Both findings answered in two commits, and the second one is a gift I did not ask for.

**c2106171** drops the stray `</content>` / `</invoke>` from the fragment — the exact two-line fix, nothing else. Verified at the head: the fragment parses clean (`kind: internal`, `area: terminal`, `change` present, no unknown keys), so `changelog-check` and the release-time compile both pass. The Changelog gate is no longer a landmine.

**0e11ba09** is a real harness bug, found the hard way: the bridge-smoke split wait filtered `kind === shell`, but every workspace pane is kind `agent` — a shell pane is an agent pane whose agent is the shell — so the filter matched nothing and the wait timed out no matter how well the split went. The fix waits for the pane the workspace did not have before the split, via the same `waitForUtilityPane` that bridge-perf and bridge-remote-hub already use, and matches the echoed token across the wrap an 80-column pane puts in it. Syntax-checked in the box; the pattern is the house one. This one has "pre-existing bug that would have bitten the next person to run the harness" written all over it — glad it tripped tonight instead of later.

Head diff vs my reviewed SHA is exactly those two files. Everything I verified last pass still holds: the machinery is gone, the renames line up, the wire did not move, the restored contract survived. Re-approving at 0e11ba09.

The one open thread is my non-blocking ledger note: `nudge_countdown.go:552` still names `attach_replay` as a suppressed input source. Dead case, no rush — sweep it the next time the daemon touches that switch. The changelog fragment, the tests, and the arc itself are done. Merge when CI agrees.

PR #925, opened 2026-08-15
fix(terminal): name the format a snapshot was written in

victorarias:
A pty-worker outlives an install. A session that was running when attn updated
is therefore served a snapshot encoded by the build it started under, and
nothing on the path could tell: the worker RPC field carrying those bytes never
changed name and the RPC minor never moved, so `IsCompatibleVersion` reports
major 1 / minor 1 on both sides while the bytes underneath went from a VT dump
to ghostty's snapshot record stream.

The client took them and threw, and the throw was routed to model-fault
recovery — a new epoch, a remount, a reattach, the same bytes. Measured on the
production install that shipped it: 421 faults in six seconds on two sessions,
421 attach round-trips behind them, each writing a DOM snapshot to
`ui-diagnostics.jsonl`, all behind a red banner.

```
PTY attach result: policy=same_app_remount ghostty_snapshot_bytes=17386
                   replay_decision=use_ghostty_snapshot
→ ghostty_model_fault operation=restoreSnapshot
                      error="ghostty_snapshot_decoder_ready failed"
```

## The upgrade path

A snapshot now names its format (`attach_result.snapshot.format`), and the
client compares that against its own decoder before decoding anything. A
foreign tag — or no tag, which is every worker running today — is treated as no
snapshot, which is already a supported attach: no reset, the client's own
watermark as the dedup baseline, the live stream painting on. The session comes
back without its earlier scrollback and keeps working; restarting it restores
full behavior. Nothing is shown to the user: a session that comes back thin is
quieter than a banner about a format tag, and the condition ends by itself.

The comparison lives in the client because the client owns the decoder. The
chain worker → remote daemon → local daemon → app can skew at any link, and
only the last link knows what it can read — so this covers a relayed remote
attach with no hub plumbing.

## The tag is derived, never typed

`scripts/snapshot-format.sh` hashes the two artifacts that decide the format —
`ghostty-vt-native.lock` (the encoder's archive) and `ghostty-vt-wasm.lock`
(the decoder's module) — plus a salt for a change to attn's own encode/decode
glue that moves neither. The Makefile feeds it to `buildinfo.SnapshotFormat`;
`vite.config.ts` calls the same script for the frontend. One implementation,
two consumers, so a bundle's worker and app agree by construction.

The pin alone would not have caught this incident: #923 changed the native
lock's `key=` (it dropped the carried patch) while `ghostty-vt.pin` stayed at
`d760ee96`. Deriving from the locks rather than from `SourceFingerprint` is
also deliberate — a tag that moved on every commit would cost every live
session its scrollback on every upgrade, for a format that did not change.

## A decode failure is not a model fault

Independent of the tag: `restoreSnapshot` now catches the throw, records it
once, and leaves the model alone. `adoptSnapshot` releases and throws before it
swaps the handle on both failure paths, so the model it declined to replace is
intact — that ordering is now load-bearing and has its own test. This is the
net under any skew the tag fails to predict, and it turns the cost of a bad
payload from an unusable session into one missing restore.

Residual, stated plainly: the pane's `reset` is emitted before the restore, so
a decode that fails *after* the tag matched leaves the pane blank until the
session next prints. The tag is what keeps the predictable skew away from it.

## Verification

Live, on a throwaway profile, simulating a real format change across an
install:

- Seeded a session, bumped the format, ran `make install PROFILE=fmtskew`. The
  worker survived the install — the incident's exact shape.
- The daemon served `snapshot_format=199ae0429d61` to a decoder built at
  `08b9b91b6b6c`.
- The pane attached, restored no history (expected), **took live input**, and
  `ghostty_model_fault` went 0 → 0.

Same-build restore is unchanged: `scenario-snapshot-scrollback-restore` still
recovers 2000 rows across a relaunch, which is what proves the Go and vite tag
computations agree in a real bundle.

Unit: `classifyAttachRestore` drops a foreign-tag and an absent-tag snapshot and
leaves the attach unreset on the client watermark; `restoreSnapshot` on
undecodable bytes records the rejection, raises no model fault, and leaves the
model taking writes; `adoptSnapshot` rejects garbage without touching the
terminal; the worker RPC round-trips the tag and decodes a legacy result to
empty; the daemon puts it on the wire. Full Go suite and 2881 frontend tests
green.

No evidence recording: the visible change here is the absence of a banner, and
a clip of a session opening normally shows nothing without a before. The fault
counts and daemon log lines above are the evidence.

## Surfaces

- **Protocol** — `AttachSnapshot.format`, `ProtocolVersion` 252 → 253
  (`generated.go`, `generated.ts`, `useDaemonSocket.ts`).
- **Worker RPC** — `AttachResult.GhosttySnapshotFormat`, additive and skew-safe;
  absent from an old worker, which reads as incompatible.
- **Backends** — worker passes it through, embedded stamps its own process's.
- **Hub** — a remote daemon built from source carries the tag in its ldflags,
  or every remote session would attach without scrollback.
- **Linux** — build plumbing, not cgo; the pure-Go stub emits no snapshot.
- **CLI** — nothing reads a snapshot.
- **Docs** — plan doc, an AGENTS.md contract line, changelog fragment.

Plan: `docs/plans/2026-08-16-snapshot-format-skew.md`.

https://claude.ai/code/session_011GhcqUUDfASUjH6YMfo4ws

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -579,6 +579,17 @@ Design and gate decisions:
   fallback — a snapshot-less attach keeps whatever the client has and dedups the
   live stream against `last_seq`. See
   [docs/plans/2026-08-16-snapshot-restore.md](docs/plans/2026-08-16-snapshot-restore.md).
+- A snapshot names the format it was written in
+  (`attach_result.snapshot.format`, derived at build time by
+  `scripts/snapshot-format.sh` from the two ghostty locks). A pty-worker
+  outlives an install, so an upgraded app is routinely offered bytes an older
+  encoder wrote; the client compares that tag against its own decoder and
+  treats anything else — a foreign tag or no tag — as no snapshot, which is
+  already a supported attach. Never decode a snapshot on the strength of the
+  field being present, and never route a decode failure to model-fault
+  recovery: the model is fine, the payload is not, and a new epoch reattaches
+  straight back into the same bytes. See
+  [docs/plans/2026-08-16-snapshot-format-skew.md](docs/plans/2026-08-16-snapshot-format-skew.md).
 - Encoding fails outright when the worker's parser sits mid-sequence and
   continuation tracking was off, so the worker enables it at construction, not at
   snapshot time.
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,9 @@ VERSION ?= $(shell bash ./scripts/version.sh)
 BUILD_TIME ?= $(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
 SOURCE_FINGERPRINT ?= $(shell bash ./scripts/source-fingerprint.sh --field fingerprint)
 GIT_COMMIT ?= $(shell bash ./scripts/source-fingerprint.sh --field commit)
+# Identity of the terminal-snapshot wire format. The frontend computes it from
+# the same script, so a bundle's worker and app agree by construction.
+SNAPSHOT_FORMAT ?= $(shell bash ./scripts/snapshot-format.sh)
 OUTPUT ?= $(BINARY_NAME)
 # Guards the macOS-only branches below (code signing, the lsof port check). It
 # was used by three recipes and defined by none, so `$(UNAME_S)` expanded to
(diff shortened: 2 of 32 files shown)


figgyster (approved):
This is the incident report from the night #923 shipped, wearing a fix. Opened a session after the 00:26 install and the worker that survived it served bytes the new decoder had never seen — `ghostty_snapshot_decoder_ready failed`, 421 model faults in six seconds, each fault remounting the pane, reattaching, and being handed the identical bytes. Two real defects stacked: the field didn't move and the version gate couldn't see it, and then a decode failure was treated as a broken model when the model was the only healthy thing in the room.

The fix goes three layers deep, and each layer earns its keep.

**The tag is derived, never typed.** `scripts/snapshot-format.sh` hashes the native encoder's lock and the wasm decoder's lock (plus a salt), and both the Makefile and `vite.config.ts` run the same script, so a bundle's worker and app agree by construction. Deriving from the locks rather than the pin is exactly the right call — this incident moved the native lock's `key=` while `ghostty-vt.pin` sat still, so a pin-born tag wouldn't have caught it.

**The comparison lives where the decoder lives.** `classifyAttachRestore` declines a snapshot whose `format` isn't this build's — foreign tag, or no tag at all — as no snapshot, falling into the already-supported snapshot-less attach: no reset, the client's own watermark, the live stream painting on. Absence-as-unknown is what makes the fix reach the workers already running out there, this incident's included.

**The net under any skew the tag can't predict.** `restoreSnapshot` catches a decode throw, records one `snapshot_decode_rejected` diag, and leaves the model alone. I checked the claim in the actual decoder: `adoptSnapshot` releases and throws before it swaps the handle, on both failure paths, so the model it declined to replace is intact and keeps taking writes.

Verified all of it in the box at bb5d49cc: protocol/pty/ptyworker/ptybackend green, the full daemon suite green (92.8s), frontend 2876/2876 with the wasm sha-checked against the lock, tsc clean, diff-check clean. The new tests are eyewitnesses, not bought ones — the GhosttyTerminal test reproduces the exact incident error string and asserts no model fault, no new epoch, and the pane still accepting live output; the terminal.snapshot test feeds real garbage to the real decoder and proves the refused decode leaves the terminal intact; the legacy-omit worker payload decodes as empty and reads as unreadable. Protocol 253 is matched on both ends (constants.go and useDaemonSocket.ts, held together by version_test), generated.go/generated.ts are regenerated, and native-ui is archived so there's no Rust version left to drift. CI's already green where it's finished; the E2E scenarios — the scrollback-restore one in particular — are the last word on the two tag computations agreeing in a real bundle.

Nothing blocking. Two notes for the ledger, both non-blocking: the salt constant in the script is the one manual knob in an otherwise derived system — the comment says when to bump it, and the failure direction of a missed bump is one lost restore, not a loop, thanks to the catch. And this upgrade deliberately costs every pre-existing session its scrollback until restart — that's the changelog's promise, and the reason this ever reaches the sessions the incident left behind.

PR #926, opened 2026-08-15
fix(opencode): show a delegated session in its pane, on the model it was pinned to

victorarias:
Two defects in `plugins/attn-opencode`: a delegated OpenCode session was invisible in its pane, and its model/effort pin was never verified — and, it turns out, the effort half of it never applied.

## The agent was invisible

The driver created the native session over HTTP and called `/tui/select-session`. That route publishes a bus event and answers `true` whether or not a TUI is attached to the event stream — proved against a headless `opencode serve`, which answers `true` with no TUI in existence. The TUI subscribes to `/event` about three seconds after the server answers `/global/health`, so a select published before that is dropped forever. The pane kept the splash while the agent worked, and there is no TUI-liveness signal anywhere in the API to wait on.

So the TUI owns the session now. A pinned launch runs `--agent build` with the model in the injected config, the delegated turn is submitted through the TUI (`/tui/clear-prompt` → `append-prompt` → `submit-prompt`), and the `session.created` event that submission produces is the receipt: it claims the run, stops the staging retry, and binds the session the TUI itself opened. The claim is taken the moment the event lands rather than after the bind's round trips, so a prompt the TUI already took is never submitted twice (that duplicate was real, and observed: two identical user messages in one session).

## The effort pin applied to nothing

The TUI reads the selected model's variant from its own XDG state (`~/.local/state/opencode/model.json`), not from the agent config. A delegation pinned to `--effort low` opened at `max` — whatever was last used interactively — and the earlier `max` runs "verified" only by coincidence. Each pinned run now gets its own `XDG_STATE_HOME` seeded with the pin (the user's display preferences are copied in, so the pane still looks like their OpenCode). That also stops delegated runs from rewriting the model selection the user made themselves.

Both ends are checked now:

- Before staging, against `/config/providers`: an effort the model does not offer fails the run and names what it does offer. MiMo declares `variants: {}`, so `--effort max` on it was silently dropped; `deepseek-v4-pro` has no `low`.
- After the session opens, against the session itself: a different provider, model, or variant sets the run degraded with the pin and what arrived, and leaves it unbound.

## Live verification — profile `ocfix`, current branch

Provider / model / variant read back from each run's own OpenCode server:

| delegated pin | session model read back | outcome |
| --- | --- | --- |
| `opencode/deepseek-v4-flash` `max` | `{opencode, deepseek-v4-flash, max}`, agent `build` | pin verified |
| `opencode/deepseek-v4-flash` `low` | `{opencode, deepseek-v4-flash, low}` | pin verified (was `max` before this change) |
| `opencode-go/deepseek-v4-flash` `max` | `{opencode-go, deepseek-v4-flash, max}` | pin verified, turn completes |
| `opencode-go/deepseek-v4-flash` `low` | footer `… OpenCode Go · low` | pin verified, turn completes |
| `opencode-go/mimo-v2.5` `max` | — | refused: `OpenCode model opencode-go/mimo-v2.5 cannot honor the delegated variant max: it offers no variants`, nothing staged |
| `opencode-go/deepseek-v4-flash` `high` | — | `OpenCode effort "high" has no verified variant mapping` at spawn, unchanged |

Visibility: the pane shows the delegated turn 2.1s after `attn delegate` returns (it never did before). Single submission per run, verified in the message list. `~/.local/state/opencode/model.json` is byte-identical before and after the whole matrix.

`opencode/*` (OpenCode Zen) models answer `APIError 401 "Model is disabled"` on this account right now — `opencode run --model opencode/deepseek-v4-flash` fails the same way outside attn, so it is an account/provider condition, not this change. The pin itself still reads back correct on those runs, which is what the pin verification is about. The `opencode/mimo-v2.5-free` leg the brief asked for is covered by the same finding: the model is both Zen-disabled and variant-less.

clip

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/22b8163f610376412552657c48343dae5e533d36/delegate-oc-fix-3f3e7a70/clip.mp4)

## Tests

`plugins/attn-opencode/test/driver.test.ts`: staging retries until the TUI attaches and binds the session it opened; stages once even while the bind is in flight; refuses an effort the pinned model does not offer; degrades loudly on a model mismatch; the launcher runs on the pinned agent with the pin in its own TUI state. Each was checked red by mutation.

https://claude.ai/code/session_01AzXBdU9mMw7yaKujfeHBgk

### DIFF
--- a/changelog.d/opencode-delegations-visible-and-pinned.yaml
+++ b/changelog.d/opencode-delegations-visible-and-pinned.yaml
@@ -1,0 +1,21 @@
+kind: fixed
+area: plugins
+change: >
+  A delegated OpenCode session now appears in its pane, on the model and effort
+  it was delegated with. The plugin used to create the session over OpenCode's
+  HTTP API and ask the TUI to select it, but that request is published to a bus
+  the TUI has not subscribed to yet when a session is starting, so it was
+  dropped and the pane sat on the OpenCode splash while the agent worked
+  unseen. The delegated turn is now submitted through the TUI itself, so the
+  pane shows the session OpenCode opened, within a couple of seconds. The
+  effort pin reached nothing at all: the TUI takes the model variant from its
+  own state, so a delegation asking for `max` quietly ran on whatever variant
+  was last used interactively. Each delegated run now gets its own OpenCode TUI
+  state seeded with the pin — which also stops delegations from rewriting the
+  model you picked yourself. A pinned model that cannot honor the requested
+  effort (MiMo declares no variants at all) now fails the run and names what
+  the model does offer, and a session that opens on a different provider, model
+  or variant than the pin is reported as degraded instead of running unnoticed.
+symptom: >
+  A delegated OpenCode session showed the OpenCode splash screen instead of its
+  work, and its `--effort` pin was not the effort it ran on.
--- a/plugins/attn-opencode/README.md
+++ b/plugins/attn-opencode/README.md
@@ -39,10 +39,21 @@ An ordinary promptless OpenCode session launches the TUI with OpenCode's own
 model and variant defaults. The plugin observes the native session OpenCode
 creates and persists its identity for resume. Delegated OpenCode sessions
 require an explicit `--model provider/model` and one of the contract-tested
-effort pins: `--effort low` or `--effort max`. The plugin sends those pins to
-OpenCode's server as `{ providerID, id, variant }`, so its staged prompt is
-bound to the selected native session. Other effort labels remain intentionally
-unsupported until they have the same adapter and live-app evidence.
+effort pins: `--effort low` or `--effort max`. Other effort labels remain
+intentionally unsupported until they have the same adapter and live-app
+evidence.
+
+A delegated launch runs the pinned model on OpenCode's `build` agent
+(`--agent build` plus the model in the injected config), and seeds its own
+`XDG_STATE_HOME` with the pinned variant, because the TUI takes the variant
+from its state rather than from the agent config — which also keeps a run from
+rewriting the user's own model selection. The delegated turn is then submitted
+through the TUI, so the pane shows the session OpenCode itself opened; the
+`session.created` event that submission produces is the receipt that stops the
+staging retry. Before staging, the plugin checks the pinned model's variants in
+OpenCode's provider catalog and fails the run when the model cannot honor the
+requested effort, and after the session opens it reads the model back and
+degrades loudly when provider, model, or variant differ from the pin.
 
 Each run gets a separate loopback port and random Basic-auth password. The
 password, staged prompt, and attn-composed launch instructions live in private
@@ -75,5 +86,5 @@ attention request, the plugin classifies the newest completed assistant prose
 using the same OpenCode model in a temporary, tool-disabled session. Ordinary
 prose questions then become `waiting_input`, completed answers become `idle`,
 and uncertain extraction or classification becomes `unknown`. The temporary
-session is deleted without selecting it in the TUI, and duplicate idle events
+session is deleted without ever reaching the TUI, and duplicate idle events
 reuse a verdict cached against the exact native message text.
--- a/plugins/attn-opencode/src/driver.ts
+++ b/plugins/attn-opencode/src/driver.ts
@@ -416,6 +416,7 @@ export class OpenCodeDriver {
   // same startup budget the server itself gets.
   private async stagePinnedTurn(client: OpenCodeHTTP, binding: NativeBinding, prompt: string, signal: AbortSignal): Promise<void> {
     const deadline = Date.now() + this.startupDeadline;
+    const before = new Set(await this.sessionIDs(client, signal));
     for (;;) {
       if (signal.aborted || binding.pinError) return;
       if (binding.claimedID) return;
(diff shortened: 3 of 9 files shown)


figgyster (changes requested):
Case closed on the invisible agent and the phantom pin — both named, both fixed, receipts attached. The TUI-ownership pivot is the right shape: no more `/tui/select-session` into a bus nobody's listening to, and the `session.created` receipt claims the run before the bind's round trips. The hanging-session-reads test is a real witness for that — staged once, even while the bind dangles.

The both-ends pin check earns its keep. MiMo declaring `variants: {}` used to mean a silently dropped `--effort max`; now the run fails and names what the model offers, and the post-open read-back degrades loudly with the pin and what arrived, leaving the session unbound. Per-run `XDG_STATE_HOME` with the kv.json copy is a clean solve for the TUI-reading-its-own-state trap — your selection stays byte-identical, the pane still wears your colors. Changelog fragment, README, cleanup path — all track.

But the suite is red at head, and it's the PR's own witness lying down on the job.

**Blocking — `test/driver.test.ts:181-185` races the bind it asserts.** The first pinned test waits for `tuiSubmissions.length === 1`, then immediately asserts `/session/status` was called exactly once. The submission only proves the TUI took the prompt — the bind (`session.created` over SSE → `getSession` → reconcile) is delivered asynchronously after the HTTP submit returns. On this box the assertion beats the event 7 of 7 full runs, plus in isolation. CI's Plugins job passed this once because the race went the other way — that's a coin flip, not a verdict. The driver itself is fine: I patched the test to wait for the `session.report_metadata` receipt first (the same wait your sibling "keeps staging" test already uses) and the whole test passes — all ten expects, including the exactly-once `/session/status` and the no-`POST /session` checks. The fix is one wait, same shape as "binding after the TUI attaches": `eventually` the metadata report before counting requests.

Two notes, neither blocking:

1. The "never submitted twice" claim has a hole the size of an SSE drop. If the stream dies between the TUI taking the staged prompt and `session.created` landing, the receipt is lost; the loop resubmits into the already-created session — and a second submit to an existing session emits no `session.created`, so the loop burns to the 20s deadline and degrades "TUI did not open a session" with N duplicate prompts in a session that very much opened. The window is a fraction of the bug it replaces, but it's the same bug class. A pre-resubmit existence probe or a last-submitted-text dedupe would close it.

2. Three long-standing tests miss their 2s `eventually` deadlines intermittently on this head (main runs 3/3 clean in the same box): "applies events buffered during reconnect", "reports unknown when native-attention reconciliation fails", and once "maps native question and permission events". The new claim→bind chain lengthens every pinned test's critical path, and those deadlines were comfortable before. Widening them or awaiting the receipt explicitly would buy margin.

Question for the record: a pinned run with model+effort but no `initial_prompt` now never stages or binds — it sits unbound until a human submits in the pane. Pre-fix it bound an empty session. `attn delegate` always composes the ticket prompt, so this looks unreachable from there; confirming that's the intent for any other spawn path that sends pins without a prompt.

Fix the race and this is a good merge — the loud-degradation story at both ends is exactly the kind of failure I like seeing in the dark.

victorarias:
Thanks — good catches, and the SSE-drop one was worth the extra round.

On the blocking finding: that race was already fixed in `a7a04f4a` ("wait for the bind, not the submission that triggers it"), which is the head you reviewed — the first pinned test waits on the `session.report_metadata` receipt, exactly the shape you describe, and the Plugins job is green on it. I think the review picked up the pre-fix content; if you see the old wait at current head, say so and I'll dig.

I also swept the same race out of the sibling tests: every `eventually` that waited on `tuiSubmissions` before driving events now waits on the metadata report.

On the SSE-drop hole — you're right, and it's fixed rather than noted. Before each resubmit the driver now looks for a session in the run's directory that carries this run's staged prompt, and binds it. Same receipt, read directly instead of waited for. Sessions are stored per project directory, so the candidate has to carry the prompt to count; a test drops `session.created` entirely and asserts one submission and a bound run.

On the deadlines: `eventually` now allows 5s. The claim→bind chain did lengthen the critical path, and 2s left no margin on a loaded runner — a passing test still finishes as fast as it ever did.

On your question: a pinned run with no initial prompt is deliberate now. The TUI still opens on the pinned model (config agent plus the seeded state), nothing is staged, and the run binds the first session the pane opens — `handleEvent`'s bind is not gated on staging. So it is not stranded; it waits for whoever is driving that pane. Pre-fix it bound an empty HTTP-created session that the pane never showed, which is the bug this PR is about. There is a test locking that path down now.

figgyster (approved):
Both finds answered, and the second one came back better than my note asked for.

a7a04f4a is the exact wait I asked for — bind receipt instead of submission — and you propagated it to every pinned test, which also retired the load-flake note I'd tacked on: the fake's `eventually` deadline went 2s→5s and the suite that was red 7/7 on the old head is green 4/4 full runs here, plus the two new witnesses, plus isolation. The old red was the race going one way; this is the race not existing.

60ca46f0 closes the SSE-loss hole properly: the receipt is no longer only an event that can be dropped — the session itself is the receipt, read directly. `before` snapshot at staging start, then a poll for sessions that match the staged prompt text, claim, bind, done. The fast path pays nothing: the poll only runs when no claim landed during the sleep. The `dropSessionCreatedEvent` test is a genuine witness for it, and the promptless-pinned test pins the answer to my question — the pane opens on the pinned model, the first session the user submits there is the run's. That's the right reading.

One small thing left, non-blocking: `userPromptText` trusts `messages[0]` to be the staged prompt's user message, but `lastAssistantTurn` sorts the same list defensively because OpenCode's message order isn't contractually chronological. If the assistant's first reply lands before the 1s poll and the list comes back newest-first, `messages[0]` is the reply, the match misses, and the loop resubmits — the exact duplicate the poll exists to prevent. Scanning the returned list for the first `role === "user"` entry instead of taking index 0 would make the receipt read order-proof. The window is thin and the poll is only reached when an event was already lost, so I'm not holding the merge on it.

CI's green across the board at the new head, diff-check clean, changelog gate passing. Case closed — the pane shows the agent, the pin is the pin, and the receipt can't be dropped twice.

figgyster (approved):
5db636d4 closes the last open note, and the witness is better than the note asked for.

On the timeline in your comment: no digging needed. My first pass was against 4ed46c1 — that's the head the case opened on and the head I fetched before verifying; the a7a04f4a push landed while I was still in the box, so the CHANGES_REQUESTED went out against the pre-fix content. The second pass at 60ca46f0 verified the fixed wait and everything around it, and this head re-verified again: old wait gone, no `tuiSubmissions`-before-bind assertions left anywhere, suite green 3/3 full runs here both times. The race is retired, not just patched.

The ordering fix is exactly the right shape: `userPromptText` scans for the first user-role message instead of trusting index 0, with the reason in the comment — and the fake now answers newest-first, so the lost-receipt test exercises the precise scenario I was worried about (reply leading, prompt second) and still binds, single submission. A witness that reproduces the adversarial ordering beats a unit test that assumes the friendly one. The poll path had to prove it could read the receipt from a list that isn't contractually chronological — now it does.

Verified at 5db636d4: diff-check clean, full plugin suite 3/3, lost-receipt and promptless-pane tests passing in isolation, CI green across the board (Plugins, Changelog, Changes, E2E). That's the whole arc closed — the pane shows the agent, the pin is the pin, and the receipt survives both an SSE drop and a hostile message order. Good merge.

PR #927, opened 2026-08-16
chore(plugins): drop the OpenCode driver

victorarias:
attn no longer supports OpenCode. This removes the bundled driver plugin, so OpenCode stops appearing in the plugin list and stops being an agent you can launch or delegate to.

## Why

Driving OpenCode needed a server on a loopback port, a private run registry, a side-channel stop classifier, and a launcher that reached around the TUI to pin the model a delegation asked for. That last part is the one that broke: OpenCode's TUI takes `-m` for the model but has no flag for the reasoning variant, and its remembered variant (in `$XDG_STATE_HOME/opencode/model.json`) beats a variant set in config. Pinning effort therefore meant handing the run a state directory of its own and depending on file formats and precedence rules OpenCode neither documents nor promises. The maintainer's call is that this is more surface than the integration is worth.

#926 landed the fix for the invisible-pane and unverified-pin defects shortly before that call. Removing the plugin takes that work back out with it; no separate revert is needed.

## What changed

- `plugins/attn-opencode/` deleted.
- `scripts/build-bundled-plugins.sh`: stage step and the `guidance-plugin.js` special case removed.
- `scripts/source-fingerprint.sh`: the plugin's two exclusion entries removed.
- `README.md`: the OpenCode row leaves the supported-agents table, and the handoff example names Codex instead.
- `plugins/attn-pi/AGENTS.md` pointed at `plugins/attn-opencode` as the reference implementation for the driver pattern — a live instruction to read a directory this PR deletes. It now names the daemon-side contract in `internal/daemon/plugin_driver.go`. attn-pi is the only driver plugin now.
- Plugin-system tests used `attn-opencode` as their example plugin name. They now use a neutral one, so the deleted plugin does not survive as the system's canonical example.
- `docs/plans/*opencode*` and `docs/grounding/pi-plugins.md` are left alone: they are dated records of what happened. Two of their lines do point into the deleted tree (`2026-07-15-opencode-prose-classifier.md:41` cites the removed classifier fixture, `2026-07-19-pi-driver-plugin.md:59` says the layout mirrors `plugins/attn-opencode`). A dated record describing the tree as it stood is still accurate about that date, so they stay as written.

## A defect this exposed

`build-bundled-plugins.sh` only ever `rm -rf`s the directory it is about to write. Nothing reaps a stage directory for a plugin that stopped being bundled, so the first rebuild after deleting the source still staged and shipped `attn-opencode` — and would have kept doing so from every checkout and CI cache that had built it before. The script now prunes any staged plugin it did not write this run.

## Existing installs need no migration

A bundled plugin's "installed" state is a name in the `installed_bundled_plugins` setting, and discovery intersects that list with what the bundle actually holds. A name with nothing behind it is ignored.

Verified rather than assumed: on a profile installed from this branch, the settings row was seeded with `["attn-opencode","attn-pi"]` and the daemon restarted. It came up with `plugin discovery … catalog=1 issues=0`, attn-pi installed and running, and no attempt to launch the missing plugin.

Existing OpenCode conversations stay in OpenCode; attn will not resume them.

## Verification

Installed profile `ocdrop` from this branch, since the change alters what ships in the app bundle.

- App bundle carries only `attn-pi`.
- Daemon catalog is `attn-pi` alone, `issues=0`.
- Stale-install-record receipt above.
- `go test` green for `internal/daemon`, `internal/store`, `internal/protocol`, `cmd/attn`.
- Frontend suite green: 2881 passed, 13 skipped.
- Profile cleaned afterwards.

Not checked visually: the Settings plugin list. The shell running the verification has no Accessibility grant for window capture. That list is a direct render of the daemon catalog verified above, and `SettingsModal.test.tsx` covers its rendering.

https://claude.ai/code/session_01AzXBdU9mMw7yaKujfeHBgk

## Follow-up on review

`build-bundled-plugins.sh` carried an invisible ordering invariant: the prune loop read a list that `stage_plugin` appended to as it ran, so a `stage_plugin` call added below the loop would have reaped its own plugin on that plugin's first build. The bundled set is now declared once above both, the sweep reads that declaration rather than a side effect, and it runs before staging so a dropped name is gone whether or not the build that follows succeeds.

Exercised against a stage root seeded with two plugins that are not in the list plus the one that is: both dead directories were removed, the survivor staged.


figgyster (approved):
This one closes a case I watched from the other side last week. 4,594 lines of deletion, and almost all of it earned.

The arithmetic holds up. The plugin was a loopback server, a private run registry, a side-channel stop classifier, and a launcher that reached around OpenCode's TUI to pin a variant the TUI won't take from config. That is a lot of surface held together by another tool's undocumented precedence rules. I spent three rounds on #926 watching you force fidelity into that surface, invisible panes, unverified pins, event-loss holes, and the fixes worked. But they were also the investigation that made the cost legible: if pinning effort means seeding a private state directory and depending on file formats OpenCode doesn't promise, the integration is a loan with a rising interest rate. Walking away a day later is not whiplash, it's the evidence paying off. And you did the honest thing with #926's changelog fragment, deleting a fix description for a feature the same release removes. I checked: the compiled CHANGELOG.md never carried that text, so no released history got rewritten.

The real find in this PR is the staging defect. A plugin that stopped being bundled kept shipping from every checkout and CI cache that had ever built it, because the script only ever cleared the directory it was about to write. The prune loop is simple and it's right; I ran the exact logic against a fake stage root, unbundled dirs died, the staged survivor stayed. And the seeded-profile witness for the no-migration claim is the right kind of evidence: settings row carrying the dead name, daemon comes up catalog=1 issues=0, no attempt to launch what isn't there. A name with nothing behind it being ignored is a clean contract.

Verified at 546b9d7c: repo-wide grep for opencode, guidance-plugin, and ATTN_OPENCODE is clean outside the compiled changelog and the dated plan docs. No Go, frontend, workflow, or build-script references survive, and the daemon's driver machinery is registration-driven so nothing hardcodes the dead agent. Scripts pass bash -n, the fragment parses strict-yaml with the right kind/area, and the source-fingerprint exclusions came out with the directory they excluded. Couldn't run the Go suites here, no toolchain in my box, so the daemon side rides on your run plus CI, which is still green-light in progress across the board.

Two non-blocking paper cuts, both for the record:

1. The PR body says the dated plan docs' citations point at daemon files that still exist. Two lines actually point into the deleted directory: docs/plans/2026-07-15-opencode-prose-classifier.md:41 cites the deleted stop-classifier-evaluation.json fixture, and docs/plans/2026-07-19-pi-driver-plugin.md:59 says the layout mirrors plugins/attn-opencode. Dated records, right to leave them be. The claim is just a half-step overbroad.

2. The prune loop carries an invisible ordering invariant: every future stage_plugin call has to land above it, or the next plugin you add gets reaped on its own first build. Works today with one call above the loop. The next editor won't know, and I'm betting someone finds out the hard way within the year.

Approve.

figgyster (approved):
2ceceeb0 answers the ordering note, and better than I asked. The bundled set is declared once at the top of the script, and the sweep now runs before staging against that declaration, so a plugin that leaves the list is reaped from the stage root even when the build that follows fails. The invisible invariant I bet on (every stage call having to land above the prune loop) is gone; adding or removing a plugin is genuinely one line now, and whatever the deletion arc grows next, the script won't be where it trips.

Verified at 2ceceeb0: bash -n clean, no stale names left in the script, and I ran the new sweep verbatim against a fake stage root — unbundled dirs die, the declared set survives, and the `name|description` split resolves both ends. Only the one script changed; the rest of the PR is exactly what I approved at 546b9d7c. CI is queued or rolling on the new head.

Re-approved at 2ceceeb0.

PR #928, opened 2026-08-16
feat(garden): make seed documents live and answerable

victorarias:
## TL;DR

Seed documents now react to live body and tender changes, and their annotation composer routes feedback to exactly one typed destination: the current tender session or the seed ledger when nobody is tending. The existing file-document annotation path is unchanged.

## What changed

`attn seed edit <id> -m <markdown>` changes only the seed body, preserving concurrent lifecycle, claim, edge, and variable changes through revision-aware retries. The daemon publishes a named `garden.body_edited` fact after the write.

The existing `garden_seeds_updated` payload already carries each seed's full body and revision. When the open seed is present, the tile now applies that payload immediately to the mounted reader, which drives the existing anchor resolver without a reload or a second annotation stack. It still performs the detail read for ledger changes and as a fallback when the bounded garden snapshot omits an older open seed; a stale detail response cannot overwrite a newer pushed body.

The composer derives its shape from daemon-owned tender truth:

| Seed state | Primary action | Secondary action |
|---|---|---|
| Live tender session | `Send N` to that session | `Note on seed` in the caret menu |
| No live tender session | `Note on seed N` | none |

`SeedDocument.tender_holds` distinguishes a live holder from stale stored tender IDs. Claim/park garden pushes update the shape immediately, and session disappearance republishes garden truth so the detail model can drop a dead holder.

Annotation submission now carries an explicit session or seed destination. The daemon requires exactly one, refuses file-to-seed or mismatched seed routing, appends seed feedback through the normal log-note path, and only clears the persisted draft after the destination accepts it.

## Review path

1. `internal/protocol/schema/main.tsp` and `internal/daemon/ws_markdown_annotations_submit.go` define and enforce the typed routing contract.
2. `internal/daemon/garden.go` and `internal/daemon/bus.go` own body editing, holder truth, and live garden projection.
3. `WorkspaceDockTile.tsx` applies pushed seed revisions and renders the two composer shapes while leaving the file branch intact.
4. The daemon and workspace-tile tests pin destination validation, live claim/park changes, stale-detail races, highlight re-anchoring, node identity, and scroll retention.

## Manual verification

In a packaged throwaway profile, I planted and opened a seed, persisted a highlight, then inserted text above it with the CLI. The mounted reader updated immediately; the highlight stayed on the same phrase without orphaning or a scroll jump.

Live seed re-anchor

[Full-quality re-anchor recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/0ac84b9c47039d55e9f8cca066070b2ab854f601/delegate-garden-bcd-63dc3858/garden-bcd-reanchor.mp4)

I sent that annotation and saw the formatted feedback arrive in the tending terminal. After parking the seed, the composer flipped to the unsplit note action; submitting another annotation cleared the draft and persisted the formatted note in the visible ledger.

Tender routing and seed note

[Full-quality routing recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/0ac84b9c47039d55e9f8cca066070b2ab854f601/delegate-garden-bcd-63dc3858/garden-bcd-routing.mp4)

## Out of scope

File-document behavior and payloads are unchanged. Delegation reporting, attach/detach verbs, ticket retirement, signposts, conversion, and the dormant artifact projection remain later garden steps.

### DIFF
--- a/internal/daemon/bus.go
+++ b/internal/daemon/bus.go
@@ -203,6 +203,10 @@ const (
 	// after the write it describes has committed, so a consumer that reads it
 	// always finds the seed.
 	FactGardenPlanted = "garden.planted"
+	// FactGardenBodyEdited: a seed's living markdown document changed. The
+	// garden snapshot carries the new body and revision so open readers can
+	// re-anchor immediately without waiting on a detail refetch.
+	FactGardenBodyEdited = "garden.body_edited"
 	// One fact per lifecycle move, rather than one `garden.changed`: the subject
 	// says which seed and the name says what happened to it, which is what a
 	// change feed and a future nudge both read. All of them project the same way
@@ -354,7 +358,12 @@ func buildWireProjections() []projection {
 		},
 		{
 			filter: bus.Filter{FactSessionUnregistered},
-			apply:  func(d *Daemon, ev bus.Event) { d.projectSessionUnregistered(ev) },
+			apply: func(d *Daemon, ev bus.Event) {
+				d.projectSessionUnregistered(ev)
+				// Session liveness is part of garden truth: it decides whether a
+				// stored tender still holds and therefore where feedback routes.
+				d.projectGardenSeeds()
+			},
 		},
 		{
 			filter: bus.Filter{FactSessionRespawned},
--- a/internal/daemon/bus.go
+++ b/internal/daemon/bus.go
@@ -314,6 +314,8 @@ func buildWireProjections() []projection {
 			filter: bus.Filter{FactSessionRegistered},
 			apply: func(d *Daemon, ev bus.Event) {
 				d.projectSessionEvent(protocol.EventSessionRegistered, ev.Subject)
+				// A new session can reactivate a stored tender with the same id.
+				d.projectGardenSeeds()
 			},
 		},
 		{
@@ -372,6 +374,8 @@ func buildWireProjections() []projection {
 					Event: protocol.EventRuntimeRespawned,
 					ID:    protocol.Ptr(ev.Subject),
 				})
+				// Keep the computed tender hold in lockstep with session rebirth.
+				d.projectGardenSeeds()
 			},
 		},
 		{
--- a/app/src/annotations/useAnnotationSend.ts
+++ b/app/src/annotations/useAnnotationSend.ts
@@ -35,7 +35,7 @@ export function useAnnotationSend<T extends AnnotationSendResult>({
     sendRef.current = send;
   }, [send]);
 
-  const sendNow = useCallback(() => {
+  const runSend = useCallback((action: () => T | null | Promise<T | null>) => {
     if (sendingRef.current) {
       return;
     }
@@ -43,7 +43,7 @@ export function useAnnotationSend<T extends AnnotationSendResult>({
 
     let result: T | null | Promise<T | null>;
     try {
-      result = sendRef.current();
+      result = action();
     } catch (error) {
       sendingRef.current = false;
       setOutcome({
@@ -81,6 +81,12 @@ export function useAnnotationSend<T extends AnnotationSendResult>({
       });
   }, []);
 
+  const sendNow = useCallback(() => runSend(sendRef.current), [runSend]);
+  const sendAlternative = useCallback(
+    (alternative: () => T | null | Promise<T | null>) => runSend(alternative),
+    [runSend],
+  );
+
   useShortcut(shortcutId, sendNow, enabled);
 
   useEffect(() => {
@@ -92,5 +98,5 @@ export function useAnnotationSend<T extends AnnotationSendResult>({
   }, [outcome, sentClearMs]);
 
   const clearOutcome = useCallback(() => setOutcome(null), []);
-  return { outcome, send: sendNow, clearOutcome };
+  return { outcome, send: sendNow, sendAlternative, clearOutcome };
 }
--- a/app/src/components/GardenPanel.reader.test.tsx
+++ b/app/src/components/GardenPanel.reader.test.tsx
@@ -30,6 +30,7 @@ function seed(overrides: Partial<Seed> = {}): Seed {
 function seedDocument(root: Seed, noteBody: string): SeedDocument {
   return {
     seed: root,
+    tender_holds: Boolean(root.tender_session || root.tender_member),
     children: [],
     notes: noteBody ? [{
       id: `n-${noteBody}`,
--- a/app/src/components/GardenPanel.tsx
+++ b/app/src/components/GardenPanel.tsx
@@ -288,6 +288,7 @@ export function GardenPanel({
             const relations = expanded ? relationsOf(index, seed.id) : [];
             const localDocument: SeedDocument = {
               seed,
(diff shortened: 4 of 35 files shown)


internal/daemon/bus.go
```diff
@@ -354,7 +358,12 @@ func buildWireProjections() []projection {
 		},
 		{
 			filter: bus.Filter{FactSessionUnregistered},
-			apply:  func(d *Daemon, ev bus.Event) { d.projectSessionUnregistered(ev) },
+			apply: func(d *Daemon, ev bus.Event) {
+				d.projectSessionUnregistered(ev)
+				// Session liveness is part of garden truth: it decides whether a
+				// stored tender still holds and therefore where feedback routes.
+				d.projectGardenSeeds()
```
figgyster, line 367: Death republishes garden truth; birth does not. This projection flips the composer when the tender session ends — right call, and the tile refetches on every snapshot so the flip lands. But the mirror is missing: `FactSessionRegistered` and `FactSessionRespawned` both skip `projectGardenSeeds`, and respawn keeps the session id, so the stored tender becomes live again without any garden fact. The tile gets no snapshot, no detail refetch fires, and the merge keeps the stale `tender_holds` from the last detail read — the composer sits on "Note on seed" while somebody actually tends. In the unsplit state there is no caret menu either, so the tending session is unreachable from the tile until the next garden fact anywhere or a reconnect. One line on the registered/respawned projections, same as here; the frontend merge already does the right thing once the refetch fires.

figgyster (approved):
The seeds are alive now, and the composer picks its one destination from daemon-owned truth instead of the tile's guess. The body push is the honest part: the bounded snapshot carries the live seed, the tile merges it without a remount, and the stale-detail guard has a witness that earns its keep — a detail read resolving *after* the push with an older body gets ignored, the highlight re-anchors on the same phrase, the scroll position survives. The typed routing is enforced at the wire, not just in the UI: exactly one of `target_session_id`/`target_seed_id`, a seed destination must match the source seed, file-to-seed is refused by name, and the draft survives every refusal — the destination matrix test plus the file-to-seed refusal are the eyewitnesses. The noted-but-clear-failed test (closing the store mid-broadcast to force the clear to fail) pins the "never risk duplicate delivery" rule properly, and the edit test's `reflect.DeepEqual` on identity/lifecycle/edges is the right way to prove a surgical edit.

Ran the frontend on the box: 422 tests across the touched files, tsc clean, diff-check clean, CI green where the wheel has turned — the generated types are genuinely in lockstep and the Changelog job passes. The Daemon and Frontend jobs are still pending, and this box has no Go toolchain, so the Go verdict rides on CI.

One comment on the wire (the liveness asymmetry, inline on bus.go), and one note, neither a blocker:

**ProtocolVersion stayed 254 while the wire changed shape.** `target_session_id` went required→optional with a new `target_seed_id`, SeedDocument gained a required `tender_holds`, `seed_edit` is a new command. Your own #915 bumped 253→254 for the same class of change. Old-end tolerates the new end, so nothing snaps — but a pre-928 app against this daemon fails seed-note submits with "target_session_id is required" instead of a version prompt. If the norm is "bump on wire changes", 255 is owed; if it's "bump only on initial-state breakage", this note dies quietly.

Also on record: the annotation note lands unattributed — no session, no member — the reader speaking for themselves. The payload names the seed, so the ledger stays readable. Fine by me.

Approved at e642ee94.

figgyster (approved):
Loop on #928: four commits since my approval, and they answer the react-doctor warnings the bot flagged — my old bet about the tile's state routing held up in the shape I didn't fully predict. `useLiveSeedDocument` now owns the seed document at the tile root: the header composer and the body read one source instead of reporting up through a callback that could go stale on retarget, and the menu open-state is derived from a `${tenderSession}:${rev}` key instead of an effect adjusting state after a prop change. That derivation is the better fix than the one the warning asked for: the old effect only closed the menu when the tender disappeared; the key closes it on *any* claim change or body push — a case the effect never covered. `useOptionalDaemonApi` scopes the seed reader so lightweight surfaces that don't carry a daemon provider get an honest "The daemon API is unavailable" instead of a hard context throw.

Verified at 992c219f: 422 frontend tests across the touched files, tsc clean, diff-check clean, CI green where finished (Generated types, Changelog, Changes, react-doctor with 0 errors — the one warning left is the pre-existing `normalizeBrowserAddress` export, not this PR's). Daemon and Frontend jobs are still pending; the Go verdict rides on CI as before.

Both notes from my first pass still stand at this head, no code moved on either:

1. **Rebirth asymmetry is still one-way.** Death republishes garden truth (bus.go unregister projection), birth doesn't — `FactSessionRegistered`/`FactSessionRespawned` still skip `projectGardenSeeds`, and nothing in the tile refetches on session events. The composer can sit on "Note on seed" while a revived tender session holds, until the next garden fact. Same line I left on bus.go:365.
2. **ProtocolVersion is still 254 with a reshaped wire.** `target_session_id` required→optional, `target_seed_id`, `seed_edit`, required `tender_holds`. If "bump on wire changes" is the norm, 255's owed; if "bump only on initial-state breakage", this note dies quietly.

Approved at 992c219f. The tile's state flow is cleaner than when I first read it — worth the four commits.

internal/daemon/bus.go
victorarias: Fixed in fdf624d9. Session registration and runtime respawn now both project garden seeds, mirroring unregister, so an open seed refetches its computed tender_holds when a stored tender comes back. The wire projection contract and flow golden cover the added snapshots; targeted daemon tests and the full make test-quick suite pass.

figgyster (approved):
Loop on #928, round three: fdf624d9 is the daemon-side half of my rebirth note, and it's the line I drew — both `FactSessionRegistered` and `FactSessionRespawned` now republish garden truth after their session event, with comments that name the exact consequence ("A new session can reactivate a stored tender with the same id"; "Keep the computed tender hold in lockstep with session rebirth"). The wire-join fixtures pin both event pairs (session event + garden push, deterministically ordered inside one apply), and the flow golden renumbers through the new pushes.

The chain closes end to end: the garden push arrives → the tile's snapshot-triggered detail refetch fires → the daemon answers `tender_holds` from live session truth → the merge keeps the fresh detail's answer when the ids haven't moved. The frontend half was already verified last round; this commit is the missing wire. Broadcost is proportional to what `session.unregistered` already paid and the tickets board pays per fact — a bounded snapshot read per session event.

Verified: diff-check clean, the two new call sites are present in the blob, and the change is daemon-only so the frontend evidence from 992c219f still stands. The Go suite is still pending on CI — the daemon verdict rides on it, as it has all night. One paper cut, non-blocking: the flip is witnessed piecewise — the wire fixtures pin the events and the seed-reader test pins the liveness math, but no single test walks a stored tender through dead→re-registered→holds. The composition is two one-line projections and a store read; I'm not booking anyone for it.

ProtocolVersion note stays open (254 with a reshaped wire) — still a convention question, still non-blocking, still dies quietly if the norm is "bump only on initial-state breakage."

Approved at fdf624d9. That's the asymmetry closed on both ends — death republishes, birth republishes, and the composer now flips both ways.

PR #929, opened 2026-08-16
feat(attn-pi): auto mode policy core

victorarias:
Slice 1 of pi auto mode ([plan](https://github.com/victorarias/attn/blob/main/docs/plans/2026-08-16-pi-auto-mode.md)):
the decision core, shipping dark.

pi has no permission system — every tool call runs. Auto mode is the pi
extension that adds one without adding dialogs: a static safety envelope
carries normal work at zero latency, a classifier judges what reaches
past it, and a denial is answered in conversation rather than by
clicking. This PR is the half that answers without asking a model.

## What is here

`plugins/attn-pi/automode/`, ~600 lines, no pi import (duck-typed against
pi 0.83.0's shapes like `suite/core.ts`, so `bun test` covers the whole
extension including its `tool_call` wiring):

- **`policy.ts`** — the decision tree, in the plan's order: hard denies →
  allow list → read-only tools (`read`, `grep`, `find`, `ls`) → `write`/
  `edit` resolving inside cwd and outside protected paths → static
  read-only bash → everything else to the classifier. A tool auto mode has
  no envelope entry for is refused with a reason naming that limit.
- **`bash.ts`** — the read-only command set. Conservative by construction:
  a command that can run another command (`env`, `xargs`, `sh`, `find`) is
  not in it at all, one that can be talked into writing carries the flags
  or subcommands that keep it read-only (`sort -o`, `git branch -D`), and
  chaining/redirect/substitution characters disqualify a line outright.
  Network is checked across the whole command line and never rides the
  envelope, however read-only a `curl` looks.
- **`paths.ts`** — where a path sits: in-envelope, outside cwd, or
  protected (`.git`, `.pi`, `.attn`, `.claude`, `.ssh`, `.env*`, shell rc
  files…). Protected wins wherever it appears, so a protected path inside
  cwd still classifies.
- **`config.ts`** — the plan's schema (`environment`, `allow`, `hard_deny`,
  models, `enabled_default`) with a loader seam; storage is slice 4, so
  slice 1 takes config as a value. A broad allow pattern (one with no
  literal characters: `*`, `**`, `?`) is refused at load with
  `AutoModeConfigError` naming the field — a blanket allow is what the
  classifier exists to replace. A broad *hard deny* is fine; that
  direction is safe.
- **`denial.ts`** — the model-facing contract, one function every later
  slice reuses: names auto mode, the blocked action, the reason, and that
  the user's explicit approval in the conversation permits a retry.
- **`session.ts`** — per-session state. Allow verdicts cache for the
  session; deny verdicts drop the moment the user speaks, because the next
  message may be the grant. Circuit breaker at 3 refusals in a row or 20
  since the user last said anything; once tripped it stops calling the
  classifier and says so, naming both limits and both counts.
- **`classifier.ts`** — the interface slice 2 implements, plus the stub the
  tests decide with. An `uncertain` verdict fails closed here; escalation
  to the 2b model is slice 2.
- **`index.ts`** — the extension factory. `tool_call` → `{ block, reason }`,
  `input` → drop the deny cache. A thrown handler blocks the tool with the
  denial contract rather than pi's error text.

## Verification

`bun test`: 126 new tests, 272 across the plugin, green. Table tests walk
every ordering boundary (hard deny over the allow list, over read-only
tools, over an in-cwd write; the allow list over the bash check and over
the envelope), the bash traps (`cd &&`, redirects, `$()`, backticks, env
prefixes, `git -C`, `sort -o`, a `curl` at the end of a pipeline), and the
cache/breaker rules. A generated-path property runs 2000 paths through
`write`/`edit` and asserts the invariant behind the envelope: a path that
does not resolve inside the working directory never returns "run" from the
static tree (with a guard so the generator cannot make it vacuous).

**Live verification exempt**, per AGENTS.md: this is a pure, isolated TS
change, fully covered by unit tests, with no daemon lifecycle, protocol,
PTY, background-runner, or UI surface — nothing imports it and
`build-bundled-plugins.sh` does not stage it (slice 3).

## Named gaps

- Path resolution is lexical (`node:path`), so a symlink inside the working
  directory pointing outside it resolves as in-envelope. Stated in
  `paths.ts`; worth closing when the live pass (slice 6) can measure what
  `realpath` on every write costs.
- The circuit breaker blocks in every mode here. The human question it is
  meant to raise (`ctx.ui.confirm`) is the UX slice; blocking everywhere is
  the fail-closed posture until then.

## Base

Carries `docs/pi: auto mode plan…`, the doc commit this branch was cut
from, which is not on main yet. If that lands separately first, this
rebases over it as a no-op.


figgyster (changes requested):
Read the whole file — the diff, the plan docs, the superdraft, the receipt, the tests. Ran the suite in the box: 272 pass, 0 fail, ~230ms, matches the PR's claim to the test. The claims check out too: `build-bundled-plugins.sh` stages `suite/index.ts` only, nothing imports `automode/`, and the AGENTS.md note is accurate. The live-verification exemption is legitimate — pure isolated TS, no daemon, no protocol, no UI.

The decision tree is in the plan's exact order and the bash set is conservative by construction: wrappers that can run other commands (`env`, `xargs`, `find`, `sh`, `sudo`) are absent, network is scanned across the whole line before the read-only check, `git -C` and bare `git` classify rather than run, protected paths win wherever they appear, broad allow patterns are refused at load, and the property test has a real guard against being vacuous. Denial contract, cache, and breaker all match the design. This is careful work, and the s7 receipt is a genuine receipt — corpus, cost, latency, per-model verdicts, reproducible on demand.

One finding, in the file the whole feature lives in.

**Blocking — `plugins/attn-pi/automode/policy.ts`, line 119: a literal NUL byte is committed in `normalizedIntent`.**

`return `${call.toolName}` + `<U+0000>` + `${callSignature(...)}`` — the byte sits exactly where the separator space belongs (compare `describeCall` in the same file, which uses a space). Verified against the committed blob via the API, not just the checkout: NUL count 1, offset 4400, right in that template literal.

Why it matters:

- Git treats the file as binary. This PR's own diff shows `Binary files differ` for the decision tree — the one file whose every line the review should be able to see. Every future PR touching `policy.ts` gets the same opaque diff, and `git grep`/`rg` skip it by default. The safety envelope of the feature is effectively undiffable until this is fixed.
- Every verdict-cache key rides the byte: `normalizedIntent` returns `bash\x00git push origin main`. It happens to work as a Map key today; it's invisible corruption waiting for the day an intent string reaches a log, a serialization, or a pattern match.
- The suite can't catch it — JS allows a NUL inside a string literal, so all 272 tests pass with the byte in place. The tests are a witness to behavior, not to the byte. That's exactly why this is the kind of thing that ships.

Fix is one byte: delete the NUL, put the space back. Suite should stay 272/272; if it does, this is an approve — the logic reads right, the traps are real, and the loop that reviews the envelope deserves to see the envelope.

victorarias:
Good catch, and a nasty one — a literal NUL byte where the separator space belongs, which is exactly the kind of thing a green suite happily carries. Fixed in cd68577a: one byte, space restored, suite still 272/272. `git grep` sees policy.ts again and the diff is text.

Re-requesting review.

figgyster (approved):
The fix landed in one commit, and I checked the byte itself, not just the suite. Between `0996278e` and `cd68577a`: one file, one byte — position 4400 of `policy.ts`, `0x00` → `0x20`. The NUL is gone, the space is back, `normalizedIntent` reads `${call.toolName} ${callSignature(...)}` like every other template in the file, and git now classifies the file as text — "JavaScript source, UTF-8" — so the decision tree diffs like a citizen again. Suite in my box at this head: 272 pass, 0 fail. CI Plugins green.

That's the whole envelope loop closed: the one byte I asked for is the one byte that changed. The tree, the traps, the property, the receipt, the dark-ship staging — all as reviewed. Approving at `cd68577a`.

PR #930, opened 2026-08-16
feat(daemon): gate the WebSocket port with a per-profile client token

victorarias:
The daemon listens on two things: a unix socket, gated by file permissions, and
a loopback WebSocket port, gated by nothing. Any local process that knew the
port spoke the full protocol — read every session, spawn agents, write files
through `fs_write`. #922 fenced profiles at the filesystem and spawn level, but
the wire carried no credential, so one profile's client could drive another
profile's daemon.

This closes it with a per-profile runtime token the daemon mints and every local
client presents in `client_hello`.

## The credential

`<data-dir>/client-token` — 32 bytes from `crypto/rand`, hex, written 0600.
Minted at `Start()` and reused on later startups, so a daemon restart does not
strand its clients. `ATTN_CLIENT_TOKEN` overrides the file for harnesses that
spawn a daemon and a client together.

Per-profile is the point: the same binary serves every profile, so the secret
lives in profile data. Nothing is baked at build time — a released binary would
ship a public secret.

Three token-shaped things now exist and do different jobs:

| | what it proves | where |
| --- | --- | --- |
| `client-token` | this process may speak the protocol at all | hello, every local client |
| `browser-host-token` | this client is the trusted Tauri main webview | hello, app only |
| `ATTN_WS_AUTH_TOKEN` | operator bearer for an exposed WS port | HTTP header, opt-in |

## Refusing is only half of it

The brief scoped this to the hello check. Building it turned up the bigger hole:
the daemon pushed `initial_state` — every session, PR, workspace, ticket and
setting — and registered the connection with the broadcast hub the moment the
socket opened, before a single byte arrived from the client. A credential that
gates writes while the whole read side streams to anyone who knows the port
would be theatre.

So hub registration and the `initial_state` push moved out of the accept path
into `admitClient`, called only once the hello passes. The hub is the sole
fan-out, so an unauthorized connection is now invisible to every broadcast, not
merely refused its commands. The app already sent hello first thing on open, so
nothing waits longer than it did.

Two things fell out of that:

- A daemon holding no token authorizes nobody. It used to match the client that
  also sent nothing — fail-open by accident. `Start()` mints before the port
  listens, so the refusal only names a daemon someone built without it.
- Hub registration became a direct locked insert (`wsHub.add`) instead of a
  handoff on the `register` channel, which had no other user. Admission is
  synchronous now, so there is no window where a client is admitted but not yet
  receiving. The unregister side stays a channel — a disconnect is discovered on
  the read pump, not asked for — and now releases the write pump even for a
  connection that was never admitted.

`/health`'s `ws_clients` consequently counts authorized clients. It has no
programmatic consumer.

## The check

`handleClientHello` compares in constant time before it records identity or
capabilities. A mismatch sends `command_error` with code `unauthorized_client`,
names the file the daemon read, then closes with a policy violation — never a
silent hangup. The refusal goes out through the send channel rather than
`conn.Close`, so the write pump drains it instead of racing the close.

One more thing had to move for that to hold. A client does not wait for a reply
to its hello, so whatever it pipelined behind it is already in the read buffer
when the refusal is queued. Answering it reaches the `workspace_sessions` gate,
which hangs up on the connection itself — closing the socket before the write
pump flushed the refusal. The live probe caught exactly that: a wrong-token
connection got `events: []` and a close reason naming the wrong problem. The
message pump now stops dispatching once a client's send channel is closed, so
the refusal is the last thing that connection hears.

Unix-socket commands never reach this handler: `daemon.go`'s socket dispatch is
a separate switch and file permissions already gate that path. The WS check is
not weakened for it.

One exemption, and it is the reason remote web still works: a connection that
cleared `ATTN_WS_AUTH_TOKEN` at the HTTP layer skips the client-token check.
That bearer is set only when the port was deliberately exposed beyond loopback
(`tailscale serve`), and the browser it serves cannot read a file on the
daemon's disk. Both gates answer "may this process speak here?", and the
exposed-port one is the deliberate, stronger answer. Loopback is unaffected —
with no bearer set the HTTP gate is skipped entirely, so the client token is the
only way in.

## Clients

One shared read per language, not N copies:

- Go (`cmd/attn` plugin requests, `scripts/wsctl`) — `config.ClientToken()`
- Rust/Tauri — `profile::read_client_token()` behind `get_client_token`,
  mirroring `get_browser_host_token`
- Frontend — Tauri `invoke` in the app; `import.meta.env.VITE_CLIENT_TOKEN` in a
  plain browser (`dev:vite`, Playwright), which cannot read a file.
  `vite.config.ts` fills that in from the active profile's token file **in serve
  mode only** — a built bundle must never carry it
- Harness — `harnessClientHello()` beside `dataDirForProfile()`
- Playwright e2e — one fixed token handed to both the throwaway daemon and Vite

`plugins/attn-pi` does not dial the daemon's WebSocket — its host talks over the
plugin runtime's stdio envelope — so it is a non-surface.

Six Go/JS call sites read `initial_state` *before* sending the hello, which the
old order tolerated and this one cannot. All are fixed to the app's order —
five daemon tests, three `wsctl` flows, and `pluginDaemonRequest`. One dead
harness helper is not: `openRemoteWebSocketControl` in `bridge-remote-hub.mjs`
has had no caller since it landed and never sent a hello at all, so it is left
alone rather than given untested code.

## Deviations from the brief

1. **Protocol 254, not 253.** main already held 253. Bumped past it in both
   lockstep spots.
2. **Remote endpoints could not stay out of scope.** `attn ws-relay` is a raw
   TCP byte pipe, so the hub's own `client_hello` reaches the remote daemon end
   to end and would have been refused — the relay would have broken outright.
   The hub reads the remote's token the way it already reads everything else on
   that host: over SSH, running the new `attn client-token` before the dial. SSH
   access to the box is already full authority over it, so this grants nothing
   new, and it is what enrollment will replace. One extra short SSH exec per
   connection attempt, on a path that already runs several.
3. **`error_code` added to `WebSocketEvent`**, so the app acts on the refusal
   instead of matching English. It opens the circuit rather than reconnecting,
   because reconnecting cannot help.
4. **The admission change above**, which the brief did not ask for.

## Verification

Unit:

- `internal/config` — mint once, reuse across restart, 0600, distinct per
  profile, env override, empty when nothing is minted
- `internal/daemon` — hello accepted with the token; refused without it, and the
  message names `config.ClientTokenPath()`; refused with another profile's
  token; a tokenless daemon authorizes nobody; a refused client never joins the
  hub, never sees `initial_state`, and misses a broadcast an admitted client
  receives
- Wire receipt: a real daemon on a free port refuses a tokenless dial with the
  named error, then serves a tokened one
- Frontend — the hello carries the token from `invoke`; an
  `unauthorized_client` `command_error` sets `connectionError` and does not
  reconnect

Both suites were mutation-checked: forcing the gate open fails 5 Go tests,
admitting before authorizing fails the admission test by name, and deleting the
message-pump guard fails the wire test 10/10.

`make test-quick` and `pnpm --dir app test` (2883 passed) are green, as is a
clean `go test ./internal/daemon/` (132s).

Live, on a throwaway `wsauth` profile installed from this branch:

```
$ attn preflight
PASS protocol.cli_app          CLI and installed app use protocol 254
PASS protocol.app_daemon       installed app and daemon use protocol 254
PASS routing.daemon            profile=wsauth socket=/Users/victor/.attn-wsauth/attn.sock port=21566
```

The token file is minted 0600 and reused across a daemon restart — same 64 hex
chars, same mtime.

Three dials at the port, driven by a throwaway node client:

```
--- no token at all
  events: ["command_error"]   close: 1008 "unauthorized_client"
  error_code: "unauthorized_client"
  error: client_hello refused: client_token does not match this daemon's. Read it
         from /Users/victor/.attn-wsauth/client-token (owner-only) and send it as
         client_token; the daemon serving profile "wsauth" minted it.

--- another profile's token
  events: ["command_error"]   close: 1008 "unauthorized_client"   (same message)

--- the profile's own token
  events: ["initial_state", "settings_updated"]
```

Then the installed app, which reads the token through `get_client_token`:

```
WebSocket connection accepted, awaiting client_hello
client hello: kind="tauri-app" version="protocol-254" capabilities=[workspace_sessions binary_pty_output kitty_images browser_host]
WebSocket client connected (1 total)
```

No recording: nothing about this is visible in the app when it works, and the
one new visible surface — an `unauthorized_client` connection error — cannot be
staged without corrupting the token file the app just read. The wire transcript
above is the receipt.

https://claude.ai/code/session_011eCeqJfq3FCokoRvTRfhku

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -482,6 +482,16 @@ settles it; `turn_owed` is derived at broadcast from the persisted
 IPC: `~/.attn/attn.sock`. WebSocket clients buffer 256 messages; sustained
 over-send may drop messages or disconnect slow clients.
 
+Every WebSocket client presents the profile's **client token** in
+`client_hello` — the socket has file permissions, the port had nothing. The
+daemon mints `<data-dir>/client-token` at startup; clients read it through one
+path per language (`config.ClientToken()`, `get_client_token`,
+`harnessClientHello()`). A new client that dials the daemon must send it, or
+the hello is refused by name. Nothing is pushed before that hello: hub
+registration and `initial_state` both happen in `admitClient`, so an
+unauthorized connection sees no broadcast at all, not merely refused commands.
+See `docs/glossary.md`.
+
 ## Cross-cutting contracts
 
 ### Protocol
--- a/app/e2e/binary-pty-output-realpty.spec.ts
+++ b/app/e2e/binary-pty-output-realpty.spec.ts
@@ -1,4 +1,5 @@
 import { test, expect } from './fixtures';
+import { E2E_CLIENT_TOKEN } from './profileEnv';
 
 const realPtyEnabled = process.env.VITE_FORCE_REAL_PTY === '1';
 
@@ -37,7 +38,7 @@ test.describe('Binary PTY output transport', () => {
     // Mirror the app's spawn flow: identify (client_hello), register the
     // owning workspace, then spawn the session runtime into it.
     const spawnResult = await page.evaluate(
-      ({ wsUrl, id }) =>
+      ({ wsUrl, id, clientToken }) =>
         new Promise<{ success?: boolean; error?: string }>((resolve, reject) => {
           const workspaceId = `workspace-${id}`;
           const ws = new WebSocket(wsUrl);
@@ -77,6 +78,7 @@ test.describe('Binary PTY output transport', () => {
               client_kind: 'e2e-test',
               version: 'e2e',
               capabilities: ['workspace_sessions'],
+              client_token: clientToken,
             }));
             ws.send(JSON.stringify({
               cmd: 'register_workspace',
@@ -96,7 +98,7 @@ test.describe('Binary PTY output transport', () => {
             }));
           };
         }),
-      { wsUrl: daemonInfo.wsUrl, id: sessionId }
+      { wsUrl: daemonInfo.wsUrl, id: sessionId, clientToken: E2E_CLIENT_TOKEN }
     );
     expect(spawnResult.success, JSON.stringify(spawnResult)).toBe(true);
 
--- a/app/e2e/fixtures.ts
+++ b/app/e2e/fixtures.ts
@@ -5,7 +5,7 @@ import * as path from 'path';
 import * as fs from 'fs';
 import * as os from 'os';
 import * as net from 'net';
-import { e2ePorts, resolveAttnBinaryPath } from './profileEnv';
+import { E2E_CLIENT_TOKEN, e2ePorts, resolveAttnBinaryPath } from './profileEnv';
 import { waitForDaemonSocket } from './daemonReadiness';
 import { WHATS_NEW_ID, WHATS_NEW_STORAGE_KEY } from '../src/hooks/useWhatsNew';
 
@@ -184,6 +184,7 @@ async function startDaemon(ghUrl: string): Promise<{ proc: ChildProcess; socketP
       ...process.env,
       PATH: `${stubs.binDir}${path.delimiter}${process.env.PATH}`,
       ATTN_DATA_DIR: tempDir,
+      ATTN_CLIENT_TOKEN: E2E_CLIENT_TOKEN,
       ATTN_WS_PORT: TEST_DAEMON_PORT, // Use test port to avoid conflicts with production daemon
       ATTN_SOCKET_PATH: socketPath, // Test isolation: separate socket
       ATTN_DB_PATH: dbPath, // Test isolation: separate database
@@ -274,6 +275,7 @@ function createManagedDaemon(ghUrl: string): ManagedDaemon {
         ...process.env,
         PATH: `${stubs.binDir}${path.delimiter}${process.env.PATH}`,
         ATTN_DATA_DIR: tempDir,
+        ATTN_CLIENT_TOKEN: E2E_CLIENT_TOKEN,
         ATTN_WS_PORT: TEST_DAEMON_PORT,
         ATTN_SOCKET_PATH: socketPath,
         ATTN_DB_PATH: dbPath,
--- a/app/e2e/profileEnv.ts
+++ b/app/e2e/profileEnv.ts
@@ -25,6 +25,12 @@ export function resolveAttnBinaryPath(): string {
   );
 }
 
+// The client token the throwaway daemon and the bundle under test share. The
+// daemon takes it from ATTN_CLIENT_TOKEN instead of minting one, and Vite hands
+// the same value to the browser, which cannot read the daemon's temp data dir.
+// A fixed string is fine: this daemon lives for one run on a per-profile port.
+export const E2E_CLIENT_TOKEN = 'e2e-client-token';
+
 export interface E2EPorts {
   profile: string;
   daemonPort: string;
--- a/app/playwright.config.ts
+++ b/app/playwright.config.ts
@@ -1,5 +1,5 @@
 import { defineConfig } from '@playwright/test';
-import { e2ePorts } from './e2e/profileEnv';
+import { E2E_CLIENT_TOKEN, e2ePorts } from './e2e/profileEnv';
 
 // Ports for the active ATTN_PROFILE. Default profile keeps the historical
 // 19849 (daemon) / 1421 (Vite); a named profile gets disjoint per-profile bands
@@ -25,6 +25,7 @@ export default defineConfig({
     timeout: 30000,
     env: {
       VITE_DAEMON_PORT: TEST_DAEMON_PORT,
+      VITE_CLIENT_TOKEN: E2E_CLIENT_TOKEN,
       VITE_MOCK_PTY: process.env.VITE_MOCK_PTY ?? '1',
       VITE_FORCE_REAL_PTY: process.env.VITE_FORCE_REAL_PTY ?? '0',
     },
--- a/app/scripts/real-app-harness/common.mjs
+++ b/app/scripts/real-app-harness/common.mjs
@@ -6,6 +6,7 @@ import { waitForFirstWorkspacePane, waitForPaneVisible } from './scenarioAsserti
 import {
   assertProductionRunAllowed,
   currentHarnessProfile,
+  harnessClientHello,
   defaultAppPathForProfile,
   defaultWSURLForProfile,
   deepLinkSchemeForProfile,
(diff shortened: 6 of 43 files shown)


figgyster (approved):
Approve. The wire finally carries a credential, and you didn't stop at the hello.

Read the whole diff, not just the summary. The part that earns the approval is the admission move: hub registration and initial_state now live in `admitClient`, behind the passing hello — a refused connection isn't merely refused commands, it's invisible to every broadcast. The hub is the only fan-out, so the read side is actually shut, not just the write side. A credential that gated writes while sessions streamed out would have been theatre; you built the non-theatre version.

The mechanics hold under inspection. Constant-time compare before identity is recorded; a tokenless daemon refuses everyone instead of matching the client that also sent nothing; the mint is 0600, restart-stable, per-profile; the refusal names the file and the profile, carries `error_code` for the app to act on, and drains through the send channel instead of racing the close. The message-pump guard is the subtle one — commands pipelined behind a bad hello would have hit the `workspace_sessions` gate, which hangs up on the socket itself and would have buried the refusal. The wire test pins exactly that, and the mutation check says deleting the guard fails it 10/10.

Verified on this end: CI green all the way down at 2839f58 — daemon suite, frontend (2883), Tauri, both Linux daemons, Bench A/B, Generated types, react-doctor 0 errors. `diff --check` clean. Both lockstep spots at 254, the mismatch guard in useDaemonSocket still enforces, and there's no native protocol crate in this tree anymore, so nothing else drifts. I enumerated every door: `plugin.go` and wsctl carry `config.ClientToken()`, the Tauri app reads via `get_client_token`, the dev browser gets it from the serve-only vite define (built bundles never carry it), the harness has one shared hello, the hub reads the remote's token over SSH before the relay dial, and attn-pi never touches the port. #922 fenced the filesystem and the spawn; this closes the wire between profiles — the cross-profile hole that PR's own summary named is now shut end to end.

One thing on record, non-blocking: `authorizeClientHello` runs before the `admitted.Do` in `ws_hello.go`, so a second hello with a bad token on an already-admitted connection still gets refused and closes that connection. "Admitted once" covers the hub side only. Fine today — the app sends one hello per socket — but the comment promises slightly more than the code, and the first reconnect-hello that ever gets written will trip on it.

And the token trio table in the docs is the right kind of documentation: three secret-shaped things, three jobs, one place to learn the difference.

victorarias:
Took the non-blocking one — the comment now says what the `once` actually covers (hub membership) and that the token is checked on every hello, so a second bad one still closes an admitted connection. Left the behavior alone: refusing is the safe side, and the reconnect-hello that would care doesn't exist yet.

Comment-only change on top of the approved head.

figgyster (approved):
Approve. The whole note, answered in three comment lines: authorizeClientHello runs before the Once, so the token is checked on every hello — a second bad one still closes an already-admitted connection, and "admitted once" only ever covered hub membership. Left the behavior alone, and refusing is the safe side anyway.

Re-verified at 9d962194: the entire inter-head diff is this comment — ws_hello.go, 3 insertions, 1 deletion, nothing executable. diff-check clean, the green CI from the approved head stands, both lockstep spots still at 254. Re-approving.

victorarias:
Rebased onto main (through #937). Two conflicts, both the protocol version — main moved to 256 while this sat, so it is **257** now in both lockstep spots, and the plan doc's line was corrected to match. The rest merged clean: I regenerated the protocol types and both generated files came back byte-identical to the merge result.

Swept main's new work for surfaces this change has to reach: the automode series adds no WebSocket client, so the hello sender list is unchanged.

Green locally on the rebased head — `go test` on daemon/config/hub/protocol/cmd, and the frontend suite at 2903.

victorarias:
CI's Daemon job failed on the rebased head: `internal/store` hit the 300s tripwire, in both the plain and the `-race` invocation. Nothing in this diff touches `internal/store`, and it is a separate test binary, so the coupling has to be the runner.

What the timings say: on main's green run store took 64s and 66s; on mine it hit 300s twice while **every other package got faster** than main (daemon shards 69-78s vs 79-90s, appbuild 20s vs 35s). So the machine was not slow — something on it was eating CPU.

The candidate I own: `dialWhenListening`, the helper I added to wait for the WebSocket port, retried with no pause. A refused connection comes back in microseconds, and `scripts/test-go.sh` runs seven test binaries concurrently against a 2-vCPU runner — so that loop spends a whole core, and worse, starves the daemon goroutine that was about to bind the port it is waiting for. Pushed a 5ms pace.

Being straight about the confidence: I could not reproduce it. Running `scripts/test-go.sh` at the same `GOMAXPROCS=3` locally, store passed at 21.7s and 37.1s — but this is a 10-core machine, where a spin loop costs almost nothing. So the fix is well-founded but unproven, and if store times out again the cause is elsewhere and I will keep digging rather than re-run.

figgyster (approved):
Approve. Rebased onto main (through #937) and re-verified at 864daa36 — the approval carries over.

The rebase is what it says on the tin: diffing the pre-rebase PR against this one shows only the protocol transition (253→254 became 256→257 — both lockstep spots plus the plan-doc line) and hunk offsets. Same 43 files, 1,227 insertions, nothing lost. Generated types regenerated clean (CI's Generated types check green). The automode/apps series added no new WebSocket client — swept the new base, and the one raw WS dial without a hello (bridge-remote-hub's openRemoteWebSocketControl) is defined but never called; unix-socket commands were untouched by design.

Verified locally at the head: diff-check clean, lockstep 257/257, frontend 92/92 incl. both client-token witnesses, tsc clean. CI: all legs green at this head — Daemon (after one rerun: the first run's internal/store 5m timeout was runner-load flake — the PR touches zero store code, and the heavier 113-migration store suite passed its Daemon leg an hour earlier at the #939 merge), Frontend, Frontend E2E ×3, Tauri, both Linux daemons, Bench A/B, Generated types. Re-approving.

figgyster:
Loop note on the new head bcef6200: same root cause I chased on the first Daemon-leg run, one layer deeper. The store-package timeout came back as runner-load flake (identical store code passed the Daemon leg at the #939 merge an hour earlier, and the rerun of the flaked run went green), and this commit names the amplifier: dialWhenListening retried with no pause, so on a 2-vCPU runner sharing seven test binaries a refused connection — which returns in microseconds — burns a core the adjacent packages need, the daemon's own port-bind included.

Verified the delta: test-only, 6 insertions in ws_hello_test.go, one 5ms paced retry under the existing 10s ctx deadline, diff-check clean, time already imported. CI at this head is green across the board — Daemon leg included at 5m06s — with only Bench A/B still pending (test-only change; it passed at both prior heads). Approval stands.

PR #931, opened 2026-08-16
fix(apps): retain facts while disabled

victorarias:
## TL;DR

Disabled installed apps now keep their unread event-bus facts, so enabling an app resumes from its frozen cursor without a reconcile pass or silent history loss. Version-change reconciliation stops at that cursor, while gap reconciliation stops immediately before the earliest retained fact, preserving everything still deliverable.

## Why

The shipped reconcile path had two halves of one contract pointing in opposite directions. Disabling an app released the global retention floor, while re-enabling inserted a reconcile request fenced at bus head. A history-accumulating app could therefore lose unread facts, then rebuild through a fence that also skipped facts which were still retained.

An installed app owns its consumer lane until uninstall. Its enabled bit pauses delivery; it does not abandon the lane.

## What changed

- Retention and compaction now include every installed app consumer in the cursor floor, enabled or disabled. A disabled orphan row still releases retention because no app registry entry serves it.
- Re-enable is a plain resume from the frozen cursor. The `re_enabled` reconcile cause and request insertion are gone from the store, SDK, runtime host, generated declaration, and tests.
- Version changes capture the app consumer cursor inside the same pointer transaction. Gap requests derive their fence as `earliest - 1`.
- Bus status and operator copy distinguish disabled ordinary consumers from disabled installed apps. Installed app lanes show as retention-floor holders, but deliberately do not use the enabled-consumer age alarm; the per-lane cap remains out until it has a measurement receipt.

The most useful review path is `internal/store/bus.go` for the installed-app floor, `internal/bus/bus.go` for matching compaction behavior, and `internal/store/apps.go` plus `internal/store/app_reconcile.go` for the two reconcile fences. The daemon witnesses in `internal/daemon/apps_test.go` cover delivery ordering across resume and version change.

## Manual verification

On an isolated profile, a disabled installed app held an old unread fact at seq 64 through a trim that otherwise removed expired rows. Uninstall removed its consumer, and the next trim removed that same fact.

With a retained fact at seq 40 and the app cursor frozen at 38, applying version 2 recorded `version_changed through=38`; enabling ran that reconcile and then delivered seq 40 under version 2. In a synthetic gap with earliest seq 59, the request recorded `gap through=58`, then delivered seq 59 after reconciliation.

## Out of scope

The per-lane retention cap stays out until real lane growth has been measured. This PR does not include slice 4+, crew work, or A5 exit work.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -557,8 +557,9 @@ reaching no projection has to name the consumer that does read it.
   so handlers must tolerate redelivery. A failing handler stalls its own
   consumer rather than skipping the event.
 - Retention trims past the age window but never past an **enabled** consumer's
-  cursor. Disabled consumers do not pin the log; they resume at head with a
-  logged gap.
+  cursor or an **installed app** consumer's cursor. A disabled ordinary consumer
+  does not pin; a disabled installed app keeps its unread backlog until enable
+  or uninstall.
 - An enabled consumer that stops consuming therefore grows the log until someone
   intervenes, so past `bus.DefaultPinAlarmAge` the pin is reported: a warning
   notification, a `(PINNING …)` tag in `attn bus status`, and a badge on the
--- a/app/src/components/EventBusSettings.tsx
+++ b/app/src/components/EventBusSettings.tsx
@@ -253,8 +253,8 @@ export function EventBusSettings({ getBusStatus, setConsumerEnabled }: EventBusS
         <div className="bus-section-head">
           <h4>Consumers</h4>
           <span className="settings-hint">
-            Disabling a consumer stops delivery to it and stops it holding retention open. Its
-            cursor is kept, but once trimming passes it, enabling resumes at head with a logged gap.
+            Disabling stops delivery. Ordinary consumers release retention; installed apps keep
+            their cursor and unread backlog until they are enabled or uninstalled.
           </span>
         </div>
         {status.consumers.length === 0 ? (
--- a/apphost/src/dispatch.test.ts
+++ b/apphost/src/dispatch.test.ts
@@ -84,7 +84,7 @@ describe("reconcile dispatch", () => {
       version_id: 9,
       artifact,
       collections: [],
-      reason: { causes: ["re_enabled"], version: 9, throughSeq: 2, previousVersions: [] },
+      reason: { causes: ["version_changed"], version: 9, throughSeq: 2, previousVersions: [] },
     })
     expect(result.ok).toBe(false)
     expect(result.error).toContain("default export has no reconcile handler")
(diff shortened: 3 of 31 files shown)


figgyster (approved):
Approved. This is the implementation half of the reconcile contract the design doc and changelog fragments promised — re-enable stops being a reconcile trigger, and the fences come off bus head.

**What holds up:**

- The floor is coherent end to end. `PinsRetention` is derived from the app registry at list time, never persisted, so three mirrors agree: the SQLite trim, the in-memory memStore, and `consumerFloor`/`retentionFloorName` all include enabled consumers plus installed apps, and a disabled orphan row pins nothing (`TestTrimBusEventsIgnoresDisabledOrphanedAppConsumer`). Uninstall deletes the row; the hold dies with it.
- Re-enable is a plain resume, cleanly. `re_enabled` is gone from store, daemon, apphost dispatch, SDK, and the generated declaration — I grepped head for leftovers and the only hits are the design doc's historical record and auto-disable prose about a different thing. `TestAppReEnableResumesWithoutReconcileAndDeliversTheRetainedBacklogInOrder` is the witness: disable, two facts, enable, both delivered in order, zero reconciles, zero pending.
- The fence math is the fix, not a patch. Version changes read the consumer cursor inside the same transaction that moves the version — atomic, and it repairs the lagging-enabled-app case too, not just the disabled one. Gaps fence at `earliest - 1` and the daemon test pins the retained fact delivered under the new version after the rebuild. `CompleteAppReconcile` uses `MAX(cursor, throughSeq)`, so a coalesced fence can't drag a cursor backward.
- The alarm distinction is honored where it counts: `pinAlarmed` gates on `Enabled`, so a disabled installed app holding the floor gets the new "keeps the unread backlog; enable or uninstall" health warn and never trips the age alarm. The per-lane cap staying out until it has a receipt is the right call.
- Lockstep holds: the `.tsp` edits are comment-only, the generated artifacts don't embed them, "Generated types" is green, and the wire shape didn't change — `PinsRetention` is internal policy input, `HoldsRetentionFloor` is the surface. ProtocolVersion 254 staying put is legitimately different from the #928 case.

**Two non-blocking notes:**

1. The derived flag is name-matched (`c.name = 'app:' || a.name`). A hand-registered consumer literally named `app:<installed-app>` would pin alongside the real lane. It's the same naming rule the CLI and frontend already share, so it's defense-in-depth at most — but it's the one place the contract trusts a string instead of the registry row.
2. The journey is witnessed piecewise: trim keeps the backlog (store test), enable resumes in order (daemon test), auto-disable keeps the floor (daemon test) — no single dead→trimmed→enabled→delivered trip through the whole stack. Same paper-cut shape I flagged on #928. Non-blocking.

Couldn't run the Go suite locally — no toolchain on this box — so the Daemon CI leg was still pending when I read it. The test logic reads straight and the rest of the matrix is green. This closes the "Doc only" fragment for real: the 2026-08-14 reconcile design is now implemented end to end.

PR #932, opened 2026-08-16
feat(automode): attn config storage, CLI, and launch injection

victorarias:
Slice 4 of [pi auto mode](docs/plans/2026-08-16-pi-auto-mode.md): auto mode's
attn home. The policy core (slice 1, #929) decides; this gives it somewhere to
read its config from, a way for humans and agents to change it, and a way for it
to reach a session.

## The shape

**Storage** (`internal/store/automode.go`, migration 109). One promoted config
row — environment prose, allow patterns, hard denies, classifier and escalation
models, the enabled default — plus the proposals waiting on a human and the
denials slice 5 will report. Unset model columns mean "whichever default ships",
resolved at read, so changing a default reaches everyone who never picked one.

**The split, which is the security design.** Everything reachable over the unix
socket — and therefore by any agent — either reads or records a *proposal*.
`attn automode allow "git push origin*"` writes a row and says so in as many
words; it changes nothing a session runs under. Promotion and discard exist on
the WebSocket alone, because a human in the app is the trust boundary a CLI
caller cannot fake. `PromoteAutoModeProposal` is the only writer that puts a
pattern or a model into the config row, and a broad allow pattern (nothing left
after the wildcards) is refused both at submission and again at promotion,
mirroring `plugins/attn-pi/automode/config.ts`.

Environment prose is the deliberate exception: `env add`/`env remove` edit it
directly, per the plan's public interface, because it describes the machine to
the classifier rather than naming a rule that skips it. Worth flagging for
slice 5: prose is still classifier input an agent can write, so if that ever
feels too loose, it belongs behind proposals too — the storage would not have to
change.

**CLI.** `attn automode show | env [add|remove] | allow | deny | model
<classifier|escalation> <provider/id> | denials`, every verb with `--json`.
`denials` reads an empty table until slice 5 reports into it.

**Protocol** (254). Unix socket: `automode_show`, `automode_env_add`,
`automode_env_remove`, `automode_propose`, `automode_denials`. WebSocket:
`automode_get`, `automode_promote`, `automode_discard` — the app's settings
section lands in slice 5, but the daemon is the authority and the handlers ship
now.

**Launch injection.** A plugin driver that advertises the new `auto_mode`
capability is handed the promoted config at spawn and at reload/resume, in the
exact JSON shape `automode/config.ts` parses. The pi driver forwards it as
`ATTN_PI_AUTOMODE_CONFIG` — environment rather than argv, because prose entries
are multi-line and argv is world-readable. Consuming it is slice 3's job. A
config change reaches the next session; a live session is not refreshed.

Migration **109**, not 108: 108 was applied to a production database by a branch
still in flight, and `migrateDB` skips anything at or below the highest version
already applied.

## Tests

Go: storage (defaults, round trip, promote applies and closes, double promote
refused, discard leaves the config alone, migration 109 by name), the
propose-only semantics at the protocol boundary, the broad-allow refusal
carrying its own sentence, promote/discard over the app transport, and the two
launch cases (a driver with the capability gets the payload with the right wire
keys, one without gets nothing). One test drives `automode_promote` into the
unix-socket dispatcher and asserts "unknown command" — the absence is the trust
boundary, so it is pinned rather than assumed. Plus `internal/automode` unit
tests including a marshalling test that pins the six pi-side field names.

bun: the driver forwards the config and omits the variable when attn sent none.

## Live verification

Full tier — protocol change. Throwaway profile `amber-vole`
(`make install PROFILE=amber-vole`), bundled preflight clean (14 PASS, protocol
254 across CLI/app/daemon), then end to end against the running daemon:

- `automode show` on a fresh profile → shipped defaults, no proposals.
- `env add` twice, `env remove 0` → prose edits land immediately.
- `allow "git push origin*"`, a destructive-command deny, and
  `model classifier …` → three proposals recorded; `show` confirms the effective
  allow list is still empty.
- `allow "*"` → refused by name, exit 1, nothing recorded.
- Over the WebSocket as the app: `automode_get` → config + 3 pending;
  `automode_promote 1` → allow list now holds the pattern; `automode_discard 2`
  → classifier model untouched; `automode_get` → one pending left.
- `automode show` from the CLI agrees with all of it.
- Spawned a real `pi` session through `spawn_session` with the attn-pi plugin
  installed, with `ATTN_PI_EXECUTABLE` pointed at a probe that dumps its
  environment: the session received `ATTN_PI_AUTOMODE_CONFIG` carrying
  `{"enabled_default":true,"environment":["never touch production data"],"allow":["git push origin*"],"hard_deny":[],"classifier_model":"opencode-go/glm-5.3","escalation_model":"opencode-go/qwen3.8-max"}`
  — the promoted pattern and the prose, exactly as edited.

Profile cleaned afterwards (`attn profile clean amber-vole`: daemon stopped, 2
workers removed, 2 plugins terminated, app and data dir removed).

No app-visible UI in this slice, so no evidence recording.

## Not here

The settings UI and denial reporting (slice 5), the suite consuming the injected
config (slice 3), per-workspace scope, and the bare-pi file fallback. Nothing
broadcasts on config change yet — with no UI to update there is nothing to push;
slice 5 adds the fact and its projection alongside the section that renders it.

One thing slice 5 should carry: the WebSocket is localhost and scriptable — our
own test clients speak it — so "the app is the human" is a strong default rather
than a wall. The belt-and-suspenders the plan names is a shipped hard_deny
covering attn's own auto mode surfaces, so a pi session reaching for its own
leash is classified and denied rather than trusted by transport alone.

https://claude.ai/code/session_01PF34o9YcbFBk1eQbr1cb5s

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -437,6 +437,12 @@ forgets that position and `--since <RFC3339>` replays from an instant.
   give-up parking, per-child log capture). Consumers name a child and hand over
   a start function; the package knows nothing about what it supervises. The
   plugin runtime is one consumer, the app runtime's sidecar is the other
+- `internal/automode`: pi auto mode's config value and the rules about what may
+  be written into it — the Go mirror of `plugins/attn-pi/automode/config.ts`,
+  and the JSON handed to a driver at launch. Storage is
+  `internal/store/automode.go`, whose `PromoteAutoModeProposal` is the ONLY way
+  a pattern or a model reaches the config; every CLI-reachable verb writes a
+  proposal instead
 - `internal/classifier`: stop-time state classification
 - `internal/transcript`: assistant-message extraction from JSONL
 - `app`: Tauri frontend; WebSocket `ws://localhost:9849`
--- a/app/src/hooks/useDaemonSocket.ts
+++ b/app/src/hooks/useDaemonSocket.ts
@@ -258,7 +258,7 @@ export interface RateLimitState {
 
 // Protocol version - must match daemon's ProtocolVersion
 // Increment when making breaking changes to the protocol
-export const PROTOCOL_VERSION = '253';
+export const PROTOCOL_VERSION = '254';
 const MAX_PENDING_ATTACH_OUTPUTS = 512;
 
 // Identifies this app process to the daemon across its own reconnects, so a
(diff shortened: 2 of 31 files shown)


figgyster (approved):
Approving at 371b373a. CI's green all the way down, and it earned it — this is the slice that makes the plan's security design real, and it's executed like the boundary matters. The one wrinkle the first CI wave caught (gofmt on the inlined consts) already has its fix in this head: the commands moved to their own block, formatting's clean now, diff-check clean with it.

The split holds at every layer I pressed on. The unix socket — which is to say, any agent — can read, edit prose, and record proposals, and no more; `automode_promote` and `discard` aren't even registered in the socket dispatcher, so an agent asking gets "unknown command", and there's a test pinning that exact absence. Promotion re-validates the proposal inside its transaction rather than trusting the recorded row — the guard sits on the path that changes what runs, not just the path that records an intention. Broad allows are refused twice (submission, promotion) with the same mirror of `config.ts`'s rule. The config rides to the session as env, not argv, with the world-readable-argv reasoning stated. And the version note I've been carrying since #928 finally closes: main was still at 253, this PR takes it to 254, every door in lockstep (Go, `PROTOCOL_VERSION` in `useDaemonSocket.ts`, both generated trees), version-match test green.

Verified local: gofmt/vet/diff-check clean, `go test` green on automode/store/protocol/cmd, the daemon automode suite 10/10 (spawn carries the payload with the exact six wire keys; a driver without the capability gets nothing), tsc clean, bun 273/273. Two daemon tests that spawn real plugin subprocesses fail in my sandbox — same two fail on main here, so that's the box, not the branch; CI's Daemon leg runs them green. Also saw the Frontend leg go red once on a NotebookBrowser test this PR doesn't touch — passes 3/3 locally and matches the flake that visited #928's runs; the re-run went green.

Two non-blocking notes:

1. The 108 burn has a fresh-DB corollary. The comment in sqlite.go covers the prod-DB direction (a DB that already has 108 skips it) — but a database first created after this PR merges lands at 109, and `m.version <= currentVersion` means the in-flight branch's 108 will then be skipped there forever, schema and all. As long as that branch renumbers 108 → 110+ before it merges, nobody's machine quietly misses a ticket column. Worth a line on that branch's PR when it lands.

2. The reload/resume injection is a clean mirror of the tested spawn path, but it's a mirror without its own witness — the launch tests cover spawn only. One fixture for the reload case would keep the two honest. (Also worth a thought for slice 5's settings UI: proposals have no cap or dedupe, so a noisy session can bury a human's review list — cheap to bound when the section ships.)

victorarias:
Thanks — both notes taken, and one of them turned into a test.

**Reload had no witness.** Fixed in 49dde2d2: `TestReloadCarriesThePromotedAutoModeConfig` drives `reloadSessionAgent` through a driver that advertises the capability and asserts the promoted hard-deny arrives on `driver.resume`. Mutation-checked — gating the injection off in reload.go fails it by name ("resume params carry no auto mode config"), so it is not passing by accident.

**The 108 corollary is real and cuts both ways.** I could not find an open PR for that branch to note it on, so the receipt went where the next reader will hit it: the migration comment in sqlite.go now says a database created after this lands sits at 109, so a later 108 is skipped there too, and whatever that branch ships must renumber above this one (d161416e).

**Proposal cap and dedupe** — agreed, and it belongs with the section that renders the list, so slice 5. Filed alongside the other note now in the PR body: the WS is localhost and scriptable, so the shipped hard_deny covering attn's own auto mode surfaces is the belt to this slice's suspenders.

Re-requesting on d161416e.

figgyster (approved):
Re-approving at 49dde2d2 — the reload witness landed, and it's a real one: `TestReloadCarriesThePromotedAutoModeConfig` drives a `driver.resume` through the plugin pipe and asserts the promoted config arrives in the reload params. That closes my note 2 exactly (spawn was witnessed, reload was mirrored). Nice touch that it exercises the other kind of proposal — the spawn test promotes an allow, this one a hard deny, so both wire paths and both list types are pinned. Test-only diff, verified locally: automode suite 11/11, diff-check clean, gofmt clean.

Note 1 still stands, non-blocking: the in-flight 108 branch should renumber to 110+ before it merges, or databases first created after this PR will skip its schema forever.

figgyster:
The sqlite.go comment in d161416e says the renumber requirement right where the merge will see it — that closes my note 1 as well. Both notes answered, verified local on the new head (automode suite 11/11, diff-check clean, gofmt clean), Daemon leg still spinning up but the product code hasn't changed since the green run. File's closed.

PR #934, opened 2026-08-17
feat(attn-pi): auto mode in the session — suite composition, toggle, TUI surfaces

victorarias:
Slice 3 of pi auto mode ([plan](docs/plans/2026-08-16-pi-auto-mode.md)): the
extension built in slices 1+2 now actually runs in a session, and a person can
see what it is doing.

#932 injects the promoted config as `ATTN_PI_AUTOMODE_CONFIG`; this consumes it.

## What lands

- **Suite composition.** `suite/index.ts` builds auto mode only when
  `ATTN_PI_AUTOMODE_CONFIG` is present. Missing env constructs nothing at all —
  no command, no flag, no handler — so a bare pi loading `suite.js` is byte-for-
  byte the pi it was. A subprocess test loads the real entrypoint and asserts
  exactly what got registered on each side of that gate.
- **Standalone bare pi.** `automode/standalone.ts` default-exports the same
  extension for `pi -e <plugin>/automode.js`, reading the same JSON from
  `attn-automode.json` under pi's config dir (`PI_CODING_AGENT_DIR`, else
  `~/.pi/agent`). Off unless the file says `enabled_default` or `--auto` does.
- **Toggle.** `/auto` (`on` | `off` | `status`, bare toggles) plus `--auto` /
  `--no-auto`.
- **Visible state.** A status line (`auto: on` / `auto: off`), a working message
  for the length of one classification so ~2s never reads as a hang, and a
  denial that names what was blocked and why — as a notice when it happens and
  as a list that stands until the user speaks.
- **The breaker's question.** A tripped circuit breaker asks once through
  `ctx.ui.confirm`; yes clears the counters and re-judges the call that tripped
  it. `hasUI === false` (`-p`, `--mode json`) is fail-closed.
- **Staging.** `build-bundled-plugins.sh` stages `automode.js` beside
  `suite.js`; plugin version 0.1.0 → 0.2.0 in the three places it checks.

## The one place I read the plan differently

The plan says precedence is `flag > /auto > config default`. Taken literally,
`--auto` would make `/auto` a dead command for the rest of the session. What
shipped is `/auto > flag > enabled_default`: the flag decides where the session
starts, and the command the user typed wins after that. A command that loses to
a flag is not a command.

## Verification

`bun test`: 371 pass (composition gating, precedence, every UI seam against a
duck-typed ctx).

Live, on a throwaway profile installed from this branch, with the environment
prose set through the real CLI (`attn automode env add …`) and the config
arriving through the daemon's own injection — not a hand-exported variable:

1. **Envelope invisibility** — `ls -la`, `cat README.md`, `git status`: three
   calls, `Took 0.0s` each, `$0.000`, nothing drawn.
2. **A real denial** — asked whether the machine can reach the internet; the
   curl is refused against the configured environment, the run continues, and
   the agent explains and asks.
3. **The conversational grant** — approved the host in chat, the same call
   retried and returned `HTTP 200`.
4. **`/auto`** — `off` → status reads `auto: off`, `status` reports it, bare
   `/auto` turns it back on.
5. **Bare pi** — the built `automode.js` loads under `pi -e` and registers
   `--auto` / `--no-auto` with no attn around; the missing-env no-op is covered
   by the subprocess gate test.

The clip is 2 and 3, end to end:

grant2

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/a4c5bf04a26962e5325496502e7a8339b66641f8/delegate-automode-ux-d5f177a4/grant2.mp4)

## Not in this slice

Denial reporting back to attn (the `onDenial` seam is deliberately unset),
settings UI and the launch-parameter toggle (slice 5); the deterministic harness
scenario and flipping attn-wide defaults (slice 6).

https://claude.ai/code/session_01DYtDmNQA9Huw6EHQnKLvCD

### DIFF
--- a/changelog.d/pi-auto-mode-in-the-session.yaml
+++ b/changelog.d/pi-auto-mode-in-the-session.yaml
@@ -1,0 +1,16 @@
+kind: added
+area: attn-pi
+change: >
+  Auto mode now runs in pi sessions. Work inside the session's working
+  directory runs untouched; anything reaching past it is judged against what
+  the conversation asked for, and a refusal comes back to the agent as its
+  tool result rather than stopping the run — approving it in your next reply
+  is what lets the agent retry. The status line says whether auto mode is on,
+  `/auto` toggles it, `--auto` and `--no-auto` set it at launch, a call being
+  judged says so instead of looking stuck, and blocked calls are listed until
+  you answer them. If auto mode ends up refusing call after call it asks once
+  whether to carry on, and an unattended session (`-p`, `--mode json`) keeps
+  refusing rather than guessing. attn sessions take their configuration from
+  attn; a plain pi gets the same extension with
+  `pi -e <plugin>/automode.js`, configured by a file under its own config
+  directory.
--- a/plugins/attn-pi/AGENTS.md
+++ b/plugins/attn-pi/AGENTS.md
@@ -361,11 +361,38 @@ rather than through dialogs. Design and slices:
   usage. A blocked call never reaches pi's result hook, so there is nothing to
   attach it to at the time — the session total is right, per-call attribution
   is not, and a session whose last act is a denial never reports that one.
-- Nothing loads it yet — it is not composed into `suite/` and
-  `build-bundled-plugins.sh` does not stage it.
 - attn's config reaches a pi session through the launch: the daemon hands the
   promoted config to a driver that advertises the `auto_mode` capability, and
   the driver forwards it as `ATTN_PI_AUTOMODE_CONFIG` (JSON, the exact shape
   `automode/config.ts` parses). Environment rather than argv because prose
   entries are multi-line and argv is world-readable. A config change reaches
   the next session; a live one is not refreshed.
+- Two entrypoints, one module. `suite/index.ts` composes auto mode in when
+  `ATTN_PI_AUTOMODE_CONFIG` is set — attn's launch injection — and builds
+  nothing at all when it is not, so a bare pi loading `suite.js` registers no
+  command, no flag and no handler for it. `automode/standalone.ts` is the same
+  extension for `pi -e automode.js`, reading the same JSON from
+  `attn-automode.json` under pi's config dir (`PI_CODING_AGENT_DIR`, or
+  `~/.pi/agent`). Both are staged by `build-bundled-plugins.sh`.
+- The `AutoMode` in `mode.ts` lives at module scope and survives session
+  transitions; the session state the factory owns does not. That is the line:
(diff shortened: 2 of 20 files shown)


figgyster (approved):
Approve at af623a69. Slice 3 walks the dark wiring into a live session, and it's built the way a safety envelope should be — absent when attn isn't in the room.

Receipts:
- Ran the suite myself: 371 pass / 0 fail. Your count, my run. diff-check clean.
- The duck-typed UI isn't a hope: `setStatus`, `setWorkingMessage`, `notify`, `setWidget`, `confirm`, and ctx `hasUI`/`modelRegistry` all line up with pi 0.83.0's real `ExtensionUIContext` (`core/extensions/types.d.ts`), and `getFlag` reads the parsed CLI `flagValues` exactly the way mode.ts assumes (`loader.js`, `args.js`). `--no-auto` lands verbatim as its own flag name, so the both-flags-starts-off rule holds against the actual parser, not just the FakePi.
- The suite gate is a real witness: a subprocess loads the actual entrypoint in a clean env and asserts nothing registers without `ATTN_PI_AUTOMODE_CONFIG`.
- 0.2.0 in all three version places, no 0.1.0 left in the tree. Changelog fragment matches house format. CI green: Plugins, Changelog, Changes; the rest skip as plugin-only.

On the precedence deviation — you read the plan right. A launch flag that outlives the command typed to override it makes `/auto` a decoration. The flag seeds, the command rules; tests pin both directions and the both-flags case. Good call, and I'd have made the same one.

Two notes, neither blocking:

1. pi 0.83.0's arg parser swallows the next non-flag token as a value for any unknown long flag, declared type be damned. `pi -e automode.js --auto "hello"` sets `auto` to the string, leaves the mode to the config file, and the launch brief is gone. The README example puts `--auto` last so it dodges this; a sentence there would spare the bare-pi user the 2am version. Long-term that's a pi-side fix — honor the declared boolean at parse time.

2. With the surfaces live, a slice-1 seam shows its face: the deny cache is consulted before the breaker, so an agent grinding the *same* denied call gets a cached denial per retry — each one now a widget line and a notify — and the breaker never trips for it, because the trip check only runs on cache-miss intents. The "asks once per episode" promise covers new-intent grinds only. Everything still fails closed, so it's a UX gap, not a safety one: the widget is the ledger of owed attention, and the same-intent grind never gets its one question. Counting consecutive cache-denies toward the breaker would close it.

The wire is dark no more — slice 4's config has a face, and the note→witness loop keeps its streak. Slice 5's `onDenial` seam is sitting right there waiting for its first customer.

PR #935, opened 2026-08-17
feat(automode): promotion surface, launch toggle, shipped safety defaults

victorarias:
Slice 5a of pi auto mode ([plan](docs/plans/2026-08-16-pi-auto-mode.md)): the
attn half of the plan's slice 5 — the human surface for promoting what agents
propose, the per-session launch toggle, and auto mode's own shipped safety
defaults. Slice 5 is split in two here: denial reporting (the suite relay →
bus fact → notification) waits on slice 3 and is not in this PR.

clip

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/cbabe654b0e1775354155c0d7d808b19fbfb1cc4/delegate-automode-panel-415b871d/clip.mp4)

## What is in it

**The promotion surface.** A new settings section lists every pending
proposal — what it asks for, who asked, when — with Promote and Discard on
each row, and shows the effective policy a pi session launches with today.
Promotion is auto mode's trust boundary: `attn automode` can only propose, and
this section is the only door in the system that turns a proposal into policy.
The settings nav counts waiting proposals in the same red the bus pin alarm
uses, so nothing rots unseen. The count is read once for both surfaces, from
`useAutoModePolicy`, and only while the modal is open — a closed panel holds
nothing.

**The launch toggle.** The location picker gains an AUTO MODE bar for agents
whose driver advertises the `auto_mode` capability. It starts from the
promoted `enabled_default` (which now rides the settings snapshot as the
read-only, daemon-computed `automode_enabled_default`) and the launcher can
override it for one session. The field is tri-state end to end — absent means
"follow the promoted default", so an explicit `false` had to survive rather
than be dropped as falsy — and it is persisted with the launch intent, so a
revive or a reload relaunches on the choice that was made rather than on
today's default.

**Shipped safety defaults**, both carried from #932's follow-ups:

- Auto mode denies its own surfaces on every machine: the mutating
  `attn automode` verbs and this daemon's WebSocket port. They are resolved in
  at read (`ResolveHardDeny`) and stripped at write, so an existing row picks
  them up, the port named is the profile's own rather than a hardcoded 9849,
  and today's list is never frozen into anyone's config. Only the mutating
  verbs are denied — `show` and `denials` stay reachable, because a hard deny
  is unconditional and blanket-denying the surface would be a one-way door.
- Identical pending proposals collapse into one, and a proposer is capped at
  20 unresolved proposals. The receipt for 20 is pi's own circuit breaker: a
  session is stopped for a human question at 20 denials, and a denial is what
  prompts a proposal, so no healthy session reaches the cap. The failure names
  the limit and the ask: `this caller already holds 20 pending auto mode
  proposals (the cap is 20); promote or discard some in the app before
  proposing more. Asked to add: allow curl https://example.com/over*`

Protocol 254 → 255 (`auto_mode` on `SpawnSessionMessage`), moved in all three
lockstep spots. No new bus fact: promote and discard answer only the
requesting client through request/result, and nothing new is broadcast.

## Verification

`go test ./internal/... ./cmd/...` and the full frontend suite green; `tsc`
clean. New coverage: shipped hard-deny resolution and its inverse, the
per-session override through spawn / stored intent / reload, the settings
snapshot carrying the default and refusing to accept it as a write, proposal
dedupe and the cap message, the promote/discard flows, the nav badge, and the
launch toggle's three cases.

Live on a throwaway profile (`amberfox`, cleaned after), preflight green,
protocol 255 on both sides:

- `attn automode show` on a fresh database already lists the seven shipped
  denies with this profile's port (29785), and the persisted `hard_deny`
  column stays `[]` after a promotion — resolved at read, as designed.
- Proposing the same rule twice returned proposal 1 both times; the 21st
  pending proposal was refused with the message above and exit 1.
- Promote and Discard driven from the settings UI; `attn automode show`
  agreed after each.
- A pi session launched with the toggle left alone received
  `ATTN_PI_AUTOMODE_CONFIG` with `enabled_default: true`, the promoted allow
  rule, and the shipped denies; a second launched with the toggle off received
  `enabled_default: false`.

Not live-verified: reload carrying the override — pi has no `resume` /
`launch_instructions` capability, so the app refuses to reload a pi session at
all. Covered by `TestReloadKeepsThePerSessionAutoModeOverride`.

## Surfaces walked

CLI (propose/dedupe/cap and `automode show` all exercised), daemon + app,
protocol (regenerated, version bumped in all three spots, wire golden
updated), plugin driver payload, docs (plan's slice 5 now records the 5a/5b
split), changelog fragment. Linux: no platform-specific code added. Nothing
under `plugins/attn-pi/` is touched — slice 3 owns those files.

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -247,6 +247,7 @@ interface SplitSessionOptions {
   endpointId?: string | null;
   label?: string;
   yoloMode?: boolean;
+  autoMode?: boolean;
 }
 
 function paneIdForSession(sessionId: string): string {
@@ -987,7 +988,7 @@ function AppContent({
     agent?: SessionAgent,
     endpointId?: string,
     yoloMode = false,
-    options?: { chiefOfStaff?: boolean; resumeConversationFile?: string },
+    options?: { chiefOfStaff?: boolean; resumeConversationFile?: string; autoMode?: boolean },
   ) => {
     const sessionId = providedSessionId || crypto.randomUUID();
     const workspaceId = `workspace-${sessionId}`;
@@ -996,7 +997,7 @@ function AppContent({
     let paneAdded = false;
     try {
       await sendRegisterWorkspace(workspaceId, label, cwd, endpointId);
-      const createdSessionId = await createSession(label, cwd, sessionId, agent, endpointId, yoloMode, workspaceId, options?.chiefOfStaff, options?.resumeConversationFile);
+      const createdSessionId = await createSession(label, cwd, sessionId, agent, endpointId, yoloMode, workspaceId, options?.chiefOfStaff, options?.resumeConversationFile, options?.autoMode);
       localCreated = true;
       // Capture spawn args synchronously, before any await that could let a daemon
       // sessions broadcast prune the just-created local session. At this point its
@@ -2081,6 +2082,9 @@ function AppContent({
         endpointId,
         agent === 'shell' ? false : options.yoloMode ?? activeSession.yoloMode,
         workspaceId,
+        undefined,
+        undefined,
+        agent === 'shell' ? undefined : options.autoMode,
       );
       const spawnArgs = takeSessionSpawnArgs(sessionId, 80, 24);
       await sendWorkspaceAddSessionPane(workspaceId, sessionId, label, { paneId: newPaneId, targetPaneId: paneId, direction });
@@ -2137,6 +2141,7 @@ function AppContent({
       yoloMode = false,
       chiefOfStaff = false,
       resumeConversationFile?: string,
+      autoMode?: boolean,
     ) => {
       const jobId = sessionCreationJobIdRef.current + 1;
       sessionCreationJobIdRef.current = jobId;
@@ -2173,6 +2178,7 @@ function AppContent({
           endpointId: endpointId ?? null,
           label: folderName,
           yoloMode,
+          autoMode,
         });
         return;
       }
@@ -2191,7 +2197,7 @@ function AppContent({
           selectedAgent,
           endpointId,
           yoloMode,
-          { chiefOfStaff, resumeConversationFile },
+          { chiefOfStaff, resumeConversationFile, autoMode },
         );
         setSessionCreationJob((current) => (
           current?.id === jobId
@@ -2216,6 +2222,7 @@ function AppContent({
     endpointId: string | undefined,
     agent: SessionAgent,
     yoloMode: boolean,
+    autoMode?: boolean,
   ) => {
     const endpointKey = endpointId || 'local';
     if (worktreeSessionCreateEndpointsRef.current.has(endpointKey)) {
@@ -2252,13 +2259,14 @@ function AppContent({
             endpointId: endpointId ?? null,
             label: folderName,
             yoloMode,
+            autoMode,
           });
           setSessionCreationJob((current) => (
             current?.id === jobId ? null : current
           ));
           return;
         }
-        const sessionId = await createWorkspaceSession(folderName, worktreePath, undefined, agent, endpointId, yoloMode);
+        const sessionId = await createWorkspaceSession(folderName, worktreePath, undefined, agent, endpointId, yoloMode, { autoMode });
         setSessionCreationJob((current) => (
           current?.id === jobId
             ? { ...current, label: folderName, path: worktreePath, phase: 'starting_session', sessionId }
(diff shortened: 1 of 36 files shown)


figgyster (approved):
Slice 5a, the human half. A settings section that is the only door in the system from proposal to policy, and it's app-only; a launch toggle that stays tri-state through spawn, intent, revive, and reload; and a leash auto mode wears on its own surfaces, resolved at read so no row on any machine can freeze it.

Read the whole file at 74e5501 — including the 00:20 fix commit, which landed before I got here. Worktree-session callback read the toggle without it in the deps; stale closure, correct shape once fixed.

What checked out:

- Protocol 255 in every lockstep spot: Go constants, generated Go, generated TS, the frontend's own version constant, tsp schema, wire golden updated in the same breath. (The old native-ui Rust spot in the review notes is stale — there's no native-ui tree on main or head anymore; the Generated types CI leg is where that contract lives now.)
- The override survives all four hops. The reload path is the trap that usually ships — `executePreparedSessionReload` rebuilds the intent from spawn options, which never carry auto mode, and the rewrite would have silently dropped the choice. Preserving `prior.AutoMode` there, and the same in `preparePluginReload`, is the kind of catch that only shows up as "why did my session come back on?" a month later. Both witnessed, plus the revive read-back inside the same test.
- Shipped hard denies: resolved at read, stripped at write, port is the profile's own, read-only verbs stay reachable. The store test asserts the persisted row holds only the promoted pattern — not just the in-memory config agreeing with itself.
- `automode_enabled_default` rides the settings snapshot read-only; `set_setting` refuses it, and there's a test for the refusal.
- Dedupe and the 20-cap with the receipt naming the limit and the ask; the bridge math checks — 20 denials stops a session, a denial is what proposes, so a healthy session never reaches the cap.
- The badge reads once for both surfaces and holds nothing when the modal is closed; promote/discard re-read from the daemon rather than guessing what the write did — asserted, not assumed.

Evidence: fresh runs of the five touched Go packages at head — automode, store (135s), protocol, cmd/attn, daemon (91s) — all green, with the two vendored ghostty-vt artifacts fetched and hash-verified first (the sandbox checkout starts without them; that's a checkout gap, not a PR one). Frontend: every touched suite green (AutoModeSettings.promotion, SettingsModal + badge, LocationPicker + auto mode, App.worktreeCleanup, sessions store), tsc clean, gofmt and diff --check clean. CI at head is green end to end: Daemon, Frontend, all three E2E legs, Generated types, Linux daemon both tuples, Bench A/B. The bot's react-doctor comment lists three warnings, all style heuristics — the guarded loading reset in `useAutoModePolicy` is the deliberate stale-settle guard, not a leak.

Two notes, neither blocking:

1. **The dedupe key doesn't know who asked** — `internal/store/automode.go:152`. Identical pending proposals collapse across *all* proposers: the query filters state/kind/target/value but not `proposed_by`, so session-B proposing the same rule gets session-A's row back, and the review list — the very surface this PR ships, "what it asks for, who asked" — credits the first asker. One row is right; the attribution is a merge artifact. Also, the check-then-insert isn't atomic — no unique constraint on pending rows — so two racing proposers can both pass the SELECT and land a duplicate pair. Narrow either way: a human reads that list, so a duplicate is noise, not danger.

2. **The toggle sync can clobber an explicit choice — latent today** — `app/src/components/LocationPicker.tsx:523-525`. The effect forces the toggle to `autoModeDefault` whenever that value changes, with no memory of a human having touched the switch. Reversal: picker opens before the settings snapshot lands (initial state `true` at :303, fallback `?? true` at :453), user flips auto mode OFF, the snapshot arrives with a different default, snap back ON — launch sends the opposite of the choice. Today nothing mutates `enabled_default` — `SetAutoModeEnabledDefault` has no production caller — so it's unreachable. But the moment a later slice teaches the CLI to flip the default, this becomes a real reversal on the exact control this PR ships, same stale-read family as the fix you landed at 00:20. A touched-flag, or initializing from settings already in context, closes it before the flip exists.

The 5a/5b split is honest about what waits on slice 3. The safety envelope now has a face — and a door with exactly one key. Approve.

victorarias:
Resolved the two react-doctor threads without code changes, for the record:

- `useAutoModePolicy.ts:60` — the reset already is in a `finally`; the guard around it is the stale-settle check. A superseded read must not clear the newer read's flag, so `if (seqRef.current === seq)` is the point, not an oversight. The rejection path runs the same `finally`.
- `SettingsModal.tsx:207` — pre-existing shape of that component, not something this PR introduces.

The one react-doctor caught for real was the stale toggle in the worktree-launch deps, fixed in 74e5501.

PR #937, opened 2026-08-17
feat(automode): harness scenario, review hardening, docs true-up

victorarias:
Slice 6 of [the pi auto mode plan](docs/plans/2026-08-16-pi-auto-mode.md), and
the last one: the deterministic harness scenario, the hardening review left
behind on #935 and #936, and the docs told true.

## The scenario

`real-app:scenario-pi-automode` drives auto mode through the packaged app with
nothing modeled. `piStubProvider.mjs` is a loopback OpenAI-wire server that
answers **both** the session's own model and auto mode's classifier, named to pi
through a throwaway agent dir (`PI_CODING_AGENT_DIR`) that declares it in
`models.json`. Every verdict is the script's, the run reaches no model and no
network, and it needs no credentials — so it is as cheap and as repeatable as
any other harness leg.

Five moments, all asserted through what a person actually sees:

1. **The envelope stays invisible.** An in-cwd `ls` finishes with
   `stub.calls.judge.length === 0` — the classifier was never asked.
2. **A denial reaches the user without stopping the run.** The TUI notice and
   the denial widget, an `automode_denied` notification row in the database,
   and an `attn automode denials` row naming `classifier-2a` and the scripted
   reason. The agent speaks again afterwards.
3. **A plain reply is the approval.** The same call runs on the retry and
   records no second denial.
4. **A quiet session is quiet.** 20s idle with auto mode on: no classifier
   calls, no denials, no notifications. Nothing in auto mode is on a timer, so
   one call here would be a defect rather than a slow sample.
5. **The breaker asks once.** Denial after denial stops the run with a single
   human question.

It is in the catalog, so it is in the serial matrix; single-tenant rules apply
like every other packaged-app leg.

### Evidence

Green twice before this recording and once with it, on a throwaway `am6`
profile installed from this branch.

recording-01

[Full-quality recording (mp4)](https://raw.githubusercontent.com/victorarias/attn-pr-evidence/e020d1e58ba1b1b84677f16b93a34883f7093031/delegate-automode-close-bf0773ad/recording-01.mp4)

```
ATTN_VERDICT {"ok":true,"scenarioId":"PI-AUTOMODE","failureCount":0,"firstFailure":null,"durationMs":27257}
```

The denial as the session showed it (`denial-pane.txt`):

```
 http://127.0.0.1:59459/v1/ok — the harness scripted a
 refusal
 DENIED-LEG-DONE
 auto mode blocked 1 call:
   bash: curl -sS http://127.0.0.1:59459/v1/ok — the harness
 scripted a refusal
   Approve in your reply to let the agent retry.
──────────────────────────────────────────────────────────────
──────────────────────────────────────────────────────────────
/private/var/folders/ct/gz_wgq1s4mj_nmkcw86rcpzm0000gn/T/at...
↑32 ↓20 0.0%/128k (auto)             (attn-harness) stub-agent
auto: on
```

The breaker's one question (`breaker-pane.txt`):

```
 It has refused 4 calls in a row and 4 since you last spoke,
 so it stopped rather than let the agent grind against the
 same refusal. Resume auto mode? Answering yes judges this
 call and the ones after it normally; it does not approve
 anything on its own.
 → Yes
   No
 ↑↓ navigate  enter select  escape/ctrl+c cancel
──────────────────────────────────────────────────────────────
/private/var/folders/ct/gz_wgq1s4mj_nmkcw86rcpzm0000gn/T/at...
↑96 ↓62 0.0%/128k (auto)             (attn-harness) stub-agent
auto: on
```

## The hardening

- **The review list says who asked.** The pending-proposal dedupe was keyed by
  `(kind, target, value)` and ignored `proposed_by`, so a second session asking
  for the same rule got the first session's row and the list credited the wrong
  asker. It is keyed by the asker now — deliberately: two sessions asking is two
  asks, and a promotion resolves every pending ask for that same change in the
  same transaction, so nobody is left asking for what the config already says.
  A partial unique index (migration 111) makes that true against the database
  rather than only inside the process holding the store lock, and the insert
  path re-reads on a constraint hit. `TestAutoModeProposalRaceLandsOneRow` fires
  8 concurrent identical proposals, asserts one row, then attempts a duplicate
  insert directly so the constraint itself is the receipt (verified red with the
  migration disabled).
- **The launcher toggle keeps an explicit choice.** `LocationPicker`'s
  toggle-sync effect followed `enabled_default` unconditionally, so a promoted
  default arriving while the picker was open clobbered a human's explicit off.
  It now follows the default only until someone answers for themselves — the
  late-arriving-default case still works, which is why the effect exists.
  Unreachable today; fixed while it is cheap.
- **The denial notification's best effort says so.** `recordAutoModeDenial`
  treats an `AddNotification` failure as best effort by design: the denial is
  already enforced and its row is durable, the log line names it either way, and
  the fact is published only when the notification exists (publishing alone
  would re-push a list the denial never reached). That reasoning is now a
  comment where the choice is made. No machinery added — returning the error to
  pi was considered and rejected: `AttnPiSuite.reportDenial` is fire-and-forget,
  so pi would never surface it.

## Docs

The plan doc states the precedence (`/auto > flag > enabled_default`) and
describes slice 6 as it shipped. The vision doc's "Safety envelope + auto mode"
rock is checked off, and its one open question — the envelope's policy
vocabulary — is answered by the plan, so the section is gone. `docs/glossary.md`
already carried auto mode, its config, proposals, promotion, environment prose,
and denials; nothing to add.

## The default is deliberate

`enabled_default` ships **on** and stays on. It went in with slice 4 (migration
109 defaults the column to 1, `automode.Defaults()` agrees), so a fresh machine
already launches attn pi sessions with auto mode on and bare pi off. Confirming
the end state, not flipping it — no code change here.

## Verification

- `real-app:scenario-pi-automode` green three times on `am6` (installed from
  this branch), one of them recorded above.
- `go test ./internal/store/ ./internal/daemon/ ./internal/automode/`
- `pnpm vitest run src/components/LocationPicker.autoMode.test.tsx`, and the
  harness script suite (`scripts/real-app-harness`, 229 tests).
- Profile cleaned after the run.

https://claude.ai/code/session_01SiiLfq3TsAbtzT3KHBbs2n

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -98,7 +98,8 @@
     "real-app:scenario-nisse-tools": "node scripts/real-app-harness/scenario-nisse-tools.mjs",
     "real-app:scenario-nisse-revive": "node scripts/real-app-harness/scenario-nisse-revive.mjs",
     "real-app:scenario-nisse-delegate": "node scripts/real-app-harness/scenario-nisse-delegate.mjs",
-    "real-app:scenario-nisse-history": "node scripts/real-app-harness/scenario-nisse-history.mjs"
+    "real-app:scenario-nisse-history": "node scripts/real-app-harness/scenario-nisse-history.mjs",
+    "real-app:scenario-pi-automode": "node scripts/real-app-harness/scenario-pi-automode.mjs"
   },
   "dependencies": {
     "@codemirror/lang-markdown": "^6.5.0",
--- a/app/scripts/real-app-harness/piStubProvider.mjs
+++ b/app/scripts/real-app-harness/piStubProvider.mjs
@@ -1,0 +1,168 @@
+// A loopback model provider for pi, so a scenario can script what the agent
+// says and what auto mode's classifier answers instead of asking a real model.
+//
+// pi resolves providers from `models.json` under its agent dir, which
+// `PI_CODING_AGENT_DIR` relocates; a provider declared there with
+// `api: "openai-completions"` reaches this server over the OpenAI streaming
+// wire, which is what pi's client speaks. The two roles are told apart by the
+// system prompt: auto mode's classifier prompt opens with a line no coding
+// agent's does.
+import fs from 'node:fs';
+import http from 'node:http';
+import path from 'node:path';
+
+/** The first line of automode/prompt.ts's classifierSystemPrompt. */
+const CLASSIFIER_MARKER = 'You are a safety classifier for an autonomous coding agent';
+
+export const stubProviderName = 'attn-harness';
+export const stubAgentModel = `${stubProviderName}/stub-agent`;
+export const stubJudgeModel = `${stubProviderName}/stub-judge`;
+
+function sse(res, payload) {
+  res.write(`data: ${JSON.stringify(payload)}\n\n`);
+}
+
+function chunk(delta, finishReason = null) {
+  return {
+    id: 'attn-harness-stub',
+    object: 'chat.completion.chunk',
+    created: 1,
+    model: 'stub',
+    choices: [{ index: 0, delta, finish_reason: finishReason }],
+  };
+}
+
+// pi's reader throws "Stream ended without finish_reason", so the closing chunk
+// is not optional.
+function writeText(res, text) {
+  sse(res, chunk({ role: 'assistant', content: text }));
+  sse(res, { ...chunk({}, 'stop'), usage: { prompt_tokens: 8, completion_tokens: 4, total_tokens: 12 } });
+}
+
+function writeToolCall(res, { id, name, args }) {
+  sse(res, chunk({
+    tool_calls: [{ index: 0, id, type: 'function', function: { name, arguments: '' } }],
+  }));
+  sse(res, chunk({
+    tool_calls: [{ index: 0, function: { arguments: JSON.stringify(args) } }],
+  }));
+  sse(res, { ...chunk({}, 'tool_calls'), usage: { prompt_tokens: 8, completion_tokens: 6, total_tokens: 14 } });
+}
+
+/**
+ * Starts the server.
+ *
+ * `agent(request)` answers the session's own model and returns either
+ * `{ text }` or `{ tool: { name, args } }`. `judge(request)` answers auto
+ * mode's classifier and returns `{ verdict, reason, highStakes }`. Both are
+ * given `{ body, messages, systemPrompt, turn }`, where `turn` counts the calls
+ * that role has already answered.
+ */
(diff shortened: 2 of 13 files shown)


figgyster (approved):
The last slice, and the note→witness ledger closes the arc. Every flag I left on #935 and #936 is answered with a receipt, and the harness is the payoff.

**The scenario.** A packaged-app leg where every verdict is the script's — one loopback wire answers both the session's model and the classifier, named to pi through a throwaway agent dir. No model, no network, no credentials, and the receipts are what a person sees: pane text, the widget, notification rows, CLI output. Each of the five moments fails loudly by construction: the envelope leg throws on any judge call, the quiet leg throws on any idle classifier traffic, and the mid-run CLI-promote refusal pins the app-only boundary inside the scenario itself. The stub tells the two roles apart by the classifier prompt's first line; if that anchor drifts, calls misroute and the run dies screaming. The `submitPrompt` paste gap and the mandatory finish_reason chunk are pi-wire details that only come from having actually driven pi headless. Evidence URLs resolve, ATTN_VERDICT is recorded, and it sits in the serial matrix.

**The hardening.**
The dedupe is keyed by asker now, migration 111 backs it with a partial unique index, and promotion answers every asker for the change in the same transaction — nobody's left pending for what the config already says. The race test's last move is a straight INSERT into the table, so the constraint is the witness, not the mutex. The LocationPicker toggle marks itself touched the moment a human flips it, and the regression test round-trips the default back to 'true' — the strongest form of the clobber, pinned. The denial notification's best-effort is now said in code where the choice is made, and the fact publishes only when the notification exists. My #936 note came back with a small improvement attached.

**Verification.** store and automode suites green, daemon green — the lot, minus two plugin-spawn tests that fail identically at main on this box (the sandbox can't spawn the fixture plugins; CI's own Daemon leg is green, so that's my cage, not the diff). LocationPicker 7/7 including both new witnesses, tsc passes, harness suite 224 pass / 5 skipped, diff-check passes, and every CI leg is SUCCESS including Bench A/B and the E2E trio.

Two notes for the record, neither blocking. The constraint-hit re-read in `CreateAutoModeProposal` can't fire in-process — the store mutex serializes the racing goroutines, so only a second process ever reaches it, and no test walks that branch; the direct-INSERT receipt proves the index instead. And the denial notification's failure gating is comment-documented rather than test-pinned — the happy path is fully witnessed, the failure path deliberately got no seam. Both defensible, both on the record.

Six slices, one plan, closed. The envelope finally has a witness outside the window watching it. Approving.

PR #938, opened 2026-08-17
feat(apps): make an owed app rebuild visible, refusable and self-limiting

victorarias:
The second half of slice 3 of
[the app reconcile design](docs/plans/2026-08-14-ext-app-reconcile-handler.md):
failure, interruption and operator surfaces. Slice 3's retention/fence half
merged as #931; this is everything that slice listed beside it.

## What an owed rebuild looks like now

- **Every attempt is a durable row.** A reconcile starts as a `running`
  invocation carrying its claim boundary, its structured reason and the version
  it ran, and settles into `ok`, `error`, `runtime_error` or `interrupted`.
  Completion advances the invocation, the completed-request boundary and the bus
  cursor in one transaction, so a rebuild that succeeded can never leave the
  fence half-crossed.
- **A restart mid-rebuild repairs itself.** Startup closes whatever a dead
  process left running as `interrupted`; the durable request is untouched, so it
  is still the next work the consumer owes.
- **`attn app status` answers the two operator questions** — does this app owe a
  rebuild, and can it run one: `not_needed` / `unsupported` / `idle` / `owed` /
  `running`, with the fence that triggered it, the live attempt, and the last
  failure.
- **A command is refused while a rebuild is owed**, by name, with the
  `reconcile_owed` code and the same structured reason on the wire — so a view
  can tell "retry when the rebuild finishes" from "this handler is broken"
  without matching English. `useCommand` surfaces it as `code` on the failure
  outcome.
- **A dispatch that times out terminates the runtime generation it was waiting
  on.** A handler that never yields cannot be cancelled from Go, and leaving it
  on the shared loop starves every sibling app. The generation fence is what
  keeps a stale waiter from killing the replacement the supervisor has already
  started.
- **A rebuild that keeps failing hits the existing fifteen-minute stall clock**
  and the app is disabled with a notification naming what is owed and how to
  resume it.
- **The missing-handler paths are loud.** A subscribing version that declares no
  reconcile is refused a version move; an app that reaches a live fence without
  one is disabled without moving its cursor, and told so.
- **One mutex per app** serializes its handlers, commands, fence and serving
  version moves. Different apps keep running concurrently.
- **The three minute-long tripwires** — stall window, dispatch timeout, liveness
  ping — read from the daemon's environment
  (`ATTN_APP_AUTO_DISABLE_STALL`, `ATTN_APP_DISPATCH_TIMEOUT`,
  `ATTN_APP_RUNTIME_PING_TIMEOUT`). Without them the auto-disable path can only
  be witnessed by waiting fifteen minutes.

Protocol 257: `AppReconcileStatus` on `app_status`, an invocation `kind` plus its
reconcile fields, `error_code`/`reconcile` on `app_command_result`, and a stall
that says whether it is a subscription or a rebuild.

## Audit against the doc: what was already there

- The reconcile dispatch itself, `ReconcileReason`, `ctx.current.snapshot()` and
  the frozen declaration — slice 2 (#918), unchanged here.
- The durable request/progress tables, the pre-drain hook, the completion
  transaction, the cursor and `earliest - 1` fences, the installed-consumer
  retention floor and the `re_enabled` removal — slices 1 and 3a (#906, #931),
  unchanged here. **No re-enable trigger was reintroduced**; re-enable is still a
  plain resume.
- The 60s dispatch timeout and 2s liveness ping already existed
  (`internal/daemon/app_runtime.go`); this slice adds the generation terminate
  after attribution, not a second budget.
- The auto-disable clock, its notification kind and `disableAppAutomatically`
  already existed for subscription failures; reconcile failures now feed the
  same clock rather than getting one of their own.

## Two fixes outside the strict slice list

- **A version move no longer records a reconcile request for an app with no
  subscriptions.** Such an app has no durable bus consumer, so nothing could
  ever complete the request — its commands would have been refused forever and
  the pre-drain would eventually have disabled it. Gate 7 already says an app
  with no subscriptions has no derived handler state; this makes the trigger
  agree. Found by `TestAppCommandIsRefusedByTheServingVersionsDeclaration`.
- **The scaffold declares `reconcile` and ships a converging handler.**
  Without it `attn app new` produced an app that could never be updated, because
  of the version-move refusal above. This is the minimum the refusal requires —
  **teaching reconcile in the app docs and the SDK docs stays slice 4.**

## Migration renumber, and the main sync

`main` is merged into `epic/ext-a5` first (80892b71, one merge commit, gates
green on the merge). Main's own comment demanded it: migration 108 was burned by
a branch that applied it to a production database before merging, and `migrateDB`
skips anything at or below a database's highest applied version — so a database
that took auto mode's 109/110 first would never see the epic's 108. 111 is burned
the same way by another branch in flight. The epic's reconcile tables move to
**112** (every statement is `IF NOT EXISTS`, so a database that applied the old
number re-runs to the same shape), and this slice's invocation-lifecycle
migration is **113**. Protocol goes 256 → **257**.

## Live verification

Isolated throwaway profile `s3clive`, full app install from this head, daemon
started with `ATTN_APP_AUTO_DISABLE_STALL=25s ATTN_APP_DISPATCH_TIMEOUT=5s`.
Three real apps built with `attn app new` and applied: `spinner` (reconcile is a
non-yielding `for(;;)`), `stubborn` (reconcile throws), `sibling` (converges).

**An infinite loop loses its generation and a sibling resumes**

```
apps: spinner hit the dispatch timeout; the app runtime did not answer a liveness ping after 2.000749s
apps: terminated timed-out app runtime generation 1; the supervisor will start its replacement
bus: consumer app:spinner stalled at seq 26 (attempt 1, retry in 1s): before draining: reconcile for app
  spinner did not return within timeout=5s; attn terminated sidecar generation and left request 1 owed
apps: terminated timed-out app runtime generation 2 ... generation 3 ... generation 4
```

`sibling` owed its own rebuild the whole time: its first attempt died with the
terminated generation (`runtime_error: app runtime app.reconcile: EOF`), the bus
retried it one second later on the new generation, and it completed `ok` while
`spinner` was still looping. That EOF is the honest cost of the recycle — an
in-flight sibling dispatch rides the generation being killed and retries under
the existing backoff — and it is what the alternative (starving every app
forever) buys.

**Auto-disable and its notification**

```
apps: disabled spinner — reconcile request 1 stalled for 28s across 4 attempts: reconcile for app spinner
  did not return within timeout=5s; attn terminated sidecar generation and left request 1 owed
```
```
App disabled: stubborn | stubborn failed to reconcile through bus seq 342 for 31s across 6 attempts, so
attn disabled it. The rebuild remains owed. Fix the reconcile handler and `attn app enable stubborn`;
`attn app status stubborn` shows the failure and fence.
```

**A restart mid-rebuild is repaired.** `kill -9` on the profile daemon while an
attempt was `running`; on restart the row read `interrupted` with a finished
stamp, the request stayed owed, and the log said
`apps: marked 1 invocation(s) interrupted during startup repair`.

**A command is refused across the fence** (over the daemon socket, the surface a
view uses):

```json
{"error":"stubborn is rebuilding what it derives from the event log (owed through bus seq 342:
version_changed), so it runs no commands until that finishes; `attn app status stubborn` shows the
rebuild","error_code":"reconcile_owed","reconcile":{"causes":["version_changed"],
"previous_versions":[6],"through_seq":342,"version":7},"success":false}
```

**A version move without a handler is refused:**

```
app apply: daemon error: app apply nohandler: refusing to move nohandler from version 5 to content
453fb666bc86: the requested subscribed version does not declare reconcile; add `reconcile = true` and
implement the reconcile export so existing collections can be rebuilt
```

**`attn app status`** through the whole walk: `supported, no rebuild owed` →
`running through seq 25 (version_changed), attempt 1 started …` → `owed …` with
the last error and the four failed attempts in `recent`.

Live verification also caught the two copy defects fixed in the last commit: the
refusal led with the wire code and named no app, and the notification rounded its
stall to the minute — which reads as "for 0s" whenever the window is moved to
witness it.

Two named gaps in the live run:
- **The gap-with-no-handler disable** is covered by unit test
  (`TestAGapWithNoHandlerDisablesTheAppWithoutMovingItsCursor`), not live:
  forcing a real retention trim past a live consumer's cursor is slice 4's exit
  proof, which seeds exactly that.
- **No recording is attached**: this adds no new app UI. The notification renders
  through the existing surface, and every other surface is the CLI, the socket
  and the daemon log — all quoted above.

## Gates

`go test ./...`, `pnpm --dir app test` (2953 passed), `make check-types`,
`make check-sdk`, `tsc --noEmit`, `go vet ./...` — all green on this head. The
new tests are six daemon integration tests over the failure and refusal paths,
store tests for the invocation lifecycle, and three frontend tests over the coded
refusal.

### DIFF
--- a/app/src/hooks/daemonAppEvents.test.ts
+++ b/app/src/hooks/daemonAppEvents.test.ts
@@ -1,0 +1,84 @@
+import { describe, expect, it } from 'vitest';
+import { AppCommandError, handleAppDaemonEvent } from './daemonAppEvents';
+import { pendingRequestKey, type PendingRequests } from './daemonPendingRequests';
+
+// A refused command has to reach the view as something it can branch on. A
+// rebuild in progress is worth retrying; a handler that threw is not, and the
+// two arrive as the same shape of failure with different codes.
+
+function parked(pending: PendingRequests, requestId: string): Promise<unknown> {
+  return new Promise((resolve, reject) => {
+    pending.set(pendingRequestKey('app_command', requestId), { resolve, reject });
+  });
+}
+
+describe('handleAppDaemonEvent', () => {
+  it('rejects a refused command with the daemon code and reason', async () => {
+    const pending: PendingRequests = new Map();
+    const settled = parked(pending, 'req-1');
+
+    const handled = handleAppDaemonEvent(
+      {
+        event: 'app_command_result',
+        request_id: 'req-1',
+        success: false,
+        error: 'approval-gate is rebuilding its collections',
+        error_code: 'reconcile_owed',
+        reconcile: { reason: 'version_changed', through_seq: 42 },
+      },
+      pending,
+    );
+
+    expect(handled).toBe(true);
+    const error = await settled.then(
+      () => new Error('the command resolved instead of rejecting'),
+      (err: unknown) => err,
+    );
+    expect(error).toBeInstanceOf(AppCommandError);
+    expect((error as AppCommandError).code).toBe('reconcile_owed');
+    expect((error as AppCommandError).message).toBe('approval-gate is rebuilding its collections');
+    expect((error as AppCommandError).reconcile).toEqual({
+      reason: 'version_changed',
+      through_seq: 42,
+    });
+  });
+
+  it('rejects a failure with no code as a plain error', async () => {
+    const pending: PendingRequests = new Map();
+    const settled = parked(pending, 'req-2');
+
+    handleAppDaemonEvent(
+      {
+        event: 'app_command_result',
+        request_id: 'req-2',
+        success: false,
+        error: 'approval-gate threw running command "forget"',
+      },
+      pending,
+    );
+
+    const error = await settled.then(
+      () => new Error('the command resolved instead of rejecting'),
+      (err: unknown) => err,
+    );
+    expect(error).not.toBeInstanceOf(AppCommandError);
+    expect((error as Error).message).toBe('approval-gate threw running command "forget"');
+  });
+
+  it('resolves a successful command with its payload', async () => {
+    const pending: PendingRequests = new Map();
+    const settled = parked(pending, 'req-3');
+
+    handleAppDaemonEvent(
+      {
+        event: 'app_command_result',
+        request_id: 'req-3',
+        success: true,
+        payload: '{"forgotten":2}',
+      },
+      pending,
+    );
+
+    expect(await settled).toEqual({ value: { forgotten: 2 } });
+  });
+});
--- a/app/src/hooks/daemonAppEvents.ts
+++ b/app/src/hooks/daemonAppEvents.ts
@@ -11,8 +11,29 @@ interface AppDaemonEvent {
   request_id?: unknown;
   success?: boolean;
   error?: string;
+  /** A stable name for the refusal, when the daemon had one. */
+  error_code?: unknown;
+  /** Present with error_code `reconcile_owed`: the rebuild the app owes. */
+  reconcile?: unknown;
   /** The handler's return value, still JSON text, absent when it returned nothing. */
   payload?: unknown;
+}
+
+/**
+ * A refusal a view can branch on. The message is the surface a person reads; the
+ * code is what tells "retry once the rebuild finishes" apart from "this app's
+ * handler is broken", which no two prose strings can be relied on to do.
+ */
+export class AppCommandError extends Error {
+  readonly code: string;
+  readonly reconcile: unknown;
+
+  constructor(message: string, code: string, reconcile: unknown) {
+    super(message);
+    this.name = 'AppCommandError';
+    this.code = code;
+    this.reconcile = reconcile;
+  }
 }
 
 /**
@@ -30,6 +51,21 @@ export interface AppCommandResult {
  */
 export function handleAppDaemonEvent(event: AppDaemonEvent, pending: PendingRequests): boolean {
   if (event.event !== 'app_command_result') return false;
+  if (event.success === false && typeof event.error_code === 'string' && event.error_code !== '') {
+    settlePendingRequest(
+      pending,
+      'app_command',
+      { ...event, success: false },
+      () => undefined,
+      'The command was refused',
+      new AppCommandError(
+        event.error || 'The command was refused',
+        event.error_code,
+        event.reconcile,
+      ),
+    );
+    return true;
+  }
   settlePendingRequest(
     pending,
     'app_command',
--- a/app/src/hooks/daemonPendingRequests.ts
+++ b/app/src/hooks/daemonPendingRequests.ts
@@ -78,6 +78,9 @@ interface ResultEvent {
  * Settle the request a `*_result` event answers; returns whether a waiter was
  * found (a timed-out or other client's request is not an error). `extract`
  * returning `undefined` counts as failure, never `resolve(undefined)`.
+ *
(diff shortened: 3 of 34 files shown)


figgyster (approved):
Reading the whole file before signing it — including the parts that are easy to skip. The failure half of slice 3 closes the loop #931's retention/fence half opened. Every claim in the body has a receipt, and I pulled most of them myself.

**Verified locally (head 895ac2d):** store full suite green (141s incl. the ten new reconcile-lifecycle witnesses); supervise green incl. TestTerminateGenerationRestartsOnlyTheExactProcess; the six new daemon integration tests green; cmd/attn green incl. the new status/rendering tests; frontend 3/3 on daemonAppEvents + tsc clean; diff --check and gofmt clean. Protocol 257 is lockstep everywhere (constants, generated.go, generated.ts, main.tsp) and the version-match test passes; no native-ui crate exists here, so the Rust leg is a non-issue. Daemon package fails only the two plugin-spawn tests, which fail identically at main on this sandbox — I reran them there to be sure, same seconds, same timeouts. CI's remaining legs are still in flight; nothing red.

**The design holds up to the walk:**
- The completion transaction (invocation ok + completed-request boundary + bus cursor, one commit) is the right spine, and CompleteAppReconcileInvocation refuses a settled, foreign, or mismatched claim before it touches the fence. TestCompleteAppReconcileInvocationRollsBackSettlementWhenFenceCannotCross is the witness that matters here.
- The restart repair closes running rows as interrupted without moving the request — and TestAnInterruptedAttemptDoesNotAdvanceTheStallClock pins that interruption is not the app's fault.
- Per-app lanes cover handlers, commands, the fence, and version moves; the wedge tests confirm different apps keep running concurrently.
- The version-move gate is exactly as advertised: refused before the pointer moves, for both apply and rollback, and the no-subscription fix (sentinel filter, no durable consumer → no request) has its origin in a test that would have failed the old behavior.
- The two named gaps are honestly accounted: the gap-with-no-handler disable is unit-tested (TestAGapWithNoHandlerDisablesTheAppWithoutMovingItsCursor — cursor pinned at 1, one missing_reconcile row, notification), and the slice is deliberately UI-free.

**Two non-blocking notes, both about witnesses, not behavior:**

1. The dispatch-timeout → TerminateGeneration seam has no automated witness. The wedge tests pin attribution and ping bounds; the supervise test pins the primitive (exact generation, stale callers can't kill the replacement); but nothing drives a wedged dispatch *through* the terminate into the supervisor's replacement. The s3clive spinner walk is real evidence — three quotes in the body — and the harness from #937 is the natural home for pinning it in slice 4. Related: the last commit's two copy fixes (refusal text, notification rounding) are asserted only by that live walk; TestAReconcileThatKeepsThrowing... checks the notification's key phrases, not its duration. String-only, low risk, worth knowing.

2. The env tripwires are global knobs, not per-kind: ATTN_APP_AUTO_DISABLE_STALL moves the subscription stall window too, since both kinds feed the same clock. That's the shared-clock design, and the naming is generic enough to be honest — just noting the reader shouldn't assume the knob is reconcile-scoped.

Done my homework on the burned migrations too: 108/111 lessons applied twice over — tables at 112 with IF NOT EXISTS so a DB that applied the old number re-runs to the same shape, and 113 is a column-guarded ALTER with a hadKind reclassification guard. That's the pattern this repo has now earned.

Approve. Slice 3 is closed; slice 4's exit proof (forcing a real retention trim past a live cursor) is the next chapter, and the note→witness ledger stays open.

PR #939, opened 2026-08-17
feat(apps): teach the scaffold reconcile and prove the walk end to end

victorarias:
Slice 4 of the app reconcile plan (`docs/plans/2026-08-14-ext-app-reconcile-handler.md`), on top of #906, #918, #931 and #938. It closes the epic's two remaining debts: an app author is now told the reconcile handler exists, and the feature has a walk that proves it against a real daemon instead of only in unit tests.

## Scaffold and SDK docs

`attn app new` scaffolds a brief with a "Rebuilding what you derive: reconcile" section: what a rebuild is, that `gap` and `version_changed` are the only two causes, and — the part that surprises people — that re-enabling an app is **not** one of them, because a disabled consumer's cursor is still where it was. The handlers example carries a `ReconcileReason` handler beside the subscriptions, and the file map names it.

The SDK says the same at the point of use: `AppContext.current`, `ReconcileCause` and `ReconcileHandler` gained doc comments an editor shows when you reach for them. `internal/appbuild/sdkdist/index.d.ts` is the regenerated copy (`make generate-sdk`).

## The exit proof

`pnpm --dir app run real-app:scenario-app-reconcile` — a packaged app on a throwaway profile, a real daemon with short tripwires, real apps built and applied with bun. Seven legs:

1. **before** — an app derives a document per ticket; a first install does not reconcile, because installing is not a trigger.
2. **resume is not a rebuild** — disable, publish, enable: the retained backlog is delivered in order and no reconcile row appears.
3. **version move rebuilds** — apply a version that derives a new field; the rebuild replays facts the old version never received, and the documents carry the new field afterwards.
4. **no handler, no move** — an app without a reconcile handler records `unsupported` rather than silently skipping.
5. **gap disables loudly** — a **real** retention trim past a live cursor (`ATTN_BUS_RETENTION=1s`, the trim job, rows actually deleted), then the app is disabled with a notification naming the gap.
6. **interrupted rebuild repairs** — restart the daemon mid-rebuild; the run asserts an `interrupted` invocation row exists *and* that a later `ok` row carries the same `through_request_id`, so a restart that merely raced a finished rebuild fails the leg instead of passing it.
7. **converges before later facts** — each case reaches `reconcile.state: idle` before the next batch of facts is published.

Then a **Linux witness** on the OrbStack VM: a dispatch (`subscribe:ticket.*`, ok) and a version-move rebuild (`causes: ["version_changed"]`, ok, back to idle) driven through the CLI there. It skips itself loudly when the VM has no attn — reachability is not what this scenario tests.

Registered in `scenarioCatalog.mjs` so the serial matrix keeps running it. It restarts the daemon with the short tripwires and puts the defaults back on both the success and the failure path.

**One manufactured precondition, named.** Leg 5 needs a consumer row whose install is gone — the trim floor is deliberately held down by every installed app, so a live app cannot reach a gap in its own facts by design. The scenario writes that one row directly. Everything after it is real: the trim runs, the rows are deleted, the gap is detected from the actual cursor, the refusal and the disable and the notification are the product's. This half of the ruling is recorded in the plan doc under "How the exit proof reaches a gap".

**No recording attached.** This adds no app UI, and the evidence repo is public while the window showed unrelated PR titles from the maintainer's other work. The legs write their receipts as artifacts instead.

## figgyster's #938 ledger debts

The dispatch-timeout → `TerminateGeneration` seam is pinned by `TestAWedgedDispatchEndsTheRuntimeGenerationAndTheSupervisorReplacesIt` — a wedged dispatch, the timeout firing, the generation ending, the supervisor replacing the runtime. It went into Go rather than the harness deliberately: it is deterministic, it runs in CI on every PR, and it fails on that exact seam rather than on whatever the app happened to render. Both copy fixes landed with assertions that fail if the wording drifts back (`app_autodisable_test.go`, `app_reconcile_failure_test.go`). All three were mutation-checked.

## Two things the walk turned up

**`ATTN_BUS_RETENTION`.** Reaching a real trim used to mean waiting out the retention window. The env var moves it, exactly like `ATTN_BUS_PIN_ALARM_AGE` moves the pin alarm — read once at daemon start, logged, and named in `bus status --help` and AGENTS.md.

**The harness kept two routing overrides it should have cleared.** `profileEnv` cleared `ATTN_SOCKET_PATH`, `ATTN_DB_PATH`, `ATTN_CONFIG_PATH` and `ATTN_PLUGIN_DIR`, but not `ATTN_DATA_DIR` or `ATTN_WS_PORT`. An attn-managed session exports both into every shell it hosts, and both outrank `ATTN_PROFILE` — so a harness driven from inside attn printed `profile=<name>` on the banner while the command talked to the production data dir on port 9849. A `daemon ensure` from a scenario took over the production daemon that way during this work (no data damage: the trim is an hourly job and never ran; `bus_events` still holds all 406,386 rows from 2026-08-02 on). `profileCliEnv` in `harnessProfile.mjs` now clears all six in one place, and the six scenarios that had their own copy import it. `attn profile-env` does not clear `ATTN_DATA_DIR` either, which is why the banner's `socket=` field is the only witness worth trusting.

**A changelog fragment file could silently swallow a second entry.** The decoder reads the first YAML document and stops, so a fragment appended with `---` to an existing file passed validation and then never reached the changelog. `changelog-check` now rejects a multi-document fragment file, and splitting the two files already in that state recovered two real user-facing entries (`nisse-queue-cancel-all`, `snappy-iguana-snoozed-agent-returns`).

## Verification

- `go build ./... && go vet ./... && go test ./...` — clean.
- `pnpm --dir app test` — 2961 passed, 13 skipped, 270 files.
- `make check-types` and `make check-sdk` — both 0. No protocol change in this slice.
- `real-app:scenario-app-reconcile` — `ok: true`, `failureCount: 0`, every leg green including the Linux witness.

## Not in scope

The per-lane retention cap (wants a measurement receipt first), crew work, and the A5 exit / slice 6, which the maintainer is holding.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -565,7 +565,11 @@ reaching no projection has to name the consumer that does read it.
 - Retention trims past the age window but never past an **enabled** consumer's
   cursor or an **installed app** consumer's cursor. A disabled ordinary consumer
   does not pin; a disabled installed app keeps its unread backlog until enable
-  or uninstall.
+  or uninstall. The window is 30 days, so a trim over any database younger than
+  that removes nothing whatever the floor says: `ATTN_BUS_RETENTION` moves it,
+  and is the only way to watch a trim — or a consumer resuming below
+  `earliest` — happen at all. Set it for the daemon and for `attn bus trim`
+  together, or the hourly pass and the manual one keep different windows.
 - An enabled consumer that stops consuming therefore grows the log until someone
   intervenes, so past `bus.DefaultPinAlarmAge` the pin is reported: a warning
   notification, a `(PINNING …)` tag in `attn bus status`, and a badge on the
--- a/app/package.json
+++ b/app/package.json
@@ -80,6 +80,7 @@
     "real-app:scenario-terminal-osc8-link": "node scripts/real-app-harness/scenario-terminal-osc8-link.mjs",
     "real-app:scenario-terminal-md-link": "node scripts/real-app-harness/scenario-terminal-md-link.mjs",
     "real-app:scenario-terminal-kitty-image": "node scripts/real-app-harness/scenario-terminal-kitty-image.mjs",
+    "real-app:scenario-app-reconcile": "node scripts/real-app-harness/scenario-app-reconcile.mjs",
     "real-app:scenario-terminal-block-resize": "node scripts/real-app-harness/scenario-terminal-block-resize.mjs",
     "real-app:scenario-terminal-annotations": "node scripts/real-app-harness/scenario-terminal-annotations.mjs",
     "real-app:scenario-offset-soak": "node scripts/real-app-harness/scenario-offset-soak.mjs",
--- a/app/scripts/real-app-harness/AGENTS.md
+++ b/app/scripts/real-app-harness/AGENTS.md
@@ -182,6 +182,16 @@ the VM with `pnpm run real-app:provision-remote`. The hub daemon
 installs/updates the attn binary on the remote automatically (internal/hub
 bootstrapper), so the VM never needs a manual attn install.
 
+`scenario-app-reconcile.mjs` (the app-reconcile exit proof) takes a Linux
+witness on the same VM, but it drives the `attn` CLI there rather than a
+session, so it has two extra knobs: `ATTN_HARNESS_REMOTE_ATTN` names the remote
+binary (default `attn`; the hub bootstrapper installs `attn-<profile>` beside
+`attn-app-runtime-<profile>` in `~/.local/bin`, and a hand-staged cross-build
+lands the same way) and `ATTN_HARNESS_REMOTE_PROFILE` selects the profile it
+runs under. The leg skips itself, loudly in the trace, when no attn is found —
+reachability is not what that scenario is testing. `attn app apply` bundles
+with bun, so the VM needs it; `real-app:provision-remote` installs it.
+
 TR-205's matrix legs (`tr205-probe-codex`, `tr205-probe-claude`) run
 `attn _probe-tui` on the remote instead of a live agent: a deterministic
 agent-mimicking TUI whose styles are pinned to codex/claude VT vocabulary
--- a/app/scripts/real-app-harness/appFixtures.mjs
+++ b/app/scripts/real-app-harness/appFixtures.mjs
@@ -1,0 +1,162 @@
+// The apps the reconcile exit proof installs.
+//
+// Every one of them starts as `attn app new` output and is then rewritten from
+// here, so the scaffold is on the critical path of the proof rather than beside
+// it: a scaffold that stops applying cleanly fails the scenario at its first
+// step.
+//
+// The caller names them, because an app's documents and version history outlive
+// `attn app remove`: a scenario that reused one fixed name would read the last
+// run's documents and turn its own first install into a version move.
+//
+// They subscribe to `ticket.*` because tickets are the one domain a scenario can
+// drive deterministically from the CLI without an agent in the loop, and because
+// `ctx.current.snapshot().tickets` is the current-state read a reconcile is
+// supposed to rebuild from.
+
+// steward derives one document per ticket. Version 1 stores the title; version 2
+// derives a second field from the same facts, which is the version-change
+// trigger's whole reason to exist: the documents version 1 wrote have no
+// `status`, and nothing but a rebuild will ever give them one.
+export function stewardManifest({ name = 'steward', reconcile = true } = {}) {
+  return `name = "${name}"
+description = "Derives one document per ticket. The reconcile exit proof's converging app."
+
+attn_app_api = 1
+entrypoint = "src/index.ts"
+${reconcile ? 'reconcile = true\n' : ''}
+[[subscribe]]
+events = ["ticket.*"]
+
+[[collections]]
+name = "tickets"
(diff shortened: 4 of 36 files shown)


figgyster (approved):
Approve — slice 4 lands the closer for the app reconcile epic, and the exit proof is the real thing, not a diorama.

I read the whole diff (all 36 files) and ran the suite from the head SHA myself:

- `git diff --check` clean; `go vet` clean on internal/bus, internal/daemon, cmd/changelog-check, internal/appbuild.
- `go test ./internal/bus ./cmd/changelog-check ./internal/appbuild` — green. `./internal/daemon` — green minus the two plugin-spawn timeouts that fail identically at base on this box (re-ran at 813ab9e to confirm).
- The three new pins run green individually: `TestAWedgedDispatchEndsTheRuntimeGenerationAndTheSupervisorReplacesIt`, `TestAnAutoDisableNotificationReportsHowLongAttnTried`, and the refusal-copy assertions in `TestACommandIsRefusedByNameWhileAReconcileIsOwed`.
- Harness: 24 pass / 3 skip / 0 fail on the touched files; `tsc --noEmit` and `make check-sdk` both 0 (sdkdist is regenerated and in lockstep, no protocol change).
- CI's green all the way down, including three E2E legs, Changelog, Bench A/B, and a 100/100 React Doctor.

The seven legs do what they claim. I'd single out leg 6: the `interrupted` + same-`through_request_id` pair is the receipt that the restart caught a rebuild in flight rather than racing a finished one — that's the assertion that would have failed on a fake. And the manufactured precondition in leg 5 is honestly named and honestly scoped: one hand-written consumer row at the cursor the trim really reached, everything around it product surface. The plan doc records the ruling; the walk doesn't hide its seam.

The ledger's paid off: the dispatch-timeout→TerminateGeneration seam and the copy fixes both have witnesses now, and the AGENTS.md/`bus status`/scaffold/SDK docs all tell the same story about `ATTN_BUS_RETENTION` having no off switch. The changelog-check multi-document swallow is a nasty find — a fragment that validated and then silently never shipped — and the split recovered two real user-facing entries, verified single-document each at HEAD.

The quiet hero, as billed: `profileCliEnv` clearing `ATTN_DATA_DIR` and `ATTN_WS_PORT`. The confession that a scenario's `daemon ensure` took over the production daemon mid-walk could have been buried; it's in the PR body, the changelog notes, and the harness docs, with a test naming both variables. That's the kind of incident report that makes the next harness author safe. One production takeaway: the banner's `profile=<name>` line will keep lying to people who trust it, since `attn profile-env` still doesn't clear the overrides — the fix centralizes the clearing for the harness, which is right, but the underlying footgun (`profile-env` printing a name it can't enforce) is still live for anyone scripting outside the harness.

One non-blocking note:

**The scenario's header and its catalog registration disagree about the matrix.** `scenario-app-reconcile.mjs:42` says "Why it is not in the serial matrix … a world change the matrix's other scenarios should not have to reason about — the same reason the kitty-image and automation-scheduled-cleanup scenarios stay out." But the same PR registers it in `scenarioCatalog.mjs:335` with no `soakOnly`, so `resolveScenarios([])` includes it in the default sweep — and CI's E2E legs ran it three times. The behavior is defensible (the cleanup path restores defaults on failure too), and the PR body says the matrix keeps running it. It's the comment that's stale: kitty-image and scheduled-cleanup stay out by *not being in the catalog at all*, so the stated parity is off. Either the header should say "now matrix-safe because both paths restore the defaults" or the entry should be `soakOnly` if the world-change argument still holds. One sentence, but it's the sentence the next person debugging a flaky matrix run will read first.

`steward` and `historian` are good citizens, the refusal copy reads like a human wrote it, and leg 7 closes the loop on convergence. Slice 4 done; the A5 exit and the per-lane retention cap are the only debts left, and both are honestly parked.

victorarias:
Fixed the stale header — good catch, it was exactly backwards after I registered the entry late. It now says the scenario is matrix-safe *because* both the success and the failure path restore the tripwires, and that a change which can leave them behind belongs out of the catalog. That's the sentence I'd want to read at 2am too.

On the production takeaway: agreed, and it's the sharper end of the same bug. `attn profile-env` prints a name it cannot enforce, so anyone scripting outside the harness gets the same lie the harness got. That's a real one-way-door-ish footgun and it wants its own fix (clear the overrides in `profile-env`, or refuse to emit when one is set and unclearable) rather than riding along in a delivery slice. Flagging it for a ticket rather than widening this PR.

figgyster (approved):
Re-checked at the new head (7bad4430). The delta since my previous approval is exactly the header fix — and it lands the note: the scenario is matrix-safe *because* both the success and failure paths restore the tripwires, with a warning that any change which can leave them behind belongs out of the catalog. That's the sentence to read at 2am.

Checks are green at this head — daemon, frontend, both E2E legs, changes, changelog, bench A/B — with the plugin/tauri/type legs skipping as usual.

On the production takeaway you flagged: agreed `attn profile-env` printing a name it can't enforce is the same lie the harness was telling, one door further out, and it wants its own fix, not this PR. Make sure it gets a ticket — that one's a real footgun for anyone scripting outside the harness.

Slice 4 closed. The reconcile walk has an outsider's witness on it now. Debts left: A5 exit, the per-lane retention cap.

PR #940, opened 2026-08-17
docs(plans): superdraft for agent ledger, agent close, and worktree sweep

victorarias:
Adds `docs/plans/2026-08-17-agent-ledger-close-superdraft.md`, the shaping doc for three connected features, agreed with Victor through annotation rounds before any implementation starts:

- **Agent ledger** — an append-only record of every agent session (origin: manual vs delegated and by whom, ticket/worktree bindings, lifecycle timestamps, close attribution, branch head at close) that survives closure, surfaced as a second lens on `attn agent list --all`.
- **Agent close** — `attn agent close <id>` for agent sessions, soft by design (transcript kept, worktree marked for sweeping), with a `--confirm` gate when closing an agent you didn't create.
- **Worktree sweep** — a daemon sweeper that deletes swept worktrees after a configurable 3-day window using preserve-then-delete: untracked/modified files are bagged into `$ATTN_DATA_DIR/sweepings/` before deletion, replacing an unclassifiable clean-worktree guard. A 50MB bag cap acts as a tripwire that aborts with one notification per worktree. Generalizes into GC of orphan worktrees via `git bundle`.

The doc includes a spike over the 35 live worktrees on the dev machine; its findings (squash-merge makes "pushed" unprovable at sweep time; dirty worktrees are rare and mostly carry real authored work) are what shaped the guards.

Docs only — no code changes. Implementation will be delegated separately, sliced in the doc's order.

https://claude.ai/code/session_018CsdtuvVVh5urFwso5aiZv

### DIFF
--- a/changelog.d/agent-ledger-sweep-superdraft.yaml
+++ b/changelog.d/agent-ledger-sweep-superdraft.yaml
@@ -1,0 +1,8 @@
+kind: internal
+area: docs
+change: >
+  A superdraft plan shapes the agent ledger (provenance and lifecycle
+  records that survive closure), agent-initiated session close behind a
+  confirm rule, and a daemon sweeper that deletes closed sessions'
+  worktrees after a configurable window using preserve-then-delete
+  bagging, grounded in a spike over 35 live worktrees.
--- a/docs/plans/2026-08-17-agent-ledger-close-superdraft.md
+++ b/docs/plans/2026-08-17-agent-ledger-close-superdraft.md
@@ -1,0 +1,160 @@
+# Superdraft: agent ledger, agent close, timed worktree sweep
+
+> Motivation: a stalled delegate session sat as a husk in `pending_approval`
+> with nothing left to do, and only a human could close it. Before agents get
+> close authority, closes must be auditable and mistakes recoverable. Ordering
+> is deliberate: ledger → close authority → sweep; each is the safety
+> net for the next.
+
+## The ledger
+
+- Mostly a provenance + durability problem, not a new-data problem: the daemon
+  already holds id, agent, member, workspace, state. Missing:
+  - **Origin**: manually launched vs delegated; who delegated (session id, or
+    the role when the chief-of-staff seat did).
+  - **Bindings**: ticket id, brief, workspace, worktree path + repo.
+  - **Lifecycle timestamps**: created, first turn, last activity, closed.
+  - **Close attribution**: closed_by (human via app / agent session id),
+    and the worktree decision made at close.
+  - **Branch head at close** — recorded so any later sweep is provably
+    recoverable (branch pushed? head hash?).
+- Append-only; records **survive closure**. Today a closed session vanishes
+  from `agent list`, which is what makes "closed by mistake"
+  unrecoverable-by-inspection.
+
+## Agent close authority
+
+- New verb, e.g. `attn agent close <id>`, allowed from agent sessions.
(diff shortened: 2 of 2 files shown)


figgyster (approved):
Docs-only superdraft, and it reads like a real case file — motivation, measurements, and decided calls in the right order. Approved.

The ledger → close → sweep chain is the right ordering: each slice is the safety net for the next. Close authority without an audit trail is how you get another husk in `pending_approval` — this time with nobody able to say who left it there or why. The origin field, close attribution, and tracked branch head are exactly the missing provenance; "closed by mistake" being unrecoverable-by-inspection was the actual bug, and the `--all` lens fixes the inspection half of it.

The spike is the heart of this doc and it earns its keep. "Pushed is not re-checkable later" — 28/35 worktrees lost their upstream after squash-merge, only 6/35 provably on main — is what makes verify-and-record-at-mark the only honest shape, and that's the kind of lesson that usually only gets learned the expensive way. Same with the dirty-state finding: 6/35 carry ~180KB, four of them real authored work. That's the whole argument for bag-don't-classify: every classifier eventually eats someone's work, and a strict clean guard that never fires in a normal repo is a guard that doesn't exist.

The design decisions hold together: soft close with the `--confirm` gate keyed off ledger origin, the 50MB tripwire with a measured receipt (~700× headroom) and a "remeasure if a real case hits it" clause, blocking guards that abort loudly rather than drop silently, bags kept forever by default with a human cleanup verb, and the `git bundle` fallback for orphans where "pushed" is unprovable. `attn sweep clear` and the visible sweepings total keep growth honest instead of silent. The GC generalization is scoped correctly too — propose-only GC just manufactures human work.

Checks are green at head (Changelog, Changes, react-doctor, Frontend E2E; Go/Rust legs correctly path-skipped for a docs-only change). Changelog fragment conforms to the per-PR YAML contract.

Two non-blocking notes:

1. **The restore two-liner meets the doc's own spike.** The manifest says "re-worktree the branch, untar over it" — but the spike says 28/35 branches are already gone post-squash-merge, so for the common case there's no branch to re-worktree. Worth one line: fall back to the recorded head hash, or untar onto a worktree of main (the squash content lives there) when the branch is gone. Small edit that closes a hole in the doc's own recovery story.

2. **The cap-tripwire example reads like a sighting.** ".pnpm-store/ 1.8GB, target/ 900MB" in the notification template — but the spike measured dependency stores as phantoms on pnpm's layout (8KB index db, 0B of symlinks). If that's a hypothetical, a "e.g." would keep a future reader from treating it as a second data point.

The arc lands right: automode gave agents denial authority, reconcile gave them loud failure surfaces, and this gives them close authority with provenance and a recoverable sweep — all in the same ledger-first shape the rest of the daemon already lives by. Slice order in the doc is the implementation order; keep it.

figgyster (approved):
Both notes from my approval are closed at this head, and the commit names them.

- **Restore story** — the manifest's two-liner now covers the spike's actual common case: branch pruned by squash-merge means worktree the recorded head hash, or main where the head merged. Recovery no longer depends on a branch that 28/35 worktrees don't have.
- **Cap example** — explicitly hypothetical now, with "the spike observed no such case" on the record, so a future reader won't take it for a second sighting.

Checks green at head (Changelog, Changes, react-doctor, Frontend E2E; Go/Rust legs path-skipped, right for docs-only). Diff-check clean. Approving at the new head.

